PR #1, opened 2017-01-17
Start testing, move CLI to exe/

kirs:
@sirupsen @KnVerey

### DIFF
--- a/exe/kubernetes-deploy
+++ b/exe/kubernetes-deploy
@@ -1,0 +1,16 @@
+#!/usr/bin/env ruby
+require 'kubernetes-deploy'
+
+options = {
+  namespace: ARGV[0],
+  context: ARGV[1],
+  environment: ENV['ENVIRONMENT'],
+  current_sha: ENV['REVISION'],
+  kubeconfig_path: ENV["KUBECONFIG"],
+}
+
+options[:template_dir] = ENV['K8S_TEMPLATE_FOLDER'] || File.expand_path("./config/deploy/#{options.fetch(:environment)}")
+
+KubernetesDeploy.with_friendly_errors do
+  KubernetesDeploy::Runner.new(options).run
+end
--- a/lib/kubernetes-deploy.rb
+++ b/lib/kubernetes-deploy.rb
@@ -1,15 +1,3 @@
-#!/usr/bin/env ruby
-
-# Usage: kubernetes-deploy <app's namespace> <kube context>
-
-# Prerequisites:
-#  - kubectl 1.5.1+ binary must be available in the shipit machine's path
-#  - ENV['KUBECONFIG'] must point to a valid kubeconfig file that includes all the contexts you want to deploy to
-#  - ENV['GOOGLE_APPLICATION_CREDENTIALS'] must point to the credentials for an authenticated service account if your user's auth provider is gcp
-
-# Optionally, the following variables can be used to override script defaults:
-#  - ENV['K8S_TEMPLATE_FOLDER']: Location of Kubernetes files to deploy. Default is config/deploy/#{environment}.
-
 require 'open3'
 require 'securerandom'
 require 'erb'
@@ -23,7 +11,34 @@ require 'active_support/descendants_tracker'
 require 'active_support/core_ext/hash/slice'
 require 'active_support/core_ext/numeric/time'
 
-class KubernetesDeploy
+module KubernetesDeploy
+  def self.logger=(value)
+    @logger = value
+  end
+
+  def self.logger
+    @logger ||= begin
+      l = Logger.new(STDOUT)
+      l.level = ENV["DEBUG"] ? Logger::DEBUG : Logger::INFO
+      l.formatter = proc do |severity, _datetime, _progname, msg|
+        case severity
+        when "FATAL", "ERROR" then "\033[0;31m[#{severity}]\t#{msg}\x1b[0m\n" # red
+        when "WARN" then "\033[0;33m[#{severity}]\t#{msg}\x1b[0m\n" # yellow
+        when "INFO" then "\033[0;36m#{msg}\x1b[0m\n" # blue
+        else "[#{severity}]\t#{msg}\n"
+        end
+      end
+      l
+    end
+  end
+
+  def self.with_friendly_errors
+    yield
+  rescue FatalDeploymentError => error
+    KubernetesDeploy.logger.fatal(error.message)
+    exit 1
+  end
+
   class FatalDeploymentError < StandardError; end
 
   # Things removed from default prune whitelist:
@@ -52,11 +67,13 @@ class KubernetesDeploy
     Pod
   )
 
-  def initialize(namespace:, environment:, current_sha:, template_folder: nil, context:)
+  class Runner
+  def initialize(namespace:, environment:, current_sha:, template_dir: nil, context:, kubeconfig_path:)
     @namespace = namespace
     @context = context
     @current_sha = current_sha
-    @template_path = File.expand_path('./' + (template_folder || "config/deploy/#{environment}"))
+    @kubeconfig_path = kubeconfig_path
+    @template_dir = template_dir
     # Max length of podname is only 63chars so try to save some room by truncating sha to 8 chars
     @id = current_sha[0...8] + "-#{SecureRandom.hex(4)}" if current_sha
   end
@@ -84,9 +101,6 @@ class KubernetesDeploy
     wait_for_completion(resources)
 
     report_final_status(resources)
-  rescue FatalDeploymentError => error
-    KubernetesDeploy.logger.fatal(error.message)
-    exit 1
   end
 
   def template_variables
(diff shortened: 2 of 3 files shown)


exe/kubernetes-deploy
```diff
@@ -0,0 +1,16 @@
+#!/usr/bin/env ruby
+require 'kubernetes-deploy'
+
+options = {
+  namespace: ARGV[0],
+  context: ARGV[1],
+  environment: ENV['ENVIRONMENT'],
+  current_sha: ENV['REVISION'],
+  kubeconfig_path: ENV["KUBECONFIG"],
+}
```
sirupsen: We should consider using proper arguments down the road with a good `--help` prompt. Not needed now, but when we add more arguments that'd be extremely helpful.

exe/kubernetes-deploy
```diff
@@ -0,0 +1,16 @@
...
+
+options[:template_dir] = ENV['K8S_TEMPLATE_FOLDER'] || File.expand_path("./config/deploy/#{options.fetch(:environment)}")
```
sirupsen: The `K8S` in the name here is somewhat redundant, and another environment variable uses `KUBE` as a prefix. I'd say just remove it.

Where does the default come from? As far as I understand, it's not used in core in this manner.

lib/kubernetes-deploy.rb
```diff
@@ -23,7 +11,34 @@
...
 
-class KubernetesDeploy
+module KubernetesDeploy
+  def self.logger=(value)
+    @logger = value
+  end
+
+  def self.logger
+    @logger ||= begin
+      l = Logger.new(STDOUT)
```
sirupsen: `$stderr` is usually the standard logging descriptor

lib/kubernetes-deploy.rb
```diff
@@ -23,7 +11,34 @@
...
+      l.level = ENV["DEBUG"] ? Logger::DEBUG : Logger::INFO
```
sirupsen: I'd consider just making this `ENV["LEVEL"]` for full control, with `ENV["DEBUG"]` implying `ENV["LEVEL"] = Logger::DEBUG`. Nitpick though

lib/kubernetes-deploy.rb
```diff
@@ -23,7 +11,34 @@
...
+      l.formatter = proc do |severity, _datetime, _progname, msg|
+        case severity
+        when "FATAL", "ERROR" then "\033[0;31m[#{severity}]\t#{msg}\x1b[0m\n" # red
+        when "WARN" then "\033[0;33m[#{severity}]\t#{msg}\x1b[0m\n" # yellow
+        when "INFO" then "\033[0;36m#{msg}\x1b[0m\n" # blue
```
sirupsen: What if the terminal doesn't support colours?

lib/kubernetes-deploy.rb
```diff
@@ -84,9 +101,6 @@ def run
     wait_for_completion(resources)
 
     report_final_status(resources)
-  rescue FatalDeploymentError => error
-    KubernetesDeploy.logger.fatal(error.message)
-    exit 1
```
sirupsen: Why did you remove this?

test/kubernetes_deploy_test.rb
```diff
@@ -1,11 +1,51 @@
 require 'test_helper'
 
 class KubernetesDeployTest < Minitest::Test
```
sirupsen: Can you share how you're planning on testing this?

exe/kubernetes-deploy
KnVerey: `$KUBECONFIG` is already handled by k8s as-is: https://kubernetes.io/docs/user-guide/kubeconfig-file/#loading-and-merging-rules.

`config/deploy/{env}` is where we drop the files for non-core apps on cloud init.

lib/kubernetes-deploy.rb
kirs: This went into `with_friendly_errors` which is only used in CLI. In tests, we don't want to `rescue` and exit on exception because we want to test these exceptions.

kirs:
> Can you share how you're planning on testing this?

There are multiple paths we can take.

1) Stub all `kubectl` calls:
```
Kernel.expects(:system).with("kubectl", "--apply", "--more-flags").returns("{ status: "ok" }")
```
This is going to be a mess counting the dozen of `kubectl` calls required for a complete deploy.

2) Use something like VCR and record real calls and output of `kubectl` just once. This is better than 1), but it won't help us to discover what breaks after an upgrade to a new version of kubectl in future

3) Use `minikube` (https://github.com/kubernetes/minikube) and perform real kubernetes testing. This requires installing `minikube` both locally and on CI. It should be the most bulletproof way of testing with a downside of being slow. Also `minikube` requires VirtualBox or KVM installation and this may be tricky on hosted CI environments like Travis and Circle.

4) Move all `kubectl` calls into another abstraction level and use Dependency Injection to simulate it. For instance, we run `kubectl config get-contexts -o name` to get the list of available contexts. We can move that into adapter and use a fake adapter in test environment:

```ruby
class RealKubectl
  def get_contexts
    system("kubectl config get-contexts -o name")
  end
end

class FakeKubectl
  def get_contexts
    ["chi2", "ash"]
  end
end

# the test injects `FakeKubectl` into the verifier
```

Personally I'd prefer 4) but that one requires rewriting most of the script. If we want a quick solution, we should probably take 2).

@sirupsen @KnVerey thoughts?

lib/kubernetes-deploy.rb
kirs: This is the initial code from the script and I've been only moving it around to separate CLI from tests. 
Long term, we should use appropriate library for coloured output that would detect terminal support. I'd leave this to another PRs.

exe/kubernetes-deploy
kirs: I agree with Simon that `K8S_TEMPLATE_FOLDER` looks redundant. @KnVerey do we use this one?

exe/kubernetes-deploy
kirs: Absolutely. I've been only moving this from the script into `exe/kubernetes-deploy` trying not to break many things.

sirupsen: Why `exe`? Isn't `bin` more common for gems?

sirupsen:
I like a variation of (3). It seems like we should be able to set up a K8s cluster with `dev` to make this a pleasant development experience. That said, I am unsure of how much benefit it will provide over (4). It would seem like there might be edge-cases to address in the future, and this is how Katrina tested the script when developing it.

I do think it's worth investing in this given the importance of this script. I am concerned about reaching a point where this keeps growing without a test suite. We've been in that situation with Capistrano, and I 😱  it again! A pro to (3) would be that it'd teach you a bunch about Kubernetes @kirs. The con is that it'd likely take a fair bit longer than (4).

Thoughts? 🤔

exe/kubernetes-deploy
KnVerey: Minimally, having an override available to point the script to another folder can useful when developing (e.g. you can run the script outside shipit and point it to a local folder). It also gives app devs flexibility in code organization; core might even need it? I'd say definitely keep this as an option.

That said, I'd be fine with moving this from an env var to a optional arg to the script if you think that's better. As of yesterday, the standard way to change the option will be via a key in your `shipit.yml`: 
```yaml
kubernetes:
  namespace: my-app-prod
  context: tier4
  template_dir: /app/config/templates
```
Right now, Shipit turns that into an env var [here](https://github.com/Shopify/shipit-engine/blob/master/app/models/shipit/deploy_spec/kubernetes_discovery.rb#L14). It is also possible to set the env var via your shipit machine env (would have been the only way until the first-class api), but I don't think preserving that is necessary.

If we keep this as an env var, making its name less explicit seems like an unnecessary 🚲 🏠  to me.

exe/kubernetes-deploy
KnVerey: http://bundler.io/blog/2015/03/20/moving-bins-to-exe.html

kirs:
Sounds good. I'll spend today investigating how we can use `minikube` for testing.

exe/kubernetes-deploy
sirupsen: I like `KUBECONFIG` if already used. `kubectl(1)` uses a combination of environment variables (for setting your environment) and arguments for whatever is very applicable to the specific command (and not multiple commands, such as the environment).

sirupsen: A combination is fine by that logic, IMO

exe/kubernetes-deploy
sirupsen: oooooh. TIL!

kirs:
Beside from getting the gem fixed, I got my hands on minikube. I could deploy the sample [project](https://github.com/knverey/trashbin) to my local cloud :tada: :tada: :tada:

I also spoke with Graeme about getting CI with enabled minikube. We can't use hosted CI services here because minikube requires a VM support. But turns our that this problem have already been solved by Railgun, which also requires xhyve for testing. We already have a Buildkite pipeline that allows to run any tests on macOS. It should be as easy as creating a template with minikube and setting up BK pipeline.

My next plan is to structure what I had to do today to deploy locally on minikube and prepare the test framework.

Notes:

* prior to using PersistentVolumeClaims, you have to manually provision PersistentVolume with a commands from minikube README

sirupsen:
Great stuff @kirs!

PR #2, opened 2017-01-18
Rip out activesupport

KnVerey:
@kirs @sirupsen @Shopify/byroot 
This rips out the dependency on ActiveSupport to get around https://github.com/Shopify/shipit/issues/124. As you can see, not having it is a very small inconvenience.

I did a quick 🎩  of a happy path deploy of my test app.

### DIFF
--- a/exe/kubernetes-deploy
+++ b/exe/kubernetes-deploy
@@ -10,7 +10,6 @@
 # Optionally, the following variables can be used to override script defaults:
 #  - ENV['K8S_TEMPLATE_FOLDER']: Location of Kubernetes files to deploy. Default is config/deploy/#{environment}.
 
-require 'bundler/setup'
 require 'open3'
 require 'securerandom'
 require 'erb'
@@ -19,10 +18,6 @@ require 'yaml'
 require 'shellwords'
 require 'tempfile'
 require 'logger'
-require 'active_support/core_ext/object/blank'
-require 'active_support/descendants_tracker'
-require 'active_support/core_ext/hash/slice'
-require 'active_support/core_ext/numeric/time'
 
 class KubernetesDeploy
   class FatalDeploymentError < StandardError; end
@@ -159,7 +154,7 @@ class KubernetesDeploy
 
   def wait_for_completion(watched_resources)
     delay_sync_until = Time.now.utc
-    while watched_resources.present?
+    while !watched_resources.empty?
       if Time.now.utc < delay_sync_until
         sleep (delay_sync_until - Time.now.utc)
       end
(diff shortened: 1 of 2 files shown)


exe/kubernetes-deploy
```diff
@@ -294,22 +289,33 @@ class KubernetesDeploy
   end
 
   class KubernetesResource
-    extend ActiveSupport::DescendantsTracker
```
KnVerey: We only need to care about direct descendants of this one class, so I believe what's below is all we need of https://github.com/rails/rails/blob/master/activesupport/lib/active_support/descendants_tracker.rb. However, please scrutinize.

exe/kubernetes-deploy
```diff
@@ -293,23 +288,48 @@ class KubernetesDeploy
     end
   end
 
+  module DescendantsTracker
```
KnVerey: I believe these are the only pieces of https://github.com/rails/rails/blob/master/activesupport/lib/active_support/descendants_tracker.rb we actually need.

exe/kubernetes-deploy
```diff
@@ -293,23 +288,48 @@ class KubernetesDeploy
...
+        DescendantsTracker.store_inherited(self, base)
+        super
+      end
+
+      def handled_type
+        name.split('::').last
+      end
+
+      def for_type(type, name, namespace, file)
+        if subclass = direct_descendants.find { |subclass| subclass.handled_type.downcase == type }
```
kirs: If what you need is getting `Service.new` from `"service"` as a string, you can do the following:

```
type = "service"
Object.const_get(type.capitalize)
```

This way you won't need `DescendantsTracker` at all.

KnVerey: True (even `KubernetesResource.const_get(type.capitalize)`, I believe?), but is that really better? It feels like a bit of a cheat (i.e. reliable because by convention constants capitalized "Randomtype" will be classes), and it would prevent us from camelizing the class names (e.g. PersistentVolumeClaim -> Persistentvolumeclaim).

kirs: This is how Ruby developers usually do (iterating over `DescendantsTracker` sounds like an antipattern in this case).

For PVL, the right thing would be:

```
irb(main):025:0> "persistent_volume_claim".classify
=> "PersistentVolumeClaim"
```

(classify comes from AS)

kirs: (assuming `persistent_volume_claim` comes from the type)

KnVerey: > (assuming persistent_volume_claim comes from the type)

Sadly no. This is coming from kubectl as `persistentvolumeclaim`.

> This is how Ruby developers usually do (iterating over DescendantsTracker sounds like an antipattern in this case). For PVL, the right thing would be: ... (classify comes from AS)

Calling it an antipattern doesn't really help me understand why you think one option is better than the other. Both approaches are available in ActiveSupport. What I want is to find the subclass that exists to support that type, so doing exactly that doesn't sound inherently bad to me.

casperisfine: I wouldn't say antipattern. But it's definitely better to strive for "strategy pattern" over iterating on all candidates to find what you are looking for.

Not that it makes a huge difference performance wise (since in this kind of cases, the candidates set is generally small), but it usually makes simpler, more direct code to read.

NB: I'm just answering this specific thread cause I saw some keywords that triggered me. But I haven't kept a close watch on that PR, so feel free to disregard my opinion.

KnVerey:
Closing. The dependency problem was fixed quickly enough that this wasn't necessary.

PR #4, opened 2017-01-20
Move exception rescue into a wrapper

kirs:
We don't want to rescue and log all exceptions in tests. `with_friendly_errors` is only used for CLI. I also added the backtrace reporting to be able to find the source of error.

Before:

screen shot 2017-01-20 at 12 22 03

After:

screen shot 2017-01-20 at 12 21 34

review @sirupsen @ibawt @KnVerey

### DIFF
--- a/lib/kubernetes-deploy/runner.rb
+++ b/lib/kubernetes-deploy/runner.rb
@@ -9,6 +9,16 @@ require 'kubernetes-deploy/kubernetes_resource'
 
 module KubernetesDeploy
   class Runner
+    def self.with_friendly_errors
+      yield
+    rescue FatalDeploymentError => error
+      KubernetesDeploy.logger.fatal <<-MSG
+#{error.class}: #{error.message}
+  #{error.backtrace && error.backtrace.join("\n  ")}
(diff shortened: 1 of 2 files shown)


lib/kubernetes-deploy/runner.rb
```diff
@@ -9,6 +9,16 @@
 
 module KubernetesDeploy
   class Runner
+    def self.with_friendly_errors
+      yield
+    rescue FatalDeploymentError => error
+      KubernetesDeploy.logger.fatal <<-MSG
+#{error.class}: #{error.message}
+  #{error.backtrace && error.backtrace.join("\n  ")}
```
sirupsen: Why would the backtrace not be there?

kirs: Dunno, but if Bundler [does that](https://github.com/bundler/bundler/blob/master/lib/bundler/friendly_errors.rb#L92) I believe they have a reason.

sirupsen: 😱

PR #5, opened 2017-01-20
Make logger configurable

kirs:
We are going to use different logger in tests and we need to make it configurable. We don't want to log to `$stderr` in tests.

review @knverey @sirupsen

### DIFF
--- a/lib/kubernetes-deploy.rb
+++ b/lib/kubernetes-deploy.rb
@@ -34,10 +34,14 @@ module KubernetesDeploy
     Pod
   )
 
+  def self.logger=(value)
+    @logger = value
+  end
+
   def self.logger
     @logger ||= begin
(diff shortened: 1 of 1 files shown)


lib/kubernetes-deploy.rb
```diff
@@ -34,10 +34,14 @@ class FatalDeploymentError < StandardError; end
...
 
+  def self.logger=(value)
+    @logger = value
+  end
+
   def self.logger
     @logger ||= begin
-      l = Logger.new(STDOUT)
-      l.level = ENV["DEBUG"] ? Logger::DEBUG : Logger::INFO
+      l = Logger.new($stderr)
```
kirs: as mentioned by @sirupsen in https://github.com/Shopify/kubernetes-deploy/pull/1/files#r96518542

PR #6, opened 2017-01-20
Unify template_dir and template_path

kirs:
Unix terminology prefers "directory" over "folder".

review @sirupsen @KnVerey

### DIFF
--- a/lib/kubernetes-deploy/runner.rb
+++ b/lib/kubernetes-deploy/runner.rb
@@ -19,11 +19,11 @@ MSG
       exit 1
     end
 
-    def initialize(namespace:, environment:, current_sha:, template_folder: nil, context:)
+    def initialize(namespace:, environment:, current_sha:, context:, template_dir: nil)
       @namespace = namespace
       @context = context
       @current_sha = current_sha
-      @template_path = File.expand_path('./' + (template_folder || "config/deploy/#{environment}"))
+      @template_dir = template_dir || File.expand_path("./config/deploy/#{environment}")
       # Max length of podname is only 63chars so try to save some room by truncating sha to 8 chars
       @id = current_sha[0...8] + "-#{SecureRandom.hex(4)}" if current_sha
     end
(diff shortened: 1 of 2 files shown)


lib/kubernetes-deploy/runner.rb
```diff
@@ -19,11 +19,11 @@ def self.with_friendly_errors
       exit 1
     end
 
-    def initialize(namespace:, environment:, current_sha:, template_folder: nil, context:)
+    def initialize(namespace:, environment:, current_sha:, context:, template_dir: nil)
       @namespace = namespace
       @context = context
       @current_sha = current_sha
-      @template_path = File.expand_path('./' + (template_folder || "config/deploy/#{environment}"))
+      @template_dir = template_dir || File.expand_path("./config/deploy/#{environment}")
```
KnVerey: Doesn't this change `ENV['K8S_TEMPLATE_FOLDER']` from being relative to being absolute? When this is being run in shipit, the current directory will always be app root, which is the reference point devs setting this option have access to. They won't know the absolute path.

kirs: It does, but for ENV variables it's always a good practice to use the absolute path (you can check `env` result on your laptop). The idea is behind ENV is that it's global, and when you set a path you don't know where from the script will execute.

> When this is being run in shipit, the current directory will always be app root, which is the reference point devs setting this option have access to. They won't know the absolute path.

Not sure I understand the concern. I couldn't find any references of `K8S_TEMPLATE_FOLDER` in Shipit too.

KnVerey: Not sure where you were looking, but this variable drives the `template_dir` option: https://github.com/Shopify/shipit-engine/blob/master/app/models/shipit/deploy_spec/kubernetes_discovery.rb#L14. I.e. if you have this in your shipit.yml:

```yaml
kubernetes:
  namespace: my-app
  context: tier4
  template_dir: config/deploy/k8s/
```
This PR would make it so you'd need to know exactly where on your Shipit machine your repo gets cloned, and if that changes, it'll need to be updated on N repos. Note that the env var can also be set directly via the `machine.environment` section.

That said, this behaviour is kinda an artifact of the pre-first-class api. I'd be fine with making  `kubernetes.template_dir` result in an argument to the script (still a relative path) instead of the setting of an env var, and changing this env var to be an override intended for local development. But that's not how it works right now.

kirs: I see. We should totally make it use a relative argument as you suggested.
I'll update the PR and shipit.

PR #7, opened 2017-01-20
Make fixtures friendly to minikube

kirs:
@KnVerey


kirs:
* Don't refer to private container registries
* Don't fake `kubeconfig.yml`

PR #3, opened 2017-01-19
Put things where they belong in a gem

KnVerey:
@kirs @sirupsen

This simply moves code around to split classes into their own files and move the files to their conventional locations for a gem. Let me know if something still isn't quite right!

The only code change is to make `KubernetesDeploy` a module and name the extracted runner class `KubernetesDeploy::Runner`, as Kir did on his branch.

PR #8, opened 2017-01-21
Reorganization follow-up

KnVerey:
@kirs @sirupsen  

This PR:
- Puts some constants I missed moving in https://github.com/Shopify/kubernetes-deploy/pull/3 where they belong.
- Addresses https://github.com/Shopify/kubernetes-deploy/pull/3#discussion_r96986742 (which I see I didn't answer--sorry about that!). The reason I put it at the bottom was that the superclass had to be loaded first, or else I'd get `uninitialized constant KubernetesDeploy::KubernetesResource (NameError)`.  Do you like this better?
- Fixes capitalization of the `ConfigMap` subclass to match the k8s norm.
- Replaces DescendantsTracker with a more straightforward solution, as previously requested.

### DIFF
--- a/lib/kubernetes-deploy/kubernetes_resource.rb
+++ b/lib/kubernetes-deploy/kubernetes_resource.rb
@@ -1,26 +1,24 @@
-require 'active_support/descendants_tracker'
 require 'json'
 require 'open3'
 require 'shellwords'
 
 module KubernetesDeploy
   class KubernetesResource
-    extend ActiveSupport::DescendantsTracker
 
     attr_reader :name, :namespace, :file
     attr_writer :type, :deploy_started
 
     TIMEOUT = 5.minutes
 
-    def self.handled_type
-      name.split('::').last
-    end
-
     def self.for_type(type, name, namespace, file)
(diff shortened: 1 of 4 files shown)


lib/kubernetes-deploy/kubernetes_resource.rb
```diff
@@ -1,26 +1,24 @@
...
-      else
-        self.new(name, namespace, file).tap { |r| r.type = type }
+      case type
+      when 'configmap' then ConfigMap.new(name, namespace, file)
+      when 'deployment' then Deployment.new(name, namespace, file)
+      when 'pod' then Pod.new(name, namespace, file)
+      when 'ingress' then Ingress.new(name, namespace, file)
+      when 'persistentvolumeclaim' then PersistentVolumeClaim.new(name, namespace, file)
+      when 'service' then Service.new(name, namespace, file)
+      else self.new(name, namespace, file).tap { |r| r.type = type }
```
sirupsen: 👍

kirs:
🚢

PR #9, opened 2017-01-23
Add cloudsql resource support

wfarr:


wfarr:
Bump

ibawt:
@Shopify/cloudplatform reviews in here please

lib/kubernetes-deploy/kubernetes_resource/cloudsql.rb
```diff
@@ -0,0 +1,28 @@
+module KubernetesDeploy
+  class Cloudsql < KubernetesResource
+    TIMEOUT = 30.seconds
```
ibawt: is this in activesupport?

wfarr: Yes.

That KubernetesResource was more or less cargo culted from the other examples.

wfarr:
Going to make some more invasive changes here.

As per https://github.com/kubernetes/kubernetes/issues/39906, going to end up moving the TPR flow to use `create`/`update` instead of `apply` until we have a better option.

ibawt:
and you'll need to own bumping the gem etc.

wfarr:
This should now skip `kubectl apply` calls for TPRs (preferring `kubectl update` instead) and use `create --dry-run` for discovery

lib/kubernetes-deploy/runner.rb
```diff
@@ -76,11 +78,14 @@ def run
...
       resources = discover_resources
 
+      tprs = resources.reject! { |r| r.tpr? }
+
       phase_heading("Checking initial resource statuses")
       resources.each(&:sync)
 
       phase_heading("Predeploying priority resources")
       predeploy_priority_resources(resources)
+      update_tprs(tprs)
```
KnVerey: This minimally needs to happen before the predeploy, which can contain rails migrations that require the db. IMO it would be better if we can keep it part of the sequence though. Otherwise, we're circumventing the initial state discovery on line 84 and also skipping (or needing to repeat the implementation of) resource polling. How about augmenting `deploy_resources` to be able to handle tprs correctly, e.g. by checking whether any in the set are `.tpr?` and adjusting the command accordingly?

lib/kubernetes-deploy/kubernetes_resource/cloudsql.rb
```diff
@@ -0,0 +1,28 @@
...
+
+    def sync
+      _, st = run_kubectl("get", type, @name)
+      @found = st.success?
+      @status = true
+      log_status
+    end
+
+    def deploy_succeeded?
+      exists?
```
wfarr: Best case is blocking on the underlying service and deployment to be created.

lib/kubernetes-deploy/runner.rb
wfarr: We don't necessarily want `update` to be used for non-TPRs (it can be destructive as it's a wholesale replacement of the resource, it doesn't do any merge strategy magic).

Right now, the `deploy_resources` appears to create a single `apply` for all resources. Is this intended? This fix would be easy if the `apply` line was distinct per resource, but the implementation is currently:

``` ruby
    def deploy_resources(resources, prune: false)
      command = ["apply", "--namespace=#{@namespace}"]
      KubernetesDeploy.logger.info("Deploying resources:")

      resources.each do |r|
        KubernetesDeploy.logger.info("- #{r.id}")
        command.push("-f", r.file.path)
        r.deploy_started = Time.now.utc
      end

      if prune
        command.push("--prune", "--all")
        PRUNE_WHITELIST.each { |type| command.push("--prune-whitelist=#{type}") }
      end

      run_kubectl(*command)
    end
```

KnVerey: The single apply is intentional, but it is only essential when `prune: true`. Perhaps something like this? i.e. still have special treatment, but without circumventing the regular flow:

```ruby
def deploy_resources(resources, prune: false)
  tprs, resources = resources.partition(&:tpr?)
  deploy_tprs(tprs) if tprs.present?
 ...
```

lib/kubernetes-deploy/kubernetes_resource/cloudsql.rb
KnVerey: > Best case is blocking on the underlying service and deployment to be created.

Does deployment can fail readiness until the instance is available? That would be great. Otherwise, still better than nothing. A resource looking for another resource isn't entirely unprecedented here, in that I've got services waiting to have endpoints before succeeding.

wfarr: > Does deployment fail readiness until the instance is available, or could it? That would be great. Otherwise, still better than nothing. A resource looking for another resource isn't entirely unprecedented here, in that I've got services waiting to have endpoints before succeeding.

The Deployment (and thus Service) cannot be created until the instance is created.

wfarr: This is special to this case, as the instance name is part of the required CLI flags passed to the pod spec in the Deployment.

lib/kubernetes-deploy/runner.rb
wfarr: 717ed99 should implement this

lib/kubernetes-deploy/kubernetes_resource.rb
```diff
@@ -69,6 +70,10 @@ def deploy_timed_out?
       !deploy_succeeded? && !deploy_failed? && (Time.now.utc - @deploy_started > self.class::TIMEOUT)
     end
 
+    def tpr?
+      type == "Cloudsql"
```
wfarr: Done.

lib/kubernetes-deploy/kubernetes_resource/cloudsql.rb
wfarr: 00b3787 should implement something basic

lib/kubernetes-deploy/kubernetes_resource/cloudsql.rb
```diff
@@ -0,0 +1,66 @@
...
+
+      log_status
+    end
+
+    def deploy_succeeded?
+      cloudsql_proxy_deployment_exists? && mysql_service_exists?
+    end
+
+    def deploy_failed?
+      !cloudsql_proxy_deployment_exists? || !mysql_service_exists?
```
KnVerey: If I'm not mistaken, these conditions mean that the deploy is incomplete, not that it failed. `deploy_failed?` should only return `true` when it the resource is doomed--it'll abort the Shipit deploy. As a result, several resource types have `false` hardcoded here. Does the cloudsql controller report anything if instance provisioning hard-fails for some reason (e.g. a instance count cap)?

lib/kubernetes-deploy/kubernetes_resource/cloudsql.rb
```diff
@@ -0,0 +1,66 @@
...
+      true
+    end
+
+    def exists?
+      @found
+    end
+
+    private
+    def cloudsql_proxy_deployment_exists?
+      deployments, st = run_kubectl("get", "deployments", "--selector", "name=cloudsql-proxy", "--namespace=#{@namespace}", "-o=json")
```
KnVerey: Since you're only interested in one, why not use `"get", "deployments, "cloudsql-proxy", "--output=json"`? Note that you don't need to send the namespace. It is always required in the context of these classes, so `run_kubectl` adds it and in fact raises if `@namespace` is blank.

(same questions/comments about the mysql version below)

lib/kubernetes-deploy/kubernetes_resource/cloudsql.rb
```diff
@@ -0,0 +1,66 @@
...
+
+      false
+    end
+
+    def mysql_service_exists?
+      services, st = run_kubectl("get", "services", "--selector", "name=mysql", "--namespace=#{@namespace}", "-o=json")
+      if st.success?
+        services_list = JSON.parse(services)
+
+        if .any? { |s| s.fetch("spec", {}).fetch("clusterIP", "") != "" }
```
KnVerey: presumably `services_list.any?`

lib/kubernetes-deploy/runner.rb
```diff
@@ -213,10 +215,22 @@ def validate_configuration
       KubernetesDeploy.logger.info("All required parameters and files are present")
     end
 
+    def update_tprs(resources)
+      resources.each do |r|
+        KubernetesDeploy.logger.info("- #{r.id}")
+        r.deploy_started = Time.now.utc
+        run_kubectl("update", "--namespace=#{@namespace}", "-f", r.file.path)
```
KnVerey: `run_kubectl` already adds the namespace

lib/kubernetes-deploy/kubernetes_resource/cloudsql.rb
```diff
@@ -0,0 +1,66 @@
...
+    TIMEOUT = 30.seconds
+
+    def initialize(name, namespace, file)
+      @name, @namespace, @file = name, namespace, file
+    end
+
+    def sync
+      _, st = run_kubectl("get", type, @name)
+      @found = st.success?
+      @status = true
```
wfarr: Do you have any suggestions for string content? This is backed by a pair of k8s resources each with their own possible descriptive string.

lib/kubernetes-deploy/kubernetes_resource/cloudsql.rb
wfarr: Only via logs sent to splunk. There is no Status field on ThirdPartyResources yet. If and when there is one, we could report it.

lib/kubernetes-deploy/runner.rb
wfarr: Ah - I copied what `deploy_resources` was doing. d10c217 should remove the redundant `--namespace` flag in both cases.

lib/kubernetes-deploy/kubernetes_resource/cloudsql.rb
KnVerey: Hmmm... Ideally, it would use the status field you explained doesn't exist yet, but maybe we could fake "Unknown"/"Provisioning"/"Provisioned"/"Available" via `@found == false`, `@found == true`, `mysql_service_exists?` and `deploy_succeeded?`? Or something like that? Or minimally "Unknown"/"Provisioned"?

wfarr:
Found (and fixed) a number of minor issues in 🎩 

One more set of 👀 and this should be good to go I think

masom:
Is there something we can do to get our app deployed while this is being worked on?

lib/kubernetes-deploy/runner.rb
```diff
@@ -220,10 +221,27 @@ def validate_configuration
...
+
+        unless st.success?
+          # it doesn't exist so we can't replace it
+          run_kubectl("create", "-f", r.file.path)
+        end
+      end
+    end
+
     def deploy_resources(resources, prune: false)
-      command = ["apply", "--namespace=#{@namespace}"]
```
kirs: why did you remove the namespace flag?

wfarr: https://github.com/Shopify/kubernetes-deploy/pull/9#discussion_r98566860

ibawt:
@masom db should be up today at some point, if not ping me or the oncall cloudplatform and we'll manually set it up.

wfarr:
Reminder: this is a public repo. The DB is and has been up, the problem is relegated _solely_ to shipit blowing up on attempting to `kubectl apply` the cloudsql resource as part of the deployment.

yfxie:
@wfarr Can you give me an explaination of this usage? I read the source code, it seems like a way to ensure DB running before other resources. Thank you!

KnVerey:
@yfxie This was Shopify adding support for our internal "Cloudsql" custom resource (CR) a long time ago. We've since implemented generic support for CRs: documentation is available here: https://github.com/Shopify/kubernetes-deploy#deploying-custom-resources. You can customize whether your CRs get predeployed (the default is `true`) using an annotation on the CRD, as described here: https://github.com/Shopify/kubernetes-deploy#customizing-behaviour-with-annotations.

PR #10, opened 2017-01-27
Allow to skip waiting for completion

kirs:
Roughly, the current workflow of the script is:

1) kubectl apply
2) watch for changes to be propagated

I'm going to start deploying Shopify core with the script and at the same time I don't want to block the current deploys with the watch & wait part if something breaks. To allow that, I'm adding the `--skip-wait` flag. I'm open to rename it to `--skip-watch` if you think it's a better name.

I'd like to ship this ASAP to start deploying today.

review @sirupsen @KnVerey

### DIFF
--- a/exe/kubernetes-deploy
+++ b/exe/kubernetes-deploy
@@ -12,13 +12,22 @@
 
 require 'kubernetes-deploy'
 
+require 'optparse'
+
+skip_wait = false
+ARGV.options do |opts|
+  opts.on("-w", "--skip-wait") { skip_wait = true }
+  opts.parse!
+end
+
 KubernetesDeploy::Runner.with_friendly_errors do
   runner = KubernetesDeploy::Runner.new(
     namespace: ARGV[0],
     context: ARGV[1],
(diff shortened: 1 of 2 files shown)


exe/kubernetes-deploy
```diff
@@ -12,13 +12,22 @@
...
+
 KubernetesDeploy::Runner.with_friendly_errors do
   runner = KubernetesDeploy::Runner.new(
     namespace: ARGV[0],
     context: ARGV[1],
     environment: ENV['ENVIRONMENT'],
     current_sha: ENV['REVISION'],
-    template_folder: ENV['K8S_TEMPLATE_FOLDER']
+    template_folder: ENV['K8S_TEMPLATE_FOLDER'],
+    wait_for_completion: !skip_wait,
```
kirs: Later, I'll move all `ENV` variables into CLI arguments. Didn't want to do this in the same PR.

KnVerey: Do not move `ENVIRONMENT` or `REVISION` please. Those appear dynamically in the shipit env.

PR #11, opened 2017-01-28
Accept --template-dir as an argument

kirs:
This is required to specify custom directory with k8s templates. It's `config/deploy/$ENVIRONMENT` by default, but for Shopify core it has to be another path. Right now the deploy fails because it can't find any k8s configs in `config/deploy/$ENVIRONMENT`.

closes https://github.com/Shopify/kubernetes-deploy/pull/6

review @sirupsen @KnVerey

### DIFF
--- a/exe/kubernetes-deploy
+++ b/exe/kubernetes-deploy
@@ -15,18 +15,28 @@ require 'kubernetes-deploy'
 require 'optparse'
 
 skip_wait = false
+template_dir = nil
 ARGV.options do |opts|
-  opts.on("-w", "--skip-wait") { skip_wait = true }
+  opts.on("--skip-wait") { skip_wait = true }
+  opts.on("--template-dir=DIR") { |v| template_dir = v }
   opts.parse!
+end
+
+if !template_dir && ENV.key?("ENVIRONMENT")
+  template_dir = "config/deploy/#{ENV["ENVIRONMENT"]}"
+end
+
+if !template_dir || template_dir.empty?
+  puts "Template directory is unknown. Either specify --template-dir argument or set $ENVIRONMENT to use config/deploy/$ENVIRONMENT as a default path."
+  exit 1
 end
 
 KubernetesDeploy::Runner.with_friendly_errors do
   runner = KubernetesDeploy::Runner.new(
     namespace: ARGV[0],
     context: ARGV[1],
-    environment: ENV['ENVIRONMENT'],
     current_sha: ENV['REVISION'],
-    template_folder: ENV['K8S_TEMPLATE_FOLDER'],
+    template_dir: template_dir,
     wait_for_completion: !skip_wait,
   )
   runner.run
--- a/lib/kubernetes-deploy/runner.rb
+++ b/lib/kubernetes-deploy/runner.rb
@@ -55,11 +55,12 @@ MSG
       exit 1
     end
 
-    def initialize(namespace:, environment:, current_sha:, context:, wait_for_completion:, template_folder: nil)
+    def initialize(namespace:, current_sha:, context:, wait_for_completion:, template_dir:)
       @namespace = namespace
       @context = context
       @current_sha = current_sha
-      @template_path = File.expand_path('./' + (template_folder || "config/deploy/#{environment}"))
+
+      @template_dir = File.expand_path(template_dir)
       # Max length of podname is only 63chars so try to save some room by truncating sha to 8 chars
       @id = current_sha[0...8] + "-#{SecureRandom.hex(4)}" if current_sha
       @wait_for_completion = wait_for_completion
@@ -119,7 +120,7 @@ MSG
 
     def discover_resources
       resources = []
-      Dir.foreach(@template_path) do |filename|
+      Dir.foreach(@template_dir) do |filename|
         next unless filename.end_with?(".yml.erb", ".yml")
 
         split_templates(filename) do |tempfile|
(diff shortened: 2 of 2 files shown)


lib/kubernetes-deploy/runner.rb
```diff
@@ -55,11 +55,12 @@ def self.with_friendly_errors
...
     end
 
-    def initialize(namespace:, environment:, current_sha:, context:, wait_for_completion:, template_folder: nil)
+    def initialize(namespace:, current_sha:, context:, wait_for_completion:, template_dir:)
       @namespace = namespace
       @context = context
       @current_sha = current_sha
-      @template_path = File.expand_path('./' + (template_folder || "config/deploy/#{environment}"))
+
+      @template_dir = Pathname.new(template_dir).realpath
```
kirs: @KnVerey this way we can accept both absolutely and relative paths; relative path will be converted to absolute.

exe/kubernetes-deploy
```diff
@@ -15,18 +15,23 @@ require 'kubernetes-deploy'
...
+template_dir = nil
 ARGV.options do |opts|
   opts.on("-w", "--skip-wait") { skip_wait = true }
+  opts.on("-tDIR", "--template-dir=DIR") { |v| template_dir = v }
   opts.parse!
 end
 
+if !template_dir || template_dir.empty?
+  raise ArgumentError, "--template-dir flag is required"
+end
```
sirupsen: Do you think this is required? The default of `config/deploy` makes sense to me 🥇

exe/kubernetes-deploy
```diff
@@ -15,18 +15,23 @@ require 'kubernetes-deploy'
 require 'optparse'
 
 skip_wait = false
+template_dir = nil
 ARGV.options do |opts|
   opts.on("-w", "--skip-wait") { skip_wait = true }
+  opts.on("-tDIR", "--template-dir=DIR") { |v| template_dir = v }
```
sirupsen: I wonder if we should have the shorthands (`-w` and `-t`). We'd always spell them out in scripts, and this shouldn't really be run by hand—even then it's prone to error. Should we consider removing them?

lib/kubernetes-deploy/runner.rb
sirupsen: What is the difference between `Pathname.new(..)` and `File.expand_path`?

```ruby
irb(main):007:0> Pathname.new("~/.bashrc").realpath
Errno::ENOENT: No such file or directory @ realpath_rec - /Users/sirup/~
	from (irb):7:in `realpath'
	from (irb):7:in `realpath'
	from (irb):7
	from /opt/rubies/2.3.3/bin/irb:11:in `<main>'
irb(main):008:0> File.expand_path("~/.bashrc")
=> "/Users/sirup/.bashrc"
```

In this I feel `File.expand_path` wins 🤔  But for anything not involving `~`, it seems to work the same.

exe/kubernetes-deploy
KnVerey: Agreed, this should not be a required arg and the default should remain on this side.

lib/kubernetes-deploy/runner.rb
KnVerey: `File.expand_path` does look preferable given those examples, but really I'm happy as long as it is possible to write `template_dir: config/my/custom/path` (relative, no tilde) into your shipit.yml.

exe/kubernetes-deploy
kirs: The thing is that we want the default to be `config/deploy/$ENVIRONMENT`, so the script would need `environment` argument just to construct the default template dir. IMO adding extra argument (environment) only to construct the template dir is not the best design.

This is the reason why I moved the default `--template-dir` construction to Shipit side where $ENVIRONMENT is available: https://github.com/Shopify/shipit-engine/pull/640/files#diff-01916dce352e57dea198b45614795a9dR17

kirs:
Thanks for review! I updated the code to use `File.expand_path` and removed shorthands of CLI arguments because I agree that we don't really need them. Please let me know what you think about setting default template dir on the Shipit side.

exe/kubernetes-deploy
KnVerey: I see your point, though I'm not convinced that the default belongs in the other repo more than it does here. You could also eliminate the arg to `Runner.new` by setting the default right here, no? I don't feel super strongly about this though.

sirupsen: Ah. Can we remove the need for `ENVIRONMENT` altogether then?

KnVerey:
That variable is always set by shipit in all deploys.

sirupsen:
> That variable is always set by shipit in all deploys.

Right! I was thinking that this script might not need to depend on it anymore if `--template-dir` was the last thing that required it. Could it makes sense to do the following:

1. If `--template-dir` is passed, used whatever is passed.
2. If `$ENVIRONMENT` is passed (always true for 🚢 as Katrina pointed out), use the `config/deploy/$ENVIRONMENT` default (and make this a strong convention).
3. Else, error asking people to either provide `$ENVIRONMENT` to infer the template directory or `--template-dir`

👍 👎 ❓

kirs:
Great suggestion, Simon! I implemented the idea.
Please 👀 when you have a minute because I'd like to ship this ASAP for the undercloud team.

lib/kubernetes-deploy/runner.rb
```diff
@@ -201,10 +202,10 @@ def validate_configuration
         errors << "Current SHA must be specified"
       end
 
-      if !File.directory?(@template_path)
-        errors << "Template path #{@template_path} doesn't exist"
-      elsif Dir.entries(@template_path).none? { |file| file =~ /\.yml(\.erb)?$/ }
-        errors << "#{@template_path} doesn't contain valid templates (postfix .yml or .yml.erb)"
+      if !File.directory?(@template_dir)
+        errors << "Template path #{@template_dir} doesn't exist"
```
KnVerey: "Template directory" in the error message for consistency?

kirs: Thanks, fixed!

PR #13, opened 2017-02-01
Don't try to apply empty resource set

KnVerey:
This PR:
- Fixes the following error:
screenshot
- Make us log the command that failed so it's easier to find out what happened in the future.

### DIFF
--- a/lib/kubernetes-deploy/runner.rb
+++ b/lib/kubernetes-deploy/runner.rb
@@ -241,13 +241,14 @@ MSG
     end
 
     def deploy_resources(resources, prune: false)
-      command = ["apply"]
       KubernetesDeploy.logger.info("Deploying resources:")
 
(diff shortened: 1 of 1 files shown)


wfarr:
> LGTM, but how did we end up with an empty resource? I'd say that maybe we should prevent developers from using empty resources if it's a mistake in YAML config

This case is encountered if the only resource being applied in a single `deploy_resources` call is a third party resource. This happens because our primary third party resource right now, Cloudsql, is in the priority list and must be deployed before the rest of the app components. When we split the TPRs out to `update_tpr`, we're left with an empty resource list in that case, which surfaces this warning.

kirs:
@wfarr thanks! 👍

KnVerey:
Just did a 🎩  against `trashbin` and it looks good

PR #14, opened 2017-02-02
Refuse to deploy to system namespaces

KnVerey:
This deploy script assumes it is in charge of managing the whole namespace it deploys to (namespace == app-env). As a result, it is never a good idea to use it in the default or system namespaces. 

@Shopify/cloudplatform @kirs

### DIFF
--- a/lib/kubernetes-deploy/runner.rb
+++ b/lib/kubernetes-deploy/runner.rb
@@ -26,6 +26,10 @@ module KubernetesDeploy
       PersistentVolumeClaim
       Pod
     )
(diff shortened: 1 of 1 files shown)


KnVerey:
> would love to hear that tests are on the horizon

Yes, Kir has been working on them 🎉  https://github.com/Shopify/kubernetes-deploy/pull/15

I'm going to merge this but wait for the bugfix in https://github.com/Shopify/kubernetes-deploy/pull/17 to bump the version / update shipit (shipit is pinned to the previous version).

PR #16, opened 2017-02-03
Allow script to run concurrently

kirs:
Right now on the script start we run `kubectl config use-context`. This command is intended to update `kubeconfig`, it also creates a lockfile around it and as a result makes it impossible to run multiple `kubernetes-deploy` commands at the same time.

**Why:** we want to deploy to multiple clusters in parallel (https://github.com/Shopify/shopify/pull/99677#issuecomment-277070450). Right now any parallel runs of `kubernetes-deploy` will fail because they will cause a race condition by concurrent `kubectl config use-context`.

**How:** pass `--context` argument to all `kubectl` invocation, as we do for `--namespace` argument.

review @ibawt @sirupsen @KnVerey @wfarr @n1koo

### DIFF
--- a/lib/kubernetes-deploy/runner.rb
+++ b/lib/kubernetes-deploy/runner.rb
@@ -264,17 +264,13 @@ MSG
     end
 
     def set_kubectl_context
-      out, err, st = run_kubectl("config", "get-contexts", "-o", "name", namespaced: false)
+      out, err, st = run_kubectl("config", "get-contexts", "-o", "name", namespaced: false, with_context: false)
       available_contexts = out.split("\n")
       if !st.success?
         raise FatalDeploymentError, err
       elsif !available_contexts.include?(@context)
         raise FatalDeploymentError, "Context #{@context} is not available. Valid contexts: #{available_contexts}"
       end
-
-      _, err, st = run_kubectl("config", "use-context", @context, namespaced: false)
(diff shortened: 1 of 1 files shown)


lib/kubernetes-deploy/runner.rb
```diff
@@ -283,12 +279,15 @@ def validate_namespace
       KubernetesDeploy.logger.info("Namespace #{@namespace} validated")
     end
 
-    def run_kubectl(*args, namespaced: true)
+    def run_kubectl(*args, namespaced: true, with_context: true)
       args = args.unshift("kubectl")
       if namespaced
-        raise FatalDeploymentError, "Namespace missing for namespaced command" unless @namespace
```
KnVerey: 💯  on the `--context` flag, but why did you remove this safeguard? It's true that we shouldn't ever hit it because of up-front validations, but it exists because leaving out the namespace would be very, very bad. Same thing for context, really.

KnVerey:
❗️  don't we need this in `run_kubectl` lib/kubernetes-deploy/kubernetes_resource.rb too? Did you 🎩  this?

lib/kubernetes-deploy/runner.rb
kirs: Because it's been validated before and I felt like we don't need to duplicate it. But I see your point and we can leave the duplicated check if you strongly feel that it's necessary to check it twice (in case someone forgets extra equal sign in `if @context = "value"`

PR #17, opened 2017-02-03
Context flag for KubernetesResource#run_kubectl

KnVerey:
@kirs @sirupsen @mkobetic this is the minimum change need to fix the problem described in https://github.com/Shopify/shipit/pull/141. Another option is to make the runner's version a utility method available to both the runner instance and the resource instances, although that wouldn't get around passing the namespace/context everywhere (suggestions on that welcome). Some background:
- The original reason for having two different versions is a difference in error handling and logging. The runner version handles critical deploy commands that could fail, and both its callers and the end user should be interested in error output as a result. The resources on the other hand are doing polling, in which context typical `!st.success?` causes are transient or even expected, e.g. 404s on a GETs for resources that aren't up yet.
- As Kir suggested in another issue, the resource instances' version should likely be replaced with using the API via a gem rather than shelling out to kubectl.

I've done a quick 🎩  already, but will also do the following on the final version after review:
- [x] context missing
- [x] context provided does not match current-context in kubeconfig
- [x] at least one success and one error deploy with the `trashbin` set of files

### DIFF
--- a/lib/kubernetes-deploy/kubernetes_resource.rb
+++ b/lib/kubernetes-deploy/kubernetes_resource.rb
@@ -5,27 +5,27 @@ require 'shellwords'
 module KubernetesDeploy
   class KubernetesResource
 
-    attr_reader :name, :namespace, :file
+    attr_reader :name, :namespace, :file, :context
     attr_writer :type, :deploy_started
 
     TIMEOUT = 5.minutes
 
-    def self.for_type(type, name, namespace, file)
+    def self.for_type(type, name, namespace, context, file)
       case type
-      when 'cloudsql' then Cloudsql.new(name, namespace, file)
-      when 'configmap' then ConfigMap.new(name, namespace, file)
-      when 'deployment' then Deployment.new(name, namespace, file)
-      when 'pod' then Pod.new(name, namespace, file)
-      when 'ingress' then Ingress.new(name, namespace, file)
-      when 'persistentvolumeclaim' then PersistentVolumeClaim.new(name, namespace, file)
-      when 'service' then Service.new(name, namespace, file)
-      else self.new(name, namespace, file).tap { |r| r.type = type }
+      when 'cloudsql' then Cloudsql.new(name, namespace, context, file)
+      when 'configmap' then ConfigMap.new(name, namespace, context, file)
+      when 'deployment' then Deployment.new(name, namespace, context, file)
+      when 'pod' then Pod.new(name, namespace, context, file)
+      when 'ingress' then Ingress.new(name, namespace, context, file)
+      when 'persistentvolumeclaim' then PersistentVolumeClaim.new(name, namespace, context, file)
+      when 'service' then Service.new(name, namespace, context, file)
+      else self.new(name, namespace, context, file).tap { |r| r.type = type }
       end
     end
 
-    def initialize(name, namespace, file)
+    def initialize(name, namespace, context, file)
       # subclasses must also set these
-      @name, @namespace, @file = name, namespace, file
+      @name, @namespace, @context, @file = name, namespace, context, file
     end
 
     def id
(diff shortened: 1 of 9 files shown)


lib/kubernetes-deploy/runner.rb
```diff
@@ -263,14 +263,15 @@ def deploy_resources(resources, prune: false)
...
-    def set_kubectl_context
+    def validate_context
       out, err, st = run_kubectl("config", "get-contexts", "-o", "name", namespaced: false, with_context: false)
       available_contexts = out.split("\n")
       if !st.success?
         raise FatalDeploymentError, err
       elsif !available_contexts.include?(@context)
         raise FatalDeploymentError, "Context #{@context} is not available. Valid contexts: #{available_contexts}"
       end
+      KubernetesDeploy.logger.info("Context #{@context} validated")
```
kirs: We do not really validate it, we only check that it exists. Let's change the wording to "context foobar found"

lib/kubernetes-deploy/runner.rb
```diff
@@ -282,10 +283,12 @@ def validate_namespace
     def run_kubectl(*args, namespaced: true, with_context: true)
       args = args.unshift("kubectl")
       if namespaced
+        raise FatalDeploymentError, "Namespace missing for namespaced command" if @namespace.blank?
         args.push("--namespace=#{@namespace}")
       end
 
       if with_context
+        raise FatalDeploymentError, "Explicit context is required to run this command" if @context.blank?
```
mkobetic: These are errors that people will see in deploy logs right? I suspect "namespace" and especially "context" might be too ambiguous as a term for most. "kubectl config context" maybe? too long?

KnVerey: > These are errors that people will see in deploy logs right? 

Yes, they could, but if they do they need to come to us to fix it. The real user-facing validations are here: https://github.com/Shopify/kubernetes-deploy/blob/master/lib/kubernetes-deploy/runner.rb#L218-L224. This is an extra safety layer because of the danger of deploying to the wrong place, especially during development.

mkobetic: 👍

lib/kubernetes-deploy/kubernetes_resource.rb
```diff
@@ -92,7 +92,9 @@ def group_name
 
     def run_kubectl(*args)
       raise FatalDeploymentError, "Namespace missing for namespaced command" if namespace.blank?
-      args = args.unshift("kubectl").push("--namespace=#{namespace}")
+      raise FatalDeploymentError, "Explicit context is required to run this command" if context.blank?
+      args = args.unshift("kubectl").push("--namespace=#{namespace}").push("--context=#{context}")
```
kirs: `push` mutates the original array which may introduce unexpected and dangerous side-effects

```ruby
args = ["original_command"]
run_kubectl(args) # you don't suppose that `run_kubectl` mutates the original argument
# but now `args` is mutated and contains three extra array items
```

kirs: The best way is to avoid mutation as much as possible, for example by:

```ruby
args = ["kubectl] + args + ["--namespace=#{namespace}", "--context=#{context}"]
```

In this case it creates a new array and doesn't mutate the original argument.

lib/kubernetes-deploy/kubernetes_resource/cloudsql.rb
```diff
@@ -2,8 +2,8 @@ module KubernetesDeploy
   class Cloudsql < KubernetesResource
     TIMEOUT = 5.minutes
 
-    def initialize(name, namespace, file)
-      @name, @namespace, @file = name, namespace, file
+    def initialize(name, namespace, context, file)
+      @name, @namespace, @context, @file = name, namespace, context, file
```
sirupsen: It seems like this could be the default `initialize` behaviour in `KubernetesResource` that this class inherits from.

KnVerey:
> Did this work for Cloud before because it's the default context on the box?

Yes, since nobody yelled, I presume we got lucky and there never was a kubeconfig/deploy target mismatch during the time the bug was rolled out on Shipit, i.e. everyone who tried to deploy during that period hit a box where the last deploy that set the context was to tier4. 

Completed the 🎩  in the description, and everything looks good.

lib/kubernetes-deploy/kubernetes_resource/cloudsql.rb
KnVerey: True. I'm going to save this one for later though (added it to my list).

sirupsen:
🤘

lib/kubernetes-deploy/kubernetes_resource.rb
KnVerey: In general I agree with you that mutation should be avoided. In this case, however, given the purpose of this array, I'm having trouble imagining a scenario where getting the command that actually ran instead of the command passed in would be a bad thing. If, for example, you're trying to log it, it would actually be more accurate.

PR #18, opened 2017-02-06
Status may not have conditions

KnVerey:
@sirupsen Fixes the error below, which someone saw in Shipit this morning. TBH I'm not sure what state leads to pods not having status conditions, but it's easy to avoid 💥 .

```
/usr/lib/ruby-shopify/ruby-shopify-2.3.1/lib/ruby/gems/2.3.0/gems/kubernetes-deploy-0.2.1/lib/kubernetes-deploy/kubernetes_resource/pod.rb:43:in `interpret_json_data': undefined method `find' for nil:NilClass (NoMethodError)
	from /usr/lib/ruby-shopify/ruby-shopify-2.3.1/lib/ruby/gems/2.3.0/gems/kubernetes-deploy-0.2.1/lib/kubernetes-deploy/kubernetes_resource/deployment.rb:27:in `block in sync'
	from /usr/lib/ruby-shopify/ruby-shopify-2.3.1/lib/ruby/gems/2.3.0/gems/kubernetes-deploy-0.2.1/lib/kubernetes-deploy/kubernetes_resource/deployment.rb:23:in `each'
	from /usr/lib/ruby-shopify/ruby-shopify-2.3.1/lib/ruby/gems/2.3.0/gems/kubernetes-deploy-0.2.1/lib/kubernetes-deploy/kubernetes_resource/deployment.rb:23:in `sync'
```

I did a 🎩  with a template set that includes pods and deployments.

### DIFF
--- a/lib/kubernetes-deploy/kubernetes_resource/pod.rb
+++ b/lib/kubernetes-deploy/kubernetes_resource/pod.rb
@@ -40,7 +40,7 @@ module KubernetesDeploy
       elsif @phase == "Terminating"
         @status = @phase
       else
-        ready_condition = pod_data["status"]["conditions"].find { |condition| condition["type"] == "Ready" }
+        ready_condition = pod_data["status"].fetch("conditions", []).find { |condition| condition["type"] == "Ready" }
         @ready = ready_condition.present? && (ready_condition["status"] == "True")
         @status = "#{@phase} (Ready: #{@ready})"
       end


lib/kubernetes-deploy/kubernetes_resource/pod.rb
```diff
@@ -40,7 +40,7 @@ def interpret_json_data(pod_data)
       elsif @phase == "Terminating"
         @status = @phase
       else
-        ready_condition = pod_data["status"]["conditions"].find { |condition| condition["type"] == "Ready" }
+        ready_condition = pod_data["status"].fetch("conditions", {}).find { |condition| condition["type"] == "Ready" }
```
sirupsen: I was a little confused at first for the `#find` but I can see in [the docs](https://kubernetes.io/docs/api-reference/v1.5/#deploymentcondition-v1beta1) that `conditions` is an array. I assume it's an array of all possible conditions, i.e. `type=ready`. What are some of the others? In this case you indeed care about that one.

It's definitely puzzling why `status` wouldn't always have conditions.

However, since `conditions` is an array shouldn't this be `#fetch("conditions", [])`? Otherwise the `#find` might behave strangely.

KnVerey: Yep, you're 100% right. 🤦‍♀️ 

Here's an example from a live pod:
```
  status:
    conditions:
    - lastProbeTime: null
      lastTransitionTime: 2017-01-23T21:35:16Z
      status: "True"
      type: Initialized
    - lastProbeTime: null
      lastTransitionTime: 2017-01-23T21:37:46Z
      status: "True"
      type: Ready
    - lastProbeTime: null
      lastTransitionTime: 2017-01-23T21:35:16Z
      status: "True"
      type: PodScheduled
```

sirupsen: 👍

PR #19, opened 2017-02-08
Add Redis TPR support

wfarr:


wfarr:
This has been 🎩  -- going to 🚢

lib/kubernetes-deploy/kubernetes_resource/redis.rb
```diff
@@ -0,0 +1,80 @@
...
+        if parsed.fetch("spec", {}).fetch("clusterIP", "") != ""
+          return true
+        end
+      end
+
+      false
+    end
+
+    def redis_resource_uuid
+      return @redis_resource_uuid if defined?(@redis_resource_uuid) && @redis_resource_uuid
```
KnVerey: Wouldn't the `@redis_resource_uuid ||= begin` version of memoization be more straightforward, since you do want to re-run the code below when it is defined but nil?

camilo:
A description would have been good to be able to grasp the problem/solution here faster/better.

PR #21, opened 2017-02-15
Tests, continued

KnVerey:
I picked up https://github.com/Shopify/kubernetes-deploy/pull/15 rather than reviewing it, since Kir is on vacation. This PR contains things I would have asked for in review + fixes to get the suite running. I've got another PR lined up with additional tests and helpers.

I'll add comments on the "why" for some details inline, but here's an outline of the broader changes:
- Separated out integration and unit tests, giving kubeclient setup to the former only. (Most of the +/- in the diff is just moving things around to this end.)
- Setup so that each integration test will operate in a different namespace.
- Renamed the fixtures to "basic" instead of "trashbin" (no need to inherit my test repo's stupid name)


test/fixtures/basic/unmanaged-pod.yml.erb
```diff
@@ -0,0 +1,21 @@
+apiVersion: v1
+kind: Pod
+metadata:
+  name: unmanaged-pod-<%= deployment_id %>
+  labels:
+    type: unmanaged-pod
+    name: unmanaged-pod-<%= deployment_id %>
+    app: basic
+spec:
+  activeDeadlineSeconds: 20
```
KnVerey: This is an easy safeguard against hanging. This pod comes up in <5s in my experience, so I think 20s is reasonable.

test/fixtures/basic/unmanaged-pod.yml.erb
```diff
@@ -0,0 +1,21 @@
...
+  restartPolicy: Never
+  containers:
+    - name: basic
+      image: hello-world:latest
+      imagePullPolicy: IfNotPresent
```
KnVerey: We usually use "Always" in production, but in this case we don't really care which version of the image we get, so I'm thinking this should be safe and save test time.

test/fixtures/basic/unmanaged-pod.yml.erb
```diff
@@ -0,0 +1,21 @@
...
+  labels:
+    type: unmanaged-pod
+    name: unmanaged-pod-<%= deployment_id %>
+    app: basic
+spec:
+  activeDeadlineSeconds: 20
+  restartPolicy: Never
+  containers:
+    - name: basic
+      image: hello-world:latest
```
KnVerey: This is a docker tutorial thing: [docker hub](https://hub.docker.com/_/hello-world/) / [repo](https://github.com/docker-library/hello-world). It just prints a message and exits, which is exactly what we need this test pod to do. Wdyt? Other ideas?

test/fixtures/basic/web.yml.erb
```diff
@@ -40,11 +40,11 @@ spec:
     metadata:
       labels:
         name: web
-        app: trashbin
+        app: basic
     spec:
       containers:
       - name: app
-        image: shopify/kubernetes-fixtures-hello-node:v1
+        image: nginx:alpine
```
KnVerey: All this really needs to do is start successfully and not exit. As with the run-once pod, other ideas welcome.

test/helpers/kubeclient_helper.rb
```diff
@@ -0,0 +1,21 @@
+module KubeclientHelper
+  def kubeclient
+    @kubeclient ||= begin
+      config = Kubeclient::Config.read(ENV["KUBECONFIG"])
+      unless config.contexts.include?("minikube")
+        raise "`minikube` context must be configured in your KUBECONFIG (#{ENV["KUBECONFIG"]})"
```
KnVerey: This was just logging before. Was there a reason not to raise? Seems to me the tests calling this method shouldn't try to run without it.

test/helpers/kubeclient_helper.rb
```diff
@@ -0,0 +1,21 @@
...
+      end
+
+      client = Kubeclient::Client.new(
+        config.context.api_endpoint,
+        config.context.api_version,
```
KnVerey: Something that sucks about this `Kubeclient` gem is that it isn't apiVersion-indifferent. Seems like you need to create a separate client to interact with the v1beta1 api endpoints, e.g. to query deployments. More on that in the next PR.

test/integration_test.rb
```diff
@@ -0,0 +1,62 @@
...
+    def teardown
+      @logger_stream.close
+      TestProvisioner.delete_namespace(@namespace)
+    end
+  end
+
+  module TestProvisioner
+    extend KubeclientHelper
+
+    def self.claim_namespace
```
KnVerey: The idea here is to let each test get its own unique namespace so they don't leak k8s state. We'll for sure want to test redeploys, but IMO that should be within one test. Suggestions for better name generation welcome (can we grab part of the test name somehow?). Why not use one namespace and wait for it to be deleted in teardown? Because namespace deletion sometimes takes many seconds.

test/integration_test.rb
```diff
@@ -0,0 +1,62 @@
...
+        kubeclient.get_persistent_volume(name)
+      rescue KubeException => e
+        raise unless e.to_s.include?("not found")
+        pv = Kubeclient::PersistentVolume.new
+        pv.metadata = { name: name }
+        pv.spec = {
+          accessModes: ["ReadWriteOnce"],
+          capacity: { storage: "1Gi" },
+          hostPath: { path: "/data/#{name}" },
+          persistentVolumeReclaimPolicy: "Recycle"
```
KnVerey: Without this, the PV was getting released after the first claim was deleted, and subsequent claims could not use it.

test/integration_test.rb
```diff
@@ -0,0 +1,62 @@
+require 'helpers/kubeclient_helper'
```
sirupsen: I think you should consider calling this `test/integration_test_helper.rb`. The test suffix I feel is often associated with an actual test to run, whereas this is more of a helper. Perhaps we should consider even putting this in the `helpers/` directory?

test/helpers/kubeclient_helper.rb
```diff
@@ -0,0 +1,21 @@
+module KubeclientHelper
+  def kubeclient
+    @kubeclient ||= begin
+      config = Kubeclient::Config.read(ENV["KUBECONFIG"])
+      unless config.contexts.include?("minikube")
+        raise "`minikube` context must be configured in your KUBECONFIG (#{ENV["KUBECONFIG"]})"
+      end
+
+      client = Kubeclient::Client.new(
```
sirupsen: Does `config.context` automatically get the current context? In that case, I think we might have to either switch the current context to `minikube`, or explicitly select the minikube context here depending on what the library supports.

test/helpers/kubeclient_helper.rb
```diff
@@ -0,0 +1,21 @@
...
+        config.context.api_endpoint,
+        config.context.api_version,
+        {
+          ssl_options: config.context.ssl_options,
+          auth_options: config.context.auth_options
+        }
+      )
+      client.discover
```
sirupsen: What does this do?

test/helpers/kubeclient_helper.rb
sirupsen: I agree this should raise!

sirupsen: Could we also add some instructions to the README on how to get this running? The error here could point towards that.

test/integration_test.rb
```diff
@@ -0,0 +1,62 @@
...
+
+module KubernetesDeploy
+  class IntegrationTest < ::Minitest::Test
+    include KubeclientHelper
+
+    def setup
+      @logger_stream = StringIO.new
+      @logger = Logger.new(@logger_stream)
+      KubernetesDeploy.logger = @logger
```
sirupsen: I still got a lot of logs to `stdout/stderr` when running the tests locally:

```
[KUBESTATUS] {"group":"Web deployment","name":"web-3766115818-x7ns1","status_string":"Pending (Ready: false)","exists":true,"succeeded":false,"failed":false,"timed_out":false}
```

And so on with the test output

test/integration_test.rb
```diff
@@ -0,0 +1,62 @@
...
+      @namespace = TestProvisioner.claim_namespace
+    end
+
+    def teardown
+      @logger_stream.close
+      TestProvisioner.delete_namespace(@namespace)
```
sirupsen: I think we should consider putting this in an `ensure` as to not pollute the cluster if forcefully killing the tests at some point.

test/integration_test.rb
sirupsen: This is a great idea Katrina, huge 👍  to this. I think you should consider prefixing the names with `k8s-deploy` so that it's clear where they're coming from if someone is inspecting the cluster while debugging a test, for example.

sirupsen: The current test name is a great idea. It looks like [`Minitest::Test#run`](https://github.com/seattlerb/minitest/blob/master/lib/minitest/test.rb#L107) will have the name of the current test set as `self.name`. The whole Minitest stacktrace can be seen in [this comment](https://github.com/seattlerb/minitest/blob/master/lib/minitest.rb#L110). Perhaps `self.name` will do the right thing here? You might have to get a bit more creative with a debugger, but if you dump the methods available with you in `pry` or `byebug` you might be able to figure it out.

test/integration_test.rb
```diff
@@ -0,0 +1,62 @@
...
+          capacity: { storage: "1Gi" },
+          hostPath: { path: "/data/#{name}" },
+          persistentVolumeReclaimPolicy: "Recycle"
+        }
+        kubeclient.create_persistent_volume(pv)
+      end
+    end
+  end
+
+  TestProvisioner.prepare_pv("pv0001")
```
sirupsen: Should this name be exposed somewhere for the tests to not hardcode it?

test/integration/kubernetes_deploy_test.rb
```diff
@@ -0,0 +1,27 @@
+require 'test_helper'
+
+class KubernetesDeployTest < KubernetesDeploy::IntegrationTest
+  def test_basic
+    runner = KubernetesDeploy::Runner.new(
+      namespace: @namespace,
+      current_sha: "abcabcabc",
+      context: "minikube",
```
sirupsen: Should we have a helper for the runner as we do with the client to pass the context and namespace automatically? I don't love that minikube is hardcoded in here, the namespace I can live with for sure, as it makes sense for the successive calls like `#get_pods`.

test/integration/kubernetes_deploy_test.rb
```diff
@@ -0,0 +1,27 @@
...
+      template_dir: File.expand_path("./test/fixtures/basic"),
+      wait_for_completion: true,
+    )
+    runner.run
+
+    pods = kubeclient.get_pods(namespace: @namespace)
+    managed_pods, unmanaged_pods = pods.partition do |pod|
```
sirupsen: It's been a while since I've seen `#partition` in the wild, nice 🎉

test/integration/kubernetes_deploy_test.rb
```diff
@@ -0,0 +1,27 @@
...
+    end
+
+    assert_equal 2, managed_pods.size
+    managed_pods.each do |pod|
+      assert_equal "Running", pod.status.phase
+    end
+
+    assert_equal 1, unmanaged_pods.size
+    assert_equal "Succeeded", unmanaged_pods.first.status.phase
+  end
```
sirupsen: 🎉

test/integration_test.rb
sirupsen: Instead of this `get` and `create` cycle in a rescue, can we create it and then get it after? I don't love having this much logic in an exception handler. I feel they should be more reserved for logging, resiliency, and that sort of thing.

test/unit/kubernetes-deploy/runner_test.rb
```diff
@@ -0,0 +1,43 @@
...
+
+  def test_that_it_has_a_version_number
+    refute_nil ::KubernetesDeploy::VERSION
+  end
+
+  def test_initializer
+    # TODO: Switch to --kubeconfig for kubectl shell out and pass env var as arg to Runner init
+    # Then fix this crappy env manipulation
+    original_env = ENV["KUBECONFIG"]
+    ENV["KUBECONFIG"] = "/this-really-should/not-exist"
```
sirupsen: I'm not exactly sure I understand your comment but it sounds like you're changing this anyway. A helper for this would be good  `with_kube_config` or `with_env(kubeconfig: path)` , but it sounds like you're planning to do something like this anyway.

test/unit/kubernetes-deploy/runner_test.rb
```diff
@@ -0,0 +1,43 @@
...
+
+class RunnerTest < Minitest::Test
+  def setup
+    @logger_stream = StringIO.new
+    @logger = Logger.new(@logger_stream)
+    KubernetesDeploy.logger = @logger
+  end
+
+  def teardown
+    @logger_stream.close
```
sirupsen: If we're going to do this in the unit tests, perhaps this should be in a default `#setup` that integration tests also inherit from?

test/unit/kubernetes-deploy/runner_test.rb
```diff
@@ -0,0 +1,43 @@
...
+    )
+    error_msg = assert_raises(KubernetesDeploy::FatalDeploymentError) do
+      runner.run
+    end.to_s
+
+    assert_includes error_msg, "Kube config not found at /this-really-should/not-exist"
+    assert_includes error_msg, "Current SHA must be specified"
+    assert_includes error_msg, "Namespace must be specified"
+    assert_includes error_msg, "Context must be specified"
+    assert_match /Template directory (\S+) doesn't exist/, error_msg
```
sirupsen: 🥇

test/fixtures/basic/unmanaged-pod.yml.erb
KnVerey: This is how we do migrations for all our apps using Cloudsql: launch a pod that runs `rake db:migrate` and exits, wait for it to finish, and dump its logs inline.

test/fixtures/basic/unmanaged-pod.yml.erb
```diff
@@ -0,0 +1,21 @@
...
+      imagePullPolicy: IfNotPresent
+      env:
+      - name: CONFIG
+        valueFrom:
+          configMapKeyRef:
+            name: configmap-data
+            key: datapoint2
```
KnVerey: Yeah, it's so this can be used to test the case where you roll out a pod depending on a configmap that you forgot to include.

test/fixtures/basic/web.yml.erb
```diff
@@ -4,10 +4,10 @@ metadata:
   name: web
   labels:
     name: web
-    app: trashbin
+    app: basic
```
KnVerey: This is usually just the name of the app itself by our conventions (e.g. the repo name).

test/helpers/kubeclient_helper.rb
KnVerey: Yeah, the default is the current context. [This method](https://github.com/abonas/kubeclient/blob/f489a35027a78ea6f410c76b5449366eadf94e85/lib/kubeclient/config.rb#L35) accepts the context name as an arg, and we should definitely pass it. Good catch.

test/helpers/kubeclient_helper.rb
KnVerey: It seems to get the list of resources that the API server handles and create methods for them: [code](https://github.com/abonas/kubeclient/blob/82f4059d117e31336102d41ccd99b905815b87f1/lib/kubeclient/common.rb#L94-L114). If I'm not mistaken, this makes the gem effectively able to pick up new resource types without code changes and to adapt to different server versions, but it also leads to the need to have a different client to talk to the `v1beta1` api as mentioned above.

It can also be done lazily on the first invocation of a discovery-requiring method. I copy-pasted this from Kir's work tbh, so not sure if he encountered issues with the lazy method.

test/integration_test.rb
KnVerey: Yep, that's because there are two loggers and the one that does kube status is not at all pluggable right now. We should definitely address it, but I'd rather leave it for later since it involves modifying the non-test code (also, while crafting tests, that output is useful anyhow :P).

test/integration_test.rb
KnVerey: Isn't the whole `teardown` run in an ensure? If it isn't, where do we put the ensure? You're right, I've definitely seen namespace hang around when I ctrl+c or forcefully exit pry during test runs.

test/fixtures/basic/unmanaged-pod.yml.erb
kirs: 👍

test/fixtures/basic/unmanaged-pod.yml.erb
kirs: Great to have `hello-world:latest`, meaning that we won't need to support our stub images.

test/helpers/kubeclient_helper.rb
kirs: Absolutely, this should have been `raise` rather than `puts`. No idea why I used `puts` 🙈

test/helpers/kubeclient_helper.rb
kirs: `discover` defines classes from the API like `Kubeclient::Service` that we need to create namespaces and stuff.

kirs: AFAIK, it also verifies that the kubernetes instance is running and auth succeeds.

test/integration_test.rb
sirupsen: I think you're right, they're run inside [`#capture_exceptions`](https://github.com/seattlerb/minitest/blob/master/lib/minitest/test.rb#L110) so they should complete.

test/helpers/kubeclient_helper.rb
sirupsen: 😭  is this relevant to the `#discover` comment?

test/integration/kubernetes_deploy_test.rb
kirs: Installing `minikube` automatically creates the `minikube` context, so this value won't change in future. Meaning that we'll always be using `minikube` context in test environment.
I'm totally 👍 for moving it into a const.

test/unit/kubernetes-deploy/runner_test.rb
kirs: We should absolutely use `with_env` here.

test/helpers/kubeclient_helper.rb
KnVerey: I'm fine with doing it eagerly here. It seems likely everything we'd be calling would require discovery anyway. Also, it's probably less confusing to have the backtrace (if auth fails for example) point here instead of to some random `get_{thing}` call.

test/integration/kubernetes_deploy_test.rb
KnVerey: I'll move it to a constant. I'd like to leave extra helper creation for the follow-up PR though. That's where I've been playing around with approaches for making test setup easy and readable.

test/integration/kubernetes_deploy_test.rb
KnVerey: This one's Kir's 😄 . I'm a fan as well though; if you wanna see more, check out `runner.rb`. 😉

test/integration_test.rb
KnVerey: Hmm, I get your point that it isn't a test to run, but it really doesn't fit with the standard definition of a helper either (i.e. a module full of related helper methods). And `SomeClassTest < IntegrationTestHelper` feels pretty wrong to me. I tried looking at what rails does, which is how I landed on this name (it's not actually under `test/` though):
- rails/activesupport/lib/active_support/test_case.rb
- rails/activerecord/test/cases/integration_test.rb

I'm not really sure what's best. I do kinda like `KubernetesDeploy::IntegrationTest < KubernetesDeploy::TestCase < ::Minitest::Test`, which I think implies `test/integration_test.rb` and `test/test_case.rb`?

test/integration_test.rb
KnVerey: It's very easy to reproduce:
```
༶ ktl get ns -w                               
NAME               STATUS    AGE
default            Active    135d
kube-system        Active    135d
NAME               STATUS    AGE
f0ea131043f23bfc   Active    6m
d934628332442544   Active    6m
```

That's two test runs with "Ctrl+C" in the middle.

https://github.com/seattlerb/minitest/issues/677
I'll give `Minitest.after_run { ... }` a try

test/integration_test.rb
KnVerey: Tests won't reference this by name. It's a cluster-admin-level resource, which is why we aren't even cleaning it up within or after runs. Here's a useful snippet on how these work:

> A user creates, ..., a PersistentVolumeClaim with a specific amount of storage requested and with certain access modes. A control loop in the master watches for new PVCs, finds a matching PV (if possible), and binds them together. ..., the user will always get at least what they asked for, but the volume may be in excess of what was requested. Once bound, PersistentVolumeClaim binds are exclusive, regardless of the mode used to bind them.

So the tests / yamls need to match the `accessMode` of an existing PV and request <= the amount of storage it has, but they don't need to know its name.

test/unit/kubernetes-deploy/runner_test.rb
KnVerey: To explain my comment, I'm planning to add a `--kubeconfig` option to the CLI, which will default to `ENV["KUBECONFIG"]`, such that the runner will always receive a path as an argument. It'll then use that path in `kubectl [things] --kubeconfig=/path/to/config` instead of relying on the env var.  Then this test can be run without manipulating the environment. I'm in favour of wrappers like that in general, but in this case it feels like putting 💄  on something shitty that's used in a single test. If a second test needs to manipulate this even temporarily, I'll add a helper.

bin/setup
```diff
@@ -1,8 +1,16 @@
...
 
-# Do any other automated setup that you need to do here
+if [ ! -x "$(which minikube)" ]; then
+  echo -e "\n\033[0;33mIf you're going to run the tests, please follow the minikube setup instructions for your operating system:\nhttps://kubernetes.io/docs/getting-started-guides/minikube/#installation\033[0m"
+fi
+
+if [ ! -x "$(which kubectl)" ]; then
+  echo -e "\n\033[0;33mPlease install kubectl version 1.5.1 or higher:\nhttps://kubernetes.io/docs/user-guide/prereqs/\033[0m"
+else
+  KUBECTL_VERSION=$(kubectl version --short --client | grep -oe "v[[:digit:]\.]\+")
```
kirs: 👍 for the snippet.
I've been planning to include client version check into the gem binary to ensure that whoever is running it (Shipit/local 🎩 ) has the right version. We can do that later.

test/test_helper.rb
```diff
@@ -2,63 +2,78 @@
...
+      TestProvisioner.delete_namespace(@namespace)
+    end
+  end
 
-module TestProvisioner
-  extend KubeclientHelpers
+  module TestProvisioner
+    extend KubeclientHelper
+
+    def self.claim_namespace(test_name:)
```
kirs: I don't think we need a keyword argument in case of a single argument.

test/unit/kubernetes-deploy/runner_test.rb
kirs: The problem now is that any of the lines below raise an error, it won't ever reach `ENV["KUBECONFIG"] = original_env`. Though I'm in favor of shipping it now as it is and following with the `--kubeconfig` PR.

KnVerey: True, added an `ensure`

README.md
```diff
@@ -22,11 +20,27 @@ Or install it yourself as:
...
 
-After checking out the repo, run `bin/setup` to install dependencies. Then, run `rake test` to run the tests. You can also run `bin/console` for an interactive prompt that will allow you to experiment.
+After checking out the repo, run `bin/setup` to install dependencies. You currently need to [manually install kubectl version 1.5.1 or higher](https://kubernetes.io/docs/user-guide/prereqs/) as well if you don't already have it.
+
+To run the tests:
+
+* [Install minikube](https://kubernetes.io/docs/getting-started-guides/minikube/#installation)
+* Start minikube (`minikube start [options]`)
+* Make sure you have a context named "minikube" in your kubeconfig. Minikube adds this context for you when you run `minikube start`; please do not rename it. You can check for it using `kubectl config get-contexts`.
+* Run `bundle exec rake test`
```
sirupsen: Is the vision here is to make this available through `dev(1)` or something?

test/test_helper.rb
```diff
@@ -2,63 +2,78 @@
...
-Minitest::Test.include(KubeclientHelpers)
+  class IntegrationTest < KubernetesDeploy::TestCase
+    include KubeclientHelper
+
+    def run
+      @namespace = TestProvisioner.claim_namespace(self.name)
+      super
+    ensure
+      TestProvisioner.delete_namespace(@namespace)
+    end
```
sirupsen: Wow, nice hacking.. 👍  I bet this is going to save someone a lot of time at some point.

README.md
kirs: Yes, but this is also an open source project and we should consider it may be used by someone who doesn't have `dev`.

PR #22, opened 2017-02-18
Enable CI

kirs:
This is the scrappy version of the CI running for master and branches. Surprisingly, it took me less than an hour to set it up (big thanks to @sander-m for help with BK and VMs) 🎉 

Important things:

* Only Shopify employees can access the CI page, however the CI status is reported back to GH and visible for everyone
* I didn't enable it for the PRs on purpose because it's a source of vulnerabilities. This is a public project and anyone could submit a PR with malicious code in `bin/ci` that would be executed on Shopify VMs. Technically the VMs are isolated but still I wouldn't like to give this opportunity to people outside of the company.

**How it works:**
Mobile CI at Shopify provides disposable Ubuntu VMs with enabled KVM. We have them integrated with BuildKite, and these computing resources are available to run CI for this project. In `bin/ci` script we install kubectl, minikube, docker kvm driver, and ruby dependencies. Them we simply run `rake test`.

It takes 2m40s to start minikube, and the complete test suite runs in 20 seconds. This sounds good enough for me, but in future we can speed it up by pre-building the VM template with minikube and all dependencies installed.

Preview:

screen shot 2017-02-17 at 20 16 56

review @sirupsen @KnVerey @ibawt @wfarr

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,5 +1,7 @@
 # Kubernetes::Deploy
 
+[![Build status](https://badge.buildkite.com/0f2d4956d49fbc795f9c17b0a741a6aa9ea532738e5f872ac8.svg)](https://buildkite.com/shopify/kubernetes-deploy-gem)
+
 Deploy script used to manage a Kubernetes application's namespace with [Shipit](https://github.com/Shopify/shipit-engine).
 
 ## Installation
@@ -44,6 +46,14 @@ To run the tests:
 
 To install this gem onto your local machine, run `bundle exec rake install`. To release a new version, update the version number in `version.rb`, and then run `bundle exec rake release`, which will create a git tag for the version, push git commits and tags, and push the `.gem` file to [rubygems.org](https://rubygems.org).
 
+## CI
+
+Buildkite will build branches as you're working on them, but as soon as you create a PR it will stop building new commits from that branch because we disabled PR builds for security reasons.
+As a Shopify employee, you can manually trigger the PR build from the Buildkite UI (just specify the branch, SHA is not required):
+
+<img width="464" alt="screen shot 2017-02-21 at 10 55 33" src="https://cloud.githubusercontent.com/assets/522155/23172610/52771a3a-f824-11e6-8c8e-3d59c45e7ff8.png">
+
+
 ## Contributing
 
 Bug reports and pull requests are welcome on GitHub at https://github.com/[USERNAME]/kubernetes-deploy.
@@ -52,4 +62,3 @@ Bug reports and pull requests are welcome on GitHub at https://github.com/[USERN
 ## License
 
 The gem is available as open source under the terms of the [MIT License](http://opensource.org/licenses/MIT).
-
--- a/bin/ci
+++ b/bin/ci
@@ -1,0 +1,32 @@
+#!/bin/bash
+set -eox pipefail
+
+MINIKUBE_VER=0.16.0
+KUBERNETES_VER=1.5.2
+DOCKER_MACHINE_KVM_VER=0.7.0
+
+echo "--- Installing dependencies"
+curl -Lo minikube https://storage.googleapis.com/minikube/releases/v$MINIKUBE_VER/minikube-linux-amd64
+chmod +x minikube
+sudo mv minikube /usr/local/bin/
+
+# minikube requires kubectl
+curl -Lo kubectl https://storage.googleapis.com/kubernetes-release/release/v$KUBERNETES_VER/bin/linux/amd64/kubectl
+chmod +x kubectl
+sudo mv kubectl /usr/local/bin/
+
+# KVM support for minikube
+curl -Lo docker-machine-driver-kvm https://github.com/dhiltgen/docker-machine-kvm/releases/download/v$DOCKER_MACHINE_KVM_VER/docker-machine-driver-kvm
+chmod +x docker-machine-driver-kvm
+sudo mv docker-machine-driver-kvm /usr/local/bin/
+
+minikube config set vm-driver kvm
+
+gem install bundler --no-ri --no-rdoc
+bundle install --jobs 4
+
+echo "--- Starting minikube"
+minikube start --cpus 2 --memory 2048 --disk-size=2gb --kubernetes-version=$KUBERNETES_VER
+
+echo "--- Running tests"
+bundle exec rake test


ibawt:
:O

bin/ci
```diff
@@ -0,0 +1,29 @@
...
+chmod +x docker-machine-driver-kvm
+sudo mv docker-machine-driver-kvm /usr/local/bin/
+
+minikube config set vm-driver kvm
+
+gem install bundler --no-ri --no-rdoc
+bundle install --jobs 4
+
+echo "--- Starting minikube"
+minikube start
```
KnVerey: Couple of things I'd like to confirm about the defaults we're using here:
- `--cpus` is 2 and `--memory` is 2048. Seems reasonable to me but ¯\\\_(ツ)_/¯
- `--disk-size` is 20g. Seems a bit excessive? `MinimumDiskSizeMB   = 2000` (from minikube code), and I've been using that much locally.
- `--kubernetes-version` is available. Perhaps we should link this to the version we download above.

bin/ci
```diff
@@ -0,0 +1,29 @@
...
+minikube status
```
KnVerey: Can we wait for this to be "Running" before trying to start the tests? Looks like it was immediately in your test runs, but perhaps that wouldn't always be the case.

bin/ci
```diff
@@ -0,0 +1,29 @@
...
+
+# minikube requires kubectl
+curl -Lo kubectl https://storage.googleapis.com/kubernetes-release/release/v1.5.2/bin/linux/amd64/kubectl
+chmod +x kubectl
+sudo mv kubectl /usr/local/bin/
+
+# KVM support for minikube
+curl -Lo docker-machine-driver-kvm https://github.com/dhiltgen/docker-machine-kvm/releases/download/v0.7.0/docker-machine-driver-kvm
+chmod +x docker-machine-driver-kvm
+sudo mv docker-machine-driver-kvm /usr/local/bin/
```
sirupsen: Can you extract all these versions to environment variables? I'd be a little concerned that these things get out of date. How about writing a test that asserts the versions used in this file are the same as we use in development?

Just a sketch in code (haven't run any of this so don't know if it works, but it illustrates an idea):

```ruby
test "check versions with CI" do
  versions = File.read("omg.sh").scan(/([\w_]+)_VERSION="([\d\.]+)"/)
  versions.each do |(util, version)|
    case util
    when "KUBECTL"
      dev_version =`kubectl version | grep -oP "(?<=v)[0-9]+\.[0-9]+\.[0-9]+"`.strip
      assert_equal dev_version, version
    when "DOCKER_MACHINE"
      # compare somehow
    when "MINIKUBE"
       # compare somehow
    else
       flunk "#{util} was not defined to compare with development..."
    end
  end
end
```

Alternatively we could enforce this somehow else, e.g. booting `test/test_helper.rb` to make sure both dev and CI runs against the same versions.

sirupsen:
I am assuming Shopify employees can trigger a build manually? Can you put that in the README as well to be required? If we ship the gem with ShipIt already, this shouldn't be an issue. I'm really concerned about this test suite getting rusty... complicated VM test suites that rely on multiple versions of things to work together are very prone to rust.

bin/ci
kirs: `minikube start` waits until the cluster is running and then [connects to it](https://github.com/kubernetes/minikube/blob/master/cmd/minikube/cmd/start.go#L154) to verify things. So the check shouldn't be required.

If you think the check is necessary, we can put `test -n "$(minikube status | grep 'Running')"` into the bash script. It will fail if it couldn't grep the status for "Running".

bin/ci
kirs: Do you mean you'd like to not rely on defaults and run it with `minikube start --cpus 2 --memory 2048 --disk-size=2gb --kubernetes-version=x.x.x`?

bin/ci
sirupsen: Why have `minikube status` at all then? In case it fails? What's the timeout on `minikube start` in case it fails?

bin/ci
sirupsen: The idea of specifying this explicitly with values that the Dev Acceleration team is comfortable with (cc @sander-m) sounds good to me. 👍  Any opinions on values Sander?

kirs:
> I am assuming Shopify employees can trigger a build manually? Can you put that in the README as well to be required? If we ship the gem with ShipIt already, this shouldn't be an issue. I'm really concerned about this test suite getting rusty... complicated VM test suites that rely on multiple versions of things to work together are very prone to rust.

Shopify employees push code to branches, and the branch builds are automatically triggered. AFAIK it also means that PRs from Shopify employees are also built automatically. It will also build git tags (== gem releases).

bin/ci
kirs: There's no timeout in `minikube start` that I could find. It may be worth adding it - I'll discuss it with the mikube maintainer.

I agree that taking the facts that I mentioned above, `minikube status` isn't necessary. I'll remove it.

kirs: It's also worth mentioning that if in the worst case `minikube start` gets crazy and starts to take forever, we have the global Buildkite timeout.

kirs:
> Can we include the pipeline.yml for this in the repo?

I had exactly the same suggestion, but then we discussed it with @sander-m and came to the conclusion that since CI is a private thing, in this case we can keep the config in BK and don't expose it to the public.

> I'm ok with doing it a bit later, but I think prebuilding the VM template will definitely be worthwhile. The suite takes 20 seconds now, but that's with a single integration test. On my branch that adds ~10 more, we're already up to 70 seconds.

Sure, working on the VM template is one of the next things in my list.

bin/ci
sander-m: - `--cpu 2` 👍  because these VM instances have 2 cores.
- `--memory 2048` The VM has 8GB RAM available, so this is fine but you could bump it if you want
- `--disk-size=2gb` You can expect roughly 20GB of free space on the VM, but we can increase this for the custom template if you need more.

sirupsen:
> > Shopify employees push code to branches, and the branch builds are automatically triggered. AFAIK it also means that PRs from Shopify employees are also built automatically. It will also build git tags (== gem releases).
> This was the only thing I was waiting to hear back on. This all looks good to me!

That's... magic. ❇️

sander-m:
LGTM. I can help out with the custom template.

bin/ci
kirs: I'm concerned about `minikube` and `kubectl` versions too. The problem is that in future they will be installed in the VM template that doesn't have the context of the repo.

I'm going to move versions to environmental variables and in the next PR I'll introduce the version check in test helper.

kirs:
I've addressed all concerns that were mentioned. Please let me know if there's any other feedback.

I have another updated about Buildkite:

BK will build branches as you're working on them, but as soon as you create a PR it will stop building new commits from that branch because it's a PR and we disabled PR builds. The workaround for that is to trigger the build manually, which is very easy:

screen shot 2017-02-21 at 10 55 33

ibawt:
fyi @stefanmb

README.md
```diff
@@ -1,5 +1,7 @@
 # Kubernetes::Deploy
 
+[![Build status](https://badge.buildkite.com/0f2d4956d49fbc795f9c17b0a741a6aa9ea532738e5f872ac8.svg)](https://buildkite.com/shopify/rubernetes-deploy-gem)
```
KnVerey: `kubernetes-deploy-gem` (I see you fixed it in buildkite)

kirs:
🎉🎉🎉

@KnVerey great suggestions. I've fixed both things.

PR #23, opened 2017-02-18
Capture all logging

KnVerey:
This makes it possible to redirect all logging during tests by making kubestatus logging (and an errant "Deploy succeeded!" puts) use a settable logger.

### DIFF
--- a/lib/kubernetes-deploy/kubernetes_resource.rb
+++ b/lib/kubernetes-deploy/kubernetes_resource.rb
@@ -4,6 +4,19 @@ require 'shellwords'
 
 module KubernetesDeploy
   class KubernetesResource
+    def self.logger=(value)
+      @logger = value
+    end
+
+    def self.logger
+      @logger ||= begin
+        l = Logger.new($stderr)
+        l.formatter = proc do |_severity, _datetime, _progname, msg|
+          "#{msg}\n"
+        end
+        l
+      end
+    end
 
     attr_reader :name, :namespace, :file, :context
     attr_writer :type, :deploy_started
@@ -112,7 +125,7 @@ module KubernetesDeploy
     end
 
     def log_status
-      STDOUT.puts "[KUBESTATUS] #{JSON.dump(status_data)}"
+      KubernetesResource.logger.info("[KUBESTATUS] #{JSON.dump(status_data)}")
     end
   end
 end
--- a/lib/kubernetes-deploy/kubernetes_resource/pod.rb
+++ b/lib/kubernetes-deploy/kubernetes_resource/pod.rb
@@ -76,7 +76,7 @@ module KubernetesDeploy
         next unless st.success? && out.present?
 
         KubernetesDeploy.logger.info "Logs from #{id} container #{container_name}:"
-        STDOUT.puts "#{out}"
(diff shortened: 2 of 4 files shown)


lib/kubernetes-deploy/kubernetes_resource.rb
```diff
@@ -4,6 +4,19 @@
...
   class KubernetesResource
+    def self.logger=(value)
+      @logger = value
+    end
+
+    def self.logger
+      @logger ||= begin
+        l = Logger.new($stderr)
+        l.formatter = proc do |_severity, _datetime, _progname, msg|
+          "#{msg}\n"
```
sirupsen: Why not print date-time and severity?

KnVerey: Because this logger is for dumping output raw: 
- `[KUBESTATUS]` lines for eventual consumption by a visualization
- Pod logs from k8s (e.g. from migrations)

(white lines in pic below)

image

Perhaps it should be named `raw_logger`?

sirupsen: Why do we not want timestamps on that?

KnVerey: I'm not convinced they'd add value rather than noise. The kube status logs happen at a defined interval, and the others are relaying logs from somewhere else, which already have their own timestamps ... imagine having another timestamp on the 2nd-to-bottom line in the screenshot. Wouldn't that just be confusing? I'd be more in favour of adding timestamps to the regular logger (coloured lines in the screenshot).

sirupsen: Yeah, double timestamps is not desirable, but why are there already double timestamps on that `ERROR`? I'd still argue we should have timestamp on `KUBESTATUS` because it's an implementation detail that it only happens at a defined interval (e.g. I didn't even know that).

KnVerey: > why are there already double timestamps on that ERROR?

That's just someone's raw logs from `rake db:migrate`... I probably copied it from a deploy of a test app whose logging is messed up. Nothing to do with the deploy script.

So you'd want three loggers then? One that's genuinely raw for log dumping, one for kube status, and one for human-friendly messages? I'm still not convinced that it's useful for kube status, which is supposed to be machine-readible json + a tag to identify the lines. I guess you're asking me to justify "why not?" and the answer comes down to that I don't see a compelling "why?" here.

sirupsen: I just think timestamps on everything is a good idea. If something is stuck, it's the type of thing that'll let me know. I don't actually think we're going to read the output by machines with the ShipIt strategy in the near future. Or do we already? I don't really understand enough about this code to comment on whether that means 3 loggers.. that seems like a lot. If you don't agree, then feel free to ignore this. I'm not going to keep advocating for this in this thread :)

KnVerey:
We don't have any disagreement on what this PR is actually changing, so I'm going to merge it. Let's continue the discussion re: whether one or both loggers should include timestamps and/or other info elsewhere.

PR #24, opened 2017-02-18
Integration tests and helpers

KnVerey:
This PR adds coverage for most of what I was looking for in the manual tophat flow with the "trashbin" repo, plus a couple bonus scenarios. Not covered:
- Error logging related to deploys that eventually time out. The current behaviour would be hard to integration test (could be covered by unit tests on the polling) and switching these to be fast failures is near the top of my todo list.
- The actual CLI invocation itself (tests trigger deploys via the Runner)

The most interesting part here was trying to devise a helper structure would make these tests easy to read, simple to write and clean to customize. My idea of what this would look like definitely changed as I wrote more tests! What I landed on here is perhaps a bit unusual, but I quite like the API it gives the tests themselves, and the encapsulation of concerns in the helpers. **I recommend checking out the integration test file first, since it is the whole point of the rest of the code.**

### Details

Here's an overview of how the helpers are set up:
- `FixtureDeployHelper` contains methods for loading and manipulating the yaml fixtures. It provides two options:
  - `deploy_fixture_set(set, subset=nil, wait: true)` to immediately deploy a full or partial set of fixtures
  - `load_fixture_data(set, subset=nil)` + `deploy_loaded_fixture_set(template_map, wait: true)` allows you to make changes/additions to the set before deploying it. The alternative to providing this option is to duplicate the fixtures for each scenario that needs a change. I find this much better because it not only reduces duplication, but also makes it super clear within the test what is special (aka broken in most cases) about the set being deployed.
- `FixtureSetAssertions::Basic` is used to encapsulate knowledge about the contents of the fixture set named "basic" (I'm kinda regretting that name... too adjectival). You instantiate it with the namespace you deployed to, then use it to assert things like "the whole set is up", "all the redis-related components are up", or "no web-related components exist".
- `FixtureSetAssertions::FixtureSet` is the superclass of `FixtureSetAssertions::Basic` and provides generic resource assertions that can be leveraged by any fixture set class (e.g. `assert_service_up(svc_name)`). I made these methods private, but it might make sense for tests to be able to use them directly as well, i.e. when they've added something to the set deployed.

### DIFF
--- a/test/fixtures/invalid/yaml-error.yml
+++ b/test/fixtures/invalid/yaml-error.yml
@@ -6,5 +6,5 @@ metadata:
     name: basic-configmap-data
     app: basic
 data:
-  datapoint1: value1
+  datapoint1: value1:
   datapoint2: value2
--- a/test/helpers/fixture_deploy_helper.rb
+++ b/test/helpers/fixture_deploy_helper.rb
@@ -1,5 +1,8 @@
 module FixtureDeployHelper
-  # use deploy_fixture_set if you are not adding to or otherwise modifying the template set
+  # Deploys the templates in the specified fixture set via KubernetesDeploy::Runner.
+  # Optionally takes an array of names of template files in that set, in which case only those files are deployed.
+  # If you need to add to or modify the fixture set before deploying,
+  # use load_fixture_data(set, subset=nil) and deploy_loaded_fixture_set(template_map, wait: true) instead.
   def deploy_fixture_set(set, subset: nil, wait: true)
     source_dir = fixture_path(set)
     return deploy_dir(source_dir) unless subset
@@ -18,7 +21,9 @@ module FixtureDeployHelper
     FileUtils.remove_dir(target_dir) if target_dir
   end
 
-  # use load_fixture_data + deploy_loaded_fixture_set to have the chance to add to / modify the template set before deploy
+  # Takes an array of templates in the format returned by load_fixture_data, saves them to a temporary directory,
+  # and invokes KubernetesDeploy::Runner on that directory with the requested options.
+  # Use this with load_fixture_data when you want to add to and/or modify the template set before deploying it.
   def deploy_loaded_fixture_set(template_map, wait: true)
     dir = Dir.mktmpdir
     files = []
@@ -35,15 +40,19 @@ module FixtureDeployHelper
     files.each { |f| File.delete(f) }
   end
 
-  # load_fixture_data return format
-  # {
-  #   file_basename: {
-  #     type_name: [
-  #       loaded_yaml_of_type,
-  #       loaded_yaml_of_type,
-  #     ]
-  #   }
-  # }
+  # Returns a hash containing a key for each template file in the set.
+  # The values of those keys are hashes containing a key for each resource type in the template file.
+  # In turn, the values of those keys are arrays of loaded kubernetes resource yaml.
+  #
+  # Example:
+  #   load_fixture_data("basic", ["web"])
+  #   => {
+  #        "web" => {
+  #          "Ingress" => [loaded_ingress_yaml],
+  #          "Service" => [loaded_service_yaml],
+  #          "Deployment" => [loaded_service_yaml]
+  #        }
+  #      }
   def load_fixture_data(set, subset=nil)
     source_dir = fixture_path(set)
     templates = {}
@@ -58,6 +67,8 @@ module FixtureDeployHelper
     templates
   end
 
+  private
+
   def deploy_dir(dir, sha: 'abcabcabc', wait: true)
     runner = KubernetesDeploy::Runner.new(
       namespace: @namespace,
@@ -68,8 +79,6 @@ module FixtureDeployHelper
     )
     runner.run
   end
-
-  private
 
   def fixture_path(set_name)
     source_dir = File.expand_path("../../fixtures/#{set_name}", __FILE__)
--- a/test/helpers/fixture_deploy_helper.rb
+++ b/test/helpers/fixture_deploy_helper.rb
@@ -1,70 +1,49 @@
 module FixtureDeployHelper
-  # Deploys the templates in the specified fixture set via KubernetesDeploy::Runner.
-  # Optionally takes an array of names of template files in that set, in which case only those files are deployed.
-  # If you need to add to or modify the fixture set before deploying,
-  # use load_fixture_data(set, subset=nil) and deploy_loaded_fixture_set(template_map, wait: true) instead.
-  def deploy_fixture_set(set, subset: nil, wait: true)
-    source_dir = fixture_path(set)
-    return deploy_dir(source_dir) unless subset
-
-    target_dir = Dir.mktmpdir
-    each_k8s_yaml(source_dir, subset) do |basename, ext, content|
-      Tempfile.open([basename, ext], target_dir) do |f|
-        f.write(content)
-      end
-    end
-
-    deploy_dir(target_dir, wait: wait)
-  ensure
-    FileUtils.remove_dir(target_dir) if target_dir
-  end
-
-  # Takes an array of templates in the format returned by load_fixture_data, saves them to a temporary directory,
-  # and invokes KubernetesDeploy::Runner on that directory with the requested options.
-  # Use this with load_fixture_data when you want to add to and/or modify the template set before deploying it.
-  def deploy_loaded_fixture_set(template_map, wait: true)
-    target_dir = Dir.mktmpdir
-    template_map.each do |file_basename, file_data|
-      data = YAML.dump_stream(*file_data.values.flatten)
-      # assume they're all erb now in case erb was added
-      Tempfile.open([file_basename, ".yml.erb"], target_dir) do |f|
-        f.write(data)
-      end
-    end
-    deploy_dir(target_dir, wait: wait)
-  ensure
-    FileUtils.remove_dir(target_dir) if target_dir
-  end
-
-  # Returns a hash containing a key for each template file in the set.
-  # The values of those keys are hashes containing a key for each resource type in the template file.
-  # In turn, the values of those keys are arrays of loaded kubernetes resource yaml.
+  # Deploys the specified set of fixtures via KubernetesDeploy::Runner.
+  #
+  # Optionally takes an array of filenames belonging to the fixture, and deploys that subset only.
+  # Example:
+  # # Deploys basic/redis.yml
+  # deploy_fixtures("basic", ["redis.yml"])
+  #
+  # Optionally yields a hash of the fixture's loaded templates that can be modified before the deploy is executed.
+  # The following example illustrates the format of the yielded hash:
+  #  {
+  #    "web.yml.erb" => {
+  #      "Ingress" => [loaded_ingress_yaml],
+  #      "Service" => [loaded_service_yaml],
+  #      "Deployment" => [loaded_service_yaml]
+  #    }
+  #  }
   #
   # Example:
-  #   load_fixture_data("basic", ["web"])
-  #   => {
-  #        "web" => {
-  #          "Ingress" => [loaded_ingress_yaml],
-  #          "Service" => [loaded_service_yaml],
-  #          "Deployment" => [loaded_service_yaml]
-  #        }
-  #      }
-  def load_fixture_data(set, subset=nil)
-    source_dir = fixture_path(set)
-    templates = {}
+  # # The following will deploy the "basic" fixture set, but with the unmanaged pod modified to use a bad image
+  #   deploy_fixtures("basic") do |fixtures|
+  #     pod = fixtures["unmanaged-pod.yml.erb"]["Pod"].first
+  #     pod["spec"]["containers"].first["image"] = "hello-world:thisImageIsBad"
+  #   end
+  def deploy_fixtures(set, subset: nil, wait: true)
+    fixtures = load_fixtures(set, subset)
+    raise "Cannot deploy empty template set" if fixtures.empty?
 
-    each_k8s_yaml(source_dir, subset) do |basename, ext, content|
-      templates[basename] = {}
-      YAML.load_stream(content) do |doc|
-        templates[basename][doc["kind"]] ||= []
-        templates[basename][doc["kind"]] << doc
-      end
-    end
-    templates
+    yield fixtures if block_given?
+
+    target_dir = Dir.mktmpdir
+    write_fixtures_to_dir(fixtures, target_dir)
+    deploy_dir(target_dir, wait: wait)
+  ensure
+    FileUtils.remove_dir(target_dir) if target_dir
   end
 
-  private
+  def fixture_path(set_name)
+    source_dir = File.expand_path("../../fixtures/#{set_name}", __FILE__)
+    raise "Fixture set #{set_name} does not exist as directory #{source_dir}" unless File.directory?(source_dir)
+    source_dir
+  end
 
+  # Deploys all fixtures in the given directory via KubernetesDeploy::Runner
+  # Exposed for direct use only when deploy_fixtures cannot be used because the template cannot be loaded pre-deploy,
+  # for example because it contains an intentional syntax error
   def deploy_dir(dir, sha: 'abcabcabc', wait: true)
     runner = KubernetesDeploy::Runner.new(
       namespace: @namespace,
@@ -76,20 +55,38 @@ module FixtureDeployHelper
     runner.run
   end
 
-  def fixture_path(set_name)
-    source_dir = File.expand_path("../../fixtures/#{set_name}", __FILE__)
-    raise "Fixture set #{set_name} does not exist as directory #{source_dir}" unless File.directory?(source_dir)
-    source_dir
-  end
+  private
 
-  def each_k8s_yaml(source_dir, subset)
-    Dir["#{source_dir}/*.yml*"].each do |filename|
-      match_data = File.basename(filename).match(/(?<basename>.*)(?<ext>\.yml(?:\.erb)?)\z/)
-      basename = match_data[:basename]
-      ext = match_data[:ext]
+  def load_fixtures(set, subset)
+    fixtures = {}
+    Dir["#{fixture_path(set)}/*.yml*"].each do |filename|
+      basename = File.basename(filename)
       next unless !subset || subset.include?(basename)
 
-      yield basename, ext, File.read(filename)
+      content = File.read(filename)
+      fixtures[basename] = {}
+      YAML.load_stream(content) do |doc|
+        fixtures[basename][doc["kind"]] ||= []
+        fixtures[basename][doc["kind"]] << doc
+      end
     end
+    fixtures
+  end
+
+  def write_fixtures_to_dir(fixtures, target_dir)
+    files = [] # keep reference outside Tempfile.open to prevent garbage collection
+    fixtures.each do |filename, file_data|
+      basename, exts = extract_basename_and_extensions(filename)
+      data = YAML.dump_stream(*file_data.values.flatten)
+      Tempfile.open([basename, exts], target_dir) do |f|
+        files << f
+        f.write(data)
+      end
+    end
+  end
+
+  def extract_basename_and_extensions(filename)
+    match_data = filename.match(/(?<basename>.*)(?<ext>\.yml(?:\.erb)?)\z/)
+    [match_data[:basename], match_data[:ext]]
   end
 end
--- a/test/helpers/fixture_deploy_helper.rb
+++ b/test/helpers/fixture_deploy_helper.rb
@@ -1,0 +1,90 @@
+module FixtureDeployHelper
+  # use deploy_fixture_set if you are not adding to or otherwise modifying the template set
+  def deploy_fixture_set(set, subset=nil, wait: true)
+    source_dir = fixture_path(set)
+    return deploy_dir(source_dir) unless subset
+
+    target_dir = Dir.mktmpdir
+    files = []
+    each_k8s_yaml(source_dir, subset) do |basename, ext, content|
+      Tempfile.open([basename, ext], target_dir) do |f|
+        files << f
+        f.write(content)
+      end
+    end
+
+    deploy_dir(target_dir, wait: wait)
+  ensure
+    files.each { |f| File.delete(f) } if files
+  end
+
+  # use load_fixture_data + deploy_loaded_fixture_set to have the chance to add to / modify the template set before deploy
+  def deploy_loaded_fixture_set(template_map, wait: true)
+    dir = Dir.mktmpdir
+    files = []
+    template_map.each do |file_basename, file_data|
+      data = YAML.dump_stream(*file_data.values.flatten)
+      # assume they're all erb now in case erb was added
+      Tempfile.open([file_basename, ".yml.erb"], dir) do |f|
+        files << f
+        f.write(data)
+      end
+    end
+    deploy_dir(dir, wait: wait)
+  ensure
+    files.each { |f| File.delete(f) }
+  end
+
+  # load_fixture_data return format
+  # {
+  #   file_basename: {
+  #     type_name: [
+  #       loaded_yaml_of_type,
+  #       loaded_yaml_of_type,
+  #     ]
+  #   }
+  # }
+  def load_fixture_data(set, subset=nil)
+    source_dir = fixture_path(set)
+    templates = {}
+
+    each_k8s_yaml(source_dir, subset) do |basename, ext, content|
+      templates[basename] = {}
+      YAML.load_stream(content) do |doc|
+        templates[basename][doc["kind"]] ||= []
+        templates[basename][doc["kind"]] << doc
+      end
+    end
+    templates
+  end
+
+  def deploy_dir(dir, sha: 'abcabcabc', wait: true)
+    runner = KubernetesDeploy::Runner.new(
+      namespace: @namespace,
+      current_sha: sha,
+      context: KubeclientHelper::MINIKUBE_CONTEXT,
+      template_dir: dir,
+      wait_for_completion: wait,
+    )
+    runner.run
+  end
+
+  # HELPER METHODS BELOW NOT INTENDED FOR INCLUDING CLASS USE
+
+  def fixture_path(set_name)
+    source_dir = File.expand_path("../../fixtures/#{set_name}", __FILE__)
+    raise "Fixture set pat #{source_dir} is invalid" unless File.directory?(source_dir)
+    source_dir
+  end
+
+  def each_k8s_yaml(source_dir, subset)
+    Dir["#{source_dir}/*.yml*"].each do |filename|
+      match_data = File.basename(filename).match(/(?<basename>.*)(?<ext>\.yml(?:\.erb)?)\z/)
+      basename = match_data[:basename]
+      ext = match_data[:ext]
+      next unless !subset || subset.include?(basename)
+
+      yield basename, ext, File.read(filename)
+    end
+  end
+end
--- a/test/helpers/fixture_deploy_helper.rb
+++ b/test/helpers/fixture_deploy_helper.rb
@@ -1,6 +1,6 @@
 module FixtureDeployHelper
   # use deploy_fixture_set if you are not adding to or otherwise modifying the template set
-  def deploy_fixture_set(set, subset=nil, wait: true)
+  def deploy_fixture_set(set, subset: nil, wait: true)
     source_dir = fixture_path(set)
     return deploy_dir(source_dir) unless subset
 
@@ -15,7 +15,7 @@ module FixtureDeployHelper
 
     deploy_dir(target_dir, wait: wait)
   ensure
-    files.each { |f| File.delete(f) } if files
+    FileUtils.remove_dir(target_dir) if target_dir
   end
 
   # use load_fixture_data + deploy_loaded_fixture_set to have the chance to add to / modify the template set before deploy
@@ -69,7 +69,7 @@ module FixtureDeployHelper
     runner.run
   end
 
-  # HELPER METHODS BELOW NOT INTENDED FOR INCLUDING CLASS USE
+  private
 
   def fixture_path(set_name)
     source_dir = File.expand_path("../../fixtures/#{set_name}", __FILE__)
--- a/test/helpers/fixture_deploy_helper.rb
+++ b/test/helpers/fixture_deploy_helper.rb
@@ -8,10 +8,8 @@ module FixtureDeployHelper
     return deploy_dir(source_dir) unless subset
 
     target_dir = Dir.mktmpdir
-    files = []
     each_k8s_yaml(source_dir, subset) do |basename, ext, content|
       Tempfile.open([basename, ext], target_dir) do |f|
-        files << f
         f.write(content)
       end
     end
@@ -25,19 +23,17 @@ module FixtureDeployHelper
   # and invokes KubernetesDeploy::Runner on that directory with the requested options.
   # Use this with load_fixture_data when you want to add to and/or modify the template set before deploying it.
   def deploy_loaded_fixture_set(template_map, wait: true)
-    dir = Dir.mktmpdir
-    files = []
+    target_dir = Dir.mktmpdir
     template_map.each do |file_basename, file_data|
       data = YAML.dump_stream(*file_data.values.flatten)
       # assume they're all erb now in case erb was added
-      Tempfile.open([file_basename, ".yml.erb"], dir) do |f|
-        files << f
+      Tempfile.open([file_basename, ".yml.erb"], target_dir) do |f|
         f.write(data)
       end
     end
-    deploy_dir(dir, wait: wait)
+    deploy_dir(target_dir, wait: wait)
   ensure
-    files.each { |f| File.delete(f) }
+    FileUtils.remove_dir(target_dir) if target_dir
   end
 
   # Returns a hash containing a key for each template file in the set.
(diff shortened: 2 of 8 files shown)


test/integration/kubernetes_deploy_test.rb
```diff
@@ -1,27 +1,118 @@
...
-      wait_for_completion: true,
-    )
-    runner.run
-
-    pods = kubeclient.get_pods(namespace: @namespace)
-    managed_pods, unmanaged_pods = pods.partition do |pod|
-      pod.metadata.ownerReferences && pod.metadata.ownerReferences.first.kind == "ReplicaSet"
+  def test_full_basic_set_deploy_succeeds
+    deploy_fixture_set("basic")
+    basic = FixtureSetAssertions::Basic.new(@namespace)
```
kirs: What's the value of wrapping assertions into `FixtureSetAssertions::Basic.new` if we can use `assert_all_up(@namespace)` as a module method, and include this assertions module into `KubernetesDeploy::IntegrationTest`?

test/helpers/fixture_sets/basic.rb
```diff
@@ -0,0 +1,63 @@
...
+    end
+
+    def assert_unmanaged_pod_statuses(status, count=1)
+      pods = kubeclient.get_pods(namespace: namespace, label_selector: "type=unmanaged-pod,app=#{app_name}")
+      assert_equal count, pods.size, "Expected to find #{count} unmanaged pod(s), found #{pods.size}"
+      assert pods.all? { |pod| pod.status.phase == status }
+    end
+
+    def refute_managed_pod_exists
+      assert_unmanaged_pod_statuses("", 0)
```
kirs: From the first look, it's not very clear what's the meaning of an empty string. I think it would way better readable to do something like `assert_equal 0, managed_pods.size`, since the purpose of this method is to make sure there are no managed pods

test/helpers/fixture_set.rb
```diff
@@ -0,0 +1,78 @@
...
+      assert_equal 1, pvc.size, "Expected 1 #{pvc_name} pvc, saw #{pvc.size}"
+      assert_equal status, pvc.first.status.phase
+    end
+
+    def assert_ingress_up(ing_name)
+      ing = v1beta1_kubeclient.get_ingresses(namespace: namespace, label_selector: "name=#{ing_name},app=#{app_name}")
+      assert_equal 1, ing.size, "Expected 1 #{ing_name} ingress, got #{ing.size}"
+    end
+
+    def assert_configmap_up(cm_name, expected_data)
```
kirs: Maybe `assert_configmap_present`? Since configmap doesn't have up/down status.

test/helpers/kubeclient_helper.rb
```diff
@@ -2,23 +2,29 @@ module KubeclientHelper
...
-    @kubeclient ||= begin
-      config = Kubeclient::Config.read(ENV["KUBECONFIG"])
-      unless config.contexts.include?(MINIKUBE_CONTEXT)
-        raise "`#{MINIKUBE_CONTEXT}` context must be configured in your KUBECONFIG (#{ENV["KUBECONFIG"]}). Please see the README."
-      end
-      minikube = config.context(MINIKUBE_CONTEXT)
+    @kubeclient ||= build_kube_client("v1")
+  end
+
+  def v1beta1_kubeclient
```
kirs: Maybe `kubeclient_v1beta1` to be consistent with `kubeclient` method?

test/integration/kubernetes_deploy_test.rb
```diff
@@ -1,27 +1,118 @@
...
+    deploy_fixture_set("basic", ["redis"])
+    basic = FixtureSetAssertions::Basic.new(@namespace)
+    basic.assert_all_redis_resources_up
+
+    deploy_fixture_set("basic", ["configmap-data"])
+    basic.assert_configmap_data_up
+    basic.refute_redis_resources_exist(expect_pvc: true)
+  end
+
+  def test_success_with_unrecognized_resource_type
```
kirs: When reading this method I'm trying to find the actual unrecognized resource type, but it's only deploying secrets.

test/integration/kubernetes_deploy_test.rb
```diff
@@ -1,27 +1,118 @@
...
+
+    @logger_stream.rewind
+    assert_match /error validating data\: found invalid field myKey for v1.ObjectMeta/, @logger_stream.read
+  end
+
+  def test_bad_container_image_on_run_once_halts_and_fails_deploy
+    fixture_set = load_fixture_data("basic")
+    pod = fixture_set["unmanaged-pod"]["Pod"].first
+    pod["spec"]["activeDeadlineSeconds"] = 3
+    pod["spec"]["containers"].first["image"] = "hello-world:elephants"
```
kirs: How would we know that `"hello-world:elephants"` is a bad container?

test/helpers/fixture_deploy_helper.rb
```diff
@@ -0,0 +1,90 @@
...
+
+  def fixture_path(set_name)
+    source_dir = File.expand_path("../../fixtures/#{set_name}", __FILE__)
+    raise "Fixture set pat #{source_dir} is invalid" unless File.directory?(source_dir)
+    source_dir
+  end
+
+  def each_k8s_yaml(source_dir, subset)
+    Dir["#{source_dir}/*.yml*"].each do |filename|
+      match_data = File.basename(filename).match(/(?<basename>.*)(?<ext>\.yml(?:\.erb)?)\z/)
```
kirs: If what we want to extract here is basename and extname, you can use `File.basename` and `File.extname`.

test/helpers/fixture_deploy_helper.rb
```diff
@@ -0,0 +1,90 @@
...
+        f.write(content)
+      end
+    end
+
+    deploy_dir(target_dir, wait: wait)
+  ensure
+    files.each { |f| File.delete(f) } if files
+  end
+
+  # use load_fixture_data + deploy_loaded_fixture_set to have the chance to add to / modify the template set before deploy
```
kirs: May be worth it converting this into RDoc with a usage example (like what Rails does)

test/fixtures/invalid/yaml-error.yml
```diff
@@ -0,0 +1,10 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: basic-configmap-data
+  labels:
+    name: basic-configmap-data
+    app: basic
+data:
+  datapoint1: value1:
```
kirs: is it intended to be `value1:`?

test/test_helper.rb
```diff
@@ -77,4 +82,5 @@ def self.prepare_pv(name)
   end
 
   TestProvisioner.prepare_pv("pv0001")
+  TestProvisioner.prepare_pv("pv0002")
```
kirs: Is `pv0002` used somewhere else or you created it to allocate more space?

test/helpers/fixture_sets/basic.rb
```diff
@@ -0,0 +1,63 @@
+module FixtureSetAssertions
+  class Basic < FixtureSet
```
kirs: TBH I'm not a big fan of naming it `Basic` in this case, because first time when you look at it you think it's a base class like `ActiveRecord::Base`, but then you see it's only about the basic scenario.

As an alternative I can think about moving basic scenario testing into `test/integration/basic_test.rb` and storing all basic helpers right in that file together with tests. This way you can limit basic assertions only to that file.

kirs:
sorry, I meant to choose this one not the approval

screen shot 2017-02-22 at 14 42 13

test/helpers/fixture_deploy_helper.rb
```diff
@@ -0,0 +1,90 @@
+module FixtureDeployHelper
+  # use deploy_fixture_set if you are not adding to or otherwise modifying the template set
+  def deploy_fixture_set(set, subset=nil, wait: true)
```
KnVerey: sure 👍

test/helpers/fixture_deploy_helper.rb
```diff
@@ -0,0 +1,90 @@
...
+      data = YAML.dump_stream(*file_data.values.flatten)
+      # assume they're all erb now in case erb was added
+      Tempfile.open([file_basename, ".yml.erb"], dir) do |f|
+        files << f
+        f.write(data)
+      end
+    end
+    deploy_dir(dir, wait: wait)
+  ensure
+    files.each { |f| File.delete(f) }
```
KnVerey: yep, definitely!

test/helpers/fixture_deploy_helper.rb
```diff
@@ -0,0 +1,90 @@
...
+      namespace: @namespace,
+      current_sha: sha,
+      context: KubeclientHelper::MINIKUBE_CONTEXT,
+      template_dir: dir,
+      wait_for_completion: wait,
+    )
+    runner.run
+  end
+
+  # HELPER METHODS BELOW NOT INTENDED FOR INCLUDING CLASS USE
```
KnVerey: hmm, good question. 😆 
This probably came from playing with moving things around... earlier on there were a bunch of different types of things in the same file and I had notes all over the place. Will fix!

test/helpers/fixture_deploy_helper.rb
KnVerey: What we need to extract is:
* `/foo/bar/baz/template.yml` -> 'template' and '.yml'
* `/foo/bar/baz/template.yml.erb` -> 'template' and '.yml.erb'

The problem I was having with just using those two is that `File.extname` only grabs the last extension, and `File.basename` needs you to say exactly which extension you're removing, whereas there are two options.

test/integration/kubernetes_deploy_test.rb
KnVerey: It'd have to be `assert_all_up('basic', @namespace)` or `assert_all_basic_up(@namespace)`, unless we consider it unlikely there'll be multiple fixture sets (hard to say--there's a second now, but we don't need any assertions for it so far). In any case, I actually started out with the latter and moved to this class-based structure as an experiment, and kept it because I found it cleaner and liked the more forceful encapsulation of the knowledge of what the fixtures contain. I'm open to switching back if you guys hate this. 😄

test/integration/kubernetes_deploy_test.rb
KnVerey: Secrets _are_ unrecognized. If you run this test with the main logger turned on, you'll see this:

image

I can add a comment I guess? I chose them because they're the one thing I don't think we should ever add explicit support for, since it means you're committing secrets to your repo.

test/integration/kubernetes_deploy_test.rb
KnVerey: Do you mean as someone reading the tests (in which case I should change it to `thisImageIsBad`) or are you worried the elephants tag could actually exist in the future (in which case I should do something like append a random string).

kirs: > Do you mean as someone reading the tests (in which case I should change it to thisImageIsBad)

Exactly that

test/integration/kubernetes_deploy_test.rb
kirs: Yeah, comment would be great. Otherwise I would expect something like `non-existing-resource-blah` to be the "right" unrecognized resource, not `secrets`

test/helpers/fixture_sets/basic.rb
KnVerey: IMO that still sounds like some sort of "basic" test rather than a group of tests that all run against a fixture called "basic". How about we ignore the fixture name in choosing the structure, and rename the fixture in a follow-up? Maybe even a suffix like "App" would help. Or perhaps "HelloWorld" or "HelloCloud", or something outlandish. I dunno.

test/fixtures/invalid/yaml-error.yml
KnVerey: yes, that's the yaml error :)

test/test_helper.rb
KnVerey: If I understand correctly how this kind of PV works (as opposed to the dynamic kind you get via storage classes), the entire PV is reserved by PVCs made against it. So in theory pv0001 could still be getting recycled from a previous test/namespace's claim the next time we need it. I didn't actually see this happen though.

test/helpers/fixture_deploy_helper.rb
kirs: it'a usually more reliable to use standard library for such things because it eliminates cases not covered by the regexp, but here I see why it wouldn't work 👍

test/helpers/fixture_sets/basic.rb
kirs: Great idea! I like `HelloCloud` 👍

test/helpers/kubeclient_helper.rb
KnVerey: That sounds weird to me tbh, 'cause it's the target api not the client that is v1/v1beta1. How would you feel about adding the prefix to `v1_kubeclient` instead?

test/helpers/fixture_deploy_helper.rb
KnVerey: I'm happy to do this, but I haven't written RDoc comments before and I'm having trouble finding the standards for usage examples. Can you point me in the right direction tomorrow?

kirs: Here is an example: https://github.com/rails/rails/blob/master/activerecord/lib/active_record/core.rb#L266-L271

test/helpers/fixture_deploy_helper.rb
sirupsen: An alternative would be:

```ruby
parts = File.basename(filename).split('.')
basename, ext = parts[0], parts[1..-1]
```

test/helpers/fixture_deploy_helper.rb
```diff
@@ -0,0 +1,90 @@
...
+  end
+
+  def each_k8s_yaml(source_dir, subset)
+    Dir["#{source_dir}/*.yml*"].each do |filename|
+      match_data = File.basename(filename).match(/(?<basename>.*)(?<ext>\.yml(?:\.erb)?)\z/)
+      basename = match_data[:basename]
+      ext = match_data[:ext]
+      next unless !subset || subset.include?(basename)
+
+      yield basename, ext, File.read(filename)
```
sirupsen: Would it be better if the whole concern of parsing the templates was handled by this method and `YAML.load_stream` was called from here as well? Then you wouldn't need the double iteration from the consuming methods. That seems to me like a nice division of responsibility.

test/helpers/fixture_set.rb
```diff
@@ -0,0 +1,76 @@
...
+      assert_equal 1, pvc.size, "Expected 1 #{pvc_name} pvc, saw #{pvc.size}"
+      assert_equal status, pvc.first.status.phase
+    end
+
+    def assert_ingress_up(ing_name)
+      ing = v1beta1_kubeclient.get_ingresses(namespace: namespace, label_selector: "name=#{ing_name},app=#{app_name}")
+      assert_equal 1, ing.size, "Expected 1 #{ing_name} ingress, got #{ing.size}"
+    end
+
+    def assert_configmap_present(cm_name, expected_data)
```
sirupsen: What's the difference between something that's `present` and something that's `up`? Why does this take an `expected_data`, and the others don't? Do you think it'd be more consistent with Minitest to call this `assert_configmap_equal`?

test/helpers/fixture_set.rb
```diff
@@ -0,0 +1,76 @@
...
+    end
+
+    def assert_deployment_up(dep_name, replicas)
+      deployments = v1beta1_kubeclient.get_deployments(namespace: namespace, label_selector: "name=#{dep_name},app=#{app_name}")
+      assert_equal 1, deployments.size, "Expected 1 #{dep_name} deployment, got #{deployments.size}"
+      available = deployments.first["status"]["availableReplicas"]
+      assert_equal replicas, available, "Expected #{dep_name} deployment to have #{replicas} available replicas, saw #{available}"
+    end
+
+    def assert_pvc_status(pvc_name, status)
```
sirupsen: Is this acronym very common in the K8s world? Just want to make sure. Otherwise we might want to spell it out. ✏️

test/helpers/fixture_set.rb
```diff
@@ -0,0 +1,76 @@
...
+      services = kubeclient.get_services(namespace: namespace, label_selector: "name=#{svc_name},app=#{app_name}")
+      assert_equal 1, services.size, "Expected 1 #{svc_name} service, got #{services.size}"
+      refute services.first["spec"]["clusterIP"].empty?, "Cluster IP was not assigned"
+
+      endpoints_obj = kubeclient.get_endpoint(svc_name, namespace)
+      num_endpoints = endpoints_obj["subsets"].first["addresses"].length
+      assert_equal 1, num_endpoints
+    end
+
+    def assert_deployment_up(dep_name, replicas)
```
sirupsen: I think you should consider making `replicas` a named parameter. It may not be obvious otherwise with the name of this method what the second argument would assert. 📠

test/helpers/fixture_set.rb
```diff
@@ -0,0 +1,76 @@
...
+      @namespace
+    end
+
+    def app_name
+      raise FixtureSetError.new("@app_name must be set in initializer") if @app_name.blank?
+      @app_name
+    end
+
+    def refute_resource_exists(type, name, beta: false)
+      client = beta ? v1beta1_kubeclient : kubeclient
```
sirupsen: Arrghhh, this is so ugly due to upstream 😔  Can we auto-detect this somehow instead of passing `beta: true`? I'm afraid of that staying around forever..

test/helpers/fixture_set.rb
```diff
@@ -0,0 +1,76 @@
+require 'helpers/kubeclient_helper'
+
+module FixtureSetAssertions
+  class FixtureSet
```
sirupsen: This name doesn't seem quite like what I'd expect.. this seems more like fairly generic resource helpers, not that tied to fixtures as the `FixtureDeployHelper`

test/helpers/fixture_sets/basic.rb
sirupsen: `HelloCloud` 👋 ☁️

sirupsen: I don't have a strong opinion on leaving the helpers here or in the integration tests. It certainly helps break up things later, I think what that comes down to is how many fixture sets do we expect to have? Do you expect 2? 5? 10? I think that would also impact some of the other work in this PR if we don't really expect to ever need more than `HelloCloud`. I think you definitely made the right call to make it easy to manipulate the set instead of creating a lot of new ones, but perhaps people will find that so convenient they won't create sets altogether. Do you see any use-case for having more than one fixture set?

test/integration/kubernetes_deploy_test.rb
```diff
@@ -1,27 +1,134 @@
...
+    container = deployment["spec"]["template"]["spec"]["activeDeadlineSeconds"] = 1
+    deploy_loaded_fixture_set(fixture_set, wait: false) # this will never succeed as pods are killed after 1s
+
+    sleep 1 # make sure to hit DeadlineExceeded on at least one pod
+
+    deploy_fixture_set("basic", subset: ["web", "configmap-data"])
+    pods = kubeclient.get_pods(namespace: @namespace, label_selector: "name=web,app=basic")
+    running_pods, not_running_pods = pods.partition { |pod| pod.status.phase == "Running" }
+    assert_equal 1, running_pods.size
+    assert not_running_pods.size >= 1
```
sirupsen: Would this be better as a helper?

test/integration/kubernetes_deploy_test.rb
```diff
@@ -1,27 +1,134 @@
...
+    fixture_set = load_fixture_data("basic", ["configmap-data"])
+    configmap = fixture_set["configmap-data"]["ConfigMap"].first
+    configmap["metadata"]["myKey"] = "uhOh"
+
+    error = assert_raises(KubernetesDeploy::FatalDeploymentError) do
+      deploy_loaded_fixture_set(fixture_set)
+    end
+    assert_match /Dry run failed for template configmap-data/, error.to_s
+
+    @logger_stream.rewind
```
sirupsen: It's very likely someone will forget to do this, can we use a shared helper?

test/integration/kubernetes_deploy_test.rb
```diff
@@ -1,27 +1,134 @@
...
+
+    live_secret = kubeclient.get_secret("test", @namespace)
+    assert_equal({ foo: "YmFy" }, live_secret["data"].to_h)
+  end
+
+  def test_invalid_yaml_fails_fast
+    error = assert_raises(KubernetesDeploy::FatalDeploymentError) do
+      deploy_fixture_set("invalid", subset: ["yaml-error"])
+    end
+    assert_match /Template \S+ cannot be parsed/, error.to_s
```
sirupsen: I think you can use `assert_raises` with a regex as the second argument here to assert on the message.

```ruby
error_message = "/Template \S+ cannot be parsed/"
assert_raises(KubernetesDeploy::FatalDeploymentError, error_message) { .. }
```

test/integration/kubernetes_deploy_test.rb
```diff
@@ -1,27 +1,134 @@
...
+    pod["spec"]["activeDeadlineSeconds"] = 1
+    pod["spec"]["containers"].first["image"] = "hello-world:elephants"
 
-    assert_equal 2, managed_pods.size
-    managed_pods.each do |pod|
-      assert_equal "Running", pod.status.phase
+    error = assert_raises(KubernetesDeploy::FatalDeploymentError) do
+      deploy_loaded_fixture_set(fixture_set)
     end
+    assert_match /1 priority resources failed to deploy/, error.to_s
```
sirupsen: Do we want to assert that something about the image not being found is in the logging output?

test/helpers/fixture_sets/basic.rb
KnVerey: > how many fixture sets do we expect to have? Do you expect 2? 5? 10? Do you see any use-case for having more than one fixture set?

I'd expect 3-5, but I'm really just speculating. We technically have a second in this PR; it contains a template with yaml errors (afaik not possible to create by dumping a ruby structure). 

Two main possibilities come to mind: 
* Shopify's closed-source ThirdPartyResources that the script explicitly supports. There are two today, and likely to be several more in the future. Having a separate "shopify" set for them could make sense. That said, these tests would be questionable to set up, as we'd need to have the minikube cluster running the actual TPR controllers or the deploys won't succeed. 
* Sets that are templated differently, if/when we finally move forward with changing the templating engine and supporting partials + a top-level "values" yaml.

lib/kubernetes-deploy/runner.rb
```diff
@@ -319,7 +319,7 @@ def phase_heading(phase_name)
     end
 
     def log_green(msg)
-      STDOUT.puts "\033[0;32m#{msg}\x1b[0m\n" # green
+      KubernetesDeploy.logger.info("\033[0;32m#{msg}\x1b[0m")
```
KnVerey: Added this to the master issue under "code" (this change is actually from the branch this is based on, which I already merged, and will disappear when I push the rebase).

test/helpers/fixture_deploy_helper.rb
```diff
@@ -0,0 +1,90 @@
+module FixtureDeployHelper
+  # use deploy_fixture_set if you are not adding to or otherwise modifying the template set
+  def deploy_fixture_set(set, subset: nil, wait: true)
```
KnVerey: Besides the efficiency aspect, I'd have one concern about that change: `deploy_loaded_fixture_set` assumes that there's ERB in the file (because you could have added some, even if there wasn't any to begin with), so if we always went through it, we'd fail to exercise the script's recognition of pure `.yml` files. Removing that assumption from `deploy_loaded_fixture_set` would require exposing the original extension somewhere in the loaded fixture hash and likely adding a bit of complexity to `deploy_loaded_fixture_set`, so there's a bit of a tradeoff. I could go either way, but I slightly favour the way it is now.

test/helpers/fixture_set.rb
KnVerey: It is; you can use it with kubectl:

```
    * componentstatuses (aka 'cs')
    * configmaps (aka 'cm')
    * daemonsets (aka 'ds')
    * deployments (aka 'deploy')
    * endpoints (aka 'ep')
    * events (aka 'ev')
    * horizontalpodautoscalers (aka 'hpa')
    * ingresses (aka 'ing')
    * jobs
    * limitranges (aka 'limits')
    * namespaces (aka 'ns')
    * networkpolicies
    * nodes (aka 'no')
    * persistentvolumeclaims (aka 'pvc')
    * persistentvolumes (aka 'pv')
    * pods (aka 'po')
    * podsecuritypolicies (aka 'psp')
    * podtemplates
    * replicasets (aka 'rs')
    * replicationcontrollers (aka 'rc')
    * resourcequotas (aka 'quota')
    * secrets
    * serviceaccounts (aka 'sa')
    * services (aka 'svc')
    * statefulsets
    * storageclasses
    * thirdpartyresources
```

test/helpers/fixture_set.rb
KnVerey: I had this as `up` for consistency, but Kir asked me to change it to `present` because that's what it's actually checking. Puristically speaking, I'd prefer for all these methods to be called `up` and the caller to not care whether that involves checks beyond presence. The data thing is not really necessary. Maybe I was thinking that we could eventually add a test that deploys some data and then redeploys a modified version + a pod that relies on the modifications. Could be a good test of the reasons behind the rollout order (as mentioned in the description, this PR strives for parity with my manual tophat flow... there's a lot more tests to add still!). I wouldn't be against having `basic.assert_configmap_data_up` call `assert_configmap_present` + `assert_configmap_equal` or something, but not sure that much extraction is really warranted for such simple checks.

test/helpers/fixture_deploy_helper.rb
KnVerey: Those comments in that file don't look super standardized to me, so I'm still not really sure what you want to see. I'll take a stab at it though.

test/helpers/fixture_set.rb
KnVerey: This class is an abstract superclass that cannot be instantiated on its own, and its subclasses each encapsulate knowledge about the contents of a fixture set... they're extremely tied to the fixtures.

test/helpers/fixture_set.rb
KnVerey: Maaaybe. Each kubeclient instance talks to the specified API to discover the resource types it can handle and dynamically creates all the resulting methods. I'd rather save this investigation for a follow-up (added to the master issue).

test/integration/kubernetes_deploy_test.rb
KnVerey: I don't think so, unless I'm misunderstanding for what... `basic.assert_some_web_pods_running(running: N, min_not_running: M)` doesn't strike me as clearer or reusable.

test/integration/kubernetes_deploy_test.rb
KnVerey: We can do `assert_logs_match(/DeadlineExceeded/)`, but note that there probably won't be anything about the bad image in the logs. A one-second deadline is not likely enough for the image pull attempt to finish in the first place; the `activeDeadlineSeconds` is the fastest way to produce these circumstances, and the bad image is just to make sure the pod won't flakily succeed if ever it does manage to finish pulling.

test/helpers/fixture_deploy_helper.rb
KnVerey: The callers do completely different things with the yielded values... one loads the stream and the other dumps the contents to file. The alternative is to make two helpers that duplicate the logic in this one.

test/helpers/fixture_deploy_helper.rb
```diff
@@ -0,0 +1,90 @@
...
+  # load_fixture_data return format
+  # {
+  #   file_basename: {
+  #     type_name: [
+  #       loaded_yaml_of_type,
+  #       loaded_yaml_of_type,
+  #     ]
+  #   }
+  # }
+  def load_fixture_data(set, subset=nil)
```
KnVerey: > I think we should consider not differing in the API between the two cases of (1) using defaults, (2) modifying fixtures. Since (1) is just the null case of (2)

💯  Absolutely on point. This gave me an idea: all (2) needs is a hook into the middle step of (1). Please check out the last commit: https://github.com/Shopify/kubernetes-deploy/pull/24/commits/f254f446a3bcd6b4f36870d1b312cb4898916ed1#diff-43c27b92ee93912981b34f7d8b622714. There's one case that has to be different (the test for unparseable yaml handling--loading the file in the helper makes the test fail there instead of during the deploy), but it works for all the others and I think looks really nice!

test/helpers/fixture_deploy_helper.rb
```diff
@@ -0,0 +1,92 @@
+module FixtureDeployHelper
+  # Deploys the specified set of fixtures via KubernetesDeploy::Runner.
```
sirupsen: This is a really awesome API Katrina. Great job.

test/integration/kubernetes_deploy_test.rb
```diff
@@ -1,27 +1,128 @@
...
-    assert_equal 2, managed_pods.size
-    managed_pods.each do |pod|
-      assert_equal "Running", pod.status.phase
+    live_secret = kubeclient.get_secret("test", @namespace)
+    assert_equal({ foo: "YmFy" }, live_secret["data"].to_h)
+  end
+
+  def test_invalid_yaml_fails_fast
+    assert_raises(KubernetesDeploy::FatalDeploymentError, /Template \S+yaml-error\S+ cannot be parsed/) do
+      deploy_dir(fixture_path("invalid"))
```
sirupsen: Since this is the only test using this, do we need to expose it? Is this for the case where the fixture path has no ERB files? I think we can just rely on them always being ERB, or make sure that whatever reads them can understand ERB and non-ERB.

KnVerey: The purpose of this test is to cover deploy behaviour in the case where invalid yaml is provided. It needs circumvent the loading of the yaml file into a ruby structure, or else the yaml parse exception will be raised in the setup helpers instead of during the deploy. So we can't get it to use `deploy_fixtures`, but I can have it explicitly use the `KubernetesDeploy::Runner` instead of exposing the helper if you prefer.

test/helpers/fixture_deploy_helper.rb
```diff
@@ -0,0 +1,92 @@
...
+  #      "Service" => [loaded_service_yaml],
+  #      "Deployment" => [loaded_service_yaml]
+  #    }
+  #  }
+  #
+  # Example:
+  # # The following will deploy the "basic" fixture set, but with the unmanaged pod modified to use a bad image
+  #   deploy_fixtures("basic") do |fixtures|
+  #     pod = fixtures["unmanaged-pod.yml.erb"]["Pod"].first
+  #     pod["spec"]["containers"].first["image"] = "hello-world:thisImageIsBad"
```
kirs: Thanks for making it a great doc!

test/helpers/fixture_set.rb
```diff
@@ -0,0 +1,76 @@
...
+    class FixtureSetError < StandardError; end
+    include KubeclientHelper
+    include Minitest::Assertions
+    attr_writer :assertions
+
+    def initialize
+      raise NotImplementedError
+    end
+
+    def assertions
```
kirs: This is required to mimic minitest assertions, right?

test/helpers/fixture_set.rb
```diff
@@ -0,0 +1,76 @@
...
+      raise FixtureSetError.new("@app_name must be set in initializer") if @app_name.blank?
+      @app_name
+    end
+
+    def refute_resource_exists(type, name, beta: false)
+      client = beta ? v1beta1_kubeclient : kubeclient
+      resources = client.public_send("get_#{type}", name, namespace) # 404s
+      flunk "#{type} #{name} unexpectedly existed"
+    rescue KubeException => e
+      raise unless e.to_s.include?("not found")
```
kirs: Looks like it would be great for a gem to provide `KubeNotFoundError` specific for 404.

test/helpers/fixture_set.rb
```diff
@@ -0,0 +1,76 @@
...
+    end
+
+    def assert_ingress_up(ing_name)
+      ing = v1beta1_kubeclient.get_ingresses(namespace: namespace, label_selector: "name=#{ing_name},app=#{app_name}")
+      assert_equal 1, ing.size, "Expected 1 #{ing_name} ingress, got #{ing.size}"
+    end
+
+    def assert_configmap_present(cm_name, expected_data)
+      configmaps = kubeclient.get_config_maps(namespace: namespace, label_selector: "name=#{cm_name},app=#{app_name}")
+      assert_equal 1, configmaps.size, "Expected 1 configmap, got #{configmaps.size}"
```
kirs: 👍 for passing the error message

test/test_helper.rb
```diff
@@ -19,10 +23,16 @@ def setup
     def teardown
       @logger_stream.close
     end
+
+    def assert_logs_match(regexp)
```
kirs: 😍

test/helpers/fixture_set.rb
KnVerey: Yes, I wish. Maybe we can dig in and PR that upstream.

test/helpers/fixture_set.rb
KnVerey: Exactly. See https://github.com/seattlerb/minitest/issues/286.

PR #25, opened 2017-02-21
Require ENV['REVISION'] on startup

kirs:
Right now having `ENV['REVISION']` missing will lead to `Current SHA must be specified` as the error message. However the message doesn't tell you what exactly is wrong and how to fix it.

This check is only the case when running the script locally. Shipit always has the `REVISION` set.

Since `Runner` has no context of `ENV['REVISION']`, I think the right place for the check would be in the bin script.

review @KnVerey @sirupsen @wfarr

### DIFF
--- a/exe/kubernetes-deploy
+++ b/exe/kubernetes-deploy
@@ -21,11 +21,16 @@ if !template_dir || template_dir.empty?
   exit 1
 end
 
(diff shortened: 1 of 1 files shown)


kirs:
That's because https://github.com/Shopify/kubernetes-deploy/pull/22 is not merged yet

PR #29, opened 2017-02-23
Log context

kirs:
Since now we may run the script in parallel to deploy to multiple clusters concurrently, it's important to see which context produced the log line.

I'd also unblock me to integrate the script with the Shipit rollout UI, since the way how it works is parsing the the log line. Right now the line like `[KUBESTATUS] {"group":"deployments","name":"web","status_string":"Waiting for rollout to finish: 132 of 150 updated replicas are available..."}` doesn't tell us which cluster/context is progressing.

review @wfarr @sirupsen @KnVerey

### DIFF
--- a/lib/kubernetes-deploy/kubernetes_resource.rb
+++ b/lib/kubernetes-deploy/kubernetes_resource.rb
@@ -125,7 +125,7 @@ module KubernetesDeploy
     end
 
     def log_status
-      KubernetesResource.logger.info("[KUBESTATUS] #{JSON.dump(status_data)}")
+      KubernetesResource.logger.info("[KUBESTATUS][#{@context}][#{@namespace}] #{JSON.dump(status_data)}")
     end
   end
 end


lib/kubernetes-deploy/kubernetes_resource.rb
```diff
@@ -112,7 +112,7 @@ def run_kubectl(*args)
     end
 
     def log_status
-      STDOUT.puts "[KUBESTATUS] #{JSON.dump(status_data)}"
+      STDOUT.puts "[KUBESTATUS #{@context}] #{JSON.dump(status_data)}"
```
KnVerey: How about `[KUBESTATUS][context]`, or even `[KUBESTATUS][context/namespace]`?

wfarr: Any reason the line as a whole just isn't a single JSON object containing context, namespace, etc?

kirs: I'm in favor of `[KUBESTATUS][context][namespace]`. Great suggestion!

> Any reason the line as a whole just isn't a single JSON object containing context, namespace, etc?

For me the idea is that the JSON object is the status of the resource and `[KUBESTATUS]` is the annotation

Issue #20, opened 2017-02-09
Failed deploy reported successful


stefanmb:
Syntax errors in the k8s template error out, but the deploy still reports as successful:

```
[WARN]	The following command failed: kubectl apply -f /tmp/testing-deployment.yml.erb20170209-12190-s9f69e --prune --all --prune-whitelist\=core/v1/ConfigMap --prune-whitelist\=core/v1/Pod --prune-whitelist\=core/v1/Service --prune-whitelist\=batch/v1/Job --prune-whitelist\=extensions/v1beta1/DaemonSet --prune-whitelist\=extensions/v1beta1/Deployment --prune-whitelist\=extensions/v1beta1/HorizontalPodAutoscaler --prune-whitelist\=extensions/v1beta1/Ingress --prune-whitelist\=apps/v1beta1/StatefulSet --namespace\=pipa-test --context\=pipa-test
[WARN]	error: unable to decode "/tmp/testing-deployment.yml.erb20170209-12190-s9f69e": [pos 290]: json: expect char '"' but got char '8'
Waiting for Deployment/pipa-test
[KUBESTATUS] {"group":"Pipa-test deployment","name":"pipa-test-1149558948-5l604","status_string":"Running (Ready: true)","exists":true,"succeeded":true,"failed":false,"timed_out":false}
...
Spent 0.58s waiting for Deployment/pipa-test
Deploy succeeded!
```

Should this error be caught in `shipit-engine` or in this gem?

KnVerey:
It's the gem's problem. Thanks for reporting.

kirs:
@KnVerey I'm wondering if we should raise the error here: https://github.com/Shopify/kubernetes-deploy/blob/1da3249/lib/kubernetes-deploy/runner.rb#L305

KnVerey:
No, because non-fatal errors can be seen there too. That method passes back the status, and the caller is supposed to deal with it. The weird thing here is that the initial dry run did not raise: https://github.com/Shopify/kubernetes-deploy/blob/1da3249/lib/kubernetes-deploy/runner.rb#L145. That implies it was the load->dump that introduced the yaml error, which is pretty strange.

In any case, the thing to do IMO is to handle the result here: https://github.com/Shopify/kubernetes-deploy/blob/1da3249/lib/kubernetes-deploy/runner.rb#L271. I.e. on success, log the output so they can see what was pruned, and raise on failure.

kirs:
[...]

PR #31, opened 2017-02-24
Fail when kubectl apply returns non-zero status

kirs:
Previously, when `kubectl apply` returned non-zero status we've been only printing the warning. This caused issues like https://github.com/Shopify/kubernetes-deploy/issues/20. We should raise an error and stop the deploy in this case.

review @KnVerey @sirupsen @wfarr

### DIFF
--- a/lib/kubernetes-deploy/runner.rb
+++ b/lib/kubernetes-deploy/runner.rb
@@ -269,7 +269,13 @@ MSG
         PRUNE_WHITELIST.each { |type| command.push("--prune-whitelist=#{type}") }
       end
 
-      run_kubectl(*command)
+      _, err, st = run_kubectl(*command)
+      unless st.success?
+        raise FatalDeploymentError, <<-MSG
+"The following command failed: #{Shellwords.join(command)}"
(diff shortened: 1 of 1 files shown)


kirs:
Closes https://github.com/Shopify/kubernetes-deploy/issues/20

kirs:
@wfarr here is the config I tried:

```
apiVersion: extensions/v1beta1
kind: ThirdPartyResource
description: "Redis Third"
metadata:
  name: "redis.shopify.io"
versions:
  - name: v1
```

The script prints this since we don't have a third party resource handler:

```
[WARN]  Don't know how to monitor resources of type thirdpartyresource. Assuming thirdpartyresource/redis.shopify.io deployed successfully.
```

and after the successful deploy I could verify that it was created:

```
$ kubectl get thirdpartyresources
NAME               DESCRIPTION   VERSION(S)
redis.shopify.io   Redis Third   v1
```

wfarr:
Ah word, I didn't look at this carefully enough. This should be safe yeah.

PR #32, opened 2017-02-24
Inspect the file from error message

kirs:
In https://github.com/Shopify/kubernetes-deploy/pull/31, I made `kubectl apply` failure a hard raise rather than a warning. The typical error message in the output would be:

```
error: unable to decode "/tmp/testing-deployment.yml.erb20170209-12190-s9f69e": [pos 290]: json: expect char '"' but got char '8'
```

The problem here is that the file with a syntax error was temporary. At the time when we'll look into the issue and inspect the file, it won't be available anymore.

My suggestion here is to catch the file path in the error message and log the file content. This should dramatically improve developer experience when hunting syntax errors.

review @sirupsen @KnVerey @wfarr

### DIFF
--- a/lib/kubernetes-deploy/runner.rb
+++ b/lib/kubernetes-deploy/runner.rb
@@ -112,6 +112,24 @@ MSG
 
     private
 
+    # Inspect the file referenced in the kubectl stderr
+    # to make it easier for developer to understand what's going on
+    def inspect_kubectl_out_for_files(stderr)
+      # Output example:
+      # error: unable to decode "/tmp/path/to/file": [pos 96]: json: expect char '"' but got char '{'
+      match = stderr.match(/error: unable to decode "(?<path>\S+)":/)
+      return unless match
+
+      path = match[:path]
+      if path.present? && File.file?(path)
+        suspicious_file = File.read(path)
+        KubernetesDeploy.logger.warn("Inspecting the file mentioned in the error message (#{path})")
+        KubernetesDeploy.logger.warn(suspicious_file)
+      else
+        KubernetesDeploy.logger.warn("Detected a file (#{path.inspect}) referenced in the kubectl stderr but was unable to inspect it")
+      end
+    end
+
     def predeploy_priority_resources(resource_list)
       PREDEPLOY_SEQUENCE.each do |resource_type|
         matching_resources = resource_list.select { |r| r.type == resource_type }
@@ -271,6 +289,7 @@ MSG
 
       _, err, st = run_kubectl(*command)
       unless st.success?
+        inspect_kubectl_out_for_files(err)
         raise FatalDeploymentError, <<-MSG
 "The following command failed: #{Shellwords.join(command)}"
 #{err}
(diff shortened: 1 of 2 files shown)


lib/kubernetes-deploy/runner.rb
```diff
@@ -112,6 +112,21 @@ def template_variables
 
     private
 
+    # Inspect the file that was mentioned in the kubectl output
+    # to make it easier for developer to understand what's going on
+    def inspect_kubectl_out_for_yaml(out)
+      # Output example:
+      # error: unable to decode "/tmp/path/to/file": [pos 96]: json: expect char '"' but got char '{'
+      if match = out.match(/error:(.+)"(?<path>\S+)":+/)
```
sirupsen: This looks pretty prone to inspecting the wrong file.. is there a directory somewhere of what the errors will look like? Could we have a regex that'll look for things that look more like files?

lib/kubernetes-deploy/runner.rb
```diff
@@ -112,6 +112,21 @@ def template_variables
...
+        path = match[:path]
+        if path.present? && File.file?(path)
```
sirupsen: Good call on checking if it's actually a file :) You may want to log that we _tried_ to look for a file, but it wasn't there.

lib/kubernetes-deploy/runner.rb
kirs: > Could we have a regex that'll look for things that look more like files?

I don't want to fix to `/tmp/*` as a pattern of temp file since the temporary directory may vary:

```
Tempfile.new.path
=> "/var/folders/9k/zxhywlyx2nx_3z3qnrzpkjd00000gn/T/20170227-55216-1goaa6q"
``` 

Any anyway we check if it's a file below.

In this case we can only inspect `error: unable to decode` issues. It's the most common scenario that happens when the YAML has syntax errors that can't be caught by `--dry-run` (https://github.com/kubernetes/kubernetes/issues/42057)

sirupsen: A+ on submitting the bug upstream

kirs:
I updated to the code to only catch `error: unable to decode` and to print a warning if the file is not readable.

Looking forward to know @KnVerey opinion before merging it.

lib/kubernetes-deploy/runner.rb
```diff
@@ -112,6 +112,24 @@ def template_variables
 
     private
 
+    # Inspect the file referenced in the kubectl stderr
+    # to make it easier for developer to understand what's going on
+    def inspect_kubectl_out_for_files(stderr)
+      # Output example:
+      # error: unable to decode "/tmp/path/to/file": [pos 96]: json: expect char '"' but got char '{'
+      match = stderr.match(/error: unable to decode "(?<path>\S+)":+/)
```
KnVerey: Why `:+` at the end of the regex? Even if there were more than one colon somehow, why should the regex cover it?

kirs: It's definitely a typo. Thanks for spotting it!

kirs:
Here is the preview:

screen shot 2017-02-27 at 20 18 59

I'm going to add the test coverage since the PR with integration tests have been merged.

kirs:
Pushed it together with tests 👌 
It's amazing how the test helpers work and how easy it is to write a test case.

PR #56, opened 2017-02-27
Add rubocop

kirs:
I wish we've started using Rubocop since the beginning of the project.
This PR introduces the Shopify Ruby styleguide (http://shopify.github.io/ruby-style-guide/) and fixes couple of files that could be fixed with `rubocop --autocorrect`.

There are many other style violations that I'm planning to fix in following PRs.

review @sirupsen @wfarr @KnVerey

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,3 +7,6 @@
 /pkg/
 /spec/reports/
 /tmp/
+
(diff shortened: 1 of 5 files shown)


sirupsen:
rubbycop

Issue #53, opened 2017-02-27
Use kubeclient gem in place of kubectl in polling code


KnVerey:
We should be sure to investigate https://github.com/Shopify/kubernetes-deploy/issues/51 before doing this.

KnVerey:
This is high-effort and would add complexity to our code base (we'd have to deal with group version discovery, which kubectl is doing for us today), so we should validate that there's enough expected reward before undertaking it. We do know that `kubectl get` (<= 1.12 anyway) is strangely slow for very large batches of resources, likely related to https://github.com/kubernetes/kubernetes/issues/69856 (as of today, the fix is up but not merged). However, we haven't had reports of other issues with kubectl, and I would not be surprised to encounter problems/gotchas using kubeclient (a much less popular tool) at scale.

dturn:
Revist when `apply` is server side so we can actually remove the dependency on kubectl.

PR #58, opened 2017-02-28
Refactor kubectl runner into KubectlWrapper

kirs:
We have two places in the codebase that build and run `kubectl` command. I think this is redundant, and we already got troubles in the past updating one place and not updating the other (https://github.com/Shopify/kubernetes-deploy/pull/17).
It makes sense to move the duplicated code into a helper (`KubectlWrapper`). The PR is also introducing test coverage for the new class.

To be merged after https://github.com/Shopify/kubernetes-deploy/pull/32

review @KnVerey @wfarr @sirupsen

### DIFF
--- a/lib/kubernetes-deploy/kubectl_wrapper.rb
+++ b/lib/kubernetes-deploy/kubectl_wrapper.rb
@@ -1,0 +1,21 @@
+module KubernetesDeploy
+  module KubectlWrapper
+    extend self
+
+    def run_kubectl(*args, context: nil, namespace: nil)
+      args = args.unshift("kubectl")
+      args.push("--context=#{context}") if context
+      args.push("--namespace=#{namespace}") if namespace
+
+      KubernetesDeploy.logger.debug Shellwords.join(args)
+      out, err, st = Open3.capture3(*args)
+      KubernetesDeploy.logger.debug(out.shellescape)
+
+      unless st.success?
+        KubernetesDeploy.logger.warn("The following command returned non-zero status: #{Shellwords.join(args)}")
+        KubernetesDeploy.logger.warn(err.shellescape)
+      end
+      [out.chomp, err, st]
+    end
+  end
+end
--- a/lib/kubernetes-deploy/kubernetes_resource.rb
+++ b/lib/kubernetes-deploy/kubernetes_resource.rb
@@ -115,13 +115,9 @@ module KubernetesDeploy
     def run_kubectl(*args)
       raise FatalDeploymentError, "Namespace missing for namespaced command" if namespace.blank?
       raise FatalDeploymentError, "Explicit context is required to run this command" if context.blank?
-      args = args.unshift("kubectl").push("--namespace=#{namespace}").push("--context=#{context}")
 
-      KubernetesDeploy.logger.debug Shellwords.join(args)
-      out, err, st = Open3.capture3(*args)
-      KubernetesDeploy.logger.debug(out.shellescape)
-      KubernetesDeploy.logger.debug("[ERROR] #{err.shellescape}") unless st.success?
-      [out.chomp, st]
+      out, _, st = KubernetesDeploy::KubectlWrapper.run_kubectl(*args, namespace: namespace, context: context)
+      [out, st]
     end
 
     def log_status
--- a/lib/kubernetes-deploy/runner.rb
+++ b/lib/kubernetes-deploy/runner.rb
@@ -279,7 +279,7 @@ MSG
     end
 
     def confirm_context_exists
-      out, err, st = run_kubectl("config", "get-contexts", "-o", "name", namespaced: false, with_context: false)
+      out, err, st = run_kubectl("config", "get-contexts", "-o", "name", with_namespace: false, with_context: false)
       available_contexts = out.split("\n")
       if !st.success?
         raise FatalDeploymentError, err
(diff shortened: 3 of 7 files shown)


lib/kubernetes-deploy/runner.rb
```diff
@@ -290,30 +290,23 @@ def confirm_context_exists
     end
 
     def confirm_namespace_exists
-      _, _, st = run_kubectl("get", "namespace", @namespace, namespaced: false)
+      _, _, st = run_kubectl("get", "namespace", @namespace, with_namespace: false)
```
sirupsen: Why rename this to `with_namespace`? `namespaced` seemed like a better name, `with_*` usually implies a block to me.

wfarr: Agreed. I'd expect something `with_namespace { ... }` with that sort of naming.

Maybe just `namespace`?

lib/kubernetes-deploy/kubernetes_resource.rb
```diff
@@ -115,13 +115,9 @@ def group_name
...
       raise FatalDeploymentError, "Namespace missing for namespaced command" if namespace.blank?
       raise FatalDeploymentError, "Explicit context is required to run this command" if context.blank?
-      args = args.unshift("kubectl").push("--namespace=#{namespace}").push("--context=#{context}")
 
-      KubernetesDeploy.logger.debug Shellwords.join(args)
-      out, err, st = Open3.capture3(*args)
-      KubernetesDeploy.logger.debug(out.shellescape)
-      KubernetesDeploy.logger.debug("[ERROR] #{err.shellescape}") unless st.success?
-      [out.chomp, st]
+      out, _, st = KubernetesDeploy::KubectlWrapper.run_kubectl(*args, namespace: namespace, context: context)
```
wfarr: Any reason we're preferring the fully resolved class name here?

wfarr: This line also uses different arguments to `.run_kubectl` than seen elsewhere in the change set

lib/kubernetes-deploy/runner.rb
kirs: Sometimes we want to run `kubectl` with a namespace flag and sometimes without it:

```
kubectl get pods --namespace blah # get pods of a namespace
kubectl get namespaces # don't need the namespace flag to get all namespaces
```

The same thing applies to `context`.

`with_context` and `with_namespace` take boolean values. I made them `with_` to differentiate that this is a opt-in/opt-out flag, and not the flag that tells what namespace you want to use. Otherwise developers may assume that the flag accepts the namespace name: `run_kubectl("get", "pods", namespace: "mynamespace")`

I'm open to rename them to something like `append_namespace` to address the concern about `with*` methods working with a block.
Does it make sense to you?

lib/kubernetes-deploy/kubernetes_resource.rb
kirs: > This line also uses different arguments to .run_kubectl than seen elsewhere in the change set

Not sure that I follow what you men

lib/kubernetes-deploy/runner.rb
KnVerey: > otherwise developers may assume that the flag accepts the namespace name: run_kubectl("get", "pods", namespace: "mynamespace")

Sure, if it were `namespace`, but it wasn't: it was `namespaced`, which sounds like it would accept a boolean and is still the best option imo.

lib/kubernetes-deploy/kubectl_wrapper.rb
```diff
@@ -0,0 +1,21 @@
...
+    def run_kubectl(*args, context: nil, namespace: nil)
+      args = args.unshift("kubectl")
+      args.push("--context=#{context}") if context
+      args.push("--namespace=#{namespace}") if namespace
+
+      KubernetesDeploy.logger.debug Shellwords.join(args)
+      out, err, st = Open3.capture3(*args)
+      KubernetesDeploy.logger.debug(out.shellescape)
+
+      unless st.success?
```
KnVerey: This will spew irrelevant warnings when the polling code uses it. For example, the return code on all 404s before the resources come up is not success. As I explained on the issue you linked to in the description:

> The original reason for having two different versions is a difference in error handling and logging. The runner version handles critical deploy commands that could fail, and both its callers and the end user should be interested in error output as a result. The resources on the other hand are doing polling, in which context typical !st.success? causes are transient or even expected, e.g. 404s on a GETs for resources that aren't up yet.

kirs:
> This contradicts the discussion in Shopify/cloudplatform#181 (comment) (related issue here: #53), re: using the kubeclient gem to replace shelling out during polling, since polling does not use any kubectl client features.

That's the long term plan, but today I'd still prefer having only one `run_kubectl` to support. Do you think that it would be easier to replace `KubernetesResource#run_kubectl` directly with the new kubeclient code?

KnVerey:
> Do you think that it would be easier to replace KubernetesResource#run_kubectl directly with the new kubeclient code?

No you're right, that will be much more involved than this PR is. I was thinking that centralizing things that we plan to take in different directions doesn't really make sense, but upon further reflection I don't see how it actually makes the migration any different in practice. Sorry for the reactionary comment. 🙏 

> we already got troubles in the past updating one place and not updating the other

I agree we should do what we can to make such troubles harder to enconter, but will these changes actually prevent that? It still looks possible to me to add a new flag to one `run_kubectl` (and the new version) but not the other, especially since the raising of FatalDeploymentErrors is still up to the duplicates. 

Just a suggestion: WDYT about making the central version a single-method module that both Runner and KubernetesResource can include? Something like:

```ruby
module RunKubectl
  # Runner includes this and uses it as-is
  def run_kubectl(*args, namespaced: true, with_context: true, log_warnings: true)
    if namespaced
      raise FatalDeploymentError, "Namespace missing for namespaced command" if @namespace.blank?
      args.push("--namespace=#{@namespace}")
    end

    if with_context
      raise FatalDeploymentError, "Explicit context is required to run this command" if @context.blank?
      args.push("--context=#{@context}")
    end
    KubernetesDeploy.logger.debug Shellwords.join(args)
    out, err, st = Open3.capture3(*args)
    KubernetesDeploy.logger.debug(out.shellescape)

    if !st.success? && log_warnings
      KubernetesDeploy.logger.warn("The following command failed: #{Shellwords.join(args)}")
      KubernetesDeploy.logger.warn(err)
    end
    [out.chomp, err.chomp, st]
  end
end

class KubernetesResource
  include RunKubectl

  def run_kubectl(*args)
    super(*args, namespaced: true, with_context: true, log_warnings: false)
  end
end
```

That'd be slightly more foolproof, I think? Normally I'm not a fan of having modules reference instance variables they're not actually in charge of, but in this case we're raising if they're used and blank, so seems pretty safe. 

Note: The exact thing written above would require the KubernetesResources to handle the `err` arg, which I'd be in favour of. Even though none of them actually need it, I kinda regret not making this consistent.

kirs:
I'm going to close this because I'd rather sit with duplicated `run_kubectl` code (before we move to  `kubeclient`) than introduce a mixin module that takes values from the instance variable.

PR #59, opened 2017-02-28
Create namespace if it does not exist

wfarr:
I thought about this when I was adding a new stack for an existing
service and found myself debating between using chat to initialize
a repository for a 5th time (it's deployed to as many clusters)
just for namespace creation, or to do it manually and out-of-band.
Neither seemed like great options to me, so I figured, why couldn't
we just create the namespace on the fly if it doesn't exist?

I don't think the tradeoff of requiring a namespace to already
exist necessarily makes sense any longer, even though in the
majority of cases at Shopify it will.

The initializer already enforces that a namespace is given,
so we can guarantee this receives a non-empty value.

This makes kubernetes-deploy a little more plug-and-play with,
IMO, minimal drawbacks.

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,7 +28,6 @@ Requirements:
 
  - kubectl 1.5.1+ binary must be available in your path
  - `ENV['KUBECONFIG']` must point to a valid kubeconfig file that includes the context you want to deploy to
- - The target namespace must already exist in the target context
  - `ENV['GOOGLE_APPLICATION_CREDENTIALS']` must point to the credentials for an authenticated service account if your user's auth provider is gcp
  - `ENV['ENVIRONMENT']` must be set to use the default template path (`config/deploy/$ENVIRONMENT`) in the absence of the `--template-dir=DIR` option
 
--- a/lib/kubernetes-deploy/runner.rb
+++ b/lib/kubernetes-deploy/runner.rb
@@ -84,7 +84,7 @@ MSG
 
       phase_heading("Identifying deployment target")
       confirm_context_exists
-      confirm_namespace_exists
+      ensure_namespace_exists
 
       phase_heading("Parsing deploy content")
       resources = discover_resources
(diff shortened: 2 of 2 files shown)


sirupsen:
What have we done in the past?

wfarr:
> What have we done in the past?

It currently will fail the deploy if the namespace does not exist.

sirupsen:
Yeah, but how have we created in the past? Manually? Some spy command? Script?

KnVerey:
> Yeah, but how have we created in the past? Manually? Some spy command? Script?

`spy cloud init` creates the namespace (and also drops an ejson keys secret in it). 

The downside is a significant decrease in control over app deployment to our tiers. I'm not sure I like the idea of making it _that_ easy for any app to create namespaces... they would effectively be able to do so in any valid tier. In the case of apps (which is what we should be optimizing this script for IMO), I think requiring explicit, external steps to initiate a namespace migration is perfectly reasonable. Presumably the thing you're deploying for the 5th time is one of our own controllers, which will be involved in the new-cluster deploy story we need to figure out.

@Shopify/cloudplatform I think we should discuss this as a team.

kirs:
@sirupsen for new namespace (e.g. staging) of existing project we created them with `kubectl create namespace walrusser` (same for undercloud)

I agree that it's a pain point right now when you create a staging stack and the namespace is missing, but at the same time I think we should be more strict about certain things and not automatically create the namespace.

If you intended to deploy to `tier4` where the namespace is ready for you but accidentally deployed to `tier1` where there's no such namespace, I'd prefer to be precautionary and fail early rather than automatically create the missing namespace.

JonPulsifer:
I'd like to start annotating these namespaces for network policies at some point, so maybe this shouldn't live here? food for thought and I rly don't care what happens in the mean time



e.g. 
```yaml
# When a namespace opts into isolation, all traffic directed
# at pods in the namespace is blocked unless explicitly allowed
# by configuration on a NetworkPolicy object
# this is arguably the k8s equiv of --icc=false

# this is configured at the namespace level
apiVersion: v1
kind: Namespace
metadata:
  name: my-new-namespace
  annotations:
    net.beta.kubernetes.io/network-policy: |
      {
        "ingress": {
          "isolation": "DefaultDeny"
        }
      }
```

KnVerey:
@JonPulsifer sounds like you'd be interested in https://github.com/Shopify/cloudplatform/issues/214

PR #60, opened 2017-02-28
Tighten timeout on deployment resource

kirs:
With the right rollout strategy, we should be ready to tighten the `Deployment` rollout. Right now it leads to deploys hanging up for 15 minutes and I'm sure that such timeout isn't necessary to find that something is not going as expected.

The new timeout is debatable and I wouldn't be surprised if at some point in future we get apps that are very slow to boot and take more than 5 minutes to roll out. However the `Shopify/shopify` monolith is a very large codebase to boot, and the rollout to hundreds of replicas manages to complete within 3 minutes.

Alternatively we can introduce a  configurable global waiting timeout as a CLI flag. This would unblock us to set it to `5.minutes` in Shipit because there we'd prefer to fail early rather than wait for 15 minutes and grow the deploy backlog.

thoughts? @KnVerey @sirupsen @wfarr

### DIFF
--- a/test/unit/kubernetes-deploy/kubernetes_resource_test.rb
+++ b/test/unit/kubernetes-deploy/kubernetes_resource_test.rb
@@ -1,0 +1,8 @@
+require 'test_helper'
+
+class KubernetesResourceTest < KubernetesDeploy::TestCase
+  def test_service_and_deployment_timeouts_are_equal
+    message = "Service and Deployment timeouts have to match since services are waiting to get endpoints from their backing deployments"
+    assert_equal KubernetesDeploy::Service.timeout, KubernetesDeploy::Deployment.timeout, message
+  end
+end
--- a/lib/kubernetes-deploy/kubernetes_resource/deployment.rb
+++ b/lib/kubernetes-deploy/kubernetes_resource/deployment.rb
@@ -1,6 +1,6 @@
 module KubernetesDeploy
   class Deployment < KubernetesResource
-    TIMEOUT = 15.minutes
+    TIMEOUT = 5.minutes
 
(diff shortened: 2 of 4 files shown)


KnVerey:
I'm in favour of this. We were being generous to start with the thought of accommodating early-stage apps doing crappy things, but in practice what I've seen is people waiting 15 minutes for doomed deploys to reach timeout on apps that normally deploy in <2m. 

You'll need to adjust the timeout on the `Service` class the same way; they should match, as services are basically waiting to get endpoints from their backing deployments. The `Pod` timeout doesn't strictly need to match, especially since we support deploying unmanaged pods (typically for running migrations), but it is probably too generous as well. Maybe 10 minutes for that one?

kirs:
> You'll need to adjust the timeout on the Service class the same way; they should match, as services are basically waiting to get endpoints from their backing deployments. 

Good to know! I've added a test to verify they match.

Also updated `Service` to 5 minutes and `Pod` to 10 minutes.

sirupsen:
Overriding this with annotations on the resource would be cool long-term, but this is good for now 👍

test/unit/kubernetes-deploy/kubernetes_resource_test.rb
```diff
@@ -0,0 +1,7 @@
+require 'test_helper'
+
+class KubernetesResourceTest < KubernetesDeploy::TestCase
+  def test_service_and_deployment_timeouts_are_equal
+    assert_equal KubernetesDeploy::Service.timeout, KubernetesDeploy::Deployment.timeout
```
KnVerey: Suggestion: make the failure message explain why this is a good idea (I'm not sure it is obvious)

Issue #43, opened 2017-02-27
Need to support a "restart" button in Shipit


KnVerey:
Right now, app maintainers need to push a new commit, triggering a new container build, to "restart" their application. We need a true restart mechanism, ideally enabling users to choose to restart only a specific deployment rather than all of them if desired.

Note that there has been some [discussion](https://github.com/kubernetes/kubernetes/issues/13488) re: adding a feature like this to kubectl. It's still open, but my read on the tl;dr there is that the maintainers are conceptually opposed to implementing it, since restarts should not be necessary when no changes have been made to the spec and both liveness and readiness probes are passing. 

As that k8s issue points out, rather than implementing selective pod deletion in accordance with the deployments rollout strategy, this can be achieved by patching the deployment's podspec, e.g. with an environment variable, a label or an annotation containing a timestamp.

kirs:
Initially I was thinking about the selective pod deletion (kind of `kubectl get pod | grep 'pod-name' | cut -d " " -f1 - | xargs -n1 -P 10 kubectl delete pod`), but you're right that we'd need to reimplement the rollout strategy. Patching the deployment spec with an annotation or label or env variable sounds more feasible here.

I also think it should be a separate command. It should accept the deployments we want to restart (web + jobs in our case) and the namespace + context:

`kubernetes-restart --context cloud1 --namespace trashbin --deployment jobs,web`

WDYT?

cc @ibawt

KnVerey:
> I also think it should be a separate command. It should accept the deployments we want to restart (web + jobs in our case) and the namespace + context:
> 
> kubernetes-restart --context cloud1 --namespace trashbin --deployment jobs,web

Sounds right to me. I would also like us to make the necessary changes on the shipit side so that any app using the first-class `kubernetes: ` syntax automatically gets a button for this task, with the ability to set the targeted deployments via env vars.

kirs:
I've done experiments and checked whether updating labels or annotations of a deployment triggers the restart.

```
EPOCH=$(date +%s)
kubectl patch deployment web -n trashbin \
  -p '{"metadata":{"labels":{"restarted_at":"'$EPOCH'"}}}'
```

This results to the deployment labels updated but no restart, which makes sense because only the metadata is changed.

```
RESTART_DATE=$(date -u '+%Y-%m-%dT%H:%M:%S%z')
kubectl patch deployment web -n trashbin \
[...]

Issue #51, opened 2017-02-27
Try to find a way around having two kubeclient helpers in tests


KnVerey:
https://github.com/Shopify/kubernetes-deploy/pull/24#discussion_r102930177

KnVerey:
The reason for needing multiple kubeclient helpers is that kubeclients are GV-specific and there's no higher-level method that selects a default GV for a kind based on discovery information. We'll probably want something like that if we do use kubeclient for resource retrieval, but for testing purposes I no longer think it's a big problem to have to use explicit clients.

PR #61, opened 2017-03-08
Restart task

kirs:
This is an implementation of the restart task according to what we discussed in the original issue (https://github.com/Shopify/kubernetes-deploy/issues/43)

TL;DR there's no way to tell Kubernetes to restart the pods and that's something that we need to have control over processed in case of an incident.

The hackish way to get the pods restarted is to update the deployment with a noop ENV variable, like `RESTARTED_AT` containing a timestamp. This can be done with a kubectl command:

```
RESTART_DATE=$(date -u '+%Y-%m-%dT%H:%M:%S%z')
kubectl patch deployment web -n trashbin \
  -p '{"spec":{"template":{"spec":{"containers":[{"name":"app","env":[{"name":"RESTARTED_AT","value":"'$RESTART_DATE'"}]}]}}}}'
```

This PR implements the restart task as a CLI tool. The usage:

```
# restarts the "web" deployment
kubernetes-deploy-restart <kube namespace> <kube context> --deployments=web
```

Under the hood it crafts the `PATCH` request to update the deployment, similar to what the kubectl command does.

fixes https://github.com/Shopify/kubernetes-deploy/issues/43

review @sirupsen @KnVerey

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -4,5 +4,5 @@ source 'https://rubygems.org'
 gemspec
 
 gem 'pry'
-gem 'kubeclient'
 gem 'rubocop'
+gem 'timecop'
--- a/README.md
+++ b/README.md
@@ -29,9 +29,16 @@ Requirements:
  - kubectl 1.5.1+ binary must be available in your path
  - `ENV['KUBECONFIG']` must point to a valid kubeconfig file that includes the context you want to deploy to
  - The target namespace must already exist in the target context
- - `ENV['GOOGLE_APPLICATION_CREDENTIALS']` must point to the credentials for an authenticated service account if your user's auth provider is gcp
+ - `ENV['GOOGLE_APPLICATION_CREDENTIALS']` must point to the credentials for an authenticated service account if your user's auth provider is GCP
  - `ENV['ENVIRONMENT']` must be set to use the default template path (`config/deploy/$ENVIRONMENT`) in the absence of the `--template-dir=DIR` option
 
+The tool also provides a task for restarting all of the pods in one or more deployments.
+It triggers the restart by touching the `RESTARTED_AT` environment variable in the deployment's podSpec.
+The rollout strategy defined for each deployment will be respected by the restart.
+
+The following command will restart all pods in the `web` and `jobs` deployments:
+
+`kubernetes-restart <kube namespace> <kube context> --deployments=web,jobs`
 
 ## Development
 
--- a/kubernetes-deploy.gemspec
+++ b/kubernetes-deploy.gemspec
@@ -21,6 +21,7 @@ Gem::Specification.new do |spec|
   spec.executables   = spec.files.grep(%r{^exe/}) { |f| File.basename(f) }
   spec.require_paths = ["lib"]
   spec.add_dependency "activesupport", ">= 4.2"
+  spec.add_dependency "kubeclient", "~> 2.3"
 
   spec.add_development_dependency "bundler", "~> 1.13"
   spec.add_development_dependency "rake", "~> 10.0"
--- a/lib/kubernetes-deploy/restart_task.rb
+++ b/lib/kubernetes-deploy/restart_task.rb
@@ -1,0 +1,130 @@
+# frozen_string_literal: true
+require 'kubernetes-deploy/kubeclient_builder'
+require 'kubernetes-deploy/ui_helpers'
+require 'kubernetes-deploy/resource_watcher'
+
+module KubernetesDeploy
+  class RestartTask
+    include UIHelpers
+    include KubernetesDeploy::KubeclientBuilder
+
+    class DeploymentNotFoundError < FatalDeploymentError
+      def initialize(name, namespace)
+        super("Deployment `#{name}` not found in namespace `#{namespace}`. Aborting the task.")
+      end
+    end
+
+    class NamespaceNotFoundError < FatalDeploymentError
+      def initialize(name, context)
+        super("Namespace `#{name}` not found in context `#{context}`. Aborting the task.")
+      end
+    end
+
+    class RestartError < FatalDeploymentError
+      def initialize(deployment_name, response)
+        super("Failed to restart #{deployment_name}. " \
+            "API returned non-200 response code (#{response.code})\n" \
+            "Response:\n#{response.body}")
+      end
+    end
+
+    HTTP_OK_RANGE = 200..299
+
+    def initialize(context:, namespace:, logger: KubernetesDeploy.logger)
+      @context = context
+      @namespace = namespace
+      @logger = logger
+      @kubeclient = build_v1_kubeclient(context)
+      @v1beta1_kubeclient = build_v1beta1_kubeclient(context)
+    end
+
+    def perform(deployments_names)
+      verify_namespace
+
+      if deployments_names.empty?
+        raise ArgumentError, "#perform takes at least one deployment to restart"
+      end
+
+      phase_heading("Triggering restart by touching ENV[RESTARTED_AT]")
+      deployments = fetch_deployments(deployments_names.uniq)
+      patch_kubeclient_deployments(deployments)
+
+      phase_heading("Waiting for rollout")
+      wait_for_rollout(deployments)
+
+      names = deployments.map { |d| "`#{d.metadata.name}`" }
+      @logger.info "Restart of #{names.join(', ')} deployments succeeded"
+    end
+
+    private
+
+    def wait_for_rollout(kubeclient_resources)
+      resources = kubeclient_resources.map { |d| Deployment.new(d.metadata.name, @namespace, @context, nil) }
+      watcher = ResourceWatcher.new(resources)
+      watcher.run
+    end
+
+    def verify_namespace
+      @kubeclient.get_namespace(@namespace)
+    rescue KubeException => error
+      if error.error_code == 404
+        raise NamespaceNotFoundError.new(@namespace, @context)
+      else
+        raise
+      end
+    end
+
+    def patch_deployment_with_restart(record)
+      @v1beta1_kubeclient.patch_deployment(
+        record.metadata.name,
+        build_patch_payload(record),
+        @namespace
+      )
+    end
+
+    def patch_kubeclient_deployments(deployments)
+      deployments.each do |record|
+        response = patch_deployment_with_restart(record)
+        if HTTP_OK_RANGE.cover?(response.code)
+          @logger.info "Triggered `#{record.metadata.name}` restart"
+        else
+          raise RestartError.new(record.metadata.name, response)
+        end
+      end
+    end
+
+    def fetch_deployments(list)
+      list.map do |name|
+        record = nil
+        begin
+          record = @v1beta1_kubeclient.get_deployment(name, @namespace)
+        rescue KubeException => error
+          if error.error_code == 404
+            raise DeploymentNotFoundError.new(name, @namespace)
+          else
+            raise
+          end
+        end
+        record
+      end
+    end
+
+    def build_patch_payload(deployment)
+      containers = deployment.spec.template.spec.containers
+      {
+        spec: {
+          template: {
+            spec: {
+              containers: containers.map do |container|
+                {
+                  name: container.name,
+                  env: [{ name: "RESTARTED_AT", value: Time.now.to_i.to_s }]
+                }
+              end
+            }
+          }
+        }
+      }
+    end
+  end
+end
--- a/test/integration/restart_task_test.rb
+++ b/test/integration/restart_task_test.rb
@@ -1,0 +1,127 @@
+# frozen_string_literal: true
+require 'test_helper'
+require 'kubernetes-deploy/restart_task'
+
+class RestartTaskTest < KubernetesDeploy::IntegrationTest
+  def test_restart
+    deploy_fixtures("hello-cloud", subset: ["configmap-data.yml", "web.yml.erb"])
+
+    refute fetch_restarted_at("web"), "no RESTARTED_AT env on fresh deployment"
+
+    restart = KubernetesDeploy::RestartTask.new(
+      context: KubeclientHelper::MINIKUBE_CONTEXT,
+      namespace: @namespace,
+    )
+    restart.perform(["web"])
+
+    assert_logs_match(/Triggered `web` restart/, 1)
+    assert_logs_match(/Restart of `web` deployments succeeded/, 1)
+
+    assert fetch_restarted_at("web"), "RESTARTED_AT is present after the restart"
+  end
+
+  def test_restart_twice
+    deploy_fixtures("hello-cloud", subset: ["configmap-data.yml", "web.yml.erb"])
+
+    refute fetch_restarted_at("web"), "no RESTARTED_AT env on fresh deployment"
+
+    restart = KubernetesDeploy::RestartTask.new(
+      context: KubeclientHelper::MINIKUBE_CONTEXT,
+      namespace: @namespace,
+    )
+    restart.perform(["web"])
+
+    assert_logs_match(/Triggered `web` restart/, 1)
+    assert_logs_match(/Restart of `web` deployments succeeded/, 1)
+
+    first_restarted_at = fetch_restarted_at("web")
+    assert first_restarted_at, "RESTARTED_AT is present after first restart"
+
+    Timecop.freeze(1.second.from_now) do
+      restart.perform(["web"])
+    end
+
+    second_restarted_at = fetch_restarted_at("web")
+    assert second_restarted_at, "RESTARTED_AT is present after second restart"
+    refute_equal first_restarted_at.value, second_restarted_at.value
+  end
+
+  def test_restart_with_same_resource_twice
+    deploy_fixtures("hello-cloud", subset: ["configmap-data.yml", "web.yml.erb"])
+
+    refute fetch_restarted_at("web"), "no RESTARTED_AT env on fresh deployment"
+
+    restart = KubernetesDeploy::RestartTask.new(
+      context: KubeclientHelper::MINIKUBE_CONTEXT,
+      namespace: @namespace,
+    )
+    restart.perform(%w(web web))
+
+    assert_logs_match(/Triggered `web` restart/, 1)
+    assert_logs_match(/Restart of `web` deployments succeeded/, 1)
+
+    assert fetch_restarted_at("web"), "RESTARTED_AT is present after the restart"
+  end
+
+  def test_restart_not_existing_deployment
+    restart = KubernetesDeploy::RestartTask.new(
+      context: KubeclientHelper::MINIKUBE_CONTEXT,
+      namespace: @namespace,
+    )
+    assert_raises(KubernetesDeploy::RestartTask::DeploymentNotFoundError) do
+      restart.perform(["web"])
+    end
+  end
+
+  def test_restart_one_not_existing_deployment
+    restart = KubernetesDeploy::RestartTask.new(
+      context: KubeclientHelper::MINIKUBE_CONTEXT,
+      namespace: @namespace,
+    )
+    assert_raises(KubernetesDeploy::RestartTask::DeploymentNotFoundError) do
+      restart.perform(%w(walrus web))
+    end
+
+    deploy_fixtures("hello-cloud", subset: ["configmap-data.yml", "web.yml.erb"])
+    refute fetch_restarted_at("web"), "no RESTARTED_AT env after failed restart task"
+  end
+
+  def test_restart_none
+    restart = KubernetesDeploy::RestartTask.new(
+      context: KubeclientHelper::MINIKUBE_CONTEXT,
+      namespace: @namespace,
+    )
+    assert_raises(ArgumentError) do
+      restart.perform([])
+    end
+  end
+
+  def test_restart_not_existing_context
+    assert_raises(KubernetesDeploy::KubeclientBuilder::ContextMissingError) do
+      KubernetesDeploy::RestartTask.new(
+        context: "walrus",
+        namespace: @namespace,
+      )
+    end
+  end
+
+  def test_restart_not_existing_namespace
+    restart = KubernetesDeploy::RestartTask.new(
+      context: KubeclientHelper::MINIKUBE_CONTEXT,
+      namespace: "walrus"
+    )
+    error = assert_raises(KubernetesDeploy::RestartTask::NamespaceNotFoundError) do
+      restart.perform(["web"])
+    end
+    assert_equal "Namespace `walrus` not found in context `minikube`. Aborting the task.", error.to_s
+  end
+
+  private
+
+  def fetch_restarted_at(deployment_name)
+    deployment = v1beta1_kubeclient.get_deployment(deployment_name, @namespace)
+    containers = deployment.spec.template.spec.containers
+    assert_equal 1, containers.size
+    containers.first.env.find { |n| n.name == "RESTARTED_AT" }
+  end
+end
--- a/test/test_helper.rb
+++ b/test/test_helper.rb
@@ -3,6 +3,7 @@ $LOAD_PATH.unshift File.expand_path('../../lib', __FILE__)
 require 'kubernetes-deploy'
 require 'kubeclient'
 require 'pry'
+require 'timecop'
 require 'minitest/autorun'
 
 require 'helpers/kubeclient_helper'
@@ -25,9 +26,14 @@ module KubernetesDeploy
       @logger_stream.close
     end
 
-    def assert_logs_match(regexp)
+    def assert_logs_match(regexp, times = nil)
       @logger_stream.rewind
-      assert_match regexp, @logger_stream.read
+      if times
+        count = @logger_stream.read.scan(regexp).count
+        assert_equal 1, count, "Expected #{regexp} to appear #{times} time(s) in the log, but appeared #{count} times"
+      else
+        assert_match regexp, @logger_stream.read
+      end
     end
   end
 
--- a/exe/kubernetes-restart
+++ b/exe/kubernetes-restart
@@ -1,0 +1,23 @@
+#!/usr/bin/env ruby
+# frozen_string_literal: true
+
+require 'optparse'
+
+require 'kubernetes-deploy'
+require 'kubernetes-deploy/restart_task'
+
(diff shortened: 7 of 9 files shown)


exe/kubernetes-deploy-restart
```diff
@@ -0,0 +1,27 @@
+#!/usr/bin/env ruby
```
sirupsen: What was the reason you decided to make a separate executable for this?

lib/kubernetes-deploy/restart_task.rb
```diff
@@ -0,0 +1,80 @@
...
+      end
+    end
+
+    private
+
+    def build_v1beta1_kubeclient(context)
+      build_kubeclient(
+        api_version: "v1beta1",
+        context: context,
+        endpoint_path: "/apis/extensions/"
```
sirupsen: Aren't we always setting `/apis/extensions` when `api_version == 'v1beta1`? Can we do this automatically?

lib/kubernetes-deploy/restart_task.rb
```diff
@@ -0,0 +1,80 @@
...
+        record.metadata.name,
+        build_patch_payload(record),
+        @namespace
+      )
+    end
+
+    def fetch_deployments(list)
+      list.map do |name|
+        records = @v1beta1_kubeclient.get_deployments(label_selector: "name=#{name}", namespace: @namespace)
+        raise "too many records: #{records.size}" if records.size > 1
```
sirupsen: This won't be unreasonably common at some point, right? Could we build some helpers to iterate through all the YAML segments?

lib/kubernetes-deploy/restart_task.rb
```diff
@@ -0,0 +1,80 @@
...
+        records.first
+      end
+    end
+
+    def build_patch_payload(deployment)
+      containers = deployment.spec.template.spec.containers
+      if containers.size > 1
+        raise "don't know how to restart > 1 containers"
+      end
```
sirupsen: Same comment here as above, we should consider supporting this out the gate—it's going to take more time for someone to discover we don't need this, and add it in, than doing it now or as a follow-up I think 🤔

lib/kubernetes-deploy/restart_task.rb
```diff
@@ -0,0 +1,80 @@
...
+
+      {
+        spec: {
+          template: {
+            spec: {
+              containers: containers.map do |container|
+                {
+                  name: container.name,
+                  env: [{ name: "RESTARTED_AT", value: Time.now.to_i.to_s }]
```
sirupsen: Someone mentioned we can do this with annotations too, why did you choose environment variables?

lib/kubernetes-deploy/restart_task.rb
```diff
@@ -0,0 +1,80 @@
...
+      if deployments.empty?
+        raise ArgumentError, "#perform takes at least one deployment to restart"
+      end
+
+      fetch_deployments(deployments).each do |record|
+        response = patch_deployment_with_restart(record)
+        if response.code == 200
+          @logger.info "Restarted #{record.metadata.name.inspect}"
+        else
+          @logger.fatal "Failed to restart #{record.metadata.name} (API returned non-200 response)"
```
sirupsen: Could you write the status code?

lib/kubernetes-deploy/restart_task.rb
```diff
@@ -0,0 +1,80 @@
...
+    end
+
+    def perform(deployments)
+      if deployments.empty?
+        raise ArgumentError, "#perform takes at least one deployment to restart"
+      end
+
+      fetch_deployments(deployments).each do |record|
+        response = patch_deployment_with_restart(record)
+        if response.code == 200
```
sirupsen: I would've accepted a 202 to be honest. 🤔

test/integration/restart_test.rb
```diff
@@ -0,0 +1,37 @@
+require 'test_helper'
+require 'kubernetes-deploy/restart_task'
+
+class KubernetesDeployTest < KubernetesDeploy::IntegrationTest
+  def test_restart
+    deploy_fixtures("basic", subset: ["configmap-data.yml", "web.yml.erb"])
```
sirupsen: Why did you need the configmap?

test/integration/restart_test.rb
```diff
@@ -0,0 +1,37 @@
...
+
+  private
+
+  def fetch_restarted_at(deployment)
+    containers = deployment.spec.template.spec.containers
+    assert_equal 1, containers.size
+    containers.first.env.find { |n| n.name == "RESTARTED_AT" }
+  end
+
+  def fetch_deployment(name)
```
sirupsen: This seems like something we should write a or already have a helper for, no?

test/integration/restart_test.rb
```diff
@@ -0,0 +1,37 @@
+require 'test_helper'
+require 'kubernetes-deploy/restart_task'
+
+class KubernetesDeployTest < KubernetesDeploy::IntegrationTest
+  def test_restart
```
sirupsen: Could we get an error test case as well, or is that too hard to reproduce?

exe/kubernetes-deploy-restart
kirs: Because it has completely different arguments and semantics, and it's more like a helper that we want to provide out of box.

lib/kubernetes-deploy/restart_task.rb
kirs: That's actually an exceptional case (two deployments with name `web` in the same namespace). I'm not sure that can ever happen in production, but I'd prefer to fail rather than proceed.

lib/kubernetes-deploy/restart_task.rb
kirs: I've looked into that a couple days ago and altering annotations or labels does not trigger the restart. Touching ENV looks like the only option. More details in https://github.com/Shopify/kubernetes-deploy/issues/43#issuecomment-284500937

test/integration/restart_test.rb
kirs: Because `web` from the fixtures is referring to config map values. That's how it works in all tests

lib/kubernetes-deploy/restart_task.rb
kirs: I'll change that to 2**

lib/kubernetes-deploy/restart_task.rb
kirs: No, because deployments are part of v1beta1 API and the rest is API v1.

kirs: Ideally the API wrapper from the gem could do it automatically but it doesn't

test/integration/restart_test.rb
kirs: That's kind of hard to reproduce. I should have mentioned it as the reason why I didn't add an error case.

Gemfile
```diff
@@ -4,5 +4,4 @@ source 'https://rubygems.org'
 gemspec
 
 gem 'pry'
-gem 'kubeclient'
```
kirs: This is now a gemspec dependency.

test/integration/restart_task_test.rb
```diff
@@ -0,0 +1,37 @@
...
+      namespace: @namespace,
+      logger: @logger
+    )
+    restart.perform(["web"])
+
+    assert_logs_match(/Restarted "web"/)
+
+    deployment = fetch_deployment("web")
+    assert fetch_restarted_at(deployment), "RESTARTED_AT is present after the restart"
+  end
```
sirupsen: Is there any helper to test restarting multiple? What happens if I pass `web` twice?

lib/kubernetes-deploy/restart_task.rb
```diff
@@ -0,0 +1,78 @@
...
+      @logger = logger
+      @v1beta1_kubeclient = build_v1beta1_kubeclient(context)
+    end
+
+    def perform(deployments)
+      if deployments.empty?
+        raise ArgumentError, "#perform takes at least one deployment to restart"
+      end
+
+      fetch_deployments(deployments).each do |record|
```
sirupsen: Does this mean each deployment is restarted in serial? How long does that take? Won't this only give the status that it's _starting_ to restart, not that it's actually successfully restarted all, say, 10,000 pods?

kirs: You're right, I should have set the message to `triggering restart of "web"` because it doesn't way for the restart to finish. The pods are recreated according to the rollout strategy that every project has defined.

Do you think the tool needs the `--wait` flag to wait for the restart? I'm not sure about that because that's something that can be always watched in kubernetes dashboard.

test/integration/restart_task_test.rb
kirs: Good point. I'll add tests for that 👍

lib/kubernetes-deploy/restart_task.rb
sirupsen: If it's hard to build, it's probably not worth it—but most devs would have no idea to look there and probably won't even have access. It would feel like you should be able to use the same code as we do for deploys to do this without too much trouble, but I'm not sure. I'll leave this up to you.

lib/kubernetes-deploy/restart_task.rb
KnVerey: Related: https://github.com/Shopify/kubernetes-deploy/issues/51 (but I haven't looked into it)

README.md
```diff
@@ -29,9 +29,13 @@ Requirements:
  - kubectl 1.5.1+ binary must be available in your path
  - `ENV['KUBECONFIG']` must point to a valid kubeconfig file that includes the context you want to deploy to
  - The target namespace must already exist in the target context
- - `ENV['GOOGLE_APPLICATION_CREDENTIALS']` must point to the credentials for an authenticated service account if your user's auth provider is gcp
+ - `ENV['GOOGLE_APPLICATION_CREDENTIALS']` must point to the credentials for an authenticated service account if your user's auth provider is GCP
  - `ENV['ENVIRONMENT']` must be set to use the default template path (`config/deploy/$ENVIRONMENT`) in the absence of the `--template-dir=DIR` option
 
+The tool also provides the task to restart all pods of a deployment by touching the ENV variable.
+The following command will fetch `web` and `jobs` deployments are make all pods to restart according to the rollout strategy:
```
KnVerey: Wording nitpicks:

> The tool also provides a task for restarting all of the pods in one or more deployments. It triggers the restart by touching the `RESTARTED_AT` environment variable in the deployment's podSpec. The rollout strategy defined for each deployment will be respected by the restart.

> The following command will restart all pods in the `web` and `jobs` deployments:

exe/kubernetes-deploy-restart
KnVerey: I agree with this being a separate executable

exe/kubernetes-deploy-restart
```diff
@@ -0,0 +1,27 @@
...
+require 'kubernetes-deploy/restart_task'
+
+raw_deployments = nil
+ARGV.options do |opts|
+  opts.on("--deployments=LIST") { |v| raw_deployments = v }
+  opts.parse!
+end
+
+if raw_deployments.nil?
+  puts "Failed: specify at least one deployment to restart with --deployments flag"
```
KnVerey: Very minor, but should we initialize the logger first and `logger.fatal` here?

KnVerey: And related to this, could this task use `KubernetesDeploy.logger` for output consistency with regular deploys, and sensitivity to `ENV["LEVEL"]` and `ENV["DEBUG"]`? It's in `lib/kubernetes-deploy.rb`, so it's actually already require'd.

lib/kubernetes-deploy/restart_task.rb
KnVerey: Should this helper also move to `kubernetes-deploy/kubeclient_builder`? The tests could use it too, and some of the Kubernetes Resource classes will probably need the same when we convert them.

lib/kubernetes-deploy/restart_task.rb
```diff
@@ -0,0 +1,78 @@
...
+      @v1beta1_kubeclient.patch_deployment(
+        record.metadata.name,
+        build_patch_payload(record),
+        @namespace
+      )
+    end
+
+    def fetch_deployments(list)
+      list.map do |name|
+        records = @v1beta1_kubeclient.get_deployments(label_selector: "name=#{name}", namespace: @namespace)
```
KnVerey: I think this code isn't handling the case where the deployment name is bad / not found for whatever reason. In any case I think we should use `@v1beta1_kubeclient.get_deployment(name, @namespace)` and possibly capture/re-raise the `KubeException` we'd get on 404 (if the resulting message would suck otherwise). Although Shopify's templates label all resources with a name label matching the resource name by convention, we don't actually control the templates, so we shouldn't rely on it (it's fine in tests since we do control those).

exe/kubernetes-deploy-restart
```diff
@@ -0,0 +1,27 @@
...
+  exit 1
+end
+
+logger = Logger.new(STDERR)
+logger.level = Logger::INFO
+logger.formatter = ->(severity, _datetime, _progname, msg) do
+  "[#{severity}] #{msg}\n"
+end
+
+restart = KubernetesDeploy::RestartTask.new(namespace: ARGV[0], context: ARGV[1], logger: logger)
```
KnVerey: Looks like if the context is invalid, we'll raise `"#{context} context must be configured in your KUBECONFIG (#{ENV['KUBECONFIG']}). "` in the client builder, but what does the failure look like when the namespace arg is either missing or invalid?

test/integration/restart_task_test.rb
KnVerey: Can we also get tests for:
- no name passed at all
- one good, one bad name passed

exe/kubernetes-deploy-restart
kirs: 👍 I'll use KubernetesDeploy.logger

Regarding the `raw_deployments` check: IMO initialization errors are different from the runtime errors that are logged with `fatal` status

README.md
kirs: Thanks!

kirs:
Thanks for the review; all the point have been really helpful. I've updated the PR and addresses all the things, including the watch for rollout and extensive tests.

lib/kubernetes-deploy/restart_task.rb
```diff
@@ -0,0 +1,121 @@
...
+
+      wait_for_rollout(deployments)
+      deployments.each do |record|
+        @logger.info "Successfully restarted #{record.metadata.name.inspect}"
+      end
+    end
+
+    private
+
+    def wait_for_rollout(deployments)
```
kirs: @KnVerey had to copypaste majority of the method to handle waiting for `Deployment` rollout

kirs: But on the bright side I didn't need to modify the resource

sirupsen: I'm a little uncomfortable with the copy-pasting here from the runner. There's some inconsistent vocabulary here in that for the deploy script, we call the class `Runner` but for restart we call it `RestartTask`. If we use the `Task` paradigm, we could rename `Runner` to `DeployTask` and then have a `Task` class they both inherit from that defines `wait_for_completion`.

lib/kubernetes-deploy/restart_task.rb
```diff
@@ -0,0 +1,121 @@
...
+      @v1beta1_kubeclient = build_v1beta1_kubeclient(context)
+    end
+
+    def perform(deployments_names)
+      if deployments_names.empty?
+        raise ArgumentError, "#perform takes at least one deployment to restart"
+      end
+      deployments = fetch_deployments(deployments_names.uniq)
+      deployments.each do |record|
+        response = patch_deployment_with_restart(record)
```
sirupsen: This is some of the first stuff that uses the API for mutation in this gem. Is that because there is a goal to move to the API completely long-term? Is there an API for, say, `kubectl apply`?

test/integration/restart_task_test.rb
```diff
@@ -0,0 +1,108 @@
+# frozen_string_literal: true
+require 'test_helper'
+require 'kubernetes-deploy/restart_task'
+
+class RestartTaskTest < KubernetesDeploy::IntegrationTest
+  def test_restart
+    deploy_fixtures("basic", subset: ["configmap-data.yml", "web.yml.erb"])
+
+    deployment = fetch_deployment("web")
+    refute fetch_restarted_at(deployment), "no RESTARTED_AT env on fresh deployment"
```
sirupsen: You have these two lines in a few places here `assert_fetched_at deployment("web")` might be convenient and a little more readable 👀

test/integration/restart_task_test.rb
```diff
@@ -0,0 +1,108 @@
...
+
+    restart = KubernetesDeploy::RestartTask.new(
+      context: KubeclientHelper::MINIKUBE_CONTEXT,
+      namespace: @namespace,
+    )
```
sirupsen: Perhaps you could instantiate this in `setup` as `@task`?

test/integration/restart_task_test.rb
```diff
@@ -0,0 +1,108 @@
...
+    assert_raises(ArgumentError) do
+      restart.perform([])
+    end
+  end
+
+  def test_restart_not_existing_context
+    assert_raises(KubernetesDeploy::KubeclientBuilder::ContextMissignError) do
```
sirupsen: There's a typo in `ContextMissgn` 😱

kubernetes-deploy.gemspec
```diff
@@ -21,6 +21,7 @@ Gem::Specification.new do |spec|
   spec.executables   = spec.files.grep(%r{^exe/}) { |f| File.basename(f) }
   spec.require_paths = ["lib"]
   spec.add_dependency "activesupport", ">= 4.2"
+  spec.add_dependency "kubeclient", "~> 2.3"
```
sirupsen: I suspect this version is different from what we used before. When I clone down this branch to attempt to run the tests I get:

```
macbookbro kubernetes-deploy (restart-task) $ rt test/integration/restart_task_test.rb 
Run options: --seed 51416

# Running:

/Users/sirup/src/github.com/Shopify/kubernetes-deploy/test/test_helper.rb:65:in `create_namespace': uninitialized constant Kubeclient::Namespace (NameError)
	from /Users/sirup/src/github.com/Shopify/kubernetes-deploy/test/test_helper.rb:57:in `claim_namespace'
	from /Users/sirup/src/github.com/Shopify/kubernetes-deploy/test/test_helper.rb:44:in `run'
	from /Users/sirup/src/github.com/Shopify/kubernetes-deploy/vendor/bundle/gems/minitest-5.10.1/lib/minitest.rb:822:in `run_one_method'
	from /Users/sirup/src/github.com/Shopify/kubernetes-deploy/vendor/bundle/gems/minitest-5.10.1/lib/minitest.rb:323:in `run_one_method'
	from /Users/sirup/src/github.com/Shopify/kubernetes-deploy/vendor/bundle/gems/minitest-5.10.1/lib/minitest.rb:310:in `block (2 levels) in run'
	from /Users/sirup/src/github.com/Shopify/kubernetes-deploy/vendor/bundle/gems/minitest-5.10.1/lib/minitest.rb:309:in `each'
	from /Users/sirup/src/github.com/Shopify/kubernetes-deploy/vendor/bundle/gems/minitest-5.10.1/lib/minitest.rb:309:in `block in run'
	from /Users/sirup/src/github.com/Shopify/kubernetes-deploy/vendor/bundle/gems/minitest-5.10.1/lib/minitest.rb:349:in `on_signal'
	from /Users/sirup/src/github.com/Shopify/kubernetes-deploy/vendor/bundle/gems/minitest-5.10.1/lib/minitest.rb:336:in `with_info_handler'
	from /Users/sirup/src/github.com/Shopify/kubernetes-deploy/vendor/bundle/gems/minitest-5.10.1/lib/minitest.rb:308:in `run'
	from /Users/sirup/src/github.com/Shopify/kubernetes-deploy/vendor/bundle/gems/minitest-5.10.1/lib/minitest.rb:158:in `block in __run'
	from /Users/sirup/src/github.com/Shopify/kubernetes-deploy/vendor/bundle/gems/minitest-5.10.1/lib/minitest.rb:158:in `map'
	from /Users/sirup/src/github.com/Shopify/kubernetes-deploy/vendor/bundle/gems/minitest-5.10.1/lib/minitest.rb:158:in `__run'
	from /Users/sirup/src/github.com/Shopify/kubernetes-deploy/vendor/bundle/gems/minitest-5.10.1/lib/minitest.rb:135:in `run'
	from /Users/sirup/src/github.com/Shopify/kubernetes-deploy/vendor/bundle/gems/minitest-5.10.1/lib/minitest.rb:62:in `block in autorun'
```

sirupsen: Could you try to blow away `vendor`, `bundle install`, and ensure it still works for you?

test/integration/restart_task_test.rb
```diff
@@ -0,0 +1,108 @@
+# frozen_string_literal: true
+require 'test_helper'
+require 'kubernetes-deploy/restart_task'
+
+class RestartTaskTest < KubernetesDeploy::IntegrationTest
```
sirupsen: Good test coverage! Suggestions for two more tests:

* `wait_for_rollout` fails to see the containers come back up (perhaps test this by making them sleep for too long?)
* Restarting after having already restarted, i.e. `RESTARTED_AT` is already set.

test/test_helper.rb
```diff
@@ -28,6 +28,12 @@ def assert_logs_match(regexp)
       @logger_stream.rewind
       assert_match regexp, @logger_stream.read
     end
+
+    def assert_logs_match_once(regexp)
+      @logger_stream.rewind
+      count = @logger_stream.read.scan(regexp).count
+      assert_equal 1, count, "Expected #{regexp} to appear only once in the log, appear #{count} times"
+    end
```
sirupsen: Instead of introducing a new helper, I think you should consider adding a `times` argument to `assert_logs_match`. That's more similar to helpers we have elsewhere internally :)

test/integration/restart_task_test.rb
```diff
@@ -0,0 +1,108 @@
...
+      )
+    end
+  end
+
+  def test_restart_not_existing_namespace
+    restart = KubernetesDeploy::RestartTask.new(
+      context: KubeclientHelper::MINIKUBE_CONTEXT,
+      namespace: "walrus"
+    )
+    assert_raises(KubernetesDeploy::RestartTask::DeploymentNotFoundError) do
```
KnVerey: Hm, I think the distinction between the namespace not existing and the deployment not existing is very useful to the user. Can we validate the namespace before attempting to access the deployment so we can make that distinction?

lib/kubernetes-deploy/restart_task.rb
kirs: The way to restart with `kubectl` is:

```
export RESTART_DATE=$(date -u '+%Y-%m-%dT%k:%M:%S%z')
kubectl patch deployment web -n trashbin \
  -p '{"spec":{"template":{"spec":{"containers":[{"name":"app","env":[{"name":"RESTART","value":"'$RESTART_DATE'"}]}]}}}}'
```

The long-term plan is to move to the API, that's the reason why I didn't use `kubectl` here.

test/integration/restart_task_test.rb
kirs: I didn't want to put this to `setup` because the arguments to `RestartTask.new` are different across test cases.

test/test_helper.rb
kirs: 👍

README.md
```diff
@@ -29,9 +29,16 @@ Requirements:
...
  - The target namespace must already exist in the target context
- - `ENV['GOOGLE_APPLICATION_CREDENTIALS']` must point to the credentials for an authenticated service account if your user's auth provider is gcp
+ - `ENV['GOOGLE_APPLICATION_CREDENTIALS']` must point to the credentials for an authenticated service account if your user's auth provider is GCP
  - `ENV['ENVIRONMENT']` must be set to use the default template path (`config/deploy/$ENVIRONMENT`) in the absence of the `--template-dir=DIR` option
 
+The tool also provides a task for restarting all of the pods in one or more deployments.
+It triggers the restart by touching the `RESTARTED_AT` environment variable in the deployment's podSpec.
+The rollout strategy defined for each deployment will be respected by the restart.
+
+The following command will restart all pods in the `web` and `jobs` deployments:<Paste>
```
kirs: My vim thing. Thanks for spotting it!

test/integration/restart_task_test.rb
kirs: Sure 👍

kirs:
It's ready for another review.
The CI is green: https://buildkite.com/shopify/kubernetes-deploy-gem/builds/100#ce42d94a-e1cd-4bc3-8f50-e9308f10750d

README.md
```diff
@@ -29,9 +29,16 @@ Requirements:
...
+ - `ENV['GOOGLE_APPLICATION_CREDENTIALS']` must point to the credentials for an authenticated service account if your user's auth provider is GCP
  - `ENV['ENVIRONMENT']` must be set to use the default template path (`config/deploy/$ENVIRONMENT`) in the absence of the `--template-dir=DIR` option
 
+The tool also provides a task for restarting all of the pods in one or more deployments.
+It triggers the restart by touching the `RESTARTED_AT` environment variable in the deployment's podSpec.
+The rollout strategy defined for each deployment will be respected by the restart.
+
+The following command will restart all pods in the `web` and `jobs` deployments:
+
+`kubernetes-deploy-restart <kube namespace> <kube context> --deployments=web,jobs`
```
KnVerey: Naming nit: I noticed @xldenis's WIP names that task  `kubernetes-run`... wdyt about this being `kubernetes-restart`?

kubernetes-deploy.gemspec
KnVerey: fwiw I just pulled the latest version of this branch and was able to run the tests fine.

kirs:
Great suggestion. `kubernetes-restart` makes more sense 👍

kirs:
This should unblock @xldenis to continue his work.

PR #62, opened 2017-03-09
Allow overriding timeout on command line

DazWorrall:
We are using kubernetes-deploy to roll out buildkite agents for CI, and would like a higher timeout value than the defaults to allow enough time for jobs to complete (matching scrooges behaviour for now). I didn't find much unit testing around this code so as it's pretty simple I've just tested locally:

```
$ kubernetes-deploy --template-dir config/deploy/testing/ buildkite-testing ci
...
Waiting for Deployment/buildkite-runner with 300s timeout
...
$ kubernetes-deploy --timeout 800 --template-dir config/deploy/testing/ buildkite-testing ci
...
Waiting for Deployment/buildkite-runner with 800s timeout
```

/cc @Shopify/pipeline

### DIFF
--- a/exe/kubernetes-deploy
+++ b/exe/kubernetes-deploy
@@ -10,6 +10,7 @@ template_dir = nil
 ARGV.options do |opts|
   opts.on("--skip-wait") { skip_wait = true }
   opts.on("--template-dir=DIR") { |v| template_dir = v }
+  opts.on("--timeout=TIMEOUT") { |v| KubernetesDeploy::KubernetesResource::timeout_override = v.to_i }
   opts.parse!
 end
 
--- a/lib/kubernetes-deploy.rb
+++ b/lib/kubernetes-deploy.rb
@@ -3,6 +3,7 @@ require 'active_support/core_ext/object/blank'
 require 'active_support/core_ext/hash/slice'
 require 'active_support/core_ext/numeric/time'
 require 'active_support/core_ext/string/inflections'
+require 'active_support/core_ext/class/attribute'
 
 require 'logger'
 require 'kubernetes-deploy/runner'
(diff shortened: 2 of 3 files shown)


exe/kubernetes-deploy
```diff
@@ -10,6 +10,7 @@ template_dir = nil
 ARGV.options do |opts|
   opts.on("--skip-wait") { skip_wait = true }
   opts.on("--template-dir=DIR") { |v| template_dir = v }
+  opts.on("--timeout=TIMEOUT") { |v| KubernetesDeploy::KubernetesResource::OVERRIDDEN_TIMEOUT = v.to_i }
```
wvanbergen: Anything capitalized in Ruby is a constant. This will override the constant, which will cause warnings and is bad practice. I would simply use a instance or class variable

wvanbergen: Sorry for some reason this comment didn't make it into the review.

casperisfine: Yes you want to store it in a class attribute, like `logger` in `KubernetesResource`.

KnVerey:
Would you mind providing more details (in Cloudplatform team's Slack channel if details are private) on what the rollout strategy is for these deployments and why they're expected to take 13min to complete? We set the individual resource limits in a fairly arbitrary 👋  manner, so I'm really curious about legitimate use cases for long timeouts. It wouldn't necessarily be out of the question to raise the limit on deployments.

Also, what environment will you be running this from? I see you tested it locally, but would you need this global timeout feature to work with the Shipit integration? (We might want to make it compatible even if you don't need it.)

DazWorrall:
> Would you mind providing more details on what the rollout strategy is for these deployments and why they're expected to take 13min to complete?

Not necessarily 13 minutes, just more than 5 :) Reached out over Slack with details.

> Also, what environment will you be running this from? I see you tested it locally, but would you need this global timeout feature to work with the Shipit integration? (We might want to make it compatible even if you don't need it.)

Yes, I was planning to make a PR for Shipit if/once this is acceptable.

KnVerey:
For the record, we concluded that the need was to be able to change the timeout more than to have it be statically higher, and that the better long-term solution for this would be to enable such an override for individual resources via annotations, as suggested in https://github.com/Shopify/kubernetes-deploy/pull/60.

PR #64, opened 2017-03-10
Fix rubocop violations in runner.rb

kirs:
review @KnVerey


KnVerey:
All violations on the repo have since been fixed.

Issue #50, opened 2017-02-27
Rename the main fixture


KnVerey:
"Basic" -> "Hello Cloud"

PR #67, opened 2017-03-16
Rename 'basic' fixture 'hello-cloud'

KnVerey:
Fixes https://github.com/Shopify/kubernetes-deploy/issues/50

### DIFF
--- a/test/integration/kubernetes_deploy_test.rb
+++ b/test/integration/kubernetes_deploy_test.rb
@@ -2,44 +2,44 @@
 require 'test_helper'
 
 class KubernetesDeployTest < KubernetesDeploy::IntegrationTest
-  def test_full_basic_set_deploy_succeeds
-    deploy_fixtures("basic")
-    basic = FixtureSetAssertions::Basic.new(@namespace)
-    basic.assert_all_up
+  def test_full_hello_cloud_set_deploy_succeeds
+    deploy_fixtures("hello-cloud")
+    hello_cloud = FixtureSetAssertions::HelloCloud.new(@namespace)
+    hello_cloud.assert_all_up
   end
 
   def test_partial_deploy_followed_by_full_deploy
-    deploy_fixtures("basic", subset: ["configmap-data.yml", "redis.yml"])
-    basic = FixtureSetAssertions::Basic.new(@namespace)
-    basic.assert_all_redis_resources_up
-    basic.assert_configmap_data_present
-    basic.refute_managed_pod_exists
-    basic.refute_web_resources_exist
+    deploy_fixtures("hello-cloud", subset: ["configmap-data.yml", "redis.yml"])
+    hello_cloud = FixtureSetAssertions::HelloCloud.new(@namespace)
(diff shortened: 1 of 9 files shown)


kirs:
There's a failing test; is it the one we detected as flaky?

KnVerey:
> There's a failing test; is it the one we detected as flaky?

It's not the only one that has failed on CI, but looking over a number of builds, it does seem to be the primary offender, yes.

KnVerey:
Hm, that failure suggests it manages to get into a state where there's only one pod running, and it is the one from the second of the two deploys in the test. If the pods from the first deploy were killed in the expected way, they would not be cleaned up... my guess is that something before the container start (e.g. image pull) is taking too long, so the pod doesn't have a chance to reach its timeout before the second deploy starts. I'll try replacing the sleep with a poll for at least one failed pod.

KnVerey:
Push a fix attempt in the last commit, and it passed on this run anyway 😅 
Quick 👀  on that?

test/integration/kubernetes_deploy_test.rb
```diff
@@ -106,13 +106,19 @@ def test_dead_pods_in_old_replicaset_are_ignored
...
     deploy_fixtures("hello-cloud", subset: ["web.yml.erb", "configmap-data.yml"])
     pods = kubeclient.get_pods(namespace: @namespace, label_selector: "name=web,app=hello-cloud")
-    running_pods, not_running_pods = pods.partition { |pod| pod.status.phase == "Running" }
-    assert_equal 1, running_pods.size
-    assert not_running_pods.size >= 1
+    running_pod_count = pods.count { |pod| pod.status.phase == "Running" }
+    final_failed_pod_count = pods.count { |pod| pod.status.phase == "Failed" }
+
+    assert_equal 1, running_pod_count
+    assert final_failed_pod_count >= initial_failed_pod_count # failed pods not cleaned up
```
kirs: Next time you can use `assert_operator 5, :<=, 4` that generates a slightly better failure message

test/integration/kubernetes_deploy_test.rb
```diff
@@ -106,13 +106,19 @@ def test_dead_pods_in_old_replicaset_are_ignored
...
+      pods = kubeclient.get_pods(namespace: @namespace, label_selector: "name=web,app=hello-cloud")
+      initial_failed_pod_count = pods.count { |pod| pod.status.phase == "Failed" }
+    end
 
     deploy_fixtures("hello-cloud", subset: ["web.yml.erb", "configmap-data.yml"])
     pods = kubeclient.get_pods(namespace: @namespace, label_selector: "name=web,app=hello-cloud")
-    running_pods, not_running_pods = pods.partition { |pod| pod.status.phase == "Running" }
-    assert_equal 1, running_pods.size
-    assert not_running_pods.size >= 1
+    running_pod_count = pods.count { |pod| pod.status.phase == "Running" }
```
kirs: 👍 for using `count`

PR #70, opened 2017-03-20
Allow deploys to protected namespaces, without pruning

KnVerey:
### What?

Re-enables deploying to `default`/`kube-system` but with pruning disabled and some large warnings (output is from the test, hence the odd namespace name):
image
image

EDIT: It will still refuse to deploy to protected namespaces unless you use `--allow-protected-ns --no-prune`, which enables the above behaviour.

### Why?

With https://github.com/Shopify/cloudplatform/issues/264, we're switching to deploying all our cluster services with Shipit/kuberentes-deploy. In most cases, we're able to move them to their own namespace. However, there's one thing we need unfortunately need to leave in default at least during a transition period. WDYT, reviewers? Is this a terrible thing to allow temporarily/permanently?

cc @Shopify/cloudplatform

### DIFF
--- a/lib/kubernetes-deploy/runner.rb
+++ b/lib/kubernetes-deploy/runner.rb
@@ -107,14 +107,10 @@ MSG
       predeploy_priority_resources(resources)
 
       phase_heading("Deploying all resources")
-      if PROTECTED_NAMESPACES.include?(@namespace)
-        KubernetesDeploy.logger.warn("Deploying to protected namespace #{@namespace} without resource pruning.")
-        # Ignoring @prune here is a redundant safeguard given the danger of pruning protected namespaces:
-        # validate_configuration should already raise an error if it is true with a protected ns
-        deploy_resources(resources, prune: false)
-      else
-        deploy_resources(resources, prune: @prune)
+      if PROTECTED_NAMESPACES.include?(@namespace) && @prune
+        raise FatalDeploymentError, "Refusing to deploy to protected namespace '#{@namespace}' with pruning enabled"
       end
+      deploy_resources(resources, prune: @prune)
 
       return unless wait_for_completion?
       wait_for_completion(resources)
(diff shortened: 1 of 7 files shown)


kirs:
It's great to see that the tool can be potentially used to deploy both apps and cluster configuration.

Maybe those could be CLI flags? For me protecting system namespaces by default sounds as a good idea. When user attempts to deploy to the protected namespace, we can tell them to use `--force` if they are 100% confident at what they are doing.

As a result to deploy the cluster services, you'd run `kubernetes-deploy --no-prune --force`.

KnVerey:
I ❤️  the flags idea. I'll redo this PR to that end.

KnVerey:
Upon further reflection, I think it would be dangerous/unhelpful to make it possible to ever prune protected namespaces. Therefore, the `--force` flag could only be used in conjunction with `--no-prune`, or would override it. I still like the idea of using a flag to enable this feature, but I've called it `--allow-protected-ns`, and enabling it results in in the originally described behaviour: the deploy is allowed, and warnings are logged about the lack of pruning. LMK what you think!

KnVerey:
Spoke with @kirs, who expressed a preference for keeping a `--no-prune` flag and requiring that it be present when using the protected namespace override (`--allow-protected-ns`). The commit I just pushed implements this. It makes the code a bit more complicated to support a feature we haven't seen a use case for, but puristically, keeping these behaviours separate does make sense.

lib/kubernetes-deploy/runner.rb
```diff
@@ -97,7 +103,14 @@ def run
       predeploy_priority_resources(resources)
 
       phase_heading("Deploying all resources")
-      deploy_resources(resources, prune: true)
+      if PROTECTED_NAMESPACES.include?(@namespace)
```
camilo: hmm yes this is a belt and suspenders situation, it feels a bit too defensive to me. If we end up in a situation where `@prune` here is anything but false there is a bug, if we want to be deffensive why not `raise` then. What I'm saying is this would silently avoid a bug, but not tell us. I rather we either trust the rest of the code and the tests or blow up entirely on a broken assumption.

KnVerey: Well put. I agree raising would be better and will update this!

PR #71, opened 2017-03-22
Add a task runner to trigger one-off jobs

xldenis:
There is a need to run one off tasks within a kubernetes infrastructure. This system allows application developers to provide a `ConfigMap` embedding a pod spec which is used as a template. The provided `kubernetes-run` executable will be able to look up the template and fill in the `args` and `command` parameters with user supplied values.

cc @Shopify/pods

### DIFF
--- a/README.md
+++ b/README.md
@@ -40,6 +40,27 @@ The following command will restart all pods in the `web` and `jobs` deployments:
 
 `kubernetes-restart <kube namespace> <kube context> --deployments=web,jobs`
 
+### Running one off tasks
+
+To trigger a one-off job such as a rake task _outside_ of a deploy, use the following command:
+
+`kubernetes-run <kube namespace> <kube context> <arguments> --entrypoint=/bin/bash`
+
+This command assumes that you've already deployed a `PodTemplate` named `task-runner-template`, which contains a full pod specification in its `data`. The pod specification in turn must have a container named `task-runner`. Based on this specification `kubernetes-run` will create a new pod with the entrypoint of the task-runner container overriden with the supplied arguments.
+
+#### Creating a PodTemplate
+
+The [`PodTemplate`](https://kubernetes.io/docs/api-reference/v1.6/#podtemplate-v1-core) object should have a field `template` containing a `Pod` specification which does not include the `apiVersion` or `kind` parameters. An example is provided in this repo in `test/fixtures/hello-cloud/template-runner.yml`. 
+
+#### Providing multiple different task-runner configurations
+
+If your application requires task runner templates you can specify which template to use by using the `--template` option. All templates are expected to provide a container called `task-runner`.
+
+#### Specifying environment variables for the container
+
+If you also need to specify environment variables on top of the arguments, you can specify the `--env-vars` flag which accepts a comma separated list of environment variables like so: `--env-vars="ENV=VAL,ENV2=VAL"`
+
+
 ## Development
 
 After checking out the repo, run `bin/setup` to install dependencies. You currently need to [manually install kubectl version 1.6.0 or higher](https://kubernetes.io/docs/user-guide/prereqs/) as well if you don't already have it.
--- a/exe/kubernetes-run
+++ b/exe/kubernetes-run
@@ -1,0 +1,31 @@
+#!/usr/bin/env ruby
+# frozen_string_literal: true
+
+require 'kubernetes-deploy'
+require 'kubernetes-deploy/runner_task'
+require 'optparse'
+
+template = "task-runner-template"
+entrypoint = nil
+env_vars = []
+
+ARGV.options do |opts|
+  opts.on("--template=TEMPLATE") { |n| template = n }
+  opts.on("--env-vars=ENV_VARS") { |vars| env_vars = n.split(",")}
+  opts.on("--entrypoint=ENTRYPOINT") { |c| entrypoint = [c] }
+  opts.parse!
+end
+
+runner = KubernetesDeploy::RunnerTask.new(
+  namespace: ARGV[0],
+  context: ARGV[1],
+)
+
+ KubernetesDeploy::Runner.with_friendly_errors do
+  runner.run(
+    task_template: template,
+    entrypoint: entrypoint,
+    args: ARGV[2..-1],
+    env_vars: env_vars
+  )
+end
--- a/lib/kubernetes-deploy.rb
+++ b/lib/kubernetes-deploy.rb
@@ -10,5 +10,12 @@ require 'kubernetes-deploy/runner'
 
 module KubernetesDeploy
   class FatalDeploymentError < StandardError; end
+
+  class NamespaceNotFoundError < FatalDeploymentError
+    def initialize(name, context)
+      super("Namespace `#{name}` not found in context `#{context}`. Aborting the task.")
+    end
+  end
+
   include Logger
 end
--- a/lib/kubernetes-deploy/runner.rb
+++ b/lib/kubernetes-deploy/runner.rb
@@ -5,7 +5,6 @@ require 'erb'
 require 'yaml'
 require 'shellwords'
 require 'tempfile'
-
 require 'kubernetes-deploy/kubernetes_resource'
 %w(
   cloudsql
@@ -16,6 +15,7 @@ require 'kubernetes-deploy/kubernetes_resource'
   pod
   redis
   service
+  pod_template
   bugsnag
 ).each do |subresource|
   require "kubernetes-deploy/kubernetes_resource/#{subresource}"
@@ -116,6 +116,7 @@ MSG
       if PROTECTED_NAMESPACES.include?(@namespace) && @prune
         raise FatalDeploymentError, "Refusing to deploy to protected namespace '#{@namespace}' with pruning enabled"
       end
+
       deploy_resources(resources, prune: @prune)
 
       return unless wait_for_completion?
--- a/lib/kubernetes-deploy/runner_task.rb
+++ b/lib/kubernetes-deploy/runner_task.rb
@@ -1,0 +1,152 @@
+# frozen_string_literal: true
+require 'tempfile'
+
+require 'kubernetes-deploy/kubeclient_builder'
+require 'kubernetes-deploy/ui_helpers'
+require 'kubernetes-deploy/kubectl'
+
+module KubernetesDeploy
+  class RunnerTask
+    include KubeclientBuilder
+    include Kubectl
+    include UIHelpers
+
+    class FatalTaskRunError < FatalDeploymentError; end
+    class TaskTemplateMissingError < FatalDeploymentError
+      def initialize(task_template, namespace, context)
+        super("Pod template `#{task_template}` cannot be found in namespace: `#{namespace}`, context: `#{context}`")
+      end
+    end
+
+    def initialize(namespace:, context:, logger: KubernetesDeploy.logger)
+      @logger = logger
+      @namespace = namespace
+      @kubeclient = build_v1_kubeclient(context)
+      @context = context
+    end
+
+    def run(task_template:, entrypoint:, args:, env_vars: [])
+      phase_heading("Validating configuration")
+      validate_configuration(task_template, args)
+
+      phase_heading("Fetching task template")
+      raw_template = get_template(task_template)
+
+      phase_heading("Constructing final pod specification")
+      rendered_template = build_pod_template(raw_template, entrypoint, args, env_vars)
+
+      validate_pod_spec(rendered_template)
+
+      phase_heading("Creating pod")
+      KubernetesDeploy.logger.info("Starting task runner pod: '#{rendered_template.metadata.name}'")
+      @kubeclient.create_pod(rendered_template)
+    end
+
+    private
+
+    def validate_configuration(task_template, args)
+      errors = []
+
+      if task_template.blank?
+        errors << "Task template name can't be nil"
+      end
+
+      if @namespace.blank?
+        errors << "Namespace can't be empty"
+      end
+
+      if args.blank?
+        errors << "Args can't be nil"
+      end
+
+      begin
+        @kubeclient.get_namespace(@namespace)
+      rescue KubeException => e
+        errors << if e.error_code == 404
+          "Namespace was not found"
+        else
+          "Could not connect to kubernetes cluster"
+        end
+      end
+
+      raise FatalTaskRunError, "Configuration invalid: #{errors.join(', ')}" unless errors.empty?
+    end
+
+    def get_template(template_name)
+      KubernetesDeploy.logger.info(
+        "Fetching task runner pod template: '#{template_name}' in namespace: '#{@namespace}'"
+      )
+
+      pod_template = @kubeclient.get_pod_template(template_name, @namespace)
+
+      pod_template.template
+    rescue KubeException => error
+      if error.error_code == 404
+        raise TaskTemplateMissingError.new(template_name, @namespace, @context)
+      else
+        raise
+      end
+    end
+
+    def build_pod_template(base_template, entrypoint, args, env_vars)
+      KubernetesDeploy.logger.info("Rendering template for task runner pod")
+
+      rendered_template = base_template.dup
+      rendered_template.kind = 'Pod'
+      rendered_template.apiVersion = 'v1'
+
+      container = rendered_template.spec.containers.find { |cont| cont.name == 'task-runner' }
+
+      raise FatalTaskRunError, "Pod spec does not contain a template container called 'task-runner'" if container.nil?
+
+      container.command = entrypoint
+      container.args = args
+      container.env ||= []
+
+      env_args = env_vars.map do |key, value|
+        key, value = env.split('=', 2)
+        { name: key, value: value }
+      end
+
+      container.env = container.env.map(&:to_h) + env_args
+
+      unique_name = rendered_template.metadata.name + "-" + SecureRandom.hex(8)
+
+      KubernetesDeploy.logger.warn("Name is too long, using '#{unique_name[0..62]}'") if unique_name.length > 63
+      rendered_template.metadata.name = unique_name[0..62]
+      rendered_template.metadata.namespace = @namespace
+
+      rendered_template
+    end
+
+    def validate_pod_spec(pod)
+      f = Tempfile.new(pod.metadata.name)
+      f.write recursive_to_h(pod).to_json
+      f.close
+
+      _out, err, status = Kubectl.run_kubectl(
+        "apply", "--dry-run", "-f", f.path,
+        namespace: @namespace,
+        context: @context
+      )
+
+      unless status.success?
+        raise FatalTaskRunError, "Invalid pod spec: #{err}"
+      end
+    end
+
+    def recursive_to_h(struct)
+      if struct.is_a?(Array)
+        return struct.map { |v| v.is_a?(OpenStruct) || v.is_a?(Array) || v.is_a?(Hash) ? recursive_to_h(v) : v }
+      end
+
+      hash = {}
+
+      struct.each_pair do |k, v|
+        recursive_val = v.is_a?(OpenStruct) || v.is_a?(Array) || v.is_a?(Hash)
+        hash[k] = recursive_val ? recursive_to_h(v) : v
+      end
+      hash
+    end
+  end
+end
--- a/test/fixtures/hello-cloud/template-runner.yml
+++ b/test/fixtures/hello-cloud/template-runner.yml
@@ -1,0 +1,23 @@
+apiVersion: v1
+kind: PodTemplate
+metadata:
+  name: hello-cloud-template-runner
+  labels:
+    name: hello-cloud-template-runner
+    app: hello-cloud
+template:
+  metadata:
+    name: task-runner
+    labels:
+      name: runner
+      app: hello-cloud
+  spec:
+    containers:
+    - name: task-runner
+      image: hello-world:latest
+      env:
+      - name: CONFIG
+        valueFrom:
+          configMapKeyRef:
+            name: hello-cloud-configmap-data
+            key: datapoint1
--- a/test/integration/runner_task_test.rb
+++ b/test/integration/runner_task_test.rb
@@ -1,0 +1,39 @@
+# frozen_string_literal: true
+require 'test_helper'
+
+require 'kubernetes-deploy/runner_task'
+class RunnerTaskTest < KubernetesDeploy::IntegrationTest
+  def test_works
+    deploy_fixtures("hello-cloud", subset: ["template-runner.yml"])
+
+    task_runner = KubernetesDeploy::RunnerTask.new(
+      context: KubeclientHelper::MINIKUBE_CONTEXT,
+      namespace: @namespace,
+    )
+
+    task_runner.run(
+      task_template: 'hello-cloud-template-runner',
+      entrypoint: ['/bin/bash'],
+      args: %w(echo "KUBERNETES-DEPLOY")
+    )
+
+    assert_logs_match(/Starting task runner/)
+  end
+
+  def test_substitutes_arguments
+    deploy_fixtures("hello-cloud", subset: ["template-runner.yml"])
+
+    task_runner = KubernetesDeploy::RunnerTask.new(
+      context: KubeclientHelper::MINIKUBE_CONTEXT,
+      namespace: @namespace,
+    )
+
+    pod = task_runner.run(
+      task_template: 'hello-cloud-template-runner',
+      entrypoint: nil,
+      args: %w(rake some_task)
+    )
+
+    assert_equal %w(rake some_task), pod.spec.containers.first.args
+  end
+end
--- a/lib/kubernetes-deploy/kubernetes_resource.rb
+++ b/lib/kubernetes-deploy/kubernetes_resource.rb
@@ -35,6 +35,7 @@ module KubernetesDeploy
       when 'ingress' then Ingress.new(name, namespace, context, file)
       when 'persistentvolumeclaim' then PersistentVolumeClaim.new(name, namespace, context, file)
       when 'service' then Service.new(name, namespace, context, file)
+      when 'podtemplate' then PodTemplate.new(name, namespace, context, file)
       else self.new(name, namespace, context, file).tap { |r| r.type = type }
       end
     end
--- a/lib/kubernetes-deploy/kubernetes_resource/pod_template.rb
+++ b/lib/kubernetes-deploy/kubernetes_resource/pod_template.rb
@@ -1,0 +1,30 @@
+# frozen_string_literal: true
+module KubernetesDeploy
+  class PodTemplate < KubernetesResource
+    def initialize(name, namespace, context, file)
+      @name = name
+      @namespace = namespace
+      @context = context
+      @file = file
+    end
+
+    def sync
+      _, _err, st = run_kubectl("get", type, @name)
+      @status = st.success? ? "Available" : "Unknown"
+      @found = st.success?
+      log_status
+    end
+
+    def deploy_succeeded?
+      exists?
+    end
+
+    def deploy_failed?
+      false
+    end
+
+    def exists?
+      @found
+    end
+  end
+end
--- a/lib/kubernetes-deploy/restart_task.rb
+++ b/lib/kubernetes-deploy/restart_task.rb
@@ -14,12 +14,6 @@ module KubernetesDeploy
       end
     end
 
-    class NamespaceNotFoundError < FatalDeploymentError
-      def initialize(name, context)
-        super("Namespace `#{name}` not found in context `#{context}`. Aborting the task.")
-      end
-    end
-
     class RestartError < FatalDeploymentError
       def initialize(deployment_name, response)
         super("Failed to restart #{deployment_name}. " \
--- a/test/helpers/fixture_set.rb
+++ b/test/helpers/fixture_set.rb
@@ -83,5 +83,13 @@ module FixtureSetAssertions
       assert_equal 1, configmaps.size, "Expected 1 configmap, got #{configmaps.size}"
       assert_equal expected_data, configmaps.first["data"].to_h
     end
+
+    def assert_pod_templates_present(cm_name)
+      pod_templates = kubeclient.get_pod_templates(
+        namespace: namespace,
+        label_selector: "name=#{cm_name},app=#{app_name}"
+      )
+      assert_equal 1, pod_templates.size, "Expected 1 podtemplate, got #{pod_templates.size}"
+    end
   end
 end
--- a/test/helpers/fixture_sets/hello-cloud.rb
+++ b/test/helpers/fixture_sets/hello-cloud.rb
@@ -11,6 +11,7 @@ module FixtureSetAssertions
       assert_all_web_resources_up
       assert_all_redis_resources_up
       assert_configmap_data_present
+      assert_podtemplate_runner_present
     end
 
     def assert_unmanaged_pod_statuses(status, count = 1)
@@ -60,6 +61,10 @@ module FixtureSetAssertions
       else
         refute_resource_exists("pvc", "redis")
       end
+    end
+
+    def assert_podtemplate_runner_present
+      assert_pod_templates_present("hello-cloud-template-runner")
     end
   end
 end
--- a/test/integration/restart_task_test.rb
+++ b/test/integration/restart_task_test.rb
@@ -110,7 +110,7 @@ class RestartTaskTest < KubernetesDeploy::IntegrationTest
       context: KubeclientHelper::MINIKUBE_CONTEXT,
       namespace: "walrus"
     )
-    error = assert_raises(KubernetesDeploy::RestartTask::NamespaceNotFoundError) do
+    error = assert_raises(KubernetesDeploy::NamespaceNotFoundError) do
       restart.perform(["web"])
     end
     assert_equal "Namespace `walrus` not found in context `minikube`. Aborting the task.", error.to_s
--- a/test/unit/kubernetes-deploy/runner_task_test.rb
+++ b/test/unit/kubernetes-deploy/runner_task_test.rb
@@ -1,0 +1,50 @@
+# frozen_string_literal: true
+require 'test_helper'
+
+class RunnerTaskUnitTest < KubernetesDeploy::TestCase
+  def test_missing_namespace
+    assert_raises(KubernetesDeploy::RunnerTask::FatalTaskRunError,
+      message: "Configuration invalid: Namespace was not found") do
+      task_runner = KubernetesDeploy::RunnerTask.new(
+        context: KubeclientHelper::MINIKUBE_CONTEXT,
+        namespace: "missing",
+      )
+
+      task_runner.run(
+        task_template: 'hello-cloud-template-runner',
+        entrypoint: nil,
+        args: 'a'
+      )
+    end
+  end
+
+  def test_missing_arguments
+    assert_raises(KubernetesDeploy::RunnerTask::FatalTaskRunError) do
+      task_runner = KubernetesDeploy::RunnerTask.new(
+        context: KubeclientHelper::MINIKUBE_CONTEXT,
+        namespace: @namespace,
+      )
+
+      task_runner.run(
+        task_template: 'hello-cloud-template-runner',
+        entrypoint: nil,
+        args: nil
+      )
+    end
+  end
+
+  def test_missing_template
+    assert_raises(KubernetesDeploy::RunnerTask::FatalTaskRunError) do
+      task_runner = KubernetesDeploy::RunnerTask.new(
+        namespace: @namespace,
+        context: KubeclientHelper::MINIKUBE_CONTEXT
+      )
+
+      task_runner.run(
+        task_template: nil,
+        entrypoint: nil,
+        args: ['a']
+      )
+    end
+  end
+end


KnVerey:
re: standard way of validating resources, there is something built into kubectl that we're already leveraging for deploys: `kubectl apply --dry-run`.

Re: kubeclient, check out @kirs work in https://github.com/Shopify/kubernetes-deploy/pull/61/files

xldenis:
The work in #61 seems very useful, when it gets merged I'll rebase my branch on it. `kubectl apply --dry-run`. It should be simple to serialize the`kubeclient` spec (`to_yaml`) and validate it. Does this provide helpful error messages? Or cryptic ones like much of `kubectl` errors?

KnVerey:
> Does this provide helpful error messages? Or cryptic ones like much of kubectl errors?

They're definitely kubectl errors, which vary considerably in clarity. As you've seen, some of the ones for syntax errors are just the worst. But for the three things that you're checking manually in this PR, `namespace` actually doesn't need to be in the template--typically we both expect it and pass it it as an arg to make sure it is _always_ set; `metadata.name` does in fact throw an error including "resource name may not be empty"; and it's fine to have multiple containers in a pod.

xldenis:
> metadata.name does in fact throw an error including "resource name may not be empty"

`kubeclient` throws even more cryptic errors than `kubectl` in some cases, having a missing name is one of those cases.

> and it's fine to have multiple containers in a pod.

Yes it is, but for the purpose of this task runner I didn't feel like it would be of value to allow more than one container per 'task' pod.

KnVerey:
> Yes it is, but for the purpose of this task runner I didn't feel like it would be of value to allow more than one container per 'task' pod.

Why not? The task pod is typically going to be the app pod, right, and I don't see a good reason to prohibit them from having sidecars.

xldenis:
> Why not? The task pod is typically going to be the app pod, right, and I don't see a good reason to prohibit them from having sidecars.

I suppose but do we perform command / argument substitutions on those? what about environment variables passed in?

KnVerey:
> I suppose but do we perform command / argument substitutions on those? what about environment variables passed in?

I'd suggest identifying the target container with an additional key in the configMap, e.g. `cmd-runner-container-name: runner`, which would only be _required_ if there are multiple containers. Or we could require a conventional name and always fail if no such container exists on the pod.

xldenis:
I've fixed the remaining test cases, I think this is ready for a quick pass to make sure I'm not going completely off course. 

It current looks either at the only container in a pod or for one called `task-runner`. It will then insert the `command` (if supplied) and the `args` (required). Then it will spawn the pod.

fw42:
Why `exe`? Can this be in `bin` or `script` or something where other stuff like this is?

wfarr:
> Why exe? Can this be in bin or script or something where other stuff like this is?

`exe` is the intended directory for executables to be installed by a ruby gem since 2015 or so to avoid collisions with binstubs etc: http://bundler.io/blog/2015/03/20/moving-bins-to-exe.html

lib/kubernetes-deploy/kubectl.rb
```diff
@@ -0,0 +1,29 @@
+# frozen_string_literal: true
+require 'kubeclient'
+
+module KubernetesDeploy
```
KnVerey: This extraction should probably be a separate PR and should cover the KubernetesResource version as well: https://github.com/Shopify/kubernetes-deploy/blob/fd557925825a3a672f62a2ff0984bee3bf1b519b/lib/kubernetes-deploy/kubernetes_resource.rb#L115

@kirs tried to do this previously in https://github.com/Shopify/kubernetes-deploy/pull/58, but we disagreed on the implementation. This version seems reasonable to me (although needs changes to be used in the additional location), but we should reopen that discussion separately.

xldenis: Yea, it would make sense in another PR but I didn't want to open that without gathering opinions on it first.

lib/kubernetes-deploy/runner_task.rb
```diff
@@ -0,0 +1,105 @@
...
+        rendered_template.spec.containers.first
+      else
+        rendered_template.spec.containers.find { |cont| cont.name == 'task-runner' }
+      end
+
+      raise FatalTaskRunError, "Pod spec does not contain a templateable container" if container.nil?
+
+      container.command = command
+
+      # TODO: handle environment variables being passed to task
```
xldenis: I've realized since then that they aren't necessary for most of our immediate applications, if an env var is passed to a rake task _after_ the task as an argument rake will take care of the handling automatically.

```
rake task ENV_KEY=VALUE
```
works perfectly.

exe/kubernetes-run
```diff
@@ -0,0 +1,24 @@
...
+end
+
+runner = KubernetesDeploy::RunnerTask.new(
+  namespace: namespace
+)
+
+runner.run(
+  task_template: ARGV[0],
+  command: command,
+  args: ARGV[1..-1]
```
xldenis: I think they can be passed wholesale to the container as the `args`. Any potential env vars could use a `--env` flag to demarcate them specifically.

lib/kubernetes-deploy.rb
```diff
@@ -43,3 +42,5 @@ def level_from_env
     end
   end
 end
+
+require 'kubernetes-deploy/runner'
```
xldenis: Yes, I don't know what the better solution would be.

exe/kubernetes-run
```diff
@@ -0,0 +1,24 @@
...
+# frozen_string_literal: true
+
+require 'kubernetes-deploy'
+require 'kubernetes-deploy/runner_task'
+require 'optparse'
+
+namespace = "default"
+command = nil
+ARGV.options do |opts|
+  opts.on("--namespace=NAMESPACE") { |n| namespace = n }
```
xldenis: huh, I had updated the script to handle the `context`. I guess I didn't push it. However, since they aren't optional you're right its best to just make them positional arugments instead.

lib/kubernetes-deploy.rb
KnVerey: Maybe the logger should get its own file that should get required by both

lib/kubernetes-deploy/kubectl.rb
```diff
@@ -0,0 +1,29 @@
...
+  module Kubectl
+    class KubectlError < FatalDeploymentError; end
+
+    def run_kubectl(*args, namespace: "", context: "", namespaced: true, with_context: true)
```
kirs: From the API part, it seems weird that you have to pass both `namespace` and `namespaced`. If you don't want it to be namespaced, just skip passing `namespace`.

lib/kubernetes-deploy/runner.rb
```diff
@@ -301,24 +303,7 @@ def confirm_namespace_exists
...
-      end
-      KubernetesDeploy.logger.debug Shellwords.join(args)
-      out, err, st = Open3.capture3(*args)
-      KubernetesDeploy.logger.debug(out.shellescape)
-      unless st.success?
-        KubernetesDeploy.logger.warn("The following command failed: #{Shellwords.join(args)}")
-        KubernetesDeploy.logger.warn(err)
-      end
-      [out.chomp, err.chomp, st]
+      super(*args, namespace: @namespace, context: @context, namespaced: namespaced, with_context: with_context)
```
kirs: TBH I see it a bit confusing to use the super method that comes from the included module... I think it would be more clear to explicitly call the helper: `Kubectl.build_command(*args, namespace: foo, context: blah)`

lib/kubernetes-deploy/kubectl.rb
kirs: I agree with Katrina and I'd also prefer moving `kubectl` helpers in a separate PR.

lib/kubernetes-deploy/runner_task.rb
```diff
@@ -0,0 +1,106 @@
...
+
+    def validate_pod_spec(pod)
+      f = Tempfile.new(pod.metadata.name)
+      f.write pod.to_hash.to_json
+      f.close
+
+      _out, _err, status = run_kubectl("apply", "--dry-run", "-f", f.path, namespace: @namespace, context: @context)
+
+      unless status.success?
+        raise FatalTaskRunError, "Invalid pod spec"
```
kirs: I think we should log the error message, otherwise it will be hard to debug and reproduce it.

test/fixtures/hello-cloud/configmap-runner.yml
```diff
@@ -0,0 +1,26 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: hello-cloud-configmap-runner
+  labels:
+    name: hello-cloud-configmap-runner
+    app: hello-cloud
+data:
```
kirs: So pod template == ConfigMap?
https://kubernetes.io/docs/concepts/workloads/pods/pod-overview/#pod-templates doesn't say anything about config maps. Maybe documenting it in the README would help to make it clear.

lib/kubernetes-deploy/kubectl.rb
xldenis: It was kept to keep comaptibility with the usage in the `Runner` class since I was really just testing this out. However, since I override it anyways, I'll change this.

lib/kubernetes-deploy/runner_task.rb
xldenis: The error is already logged by `run_kubectl`.

test/fixtures/hello-cloud/configmap-runner.yml
xldenis: This is not the same thing as the pod template resource. That one is not really open for arbitrary manipulation, you can't create one directly. Instead we have to 'hack' one in via `ConfigMap`s. I'll add information to the readme though.

wfarr:
I'm a little confused as to what's producing the need for this in its current form. I have a few questions to try and understand the why.

* For running one-off tasks, are you finding yourself frequently running tasks in images besides the general application one for your repository?
* What makes this fundamentally different from the `db-migrate` and `cmd-runner` Pod patterns that already exist in repositories? Why do we feel this is better?

The vibe I got scanning through how this works and reading the history feels like a generic `cmd-runner` Pod in the repository that we munge the container command into on the fly would be sufficient, and that the pain being surfaced here is we might benefit from implementing some rudimentary form of stages with before and after hooks (a la Capistrano/Rake/etc).

sirupsen:
@wfarr's comment makes sense to me, @xldenis have you given this more thought?

xldenis:
> @wfarr's comment makes sense to me, @xldenis have you given this more thought?

We've discussed this a lot offline and decided that using the already existing `PodTemplate` resource was the best approach. `PodTemplates` are used throughout parts of Kubernetes and it seems like they will at least partially converge with our usage.

I've already updated the PR a while ago with the switch to `PodTemplates`, and merged the PR adding the `kubectl` helper last week.

test/fixtures/hello-cloud/template-runner.yml
```diff
@@ -0,0 +1,23 @@
...
+  spec:
+    containers:
+    - name: task-runner
+      image: hello-world:latest
+      env:
+      - name: CONFIG
+        valueFrom:
+          configMapKeyRef:
+            name: hello-cloud-configmap-runner
+            key: datapoint1
```
sirupsen: Where is the `hello-cloud-configmap-runner ` key defined?

test/integration/runner_task_test.rb
```diff
@@ -0,0 +1,84 @@
+# frozen_string_literal: true
+require 'test_helper'
+
+require 'kubernetes-deploy/runner_task'
+class RunnerTaskTest < KubernetesDeploy::IntegrationTest
```
sirupsen: When running this, I'm getting:

```
/Users/sirup/src/github.com/Shopify/kubernetes-deploy/test/test_helper.rb:68:in `create_namespace': uninitialized constant Kubeclient::Namespace (NameError)
```

I had to change `Kubernetes::Namespace` there to `Kubernetes::Resource`. How does this work? 🤔

sirupsen: When that was fixed, the tests run for me locally.

exe/kubernetes-run
```diff
@@ -0,0 +1,29 @@
...
+  opts.parse!
+end
+
+runner = KubernetesDeploy::RunnerTask.new(
+  namespace: ARGV[0],
+  context: ARGV[1],
+)
+
+runner.run(
+  task_template: ARGV[],
```
sirupsen: What's `ARGV[]`? It doesn't work for me locally.

lib/kubernetes-deploy/runner_task.rb
```diff
@@ -0,0 +1,132 @@
...
+      @context = context
+    end
+
+    def run(task_template:, command:, args:, env_vars: [])
+      validate_configuration(task_template, @namespace, args)
+
+      raw_template = get_template(task_template, namespace: @namespace)
+      rendered_template = build_pod_template(raw_template, command, args, env_vars)
+
+      KubernetesDeploy.logger.info("Starting task runner pod #{rendered_template.metadata.name}")
```
sirupsen: Can you print a screenshot of what all the logging looks like? Is it consistent with the other commands?

lib/kubernetes-deploy/runner_task.rb
```diff
@@ -0,0 +1,132 @@
...
+    end
+
+    def build_pod_template(base_template, command, args, env_vars)
+      KubernetesDeploy.logger.info("Rendering template for task runner pod")
+
+      rendered_template = base_template.dup
+      rendered_template.kind = 'Pod'
+      rendered_template.apiVersion = 'v1'
+      rendered_template.spec.containers
+        .first.args = args
```
sirupsen: Would this ever need to support ERB?

test/fixtures/hello-cloud/template-runner.yml
```diff
@@ -0,0 +1,23 @@
...
+template:
+  metadata:
+    name: task-runner
+    labels:
+      name: runner
+      app: hello-cloud
+  spec:
+    containers:
+    - name: task-runner
+      image: hello-world:latest
```
sirupsen: This means it'll run the latest production build, not the latest deployed one. Perhaps we should convert this into a template and allow running an arbitrary SHA and default to `latest`, or something like this..

lib/kubernetes-deploy/runner_task.rb
```diff
@@ -0,0 +1,132 @@
...
+
+      container = rendered_template.spec.containers.find { |cont| cont.name == 'task-runner' }
```
sirupsen: Why would it ever contain multiple containers? Should we raise if it containers more than one?

lib/kubernetes-deploy/runner_task.rb
```diff
@@ -0,0 +1,132 @@
...
+    def validate_pod_spec(pod)
+      f = Tempfile.new(pod.metadata.name)
+      f.write recursive_to_h(pod).to_json
+      f.close
+
+      _out, err, status = Kubectl.run_kubectl(
+        "apply", "--dry-run", "-f", f.path,
+        namespace: @namespace,
+        context: @context
+      )
```
sirupsen: Is it worth putting this on `Kubectl`? We also do this in `runner.rb`. Do they have to be different..?

lib/kubernetes-deploy/runner_task.rb
```diff
@@ -0,0 +1,132 @@
...
+      rendered_template.metadata.namespace = @namespace
+
+      validate_pod_spec(rendered_template)
+
+      rendered_template
+    end
+
+    def validate_pod_spec(pod)
+      f = Tempfile.new(pod.metadata.name)
+      f.write recursive_to_h(pod).to_json
```
sirupsen: Why do we need this? `ActiveSupport` is a dependency, would it help?

README.md
```diff
@@ -40,6 +40,23 @@ The following command will restart all pods in the `web` and `jobs` deployments:
 
 `kubernetes-restart <kube namespace> <kube context> --deployments=web,jobs`
 
+### Running one off tasks
+
+To trigger a single use job such as a rake task, use the following command:
+
+`kubernetes-run <kube namespace> <kube context> --command=rake <arguments>`
+
+This command assumes that there is a `ConfigMap` named `task-runner-template`, which contains a full pod specification in it's `data`. The pod specification in turn must have a container named `task-runner`. This container will be passed the arguments and command passed to `kubernetes-run`.
```
sirupsen: Could we provide a piece of documentation on how to put the `task-runner-template` in `ConfigMap`?

lib/kubernetes-deploy/runner_task.rb
```diff
@@ -0,0 +1,132 @@
...
+      end
+
+      raise FatalTaskRunError, "Configuration invalid: #{errors.join(', ')}" unless errors.empty?
+    end
+
+    def get_template(template_name, namespace:)
+      KubernetesDeploy.logger.info("Fetching task runner pod template: '#{template_name}' in namespace: '#{namespace}'")
+
+      pod_template = begin
+        @kubeclient.get_pod_template(template_name, namespace)
```
sirupsen: Curious how you found this method and how you knew it grabbed it from the `ConfigMap`?

xldenis:
> Overall, this approach makes a lot of sense to me. I'm curious why we went with the ConfigMap, instead of just storing it as a template in the appropriate repository. Is that to avoid people having to run it in the directory of the app? Is that necessary?

Yes, exactly, this allows teams to trigger jobs from other locations (such as on server in the datacenter), it also means that it will use the correct version of the configuration (since it's tied to whats deployed, not the commit locally available).

test/integration/runner_task_test.rb
xldenis: That's very weird, did you do anything weird when running tests? I ran this in CI and it worked (as well as working locally).

lib/kubernetes-deploy/runner_task.rb
xldenis: Not for the forseeable future, the UI for this is via the CLI where using something like ERB would be massively impractical.

lib/kubernetes-deploy/runner_task.rb
xldenis: This was addressed in previous comments. Sidecar containers are perfectly normal in kubernetes and there is no reason to disallow them.

lib/kubernetes-deploy/runner_task.rb
xldenis: 👍 could be `Kubectl.validate_file` or something related.

lib/kubernetes-deploy/runner_task.rb
xldenis: Unfortunately no, this is because `OpenStruct` doesn't recursively call `to_h` which means we end up serializing nested hashes as strings in the output yaml.

lib/kubernetes-deploy/runner_task.rb
xldenis: We are actually using a `PodTemplate` so it will grab the appropriate `PodTemplate.

exe/kubernetes-run
xldenis: fixed

test/fixtures/hello-cloud/template-runner.yml
xldenis: It isn't I've moved it to a real key, however, that doesn't matter for the purposes of our tests.

test/fixtures/hello-cloud/template-runner.yml
xldenis: Each service will define their own `template-runner`s which will be uploaded at each deploy. They will therefore include specific sha's

README.md
xldenis: done.

lib/kubernetes-deploy/runner_task.rb
xldenis: What are you looking for wrt consistency?

lib/kubernetes-deploy/runner_task.rb
sirupsen: That's not what the README says, I'm a bit confused, maybe the README needs an update?

test/integration/runner_task_test.rb
sirupsen: Nope, clean slate, `dev up`.

lib/kubernetes-deploy/runner_task.rb
sirupsen: Just that: Consistency.

test/fixtures/hello-cloud/template-runner.yml
sirupsen: How does the deploy upload a new `template-runner`? Where does that happen? I don't see the code for it

lib/kubernetes-deploy/runner_task.rb
xldenis: Yep I updated it in the last commit after I saw your confusion.

test/integration/runner_task_test.rb
xldenis: this isn't a `dev` project. what do you get from `bundle exec rake test` ?

test/fixtures/hello-cloud/template-runner.yml
xldenis: A template is added to the relevant project. For example Shopify would include a template specifying what a shopify pod looks like. This infrastructure can then use the template to actually run tasks.

lib/kubernetes-deploy/runner_task.rb
xldenis: After trying to implment this, I think it's not as useful as I first thought. I'd rather take a stab at it in a separate PR that can cleanup other validation related things at the same time.

test/integration/runner_task_test.rb
sirupsen: I ran the test individually with `bundle exec ruby -Itest ..`, because it'd take too long to run the suite.

lib/kubernetes-deploy/runner_task.rb
xldenis:
```
λ bundle exec ./exe/kubernetes-run default minikube --template=hello-cloud-template-runner --command=rake test
[INFO][2017-04-19 14:53:04 -0500] Fetching task runner pod template: 'hello-cloud-template-runner' in namespace: 'default'
[INFO][2017-04-19 14:53:04 -0500] Rendering template for task runner pod
[INFO][2017-04-19 14:53:04 -0500] Starting task runner pod task-runner-f6d09899f2fc5093
``` 
example output.

lib/kubernetes-deploy/runner_task.rb
```diff
@@ -0,0 +1,132 @@
...
+      rendered_template.spec.containers
+        .first.args = args
+
+      container = rendered_template.spec.containers.find { |cont| cont.name == 'task-runner' }
+
+      raise FatalTaskRunError, "Pod spec does not contain a template container called 'task-runner'" if container.nil?
+
+      container.command = command
+      container.env ||= []
+      container.env = container.env.map(&:to_h) + env_vars.map { |env| env.split('=') }.map do |key, value|
```
KnVerey: This is a bit hard to read, and you don't need to map twice in the second part. Suggestion:
```ruby
new_env_vars = env_vars.map do |env|
  key, value = env.split("=", 2)
  { name: key, value: value }
end

container.env = (container.env || []).map(&:to_h) + new_env_vars
```

lib/kubernetes-deploy/runner_task.rb
```diff
@@ -0,0 +1,132 @@
...
+        "apply", "--dry-run", "-f", f.path,
+        namespace: @namespace,
+        context: @context
+      )
+
+      unless status.success?
+        raise FatalTaskRunError, "Invalid pod spec: #{err}"
+      end
+
+      true
```
KnVerey: Nothing is actually checking this value

lib/kubernetes-deploy/runner_task.rb
KnVerey: Ugh, this is a pain. I'm thinking it would be much easier/cleaner to use kubectl in place of `get_pod_template` and manipulate the json/hash. It seems like the gem is getting in the way FAR more than it is helping here.

I've encountered similar pains in the feature I'm working on right now (needed to `.to_h.stringify_keys` some things). cc @kirs between this and your simple patch for the 404 error still not being accepted, I'm having second thoughts about favouring the gem over kubectl. 😞

test/fixtures/hello-cloud/template-runner.yml
```diff
@@ -0,0 +1,23 @@
...
+metadata:
+  name: hello-cloud-template-runner
+  labels:
+    name: hello-cloud-template-runner
+    app: hello-cloud
+template:
+  metadata:
+    name: task-runner
+    labels:
+      name: runner
```
KnVerey: nit: this name should match `metadata.name`

test/integration/runner_task_test.rb
```diff
@@ -0,0 +1,84 @@
...
+  def test_works
+    deploy_fixtures("hello-cloud")
```
KnVerey: Please add an assertion that the template exists to `FixtureSetAssertions::HelloCloud#assert_all_up` and then call that method here.

test/integration/runner_task_test.rb
KnVerey: This runs fine for me locally with `bundle exec ruby -I test` or within the whole suite. ¯\\\_(ツ)_/¯

test/integration/runner_task_test.rb
```diff
@@ -0,0 +1,84 @@
...
+
+    task_runner = KubernetesDeploy::RunnerTask.new(
+      context: KubeclientHelper::MINIKUBE_CONTEXT,
+      namespace: @namespace,
+    )
+
+    task_runner.run(
+      task_template: 'hello-cloud-template-runner',
+      command: nil, # idk
```
KnVerey: why `# idk`? Shouldn't we be validating that this isn't nil, actually?

test/integration/runner_task_test.rb
KnVerey: Acutally, don't deploy the whole thing. It'll be way slower than necessary. Use `subset: [filename]` kwarg to select just the configmap and podtemplate you need. Still do add a helper to `FixtureSetAssertions::HelloCloud` (and call it from `assert_all_up`, since this is now part of that deploy set) and call that + the existing one for the configmap here.

test/integration/runner_task_test.rb
```diff
@@ -0,0 +1,84 @@
...
+      args: ['a']
+    )
+
+    assert_logs_match(/Starting task runner/)
```
KnVerey: These tests are expensive, so I'd suggest doing one big test for the success case, that checks all the expected logs, and the command, and the args, etc.

test/integration/runner_task_test.rb
```diff
@@ -0,0 +1,84 @@
...
+      task_template: 'hello-cloud-template-runner',
+      command: nil, # idk
+      args: ['a']
+    )
+
+    assert_equal ['a'], pod.spec.containers.first.args
+  end
+
+  def test_missing_namespace
+    assert_raises(KubernetesDeploy::RunnerTask::FatalTaskRunError) do
```
KnVerey: Please add assertions on the actual error messages here, to make sure they are helpful.

test/integration/runner_task_test.rb
```diff
@@ -0,0 +1,84 @@
...
+
+      task_runner.run(
+        task_template: 'hello-cloud-template-runner',
+        command: nil, # idk
+        args: 'a'
+      )
+    end
+  end
+
+  def test_missing_arguments
```
KnVerey: These last two should be unit tests. Since they don't rely on a live namespace, let's avoid the overhead of creating/destroying one (which happens for every integration test case).

KnVerey: Btw it needs a rebase (hopefully I'll have a chance to do that imminently), but [this branch](https://github.com/Shopify/kubernetes-deploy/compare/secret_ejson) has some examples of using a mock kubeclient to test sad paths / edge cases: https://github.com/Shopify/kubernetes-deploy/blob/33d8d2ebd31997e6f62e9e632a939fea344008e5/test/unit/kubernetes-deploy/ejson_secret_provisioner_test.rb

README.md
```diff
@@ -40,6 +40,27 @@ The following command will restart all pods in the `web` and `jobs` deployments:
...
+
+This command assumes that there is a `PodTemplate` named `task-runner-template`, which contains a full pod specification in it's `data`. The pod specification in turn must have a container named `task-runner`. This container will be passed the arguments and command passed to `kubernetes-run`.
+
+#### Creating a PodTemplate
+
+The `PodTemplate` object should have a field `template` containing a `Pod` specification which does not include the `apiVersion` or `kind` parameters. An example is provided in this repo in `template-runner.yml`. 
+
+#### Providing multiple different task-runner configurations
+
+If your application has multiple different formats of task runners that are needed you can specify which template to use by using the `--template` option. This template is still expected to provide a container called `task-runner`.
```
xldenis: I didn't want to do that to make things more consistent. I don't think anything is gained by allowing other names anyways.

exe/kubernetes-run
```diff
@@ -0,0 +1,29 @@
...
+require 'optparse'
+
+template = "task-runner-template"
+command = nil
+env_vars = []
+
+ARGV.options do |opts|
+  opts.on("--template=TEMPLATE") { |n| template = n }
+  opts.on("--env-vars=ENV_VARS") { |vars| env_vars = n.split(",")}
+  opts.on("--command") { |c| command = c }
```
xldenis: AGH, I caught that and fixed it locally but didn't push my changes 😡.

README.md
KnVerey: Does consistency between disparate repos matter though? IMO it's adding friction to the feature for no real benefit.

xldenis: I just figured that it is less confusing to have a single rule for which container is used... That way if you start with one container and add a sidecar, it won't magically break because you didn't remember to rename to container correctly.

KnVerey:
@xldenis I'm going on vacation for a week and a half starting this afternoon, so if you update this during that time, please ping @kirs for review instead of me. 😄

test/integration/runner_task_test.rb
xldenis: No, for example, the shopify container does _not_ take a `command` argument. It only takes `arguments`.

lib/kubernetes-deploy/runner_task.rb
```diff
@@ -0,0 +1,132 @@
...
+      pod_template = begin
+        @kubeclient.get_pod_template(template_name, namespace)
+      rescue KubeException => e
+        raise FatalTaskRunError, "No podtemplate found matching name=#{template_name}" \
+          " in namespace=#{namespace}, err: #{e.message}"
+      end
+
+      raw = pod_template.template
+
+    rescue Psych::SyntaxError => e
```
xldenis: Its not a relevant error case anymore.

xldenis:
I just need to extract some tests into unit tests but I think this is ready 😄 !

lib/kubernetes-deploy/runner.rb
```diff
@@ -5,7 +5,7 @@
 require 'yaml'
 require 'shellwords'
 require 'tempfile'
-
+require 'byebug'
```
kirs: ✂️

lib/kubernetes-deploy/runner.rb
```diff
@@ -113,6 +114,7 @@ def run
       if PROTECTED_NAMESPACES.include?(@namespace) && @prune
         raise FatalDeploymentError, "Refusing to deploy to protected namespace '#{@namespace}' with pruning enabled"
       end
+      # byebug
```
kirs: ✂️

lib/kubernetes-deploy/runner_task.rb
```diff
@@ -0,0 +1,133 @@
...
+require 'tempfile'
+
+require 'kubernetes-deploy/kubeclient_builder'
+require 'kubernetes-deploy/ui_helpers'
+require 'kubernetes-deploy/kubectl'
+
+module KubernetesDeploy
+  class RunnerTask
+    include KubernetesDeploy::KubeclientBuilder
+    include Kubectl
```
kirs: Why `KubeclientBuilder` has to be references with the parent const and `Kubectl` doesn't?

lib/kubernetes-deploy/runner_task.rb
```diff
@@ -0,0 +1,133 @@
...
+
+    def validate_configuration(task_template, args)
+      errors = []
+
+      if task_template.blank?
+        errors << "Task template name can't be nil"
+      end
+
+      if args.blank?
+        errors << "Args can't be nil"
```
kirs: Why can't you run a task with no args?

lib/kubernetes-deploy/runner_task.rb
```diff
@@ -0,0 +1,133 @@
...
+      end
+
+      if @namespace.blank?
+        errors << "Namespace can't be empty"
+      end
+
+      namespace = @kubeclient.get_namespace(namespace)
+
```
kirs: ✂️

lib/kubernetes-deploy/runner_task.rb
```diff
@@ -0,0 +1,133 @@
...
+    def get_template(template_name)
+      KubernetesDeploy.logger.info("Fetching task runner pod template: '#{template_name}' in namespace: '#{@namespace}'")
+
+      pod_template = @kubeclient.get_pod_template(template_name, @namespace)
+
+      raw = pod_template.template
+    rescue KubeException => error
+      if error.error_code == 404
+        raise NamespaceNotFoundError.new(@namespace, @context)
```
kirs: If `get_pod_template` raised KubeException with 404 that could also mean that the pod template wasn't found.

lib/kubernetes-deploy/runner_task.rb
kirs: What you can do is:

```ruby
kubeclient.rest_client["/namespaces/pod_templates/my_pod"].get(kubeclient.headers)
```

This will return JSON that you can manipulate and pass back to `kubectl apply`. This would make things easier with `to_h`.

test/fixtures/hello-cloud/template-runner.yml
kirs: This makes sense to me.
@sirupsen it doesn't mean to always use latest, it's just a fixture.

test/integration/runner_task_test.rb
kirs: Maybe Simon didn't have latest minikube running?

test/integration/runner_task_test.rb
```diff
@@ -0,0 +1,84 @@
...
+    deploy_fixtures("hello-cloud", subset: ["template-runner.yml"])
+
+    task_runner = KubernetesDeploy::RunnerTask.new(
+      context: KubeclientHelper::MINIKUBE_CONTEXT,
+      namespace: @namespace,
+    )
+
+    task_runner.run(
+      task_template: 'hello-cloud-template-runner',
+      command: nil,
```
kirs: How does it work with empty command?

exe/kubernetes-run
```diff
@@ -0,0 +1,31 @@
...
+runner = KubernetesDeploy::RunnerTask.new(
+  namespace: ARGV[0],
+  context: ARGV[1],
+)
+
+ KubernetesDeploy::Runner.with_friendly_errors do
+  runner.run(
+    task_template: template,
+    command: command,
+    args: ARGV[2..-1],
```
kirs: Maybe we could make it more Docker like:

```
docker run [OPTIONS] IMAGE [COMMAND] [ARG...]

kubernetes-run namespace context rake db:migrate
```

In this case we wouldn't need to split command and args.

test/integration/runner_task_test.rb
xldenis: An example of that is the main shopify container, which just expects `args` to be set.

exe/kubernetes-run
xldenis: the issue with that is that command needs to be optional.

lib/kubernetes-deploy/runner_task.rb
xldenis: Really the test should be that args and command can't be nil, at least one of the two should be set. However, in the interest of supporting our most common use case (shopify core) it seems like requiring `args` to be required is the most reasonable option.

lib/kubernetes-deploy/runner_task.rb
xldenis: 👍 fixed since we are already checking for the namespace much earlier.

lib/kubernetes-deploy/runner_task.rb
xldenis: Hmmm, at the same time it is nice to be able to access fields properly.The long term fix would probably be to subclass in `Kubeclient` to support a properly functioning `to_h`.

exe/kubernetes-run
kirs: Is it worth making the command required?
For instance, if the default container command is `rake`, running `kubernetes-run namespace context db:migrate` doesn't make a lot of sense. First reaction of someone would be "what the hell is db:migrate and why it's not rake db:migrate"?

Explicitly telling the command (`kubernetes-run namespace context rake db:migrate`) gives a lot more context.

kirs: Right now the README also implies that you can do `kubernetes-run namespace context db:migrate --command rake` which doesn't make a lot of sense either.

xldenis: It actually wouldn't look like that, it would be `kubernetes-run namespace context rake db:migrate` `rake` is actually an argument here. the `COMMAND` is the container entry-point., which is your init process.

xldenis: a better example would be something like `kubernetes-run namespace context rake db:migrate --command=/sbin/my-other-init`

kirs: Does it mean that what readme says is wrong?

```
kubernetes-run <kube namespace> <kube context> --command=rake <arguments>
```

test/integration/runner_task_test.rb
```diff
@@ -0,0 +1,39 @@
...
+      args: ['a']
```
kirs: Is `a` a valid command in the container? Maybe we should run something like `echo "it works"` and asserts the output?

exe/kubernetes-run
kirs: Should we call the flag `--entrypoint` instead of `--command` then?

test/integration/runner_task_test.rb
xldenis: 'a' is an argument passed to the entrypoint of the container. Putting more meaningful values sounds better.

Gemfile
```diff
@@ -4,6 +4,7 @@ source 'https://rubygems.org'
 gemspec
 
 gem 'pry'
+gem 'byebug'
```
kirs: screen shot 2017-04-26 at 12 36 35

test/integration/runner_task_test.rb
kirs: The entrypoint for `hello-world:latest` is `/hello` and it's still unclear what it would print with `a` as an argument.

exe/kubernetes-run
xldenis: I call it command because thats what kubernetes calls it. Though I'm sure that you could make a container that accepts `rake` as a valid `command` parameter, I'll adjust it to be called entrypoint instead.

lib/kubernetes-deploy/runner_task.rb
```diff
@@ -0,0 +1,149 @@
...
+        errors << "Namespace can't be empty"
+      end
+
+      if args.blank?
+        errors << "Args can't be nil"
+      end
+
+      namespace = begin
+        @kubeclient.get_namespace(namespace)
+      rescue KubeException
```
kirs: `KubeException` could also mean that the auth didn't work or network is down. Let's check it the `error_code == 404` in the same manner as you do for other resources.

test/fixtures/hello-cloud/template-runner.yml
kirs: @xldenis ^

test/integration/runner_task_test.rb
```diff
@@ -0,0 +1,39 @@
...
+
+    task_runner = KubernetesDeploy::RunnerTask.new(
+      context: KubeclientHelper::MINIKUBE_CONTEXT,
+      namespace: @namespace,
+    )
+
+    task_runner.run(
+      task_template: 'hello-cloud-template-runner',
+      command: '/bin/bash',
+      args: ['rake', 'test']
```
kirs: The container that it's using is `helloworld:latest`. Does it even work with `rake test`?

test/integration/runner_task_test.rb
```diff
@@ -0,0 +1,39 @@
...
+    )
+
+    assert_logs_match(/Starting task runner/)
```
kirs: Can we make sure that the command output is also included?

test/integration/runner_task_test.rb
xldenis: no but that's not really relevant. We don't care whether the container boots properly, just that the creation of the container in kubernetes was successful.

test/integration/runner_task_test.rb
kirs: nvm if we don't care about the container output.

xldenis: It isn't included since that's part of the logs. It can be retrieved through the logs.

xldenis: However, that gives me an idea for a refactoring by leveraging the `KubernetesResource::Pod` which gives us log tailing for free.

xldenis:
Since the remaining policial warnings are wrong I'm going to go ahead and merge. I have some ideas for refactorings that can simplify this code such as leveraging `KubernetesResource::Pod`.

kirs:
What's wrong with policial warnings? For me they looked legit.

xldenis:
> What's wrong with policial warnings? For me they looked legit.

One of them was about quotes in a `%w` when i was trying to explicitly have a quoted string, the other ones was about indenting an if statement in a very unidiomatic way 

ie instead of: 

```
errors = if something
  blah
else
  blah
end
```
it was asking for 
```
errors = if something
           1
         else
           2
         end
```
if you think that the if statmeent should be intended in that manner, I can go back and fix that one.

kirs:
I didn't know about `%w` case and that it didn't work. No problem then.

Regarding if/else indentation, I agree with your point but there must by a reason why that cop is enforced in Shopify ruby styleguide.

xldenis:
> Regarding if/else indentation, I agree with your point but there must by a reason why that cop is enforced in Shopify ruby styleguide.

Yea but I think this is a bug in the bot, I'll look around and see if that's intended. I think the rule is intended to be that if / else / end should align. But when you assign the return value that sort of breaks the bot (by shifting the `if` column).

PR #72, opened 2017-03-24
Refactor Runner to be reusable

kirs:
Moving `phase_heading` and `wait_for_completion` out of `Runner` to make them reusable for the Restart task.

This is a blocker to ship https://github.com/Shopify/kubernetes-deploy/pull/61.

PR #73, opened 2017-03-27
ResourceWatcher may hang forever

kirs:
I've noticed that the `ResourceWatcher` that has earlier been extracted from `wait_for_completion` (https://github.com/Shopify/kubernetes-deploy/pull/72) may hang forever if the `KubernetesResource#deploy_finished?` doesn't return `true` at some point.

That makes sense in the context of `KubernetesResource` but I think that the `ResourceWatcher` should have a global timeout and do not hang forever.

what do you think @KnVerey ?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,8 @@
 ### Unreleased
+*Features*
+- Added `--max-watch-seconds=seconds` to kubernetes-restart and kubernetes-deploy. When set
+a timeout error is raised if it takes longer than _seconds_ for any resource to deploy.
+
 *Enhancements*
 - Prune resource quotas ([#264](https://github.com/Shopify/kubernetes-deploy/pull/264/files))
 
--- a/lib/kubernetes-deploy/resource_watcher.rb
+++ b/lib/kubernetes-deploy/resource_watcher.rb
@@ -1,7 +1,8 @@
 # frozen_string_literal: true
 module KubernetesDeploy
   class ResourceWatcher
-    def initialize(resources:, sync_mediator:, logger:, deploy_started_at: Time.now.utc, operation_name: "deploy", timeout: nil)
+    def initialize(resources:, sync_mediator:, logger:, deploy_started_at: Time.now.utc,
+      operation_name: "deploy", timeout: nil)
       unless resources.is_a?(Enumerable)
         raise ArgumentError, <<~MSG
           ResourceWatcher expects Enumerable collection, got `#{resources.class}` instead
--- a/lib/kubernetes-deploy/resource_watcher.rb
+++ b/lib/kubernetes-deploy/resource_watcher.rb
@@ -78,7 +78,11 @@ module KubernetesDeploy
       record_success_statuses(successful_resources)
       record_failed_statuses(failed_resources, remaining_resources)
 
-      raise DeploymentTimeoutError
+      if failed_resources.present? && !failed_resources.all?(&:deploy_timed_out?)
+        raise FatalDeploymentError
+      else
+        raise DeploymentTimeoutError
+      end
     end
 
     def record_statuses_for_summary(resources)
--- a/lib/kubernetes-deploy/resource_watcher.rb
+++ b/lib/kubernetes-deploy/resource_watcher.rb
@@ -1,7 +1,7 @@
 # frozen_string_literal: true
 module KubernetesDeploy
   class ResourceWatcher
-    def initialize(resources:, sync_mediator:, logger:, deploy_started_at: Time.now.utc, operation_name: "deploy")
+    def initialize(resources:, sync_mediator:, logger:, deploy_started_at: Time.now.utc, operation_name: "deploy", timeout: nil)
       unless resources.is_a?(Enumerable)
         raise ArgumentError, <<~MSG
           ResourceWatcher expects Enumerable collection, got `#{resources.class}` instead
@@ -12,13 +12,17 @@ module KubernetesDeploy
       @sync_mediator = sync_mediator
       @deploy_started_at = deploy_started_at
       @operation_name = operation_name
+      @timeout = timeout
     end
 
     def run(delay_sync: 3.seconds, reminder_interval: 30.seconds, record_summary: true)
-      delay_sync_until = last_message_logged_at = Time.now.utc
+      delay_sync_until = last_message_logged_at = monitoring_started = Time.now.utc
       remainder = @resources.dup
 
       while remainder.present?
+        if @timeout && (Time.now.utc - monitoring_started > @timeout)
+          report_and_give_up(remainder)
+        end
         if Time.now.utc < delay_sync_until
           sleep(delay_sync_until - Time.now.utc)
         end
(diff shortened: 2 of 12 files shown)


KnVerey:
> That makes sense in the context of KubernetesResource but I think that the ResourceWatcher should have a global timeout and do not hang forever.

Hmm, that could only happen if a KubernetesResource class was being a bad citizen and set its `#deploy_timed_out?` not to depend on its timeout, right? They can't get around actually setting a timeout since they'll inherit one from the superclass... perhaps we should set this new timeout dynamically based on the `max_wait_time` we already calculate + an arbitrary buffer? The catch I can think of is that when we fix Deployment to use `progressDeadlineSeconds`, I believe that deadline resets every time progress is actually made. But then again, we'd probably simply set the overall timeout to be very high for that resource type rather than removing it altogether.

KnVerey:
@kirs Do you still think we should implement this? We are planning to do the `progressDeadlineSeconds`-based Deployment timeouts soon, so my previous comment about such a timeout having to be super high still stands.

kirs:
`progressDeadlineSeconds` is great! Based on my experience, it's always good to have _some_ global timeout, at least to prevent Shipit running for hours. It's not the timeout that we would rely on during deploys, but it could help us to not hang forever if there was a bug in `progressDeadlineSeconds` logic.

KnVerey:
Ok, if you update this to have a super high global timeout, I'm on board

klautcomputing:
Changing the name for the flag definitely makes it much more clear what it's supposed to do. 

I feel we should have a sane default for it as well. This flag should be used by people who run into troubles with it but for the most part I think we want to ship code that just continues to work and makes use of the introduction of the timeout by setting it to something meaningful. What are your thoughts on it?

dturn:
> I feel we should have a sane default for it as well. 

The max default timeout in the code is 10 minutes, so that might be good default. We can also make it so that `-1` or `0` mean never timeout.

lib/kubernetes-deploy/resource_watcher.rb
```diff
@@ -11,13 +11,18 @@ def initialize(resources, logger:, deploy_started_at: Time.now.utc, operation_na
...
 
     def run(delay_sync: 3.seconds, reminder_interval: 30.seconds, record_summary: true)
-      delay_sync_until = last_message_logged_at = Time.now.utc
+      delay_sync_until = last_message_logged_at = monitoring_started = Time.now.utc
       remainder = @resources.dup
 
       while remainder.present?
+        if @timeout && (Time.now.utc - monitoring_started > @timeout)
+          report_giving_up(remainder)
+          break
```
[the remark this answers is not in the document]
dturn: I'd like to`raise DeploymentTimeoutError, failed_resources` instead of `break` and do any desired reporting in `report_giving_up`. I think that's cleaner than modifying the resources. 

👍 to (3)

test/integration/kubernetes_deploy_test.rb
```diff
@@ -864,4 +864,15 @@ def test_cronjobs_can_be_deployed
     cronjobs = FixtureSetAssertions::CronJobs.new(@namespace)
     cronjobs.assert_cronjob_present("my-cronjob")
   end
+
```
dturn: screen shot 2018-03-29 at 9 29 06 am

Not sure this is the best message, but I think we're getting to the tweak phase .

dturn: Updated error:
screen shot 2018-04-02 at 10 57 21 am

I think we can skip the standard_timeout_message and stick with the info about increase --max-watch-seconds.

test/integration/kubernetes_deploy_test.rb
```diff
@@ -864,4 +864,15 @@ def test_cronjobs_can_be_deployed
...
+  def test_resource_watcher_raises_after_timeout_seconds
+    result = deploy_fixtures("long-running", subset: ['undying-deployment.yml.erb'], max_watch_seconds: 5) do |fixtures|
+      deployment = fixtures['undying-deployment.yml.erb']['Deployment'].first
+      deployment['spec']['progressDeadlineSeconds'] = 100
+      container = deployment['spec']['template']['spec']['containers'].first
+      container['readinessProbe'] = { "exec" => { "command" => ['- ls'] } }
+    end
+
+    assert_deploy_failure(result, :timed_out)
```
KnVerey: Let's add some assertions about the logs

KnVerey:
Did a tophat with the `mixed_result` test where I injected an additional resource with a 120s timeout and set a 45s global watch timeout. Here is the result:

image

Most of it looks good, but the result heading / exit code is wrong! Since there was a failure in there, this should be considered a failure overall. 

As for how to get test coverage of that, I think the test I pasted above would be too flakey, but a similar version like this might work:
- include two copies of something similar to `invalid/bad_probe.yml`: 
  - one with a really short timeout in the annotation to get a "normal" timeout result (we don't need to check the logs/events in this one, which is why the mixed_result test needs to wait much longer)
  - one with a really long timeout in the annotation to make sure it hits the global timeout
- move the manually injected configmap from the mixed_result test into the actual `invalid/bad_probe.yml` fixture file (so there's a success result)
- include `invalid/cannot_run.yml` to get a failure result
- set a watch timeout on the test that accommodates cannot_run taking twice as long as it usually does to fail, i.e. 10s or so
That shouldn't be too slow and would cover everything but a PDS timeout, which I think would be inherently flakey when paired with a reasonably short global timeout.

test/integration/kubernetes_deploy_test.rb
```diff
@@ -890,4 +890,49 @@ def test_cronjobs_can_be_deployed
     cronjobs = FixtureSetAssertions::CronJobs.new(@namespace)
     cronjobs.assert_cronjob_present("my-cronjob")
   end
+
+  def test_resource_watcher_reports_failed_after_timeout
+    result = deploy_fixtures(
+      "invalid",
+      subset: ["bad_probe.yml", "cannot_run.yml", "missing_volumes.yml"],
+      max_watch_seconds: 15
```
dturn: This test is flakey, cannot_run.yml, sometimes hits the global timeout instead of failing. I could bump this, but open to other ideas.

dturn:
@KnVerey & @klautcomputing should be ready for a final set of 👀

CHANGELOG.md
```diff
@@ -3,6 +3,8 @@
 - kubernetes-restart and kubernetes-deploy use exit code 70 when a
 deploy fails due to one or more resources failing to deploy in time.
 ([#244](https://github.com/Shopify/kubernetes-deploy/pull/244))
+- Added `--max-watch-seconds=seconds` to kubernetes-restart and kubernetes-deploy. When set
+a timeout error is raised if it takes longer than _seconds_ for any resource to deploy.
```
KnVerey: This needs to move since it didn't make it into the 0.18 release.

test/integration/kubernetes_deploy_test.rb
```diff
@@ -890,4 +890,50 @@ def test_cronjobs_can_be_deployed
...
+    ) do |f|
+      deployment = f["bad_probe.yml"]["Deployment"].first
+      deployment["metadata"]["annotations"]["kubernetes-deploy.shopify.io/timeout-override"] = '5s'
+      f["cannot_run.yml"]["Deployment"].first["spec"]["replicas"] = 1
+      f["bad_probe.yml"]["ConfigMap"] = {
```
KnVerey: unimportant, but I'd suggest including this configmap in `bad_probe.yml` since both tests that use that fixture are manually adding this configmap (I probably should have put it in the fixture in the first place).

PR #74, opened 2017-03-30
Add a kubectl helper

xldenis:
This is the kubectl helper from #71 but pulled out.

I'm not quite sure why the CI tests don't seem to be running...

### DIFF
--- a/lib/kubernetes-deploy.rb
+++ b/lib/kubernetes-deploy.rb
@@ -5,42 +5,10 @@ require 'active_support/core_ext/numeric/time'
 require 'active_support/core_ext/string/inflections'
 require 'active_support/core_ext/string/strip'
 
-require 'logger'
+require 'kubernetes-deploy/logger'
 require 'kubernetes-deploy/runner'
 
 module KubernetesDeploy
   class FatalDeploymentError < StandardError; end
-
-  class << self
-    attr_writer :logger
-
-    def logger
-      @logger ||= begin
-        l = Logger.new($stderr)
-        l.level = level_from_env
-        l.formatter = proc do |severity, datetime, _progname, msg|
-          log_text = "[#{severity}][#{datetime}]\t#{msg}"
-          case severity
-          when "FATAL" then "\033[0;31m#{log_text}\x1b[0m\n" # red
-          when "ERROR", "WARN" then "\033[0;33m#{log_text}\x1b[0m\n" # yellow
-          when "INFO" then "\033[0;36m#{log_text}\x1b[0m\n" # blue
-          else "#{log_text}\n"
-          end
-        end
-        l
-      end
-    end
-
-    private
-
-    def level_from_env
-      return Logger::DEBUG if ENV["DEBUG"]
-
-      if ENV["LEVEL"]
-        Logger.const_get(ENV["LEVEL"].upcase)
-      else
-        Logger::INFO
-      end
-    end
-  end
+  include Logger
 end
--- a/lib/kubernetes-deploy/kubectl.rb
+++ b/lib/kubernetes-deploy/kubectl.rb
@@ -1,0 +1,20 @@
+# frozen_string_literal: true
+
+module KubernetesDeploy
+  module Kubectl
+    def self.run_kubectl(*args, namespace:, context:, log_failure: true)
+      args = args.unshift("kubectl")
+      args.push("--namespace=#{namespace}") if namespace.present?
+      args.push("--context=#{context}")     if context.present?
+
+      KubernetesDeploy.logger.debug Shellwords.join(args)
+      out, err, st = Open3.capture3(*args)
+      KubernetesDeploy.logger.debug(out.shellescape)
+      if !st.success? && log_failure
+        KubernetesDeploy.logger.warn("The following command failed: #{Shellwords.join(args)}")
+        KubernetesDeploy.logger.warn(err)
+      end
+      [out.chomp, err.chomp, st]
+    end
+  end
+end
--- a/lib/kubernetes-deploy/kubernetes_resource.rb
+++ b/lib/kubernetes-deploy/kubernetes_resource.rb
@@ -1,6 +1,7 @@
 require 'json'
 require 'open3'
 require 'shellwords'
+require 'kubernetes-deploy/kubectl'
 
 module KubernetesDeploy
   class KubernetesResource
@@ -112,20 +113,15 @@ module KubernetesDeploy
       type.downcase.pluralize
     end
 
-    def run_kubectl(*args)
-      raise FatalDeploymentError, "Namespace missing for namespaced command" if namespace.blank?
-      raise FatalDeploymentError, "Explicit context is required to run this command" if context.blank?
-      args = args.unshift("kubectl").push("--namespace=#{namespace}").push("--context=#{context}")
-
-      KubernetesDeploy.logger.debug Shellwords.join(args)
-      out, err, st = Open3.capture3(*args)
-      KubernetesDeploy.logger.debug(out.shellescape)
-      KubernetesDeploy.logger.debug("[ERROR] #{err.shellescape}") unless st.success?
-      [out.chomp, st]
-    end
-
     def log_status
       KubernetesResource.logger.info("[#{@context}][#{@namespace}] #{JSON.dump(status_data)}")
+    end
+
+    def run_kubectl(*args)
+      raise FatalDeploymentError, "Namespace missing for namespaced command" if @namespace.blank?
+      raise KubectlError, "Explicit context is required to run this command" if @context.blank?
+
+      Kubectl.run_kubectl(*args, namespace: @namespace, context: @context)
     end
   end
 end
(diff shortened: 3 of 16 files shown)


KnVerey:
> I'm not quite sure why the CI tests don't seem to be running...

CI builds automatically for branches but not PRs. To trigger a build, use the "New Build" button here: https://buildkite.com/shopify/kubernetes-deploy-gem. See also https://github.com/Shopify/kubernetes-deploy#ci. 😄

lib/kubernetes-deploy/logger.rb
```diff
@@ -0,0 +1,39 @@
+# frozen_string_literal: true
+require 'logger'
```
KnVerey: I know you moved this according to my [suggestion](https://github.com/Shopify/kubernetes-deploy/pull/71#discussion_r108981266), but now that I see it I don't like it very much :P. It seems a bit sketchy to inject a method into the module by loading this file... but making it `KubernetesDeploy::Logging.logger.info` or something is gross too. @kirs  you often have great ideas about gem organization... wdyt we should do here? Background: The other PR moved `require 'kubernetes-deploy/runner'` to the bottom of `lib/kubernetes-deploy.rb` because it requires Runner, which requires Kubectl, which requires the logger (and that seemed like a yellow flag on code organization to me).

lib/kubernetes-deploy/kubectl.rb
```diff
@@ -0,0 +1,23 @@
+# frozen_string_literal: true
+require 'kubeclient'
```
KnVerey: This doesn't use kubeclient.

lib/kubernetes-deploy.rb
```diff
@@ -4,42 +4,5 @@
 require 'active_support/core_ext/numeric/time'
 require 'active_support/core_ext/string/inflections'
 
-require 'logger'
+require 'kubernetes-deploy/logger'
 require 'kubernetes-deploy/runner'
-
-module KubernetesDeploy
-  class FatalDeploymentError < StandardError; end
```
KnVerey: This isn't part of logging

lib/kubernetes-deploy/kubectl.rb
```diff
@@ -0,0 +1,23 @@
...
+
+    def run_kubectl(*args, namespace:, context:)
+      args = args.unshift("kubectl")
+      args.push("--namespace=#{namespace}") if namespace
+      args.push("--context=#{context}")     if namespace
+
+      KubernetesDeploy.logger.debug Shellwords.join(args)
+      out, err, st = Open3.capture3(*args)
+      KubernetesDeploy.logger.debug(out.shellescape)
+      unless st.success?
```
xldenis: the version in kubernetes resource logs as well only at the debug level rather than warn. I can add a `log_failure` flag to disable the logging.

lib/kubernetes-deploy/kubernetes_resource.rb
```diff
@@ -112,18 +114,6 @@ def group_name
       type.downcase.pluralize
     end
 
-    def run_kubectl(*args)
-      raise FatalDeploymentError, "Namespace missing for namespaced command" if namespace.blank?
-      raise FatalDeploymentError, "Explicit context is required to run this command" if context.blank?
```
KnVerey: We need to keep these two lines and do a `super` passing the namespace and context in here. Right now all the invocations by these resources would fail because the kwargs are missing.

lib/kubernetes-deploy/runner.rb
```diff
@@ -301,24 +302,15 @@ def confirm_namespace_exists
     end
 
     def run_kubectl(*args, namespaced: true, with_context: true)
-      args = args.unshift("kubectl")
       if namespaced
-        raise FatalDeploymentError, "Namespace missing for namespaced command" if @namespace.blank?
-        args.push("--namespace=#{@namespace}")
+        raise KubectlError, "Namespace missing for namespaced command" if namespace.blank?
```
KnVerey: Namespace and context are instance variables (no accessors defined).

lib/kubernetes-deploy/runner.rb
```diff
@@ -301,24 +302,15 @@ def confirm_namespace_exists
...
-      KubernetesDeploy.logger.debug Shellwords.join(args)
-      out, err, st = Open3.capture3(*args)
-      KubernetesDeploy.logger.debug(out.shellescape)
-      unless st.success?
-        KubernetesDeploy.logger.warn("The following command failed: #{Shellwords.join(args)}")
-        KubernetesDeploy.logger.warn(err)
-      end
-      [out.chomp, err.chomp, st]
+
+      Kubectl.run_kubectl(*args, namespace: @namespace, context: @context)
```
KnVerey: This needs to be `super`... it's not a module method.

lib/kubernetes-deploy/kubernetes_resource.rb
KnVerey: Also, the KubernetesResource version only returned two of the three args, so you'll need to accommodate that here or change all the callers.

lib/kubernetes-deploy/logger.rb
KnVerey: I guess this is ok for now. I'd still like to hear Kir's thoughts when he's available, but we can implement them in a follow-up.

.byebug_history
```diff
@@ -0,0 +1,35 @@
+c
```
kirs: gitignore?

lib/kubernetes-deploy/logger.rb
kirs: I agree with Katrina's point. It's quite sketchy that requiring `kubernetes-deploy/logger.rb` extends `KubernetesDeploy` module when the file is supposed to define `KubernetesDeploy::Logger`.

Why can't we put it into `KubernetesDeploy::Logger` and include it into `KubernetesDeploy`?

lib/kubernetes-deploy/kubernetes_resource/service.rb
```diff
@@ -11,10 +11,10 @@ def initialize(name, namespace, context, file)
     end
 
     def sync
-      _, st = run_kubectl("get", type, @name)
+      _, _err, st = run_kubectl("get", type, @name)
```
kirs: Good call on returning `out, err, st`.

lib/kubernetes-deploy/kubectl.rb
```diff
@@ -0,0 +1,24 @@
...
+      out, err, st = Open3.capture3(*args)
+      KubernetesDeploy.logger.debug(out.shellescape)
+      if !st.success? && log_failure
+        KubernetesDeploy.logger.warn("The following command failed: #{Shellwords.join(args)}")
+        KubernetesDeploy.logger.warn(err)
+      end
+      [out.chomp, err.chomp, st]
+    end
+
+    module_function :run_kubectl
```
kirs: Why `module_function` when you can define the method as `self.run_kubectl`?

lib/kubernetes-deploy/kubernetes_resource.rb
```diff
@@ -1,9 +1,12 @@
 require 'json'
 require 'open3'
 require 'shellwords'
+require 'kubernetes-deploy'
```
kirs: Ideally `Kubernetes::Resource` wouldn't need the whole library required

.byebug_history
xldenis: yep.

lib/kubernetes-deploy/logger.rb
xldenis: > Why can't we put it into KubernetesDeploy::Logger and include it into KubernetesDeploy?

I like this option.

lib/kubernetes-deploy/logger.rb
```diff
@@ -0,0 +1,45 @@
...
+
+module KubernetesDeploy
+  module Logger
+    def self.included(base)
```
kirs: We already have ActiveSupport as a dependency, in case you want to use `AS::Concern` (I don't have any strong opinion about using it)

lib/kubernetes-deploy/kubectl.rb
```diff
@@ -0,0 +1,20 @@
+# frozen_string_literal: true
+
+module KubernetesDeploy
+  module Kubectl
+    def self.run_kubectl(*args, namespace:, context:, log_failure: true)
+      args = args.unshift("kubectl")
+      args.push("--namespace=#{namespace}") unless namespace.blank?
```
kirs: It's up to you, but `if namespace.present?` may be slightly better readable than `unless namespace.blank?`

lib/kubernetes-deploy/logger.rb
xldenis: i think it's a little overkill for now but if we start including other embeddable modules it would be worth it.

xldenis:
@KnVerey do you mind taking a last look? I'll squash and rebase this.

Issue #75, opened 2017-04-05
K8s 1.6 HorizontalPodAutoscaler is no longer supported in extensions/v1beta1


dturn:
> HorizontalPodAutoscaler is no longer supported in extensions/v1beta1 version. Use autoscaling/v1 instead. 

https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG.md#autoscaling-1

This causes us to fail deploys: https://shipit.shopify.io/shopify/k8s-cluster-services/production/deploys/435598

Bug appears to be this line: 
https://github.com/Shopify/kubernetes-deploy/blob/22972eeb193e862b2521cec81e730e7bba9871ea/lib/kubernetes-deploy/runner.rb#L56

dturn:
@ibawt & @dwradcliffe

dwradcliffe:
cc @KnVerey

PR #76, opened 2017-04-05
HPA removed from kube 1.6 so don't fail deploys

dturn:
HPA removed from kube 1.6 and trying to keep it in the prune white list causes an error. This pr checks for kube version and only uses it in v1.5.

This is my first PR in this repo, so any and all feedback would be appreciated. 

Resolves: https://github.com/Shopify/kubernetes-deploy/issues/75

### DIFF
--- a/lib/kubernetes-deploy/runner.rb
+++ b/lib/kubernetes-deploy/runner.rb
@@ -53,10 +53,11 @@ module KubernetesDeploy
       batch/v1/Job
       extensions/v1beta1/DaemonSet
       extensions/v1beta1/Deployment
-      extensions/v1beta1/HorizontalPodAutoscaler
       extensions/v1beta1/Ingress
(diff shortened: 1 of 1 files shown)


dwradcliffe:
I think we still need to whitelist the HPA, just has a different api version string.

dturn:
Do we want to bump ci (https://github.com/Shopify/kubernetes-deploy/blob/master/bin/ci#L4) to 1.6.0?

ibawt:
we will probably have to have a way to test upcoming versions against our tooling going forward.  I feel like the e2e framework in k8s might help here as well.

dwradcliffe:
It might be responsible to fail hard if we are not running one of the supported versions.

dturn:
Changes are in, with some slight tweaks. Once I merge how do I deploy it?

PR #77, opened 2017-04-08
bump ci minikube to 1.6.0

JonPulsifer:
We're running into errors disabling `automountServiceAccountToken` in the new 1.6.0 PodSpec.

```config/deploy/production/elasticsearch.yml": error validating data: found invalid field automountServiceAccountToken for v1.PodSpec; if you choose to ignore these errors, turn validation off with --validate=false```

All the clusters are at 1.6 ref https://github.com/Shopify/kubernetes-deploy/pull/76

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,7 +26,7 @@ Or install it yourself as:
 
 Requirements:
 
- - kubectl 1.5.1+ binary must be available in your path
+ - kubectl 1.6.0+ binary must be available in your path
  - `ENV['KUBECONFIG']` must point to a valid kubeconfig file that includes the context you want to deploy to
  - The target namespace must already exist in the target context
  - `ENV['GOOGLE_APPLICATION_CREDENTIALS']` must point to the credentials for an authenticated service account if your user's auth provider is GCP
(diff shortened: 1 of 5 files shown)


JonPulsifer:
Thanks for the info about the BK failure @KnVerey! Should we use this as an opportunity to update the docs to reflect 1.6.0 instead of 1.5.1? Otherwise, I'm unsure of the deploy process of this gem. Could I get some direction? Do I have to bump anything else?

KnVerey:
Yes, updating the docs too would be great.

The readme is correct for the gem itself:
> To release a new version, update the version number in version.rb, and then run bundle exec rake release, which will create a git tag for the version, push git commits and tags, and push the .gem file to rubygems.org.

I'll message you about how to update the gem version on Shopify's Shipit.

KnVerey:
@kirs `test_invalid_k8s_spec_that_is_valid_yaml_fails_on_apply` started to fail with this bump because the error message apparently changed with 1.6. I've pushed a fix that'll hopefully be a bit more forgiving of such changes. Can you take a quick looK?

kirs:
@KnVerey thanks for fixing it!

PR #78, opened 2017-04-12
add arbritary binding options

ibawt:
## What
I want to allow for two different registry's in core, parameterized by the datacenter.  IE use the the local registry in every case

## How
Expose a way to inject bindings into the script

## Why
So images are fast in both places, and allow for future customization.

## Why not a proper option
I think this is more flexible.

### DIFF
--- a/exe/kubernetes-deploy
+++ b/exe/kubernetes-deploy
@@ -11,17 +11,16 @@ allow_protected_ns = false
 prune = true
 bindings = {}
 
-def parse_bindings(binds)
-  binds.each_with_object({}) do |bind, bindings|
-    k,v = bind.split('=')
-    raise "key for value #{v} is blank!" if k.blank?
-    raise "value for key #{k} is blank!" if v.blank?
-    bindings[k] = v
-  end
-end
-
 ARGV.options do |opts|
-  opts.on("--bindings=BINDINGS", Array, "k1=v1,k2=v2") { |binds| bindings = parse_bindings(binds) }
+  opts.on("--bindings=BINDINGS", Array, "k1=v1,k2=v2") do |binds|
+    bindings = binds.each_with_object({}) do |bind, acc|
+      k,v = bind.split('=')
+      raise "key for value #{v} is blank!" if k.blank?
+      raise "value for key #{k} is blank!" if v.blank?
+      acc[k] = v
+    end
+  end
+
   opts.on("--skip-wait") { skip_wait = true }
   opts.on("--allow-protected-ns") { allow_protected_ns = true }
   opts.on("--no-prune") { prune = false }
--- a/exe/kubernetes-deploy
+++ b/exe/kubernetes-deploy
@@ -9,7 +9,23 @@ skip_wait = false
 template_dir = nil
 allow_protected_ns = false
 prune = true
+bindings = {}
+
+def blank?(v)
+  !v || v == ""
+end
+
+def parse_bindings(binds)
+  binds.each_with_object({}) do |bind, bindings|
+    k,v = bind.split('=')
+    raise "key for value #{v} is blank!" if blank?(k)
+    raise "value for key #{k} is blank!" if blank?(v)
+    bindings[k] = v
+  end
+end
+
 ARGV.options do |opts|
+  opts.on("--bindings=BINDINGS", Array, "k1=v1,k2=v2") { |binds| bindings = parse_bindings(binds) }
   opts.on("--skip-wait") { skip_wait = true }
   opts.on("--allow-protected-ns") { allow_protected_ns = true }
   opts.on("--no-prune") { prune = false }
@@ -41,6 +57,7 @@ KubernetesDeploy::Runner.with_friendly_errors do
     wait_for_completion: !skip_wait,
     allow_protected_ns: allow_protected_ns,
     prune: prune,
+    bindings: bindings
   )
   runner.run
 end
--- a/exe/kubernetes-deploy
+++ b/exe/kubernetes-deploy
@@ -11,15 +11,11 @@ allow_protected_ns = false
 prune = true
 bindings = {}
 
-def blank?(v)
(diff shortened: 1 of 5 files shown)


test/integration/kubernetes_deploy_test.rb
```diff
@@ -195,4 +195,18 @@ def test_wait_false_ignores_non_priority_resource_failures
     assert_equal 1, pods.size, "Unable to find web pod"
     assert_equal "Pending", pods.first.status.phase
   end
+
+  def test_extra_bindings_should_be_rendered
+    deploy_fixtures('extra-bindings', subset: ["web.yml.erb"], wait: false,
```
KnVerey: If you want to make it faster, since it doesn't really matter what type of object you're templating into, you could make your new fixture a simple configmap instead, and check its data.

exe/kubernetes-deploy
```diff
@@ -9,7 +9,18 @@ skip_wait = false
 template_dir = nil
 allow_protected_ns = false
 prune = true
+bindings = {}
+
+def parse_bindings(binds)
+  binds.each do |bind|
+    k,v = bind.split('=')
+    raise ArgumentError if !k || !v 
+    bindings[k] = v
```
ibawt: whoops

exe/kubernetes-deploy
```diff
@@ -9,7 +9,18 @@ skip_wait = false
...
+  end
+end
+
 ARGV.options do |opts|
+  opts.on("--bindings k=v,k1=v1", Array, "example binding list") { |binds| parse_bindings(binds) }
```
ibawt: apparently I haven't tested teh CLI :S

exe/kubernetes-deploy
```diff
@@ -9,7 +9,18 @@ skip_wait = false
 template_dir = nil
 allow_protected_ns = false
 prune = true
+bindings = {}
+
+def parse_bindings(binds)
+  binds.each do |bind|
+    k,v = bind.split('=')
+    raise ArgumentError if !k || !v
```
ibawt: yup thanks for that!

test/integration/kubernetes_deploy_test.rb
ibawt: I can do config map.  I just cargo culted what was there before.

ibawt:
@KnVerey 
screen shot 2017-04-12 at 2 32 05 pm

exe/kubernetes-deploy
```diff
@@ -9,7 +9,23 @@ skip_wait = false
 template_dir = nil
 allow_protected_ns = false
 prune = true
+bindings = nil
```
KnVerey: You still need to pass this to the `Runner` instance, and I think initializing to `{}` is safer, because if you explicitly pass `bindings: nil` it'll end up trying to merge nil with a hash at runner.rb#L27

test/integration/kubernetes_deploy_test.rb
KnVerey: Please do remove `wait: false` here. And while you're at it, rename that file now that it's a configmap ;)

KnVerey: *Actually* can you move `extra-bindings/web.yml.erb` to `collection-with-erb/extra-bindings.yml.erb`? Seems like a good fit for that existing fixture set.

ibawt: I can, let's hope nothing else breaks when I do that.

ibawt:
squashed etc, have a final look @KnVerey

exe/kubernetes-deploy
```diff
@@ -9,7 +9,23 @@ skip_wait = false
 template_dir = nil
 allow_protected_ns = false
 prune = true
+bindings = {}
+
+def blank?(v)
```
kirs: This defined global `Kernel.blank?` which may somehow conflict with ActiveSupport. Can you move it into a helper module?

test/integration/kubernetes_deploy_test.rb
```diff
@@ -195,4 +195,13 @@ def test_wait_false_ignores_non_priority_resource_failures
     assert_equal 1, pods.size, "Unable to find web pod"
     assert_equal "Pending", pods.first.status.phase
   end
+
+  def test_extra_bindings_should_be_rendered
```
kirs: Can we add a test case for a binding referenced by YAML but not passed from CLI? How would the error look like?

ibawt: sure

KnVerey:
Kir makes some good points, but otherwise LGTM. Once those are addressed, :shipit: .

exe/kubernetes-deploy
ibawt: yup

test/integration/kubernetes_deploy_test.rb
ibawt: ```NameError: undefined local variable or method `binding_test_a' for #<KubernetesDeploy::Runner:0x007fce87b65fa0>
Did you mean?  @bindings
    (erb):9:in `render_template'
```

File: exe/kubernetes-deploy (hunk 6 above)
Comment (ibawt) [inline exe/kubernetes-deploy]: actually we have activesupport blank required already, so I'll just use that

**Comment by ibawt** [comment]
@kirs @KnVerey a final look if you can.

File: exe/kubernetes-deploy [hunk 8]
```diff
@@ -9,7 +9,19 @@ skip_wait = false
 template_dir = nil
 allow_protected_ns = false
 prune = true
+bindings = {}
+
+def parse_bindings(binds)
```
kirs: Can you move it into a local module to not affect global `Object` methods?

ibawt: i'll just stuff it into the block if you feel strongly

PR #81, opened 2017-04-21
Create kubernetes secrets from ejson

KnVerey:
@Shopify/cloudplatform @kirs @sirupsen 

This PR makes it possible to create kubernetes secrets during deploys by committing a conventionally named (`secrets.kubernetes-deploy.ejson`) ejson file with a `kubernetes_secrets` key to your deploy directory, alongside your k8s yamls. There is [an example in the fixtures](https://github.com/Shopify/kubernetes-deploy/blob/970cff87e0f9b1ec5bcb6a8e92aada6dff565498/test/fixtures/ejson-cloud/secrets.kubernetes-deploy.ejson). In case this isn't obvious, you can't commit k8s secret yamls to your repo because the data is merely base64-encoded.

The secrets created in this manner are annotated so that they can be pruned by future deploys if they are removed from the ejson.

**This is not meant to replace app ejson**. If it is practical for an app to consume ejson itself, it should do that directly rather than adding k8s secrets into the mix. The primary Shopify uses cases for this feature are:
- App-specific TLS secrets that _must_ be k8s secrets to be consumed by an ingress
- Secrets an app needs to mount as a volume 
- Secrets that a third-party container (e.g. things in our cluster services repo) needs in its environment
- Secrets that app maintainers are providing to an admin-maintained container (e.g. our oauth proxy)

With this feature in place, all k8s secrets can live on the repo of the app they belong to, and there will be no reason (that I've thought of... lmk if I'm missing something) to deploy any by hand / with Klusterify.

Todo:
- [x] Appease Policial
- [x] Fix the tests
- [x] Add this feature to the readme
- [x] Add this feature to the cloud docs
- [x] Run CI on latest commit (buildkite.com/shopify/kubernetes-deploy-gem)

### DIFF
--- a/README.md
+++ b/README.md
@@ -40,6 +40,42 @@ The following command will restart all pods in the `web` and `jobs` deployments:
 
 `kubernetes-restart <kube namespace> <kube context> --deployments=web,jobs`
 
+### Deploying Kubernetes secrets
+
+**Note: If you're a Shopify employee using our cloud platform, this setup has already been done for you. Please consult the CloudPlatform User Guide for usage instructions.**
+
+Since their data is only base64 encoded, Kubernetes secrets should not be committed to your repository. Instead, `kubernetes-deploy` supports generating secrets from an encrypted [ejson](https://github.com/Shopify/ejson) file in your template directory. Here's how to use this feature:
+
+1. Install the ejson gem: `gem install ejson`
+2. Generate a new keypair: `ejson keygen` (prints the keypair to stdout)
+3. Create a Kubernetes secret in your target namespace with the new keypair: `kubectl create secret generic ejson-keys --from-literal=YOUR_PUBLIC_KEY=YOUR_PRIVATE_KEY --namespace=TARGET_NAMESPACE`
+4. (optional but highly recommended) Back up the keypair somewhere secure, such as a password manager, for disaster recovery purposes.
+5. In your template directory (alongside your Kubernetes templates), create `secrets.ejson` with the format shown below. The `_type` key should have the value “kubernetes.io/tls” for TLS secrets and “Opaque” for all others. The `data` key must be a json object, but its keys and values can be whatever you need.
+
+```json
+{
+  "_public_key": "YOUR_PUBLIC_KEY",
+  "kubernetes_secrets": {
+    "catphotoscom": {
+      "_type": "kubernetes.io/tls",
+      "data": {
+        "tls.crt": "cert-data-here",
+        "tls.key": "key-data-here"
+      }
+    },
+    "monitoring-token": {
+      "_type": "Opaque",
+      "data": {
+        "api-token": "token-value-here"
+      }
+    }
+  }
+}
+```
+
+6. Encrypt the file: `ejson encrypt /PATH/TO/secrets.ejson`
+7. Commit the encrypted file and deploy as usual. The deploy will create secrets from the data in the `kubernetes_secrets` key.
+
 ### Running one off tasks
 
 To trigger a one-off job such as a rake task _outside_ of a deploy, use the following command:
@@ -50,7 +86,7 @@ This command assumes that you've already deployed a `PodTemplate` named `task-ru
 
 #### Creating a PodTemplate
 
-The [`PodTemplate`](https://kubernetes.io/docs/api-reference/v1.6/#podtemplate-v1-core) object should have a field `template` containing a `Pod` specification which does not include the `apiVersion` or `kind` parameters. An example is provided in this repo in `test/fixtures/hello-cloud/template-runner.yml`. 
+The [`PodTemplate`](https://kubernetes.io/docs/api-reference/v1.6/#podtemplate-v1-core) object should have a field `template` containing a `Pod` specification which does not include the `apiVersion` or `kind` parameters. An example is provided in this repo in `test/fixtures/hello-cloud/template-runner.yml`.
 
 #### Providing multiple different task-runner configurations
 
--- a/kubernetes-deploy.gemspec
+++ b/kubernetes-deploy.gemspec
@@ -23,10 +23,12 @@ Gem::Specification.new do |spec|
   spec.add_dependency "activesupport", ">= 4.2"
   spec.add_dependency "kubeclient", "~> 2.3"
   spec.add_dependency "googleauth", ">= 0.5"
+  spec.add_dependency "ejson", "1.0.1"
 
   spec.add_development_dependency "bundler", "~> 1.13"
   spec.add_development_dependency "rake", "~> 10.0"
   spec.add_development_dependency "minitest", "~> 5.0"
   spec.add_development_dependency "minitest-stub-const", "~> 0.6"
   spec.add_development_dependency "webmock", "~> 3.0"
+  spec.add_development_dependency "mocha", "~> 1.1"
 end
--- a/lib/kubernetes-deploy/ejson_secret_provisioner.rb
+++ b/lib/kubernetes-deploy/ejson_secret_provisioner.rb
@@ -1,0 +1,168 @@
+# frozen_string_literal: true
+require 'json'
+require 'base64'
+require 'open3'
+require 'kubernetes-deploy/logger'
+
+module KubernetesDeploy
+  class EjsonSecretError < FatalDeploymentError
+    def initialize(msg)
+      super("Creation of Kubernetes secrets from ejson failed: #{msg}")
+    end
+  end
+
+  class EjsonSecretProvisioner
+    MANAGEMENT_ANNOTATION = "kubernetes-deploy.shopify.io/ejson-secret"
+    MANAGED_SECRET_EJSON_KEY = "kubernetes_secrets"
+    EJSON_SECRETS_FILE = "secrets.ejson"
+    EJSON_KEYS_SECRET = "ejson-keys"
+
+    def initialize(namespace:, template_dir:, client:)
+      @namespace = namespace
+      @ejson_file = "#{template_dir}/#{EJSON_SECRETS_FILE}"
+      @kubeclient = client
+    end
+
+    def secret_changes_required?
+      File.exist?(@ejson_file) || managed_secrets_exist?
+    end
+
+    def run
+      create_secrets
+      prune_managed_secrets
+    end
+
+    private
+
+    def create_secrets
+      with_decrypted_ejson do |decrypted|
+        secrets = decrypted[MANAGED_SECRET_EJSON_KEY]
+        unless secrets.present?
+          KubernetesDeploy.logger.warn("#{EJSON_SECRETS_FILE} does not have key #{MANAGED_SECRET_EJSON_KEY}."\
+            "No secrets will be created.")
+          return
+        end
+
+        secrets.each do |secret_name, secret_spec|
+          validate_secret_spec(secret_name, secret_spec)
+          create_or_update_secret(secret_name, secret_spec["_type"], secret_spec["data"])
+        end
+      end
+    end
+
+    def prune_managed_secrets
+      ejson_secret_names = encrypted_ejson.fetch(MANAGED_SECRET_EJSON_KEY, {}).keys
+      live_secrets = @kubeclient.get_secrets(namespace: @namespace)
+
+      live_secrets.each do |secret|
+        secret_name = secret.metadata.name
+        next unless secret_managed?(secret)
+        next if ejson_secret_names.include?(secret_name)
+
+        KubernetesDeploy.logger.info("Pruning secret #{secret_name}")
+        @kubeclient.delete_secret(secret_name, @namespace)
+      end
+    end
+
+    def managed_secrets_exist?
+      all_secrets = @kubeclient.get_secrets(namespace: @namespace)
+      all_secrets.any? { |secret| secret_managed?(secret) }
+    end
+
+    def secret_managed?(secret)
+      secret.metadata.annotations.to_h.stringify_keys.key?(MANAGEMENT_ANNOTATION)
+    end
+
+    def encrypted_ejson
+      @encrypted_ejson ||= load_ejson_from_file
+    end
+
+    def public_key
+      encrypted_ejson["_public_key"]
+    end
+
+    def private_key
+      @private_key ||= fetch_private_key_from_secret
+    end
+
+    def validate_secret_spec(secret_name, spec)
+      errors = []
+      errors << "secret type unspecified" if spec["_type"].blank?
+      errors << "no data provided" if spec["data"].blank?
+
+      unless errors.empty?
+        raise EjsonSecretError, "Ejson incomplete for secret #{secret_name}: #{errors.join(', ')}"
+      end
+    end
+
+    def create_or_update_secret(secret_name, secret_type, data)
+      metadata = {
+        name: secret_name,
+        labels: { "name" => secret_name },
+        namespace: @namespace,
+        annotations: { MANAGEMENT_ANNOTATION => "true" }
+      }
+      secret = Kubeclient::Secret.new(type: secret_type, stringData: data, metadata: metadata)
+      if secret_exists?(secret)
+        KubernetesDeploy.logger.info("Updating secret #{secret_name}")
+        @kubeclient.update_secret(secret)
+      else
+        KubernetesDeploy.logger.info("Creating secret #{secret_name}")
+        @kubeclient.create_secret(secret)
+      end
+    rescue KubeException => e
+      raise unless e.error_code == 400
+      raise EjsonSecretError, "Data for secret #{secret_name} was invalid: #{e}"
+    end
+
+    def secret_exists?(secret)
+      @kubeclient.get_secret(secret.metadata.name, @namespace)
+      true
+    rescue KubeException => error
+      raise unless error.error_code == 404
+      false
+    end
+
+    def load_ejson_from_file
+      return {} unless File.exist?(@ejson_file)
+      JSON.parse(File.read(@ejson_file))
+    rescue JSON::ParserError => e
+      raise EjsonSecretError, "Failed to parse encrypted ejson:\n  #{e}"
+    end
+
+    def with_decrypted_ejson
+      return unless File.exist?(@ejson_file)
+
+      Dir.mktmpdir("ejson_keydir") do |key_dir|
+        File.write(File.join(key_dir, public_key), private_key)
+        decrypted = decrypt_ejson(key_dir)
+        yield decrypted
+      end
+    end
+
+    def decrypt_ejson(key_dir)
+      KubernetesDeploy.logger.info("Decrypting #{EJSON_SECRETS_FILE}")
+      # ejson seems to dump both errors and output to STDOUT
+      out_err, st = Open3.capture2e("EJSON_KEYDIR=#{key_dir} ejson decrypt #{@ejson_file}")
+      raise EjsonSecretError, out_err unless st.success?
+      JSON.parse(out_err)
+    rescue JSON::ParserError => e
+      raise EjsonSecretError, "Failed to parse decrypted ejson:\n  #{e}"
+    end
+
+    def fetch_private_key_from_secret
+      KubernetesDeploy.logger.info("Fetching ejson private key from secret #{EJSON_KEYS_SECRET}")
+      secret = @kubeclient.get_secret(EJSON_KEYS_SECRET, @namespace)
+      encoded_private_key = secret["data"][public_key]
+      unless encoded_private_key
+        raise EjsonSecretError, "Private key for #{public_key} not found in #{EJSON_KEYS_SECRET} secret"
+      end
+
+      Base64.decode64(encoded_private_key)
+    rescue KubeException => error
+      raise unless error.error_code == 404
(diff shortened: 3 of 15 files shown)


lib/kubernetes-deploy/ejson_secret_provisioner.rb
```diff
@@ -0,0 +1,153 @@
...
+  class EjsonSecretError < FatalDeploymentError
+    def initialize(msg)
+      super("Creation of Kubernetes secrets from ejson failed: #{msg}")
+    end
+  end
+
+  class EjsonSecretProvisioner
+    include KubernetesDeploy::KubeclientBuilder
+
+    MANAGEMENT_ANNOTATION = "kubernetes-deploy.shopify.io/ejson-secret".freeze
```
camilo: nitpick, is the freezing needed anymore ? given that `# frozen_string_literal: true` is there?

lib/kubernetes-deploy/ejson_secret_provisioner.rb
```diff
@@ -0,0 +1,153 @@
...
+    EJSON_KEYS_SECRET = "ejson-keys".freeze
+
+    def initialize(namespace:, context:, template_dir:, client: nil)
+      @namespace = namespace
+      @context = context
+      @ejson_file = "#{template_dir}/#{EJSON_SECRETS_FILE}"
+      @kubeclient = client || build_v1_kubeclient(@context)
+    end
+
+    def secrets_requested?
```
kirs: From the content of the method I think `secrets_provised?` would better describe it. What do you think?

lib/kubernetes-deploy/ejson_secret_provisioner.rb
```diff
@@ -0,0 +1,153 @@
...
+      super("Creation of Kubernetes secrets from ejson failed: #{msg}")
+    end
+  end
+
+  class EjsonSecretProvisioner
+    include KubernetesDeploy::KubeclientBuilder
+
+    MANAGEMENT_ANNOTATION = "kubernetes-deploy.shopify.io/ejson-secret".freeze
+    MANAGED_SECRET_EJSON_KEY = "kubernetes_secrets".freeze
+    EJSON_SECRETS_FILE = "secrets.kubernetes-deploy.ejson".freeze
```
kirs: Do you think that just picking `*.ejson` file from the `template_dir` would be better user experience rather than hardcoding the name?

lib/kubernetes-deploy/ejson_secret_provisioner.rb
```diff
@@ -0,0 +1,153 @@
...
+      File.exist?(@ejson_file)
+    end
+
+    def create_secrets
+      raise EjsonSecretError, "#{EJSON_SECRETS_FILE} not found" unless secrets_requested?
+
+      with_decrypted_ejson do |decrypted|
+        secrets = decrypted[MANAGED_SECRET_EJSON_KEY]
+        unless secrets.present?
+          KubernetesDeploy.logger.warn("#{EJSON_SECRETS_FILE} does not have key #{MANAGED_SECRET_EJSON_KEY}. No secrets will be created.")
```
kirs: Why is it a warning not a hard failure? TBH I don't think developers will notice it in the huge deploy logs.

lib/kubernetes-deploy/ejson_secret_provisioner.rb
```diff
@@ -0,0 +1,153 @@
...
+
+  class EjsonSecretProvisioner
+    include KubernetesDeploy::KubeclientBuilder
+
+    MANAGEMENT_ANNOTATION = "kubernetes-deploy.shopify.io/ejson-secret".freeze
+    MANAGED_SECRET_EJSON_KEY = "kubernetes_secrets".freeze
+    EJSON_SECRETS_FILE = "secrets.kubernetes-deploy.ejson".freeze
+    EJSON_KEYS_SECRET = "ejson-keys".freeze
+
+    def initialize(namespace:, context:, template_dir:, client: nil)
```
camilo: This `client:` argument seems to be there only as a convenience for testing? I don't love that but maybe there is another use case? If not, would it be more consistent to make this argument mandatory and _always_ pass the client?

lib/kubernetes-deploy/errors.rb
```diff
@@ -0,0 +1,3 @@
+module KubernetesDeploy
+  class FatalDeploymentError < StandardError; end
```
kirs: Is it already defined in https://github.com/Shopify/kubernetes-deploy/blob/master/lib/kubernetes-deploy.rb#L12 ?

lib/kubernetes-deploy/runner.rb
```diff
@@ -37,6 +38,7 @@ class Runner
     PROTECTED_NAMESPACES = %w(
       default
       kube-system
+      kube-public
```
kirs: Is it related to the secrets too?

lib/kubernetes-deploy/runner.rb
```diff
@@ -106,6 +108,13 @@ def run
       phase_heading("Checking initial resource statuses")
       resources.each(&:sync)
 
+      ejson = EjsonSecretProvisioner.new(namespace: @namespace, context: @context, template_dir: @template_dir)
+      if ejson.secrets_requested?
```
camilo: I find this a bit odd,  why do we have to create the object, ask it `hey do you want to do things?` and then call methods on it? instead of encapsulating all it in the `ejson` object.

```
# in this file just:
ejson.run()

# in the object

def run 
 return unless secrets_requested?
  ejson.create_secrets
  ejson.prune_managed_secrets
 end
end
```

camilo: Also what if someone just deletes the file, then we never prune the secrets?

test/unit/kubernetes-deploy/ejson_secret_provisioner_test.rb
```diff
@@ -0,0 +1,120 @@
+# frozen_string_literal: true
+require 'test_helper'
+
+class EjsonSecretProvisionerTest < KubernetesDeploy::TestCase
+  include FixtureDeployHelper
```
kirs: I thought `FixtureDeployHelper` was intended for integration tests that require deploying resources. It's a bit sketchy that the unit test will have access to helpers like `deploy_dir`.
Maybe we could move `fixture_path` that you're using in this test into `KubernetesDeploy::TestCase`?

lib/kubernetes-deploy/ejson_secret_provisioner.rb
```diff
@@ -0,0 +1,153 @@
...
+    def prune_managed_secrets
+      ejson_secret_names = encrypted_ejson.fetch(MANAGED_SECRET_EJSON_KEY, {}).keys
+      live_secrets = @kubeclient.get_secrets(namespace: @namespace)
+
+      live_secrets.each do |secret|
+        secret_name = secret.metadata.name
+        next unless secret.metadata.annotations.to_h.stringify_keys.key?(MANAGEMENT_ANNOTATION)
+        next if ejson_secret_names.include?(secret_name)
+
+        KubernetesDeploy.logger.info("Pruning secret #{secret_name}")
```
camilo: Bump this to `warn` log lvl maybe?

kubernetes-deploy.gemspec
```diff
@@ -22,9 +22,11 @@ Gem::Specification.new do |spec|
   spec.require_paths = ["lib"]
   spec.add_dependency "activesupport", ">= 4.2"
   spec.add_dependency "kubeclient", "~> 2.3"
+  spec.add_dependency "ejson", "1.0.1"
```
kirs: It seems that `ejson` gem is only a wrapper around the tool in Go (https://github.com/Shopify/ejson/blob/master/rubygem/bin/ejson). Do we still it as a dependency?

lib/kubernetes-deploy/ejson_secret_provisioner.rb
KnVerey: You're correct... I did this out of habit, and Policial is complaining about it too!

lib/kubernetes-deploy/ejson_secret_provisioner.rb
KnVerey: Correct, it is always for testing. I don't love it either. The runner doesn't use kubeclient yet, but I'll see how clean it could be to make it pass an instance anyway.

lib/kubernetes-deploy/ejson_secret_provisioner.rb
KnVerey: Why? It's intended behaviour based on a committed change to the repo.

lib/kubernetes-deploy/runner.rb
KnVerey: I did it this way to be able to avoid printing the heading when there's nothing to do, i.e. hide the feature from the average deploy user. 

> Also what if someone just deletes the file, then we never prune the secrets?

Good point. `secrets_requested?` would need to be `secret_changes_requested?` and look at live annotated secrets as well. I should add a test for this regardless of the heading issue.

lib/kubernetes-deploy/ejson_secret_provisioner.rb
KnVerey: What should we do if there's more than one in that case? It seemed simpler to me to have a conventional name, which ejson files usually do anyways (e.g. secrets.ENV.ejson).

lib/kubernetes-deploy/ejson_secret_provisioner.rb
KnVerey: It shouldn't hard fail because there could still be secrets that need to be pruned. It's mostly there in case they're deploying secrets for the first time and typo'd the key name, in which case hopefully they'd be on the lookout for logs about why it isn't working?

lib/kubernetes-deploy/errors.rb
KnVerey: I was intending to move it, but apparently it got lost in the rebase, and now I don't remember why it was necessary. I'll put it back.

lib/kubernetes-deploy/runner.rb
KnVerey: er, not really. It's a new system ns in 1.6. I can move this commit out if you prefer.

test/unit/kubernetes-deploy/ejson_secret_provisioner_test.rb
KnVerey: > Maybe we could move fixture_path that you're using in this test into KubernetesDeploy::TestCase?

👍

lib/kubernetes-deploy/ejson_secret_provisioner.rb
```diff
@@ -0,0 +1,153 @@
...
+      end
+    end
+
+    def prune_managed_secrets
+      ejson_secret_names = encrypted_ejson.fetch(MANAGED_SECRET_EJSON_KEY, {}).keys
+      live_secrets = @kubeclient.get_secrets(namespace: @namespace)
+
+      live_secrets.each do |secret|
+        secret_name = secret.metadata.name
+        next unless secret.metadata.annotations.to_h.stringify_keys.key?(MANAGEMENT_ANNOTATION)
```
ibawt: lol

kubernetes-deploy.gemspec
kirs: It turns out that the gem includes the Go binary for linux and darwin:

```
$ wget https://rubygems.org/downloads/ejson-1.0.2.gem
$ gem unpack /Users/kir/Downloads/ejson-1.0.2.gem
Unpacked gem: '/Users/kir/Projects/opensource/kubernetes-deploy/ejson-1.0.2'
$ ls -la ejson-1.0.2/build
drwxr-xr-x  3 kir  staff  102 Apr 21 15:04 darwin-amd64
drwxr-xr-x  3 kir  staff  102 Apr 21 15:04 linux-amd64
```

So ignore my point.

test/test_helper.rb
```diff
@@ -6,14 +6,15 @@
...
 
-require 'helpers/kubeclient_helper'
-require 'helpers/fixture_deploy_helper'
-require 'helpers/fixture_set'
-require 'helpers/fixture_sets/hello-cloud'
-
+Dir.glob(File.expand_path("../helpers/**/*.rb", __FILE__)).each { |file| require file }
 ENV["KUBECONFIG"] ||= "#{Dir.home}/.kube/config"
 
+Mocha::Configuration.prevent(:stubbing_method_unnecessarily)
```
ibawt: romg they do have it

ibawt:
If I'm reading this correctly for a use case such as Oauth tokens for the proxy, they will be have to share the same oauth?  I think that's ok, but the dev will have to add the callback url whitelist to both I think?  We should make that blatantly obvious somehow.

test/test_helper.rb
KnVerey: yep! I didn't know either until recently. IIRC was chatting with @sirupsen about whether to use rspec or minitest for this project, and he pointed it out to me.

KnVerey:
> If I'm reading this correctly for a use case such as Oauth tokens for the proxy, they will be have to share the same oauth?

Nope, each app puts this ejson file in the deploy folder in their own repo, so the secrets created for each can (and should) be unique.

ibawt:
that's what I was missing.  so this is more of a transparent .ejson -> k8s yml mapping.  👍

lib/kubernetes-deploy/ejson_secret_provisioner.rb
mkobetic: I agree that the `kubernetes-deploy` feels a bit superfluous. It's not really an environment either. I think `secrets.ejson` would be fine, sort of fitting with the rest of the files instead of yaml it's ejson.

lib/kubernetes-deploy/ejson_secret_provisioner.rb
camilo: you are not wrong, but this is an essentially destructive op, so maybe bending the rules a bit to make sure we are able to see it is not the worst.

lib/kubernetes-deploy/ejson_secret_provisioner.rb
KnVerey: I like `secrets.ejson`

lib/kubernetes-deploy/runner.rb
camilo: > I did it this way to be able to avoid printing the heading when there's nothing to do, i.e. hide the feature from the average deploy user.

I see that, still makes for fairly awkward code see how that check `secrets_requested?` repeats on the inner functions. and IMO breaks encapsulation, why not have the hypotetycal `run` return a value and print based on that?

lib/kubernetes-deploy/ejson_secret_provisioner.rb
KnVerey: I'd kinda rather that warnings stay things they should worry about when they see. If warnings are expected, I'd expect people will start ignoring them. Like they do on core deploys. Worth noting that we do not log warnings when pruning other resources (we don't even say what we pruned, though that we definitely should do).

KnVerey:
I think I've addressed all the comments, but the last commit will not be green; there's an integration test to write and some unit tests to fix to use the API changes I made following Camilo's comments. There was also a weird CI-only failure [here](https://buildkite.com/shopify/kubernetes-deploy-gem/builds/139#17cd24e5-f6c0-4419-94cb-c4b7327f3bf0) that I didn't get to the bottom of. Unfortunately I'm out of time to work on this before my vacation. If someone wants to pick this up while I'm gone, I've left a todo list on the PR description.

cc @camilo

KnVerey:
I just finished all the outstanding action items for this PR:
- Rebased 
- reworked the unit tests for the new `EjsonSecretProvisioner#run` API
- Fixed the test that was only failing on CI. It turned out CI was the one behaving normally... the test wrote `"}".to_json` to file, and later read it back and parsed it. Locally, that parsing raised an error. On CI, it was [properly parsed as a string](https://buildkite.com/shopify/kubernetes-deploy-gem/builds/189#cf32a2c4-7d38-497b-bf0d-cadfea37f576).
- Added a README update
- Added the feature to the cloud docs as a suggestion
- Successfully ran CI: https://buildkite.com/shopify/kubernetes-deploy-gem/builds/190#272382b7-74fd-4a98-8868-7471134ce8c3

Final 👀  please!

lib/kubernetes-deploy/ejson_secret_provisioner.rb
```diff
@@ -0,0 +1,168 @@
...
+      Dir.mktmpdir("ejson_keydir") do |key_dir|
+        File.write(File.join(key_dir, public_key), private_key)
+        decrypted = decrypt_ejson(key_dir)
+        yield decrypted
+      end
+    end
+
+    def decrypt_ejson(key_dir)
+      KubernetesDeploy.logger.info("Decrypting #{EJSON_SECRETS_FILE}")
+      # ejson seems to dump both errors and output to STDOUT
```
camilo: we should probably patch that, was that intended ? @burke ?

lib/kubernetes-deploy/ejson_secret_provisioner.rb
```diff
@@ -0,0 +1,168 @@
...
+      out_err, st = Open3.capture2e("EJSON_KEYDIR=#{key_dir} ejson decrypt #{@ejson_file}")
```
camilo: Is there no way this can be done via a function call to the gem?

lib/kubernetes-deploy/ejson_secret_provisioner.rb
```diff
@@ -0,0 +1,168 @@
...
+
+    def run
+      create_secrets
+      prune_managed_secrets
+    end
+
+    private
+
+    def create_secrets
+      return unless File.exist?(@ejson_file)
```
camilo: IMO this check belongs inside `with_decrypted_ejson`

camilo:
What happens if the secret exists AND there is no file, do we nuke it?

README.md
```diff
@@ -40,6 +40,40 @@ The following command will restart all pods in the `web` and `jobs` deployments:
 
 `kubernetes-restart <kube namespace> <kube context> --deployments=web,jobs`
 
+### Deploying Kubernetes secrets
+
+Since their data is only base64 encoded, Kubernetes secrets should not be committed to your repository. Instead, `kubernetes-deploy` supports generating secrets from an encrypted [ejson](https://github.com/Shopify/ejson) file in your template directory. Here's how to use this feature:
+
+1. Install the ejson gem: `gem install ejson`
+2. Generate a new keypair: `ejson keygen` (prints the keypair to stdout)
+3. Create a Kubernetes secret in your target namespace with the new keypair: `kubectl create secret generic ejson-keys --from-literal=YOUR_PUBLIC_KEY=YOUR_PRIVATE_KEY --namespace=TARGET_NAMESPACE`
```
mkobetic: Most won't be able to do this, why not reuse the existing ejson key?

README.md
```diff
@@ -40,6 +40,40 @@ The following command will restart all pods in the `web` and `jobs` deployments:
 
 `kubernetes-restart <kube namespace> <kube context> --deployments=web,jobs`
 
+### Deploying Kubernetes secrets
+
```
mkobetic: Did we want to emphasize to only use this when actual K8S secrets are needed, and to stick with normal ejson workflow otherwise?

README.md
mkobetic: Just realized that this is the same key name as the GC generated keys, I'm worried someone might end up overwriting that one based on these instructions.

KnVerey:
> What happens if the secret exists AND there is no file, do we nuke it?

Yes. The test for this scenario is here: https://github.com/Shopify/kubernetes-deploy/blob/secret_ejson/test/integration/kubernetes_deploy_test.rb#L279-L296

README.md
KnVerey: I wrote these instructions from the perspective of someone trying to use `kubernetes-deploy` directly, without any of the magic from our platform. Our users wouldn't be able to do this even if they wanted to, since they don't have `kubectl` access. The draft docs for our users are in the Cloudplatform User Guide as suggestions. How about I add a line at the top of this section saying that Shopify employees using our Cloudplatform should consult the user guide instead?

README.md
KnVerey: The Cloudplatform docs version does do this. :)

lib/kubernetes-deploy/ejson_secret_provisioner.rb
KnVerey: AFAICT installing [the gem](https://github.com/Shopify/ejson) just installs the Go binary... [capistrano-ejson](https://github.com/Shopify/capistrano-ejson/blob/51665d2dc7c71a7a3ad19070be993582f19b949b/lib/capistrano/tasks/ejson.cap#L14) is also shelling out to use it, as are other apps that turn up in a Github search (though I only checked a few results).

ibawt: the gem is literally that, welp thats what happens when you can't ffi to go easily :P

README.md
mkobetic: sounds good.

KnVerey:
Addressed the latest comments and did a tophat showing the logging for this feature:

image

Anything else @camilo @mkobetic @ibawt @kirs ?

mkobetic:
🚀

camilo:
🐑

stefanmb:
Thank you for this!

PR #82, opened 2017-04-21
Fix assertions on exception messages

KnVerey:
I realized while working on a feature on this repo that we've been misusing `assert_raises`. We have a bunch of tests that assume the second arg will be asserted as the error message, whereas it is in fact just used as the assertion failure message. I wrote most of these mistakes myself, but I don't think I'm the only one with the misimpression (it was actually [originally suggested to me by a reviewer](https://github.com/Shopify/kubernetes-deploy/pull/24#discussion_r102931248)). This PR fixes the misuse, introduces a helper with the desired behaviour, and tries to prevent future confusion.

### DIFF
--- a/test/test_helper.rb
+++ b/test/test_helper.rb
@@ -42,7 +42,8 @@ module KubernetesDeploy
     def assert_raises(*args)
       case args.last
       when Regexp, String
-        flunk("Please use assert_raises_msg to check the exception message. That is not what the last argument of assert_raises does.")
+        flunk("Please use assert_raises_msg to check the exception message.\n"\
+          "That is not what the last argument of assert_raises does.")
       else
         orig_assert_raises(*args) { yield }
       end
--- a/test/test_helper.rb
+++ b/test/test_helper.rb
@@ -37,6 +37,22 @@ module KubernetesDeploy
         assert_match regexp, @logger_stream.read
       end
     end
+
+    alias_method :orig_assert_raises, :assert_raises
+    def assert_raises(*args)
+      case args.last
+      when Regexp, String
+        flunk("Please use assert_raises_msg to check the exception message. That is not what the last argument of assert_raises does.")
+      else
+        orig_assert_raises(*args) { yield }
+      end
+    end
+
+    def assert_raises_msg(exception_class, exception_message)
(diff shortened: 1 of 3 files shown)


test/test_helper.rb
```diff
@@ -37,6 +37,23 @@ def assert_logs_match(regexp, times = nil)
         assert_match regexp, @logger_stream.read
       end
     end
+
+    alias_method :orig_assert_raises, :assert_raises
+    def assert_raises(*args)
+      case args.last
```
kirs: I see the value of the override but we should keep it mind that we could no longer use the last argument feature (http://www.rubydoc.info/gems/minitest/5.10.1/Minitest/Assertions#assert_raises-instance_method):

```ruby
assert_raises(WhateverError, "does not raise when logging is disabled") { ... }
```

KnVerey:
> I see the value of the override but we should keep it mind that we could no longer use the last argument feature 

True, and it would be easy enough to make that possible. Pushed another commit that lets you use a kwarg if you're using that feature on purpose. LMK what you think. Here are some examples:

```ruby
  def test_assert_raises
    assert_raises(StandardError, "foobar") do
    end
  end

  1) Failure:
RunnerTest#test_assert_raises [test/unit/kubernetes-deploy/runner_test.rb:10]:
Please use assert_raises_msg to check the exception message,
or use the msg keyword argument to specify a message to use when the assertion fails.
```

```ruby
  def test_assert_raises
    assert_raises(StandardError, msg: "This is an error message.") do
    end
  end

  1) Failure:
RunnerTest#test_assert_raises [test/unit/kubernetes-deploy/runner_test.rb:10]:
This is an error message.
StandardError expected but nothing was raised.
```

```ruby
  def test_assert_raises
    assert_raises(StandardError) do
    end
  end

  1) Failure:
RunnerTest#test_assert_raises [test/unit/kubernetes-deploy/runner_test.rb:10]:
StandardError expected but nothing was raised.
```

kirs:
Great idea!
What about using `message` instead of `msg`? That would keep parity with other test helpers at Shopify that use `message` as a keyword argument.

KnVerey:
Pushed another commit that makes this a duplicate of what I just proposed in core.

PR #84, opened 2017-04-22
Test long running pods that don't shutdown

kirs:
This adds test coverage for pods that don't shutdown properly and rely on `terminationGracePeriodSeconds`, which is the case for long-running jobs and LHMs.

### DIFF
--- a/test/fixtures/long-running/jobs.yml.erb
+++ b/test/fixtures/long-running/jobs.yml.erb
@@ -1,0 +1,20 @@
+apiVersion: extensions/v1beta1
+kind: Deployment
+metadata:
+  name: jobs
+spec:
+  replicas: 2
+  template:
+    metadata:
+      labels:
+        name: jobs
+        app: fixtures
+    spec:
+      containers:
+      - name: app
+        image: busybox
+        command: ["sleep", "40"]
+        env:
+        - name: GITHUB_REV
+          value: <%= current_sha %>
+      terminationGracePeriodSeconds: 30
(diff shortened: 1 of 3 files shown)


test/fixtures/long-running/jobs.yml.erb
```diff
@@ -0,0 +1,20 @@
...
+  replicas: 2
+  template:
+    metadata:
+      labels:
+        name: jobs
+        app: fixtures
+    spec:
+      containers:
+      - name: app
+        image: kirshatrov/sleep-forever:latest
```
kirs: This is https://github.com/kirs/docker-run-forever/blob/master/sleep.rb packed into Docker image.

test/integration/kubernetes_deploy_test.rb
```diff
@@ -213,4 +213,28 @@ def test_should_raise_if_required_binding_not_present
...
+  def test_long_running_deployment
+    2.times do
+      deploy_fixtures('long-running')
+    end
+
+    pods = kubeclient.get_pods(namespace: @namespace, label_selector: 'name=jobs,app=fixtures')
+    assert_equal 4, pods.size
+
+    count = count_by_revisions(pods)
+    assert_equal [2, 2], count.values
```
kirs: We assert that both revisions are still running, after the latest one has been deployed.

test/fixtures/long-running/jobs.yml.erb
ibawt: I would just done image: busybox and then use command: ["sleep", "whatevs"]

kirs: 👍 👍 👍

PR #85, opened 2017-04-24
Test long-running pods together with revisionHistoryLimit

kirs:
Adding `revisionHistoryLimit=1` to the long-running deployment fixture to make sure that it doesn't conflict with multiple revisions in the terminating state.

cc @ibawt


ibawt:
I think this is enough. 👍

PR #86, opened 2017-04-24
Add "all" as an option to restart all deployments

kirs:
I'm working on adding support for restarting deployments from the Shipit UI, and I think it would be useful to have an option to restart all of them (for instance web, jobs and redis).

### DIFF
--- a/lib/kubernetes-deploy/restart_task.rb
+++ b/lib/kubernetes-deploy/restart_task.rb
@@ -45,6 +45,11 @@ module KubernetesDeploy
         raise ArgumentError, "#perform takes at least one deployment to restart"
       end
 
+      if deployments_names == "all"
+        deployments_names = @v1beta1_kubeclient.get_deployments(namespace: @namespace).map { |d| d.metadata.name }
+        raise ArgumentError, "no deployments found in namespace #{@namespace}" if deployments_names.none?
+      end
+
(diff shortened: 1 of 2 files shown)


lib/kubernetes-deploy/restart_task.rb
```diff
@@ -45,6 +45,11 @@ def perform(deployments_names)
         raise ArgumentError, "#perform takes at least one deployment to restart"
       end
 
+      if deployments_names == "all"
```
kirs: Shipit has no data about what deployments in the app so kubernetes-deploy would have to query the list of all deployments first.

ibawt:
redis very dangerous to restart.  ES as well, there's a few of them.

Can we do this for deployments with containers that match the repo name? wdyt

kirs:
@ibawt here is the use case: the app owner wants to restart in Shipit:

screen shot 2017-04-24 at 17 03 39

But they don't know anything about deployments. They just want to restart it. My idea was to provide two generic resources (web and jobs) and "all" as an option just in case.

I'm open to suggestions what would be the best experience for developers. I'm not sure how repo name matching is relevant here.

ibawt:
you can filter the deployments by the predicate of their image matching the code they ship.

kirs:
Implemented in https://github.com/Shopify/kubernetes-deploy/issues/90

PR #87, opened 2017-04-27
Bump gem version

xldenis:
After merging my pr adding support for `PodTemplate` based task runners, a new version of the gem needs to be deployed.

r: @wfarr (since you are an owner on rubygems) / @kirs


kirs:
I released 0.6.0 on rubygems.

PR #88, opened 2017-05-01
Add google cloud configuration object for kubeclient

xldenis:
Kubeclient doesn't support GCE authentication and there is currently no plan to merge the PR that would add support for it. To solve our problem I've added a small wrapper class based off the kubeclient PR. It provides a subclass of the `Kubeclient::Config` which will look for the approriate auth type and if found will fetch a new (1 hour lifetime) token. 

[Kubeclient GCE PR](https://github.com/abonas/kubeclient/pull/213#issuecomment-298055112)

@jeremywadsack, since you were the author of the original PR you may be interested in this.

### DIFF
--- a/lib/kubernetes-deploy/kubeclient_builder.rb
+++ b/lib/kubernetes-deploy/kubeclient_builder.rb
@@ -1,5 +1,6 @@
 # frozen_string_literal: true
 require 'kubeclient'
+require 'kubernetes-deploy/kubeclient_builder/google_friendly_config'
 
 module KubernetesDeploy
   module KubeclientBuilder
@@ -28,7 +29,7 @@ module KubernetesDeploy
     end
 
     def _build_kubeclient(api_version:, context:, endpoint_path: nil)
-      config = Kubeclient::Config.read(ENV.fetch("KUBECONFIG"))
+      config = GoogleFriendlyConfig.read(ENV.fetch("KUBECONFIG"))
       unless config.contexts.include?(context)
         raise ContextMissingError, context
       end
--- a/test/unit/kubernetes-deploy/google_friendly_config_test.rb
+++ b/test/unit/kubernetes-deploy/google_friendly_config_test.rb
@@ -1,0 +1,107 @@
+# frozen_string_literal: true
+require 'test_helper'
+
+class GoogleFriendlyConfigTest < KubernetesDeploy::TestCase
+  def setup
+    WebMock.disable_net_connect!
+    set_google_env_vars
+  end
+
+  def teardown
+    WebMock.allow_net_connect!
+  end
+
+  def test_auth_use_default_gcp_success
+    config = KubernetesDeploy::KubeclientBuilder::GoogleFriendlyConfig.new(kubeconfig, "")
+
+    stub_request(:post, 'https://www.googleapis.com/oauth2/v3/token')
+      .to_return(
+        headers: { 'Content-Type' => 'application/json' },
+        body: {
+          "access_token": "bearer_token",
+          "token_type": "Bearer",
+          "expires_in": 3600,
+          "id_token": "identity_token"
+        }.to_json,
+        status: 200
+      )
+
+    context = config.context("google")
+    assert_equal 'bearer_token', context.auth_options[:bearer_token]
+  end
+
+  def test_auth_use_default_gcp_failure
+    config = KubernetesDeploy::KubeclientBuilder::GoogleFriendlyConfig.new(kubeconfig, "")
+
+    stub_request(:post, 'https://www.googleapis.com/oauth2/v3/token')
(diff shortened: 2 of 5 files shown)


lib/kubernetes-deploy/kubeclient_builder.rb
```diff
@@ -28,7 +29,7 @@ def build_v1beta1_kubeclient(context)
     end
 
     def _build_kubeclient(api_version:, context:, endpoint_path: nil)
-      config = Kubeclient::Config.read(ENV.fetch("KUBECONFIG"))
+      config = GoogleConfig.read(ENV.fetch("KUBECONFIG"))
```
kirs: I'm not sold about naming because someone reading this code could think that it only supports GKE config. What about `KubeconfigWithGoogleSupport` or whatever?

lib/kubernetes-deploy/kubeclient_builder/google_config.rb
```diff
@@ -0,0 +1,42 @@
...
+
+      def self.read(filename)
+        new(YAML.load_file(filename), File.dirname(filename))
+      end
+
+      def get_token
+        scopes = ['https://www.googleapis.com/auth/cloud-platform']
+        authorization = Google::Auth.get_application_default(scopes)
+
+        authorization.apply({})
```
kirs: Why does it need an empty apply? Can you add a comment to the code with the reason?

test/unit/kubernetes-deploy/google_config_test.rb
```diff
@@ -0,0 +1,46 @@
...
+class GoogleConfigTest < KubernetesDeploy::TestCase
+  def setup
+    WebMock.disable_net_connect!
+  end
+
+  def teardown
+    WebMock.allow_net_connect!
+  end
+
+  def test_auth_use_default_gcp_success
```
kirs: I think it also makes sense to check that the config works as usual if auth provider isn't GCP.

lib/kubernetes-deploy/kubeclient_builder/google_config.rb
```diff
@@ -0,0 +1,42 @@
+require 'googleauth'
+module KubernetesDeploy
+  module KubeclientBuilder
+    class GoogleConfig < Kubeclient::Config
+      def fetch_user_auth_options(user)
+        if user['auth-provider'] && (user['auth-provider']['name'] == 'gcp')
+          { bearer_token: self.get_token }
```
kirs: Is `self` required here?

xldenis: not anymore.

test/unit/kubernetes-deploy/google_config_test.rb
xldenis: 👍

lib/kubernetes-deploy/kubeclient_builder/google_config.rb
xldenis: Well `apply` adds the access token to a set of headers in a hash. Since we just want to recover the token we don't need to pass in anything.

lib/kubernetes-deploy/kubeclient_builder.rb
xldenis: I dont have any particular attachment to the name, though I'm not a huge fan of enormous names either.

kirs: What do you think about `GoogleConfig` incorrectly telling someone that the tool only supports google auth?

xldenis: What about just `ConfigWithGoogle` ? I'm just not a huge fan of having `KubernetesDeploy::KubeclientBuilder::KubeconfigWithGoogleSupport`

xldenis:
I've addressed the review comments the configuration is now called `GoogleFriendlyConfig`.

test/unit/kubernetes-deploy/google_friendly_config_test.rb
```diff
@@ -0,0 +1,107 @@
...
+            'password' => 'test',
+            'username' => 'admin'
+          }
+        }
+      ]
+    }.stringify_keys
+  end
+
+  def set_google_env_vars
+    ENV["GOOGLE_PRIVATE_KEY"] ||= "FAKE"
```
kirs: There's probably a reason why it's necessary, but I'd prefer them to be reset on teardown

xldenis: Its because the `googleauth` gem does a bunch of magic inspection to determine how to load configuration for a host (ie checking if creds are locally stored on disk, whether this is a GCE host, etc...).

PR #90, opened 2017-05-02
Restart deployments with annotation

kirs:
This is alternative to https://github.com/Shopify/kubernetes-deploy/pull/86

The problem: we want to have the magic restart button in Shipit but the app may have many deployments (web, jobs, redis, memcache) and it's not safe to restart all of them. We need a way to determine app deployments (web & jobs) that are safe to restart.

When I discussed it with @ibawt he had an idea of matching the container and app names. For instance to restart the `hackdays` app, we would fetch all deployments and see which of them have `hackdays` substring in the container address.

This is somewhat fuzzy and won't work for apps with generic name like "shopify". I suggest that we use special annotation to tell Shipit if the deployment is safe to restart.

cc @Shopify/cloudplatform

### DIFF
--- a/exe/kubernetes-restart
+++ b/exe/kubernetes-restart
@@ -8,16 +8,11 @@ require 'kubernetes-deploy/restart_task'
 
 raw_deployments = nil
 ARGV.options do |opts|
-  opts.on("--deployments=LIST") { |v| raw_deployments = v }
+  opts.on("--deployments=LIST") { |v| raw_deployments = v.split(",") }
   opts.parse!
-end
-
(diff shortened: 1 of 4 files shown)


ibawt:
if you used a label you could have gotten a filtered list back, but I'm not sure that's a great use for a label.

kirs:
> if you used a label you could have gotten a filtered list back, but I'm not sure that's a great use for a label. and N will be fairly small.

I was deciding between label and annotation and annotation seemed like the better fit in terms of semantics.

PR #91, opened 2017-05-08
Use kubectl in EjsonSecretProvisioner

KnVerey:
tl;dr `kubeclient` isn't working for some of our production clusters, so this switches `EjsonSecretProvisioner` over to `kubectl`--which is what the rest of the deploy process uses--for the time being. 

---

When I rolled out https://github.com/Shopify/kubernetes-deploy/pull/81 in shipit, I discovered that kubeclient does not seem to be working with our GCP clusters. It is failing with an SSL error on `client.discover` ([see this deploy](https://shipit.shopify.io/shopify/hackdays-gke-rails5/production/deploys/445001)). This may be a further piece to the work @xldenis already did in https://github.com/Shopify/kubernetes-deploy/pull/88; he is investigating.

The most meaningful change is that we need to write a tempfile to use `kubectl apply` for secret creation/update. Note that although the unit tests required surgery, the integration tests all passed as-is.

**Why not wait for the kubeclient fix?** 
The current consensus is still that we want to use a gem rather than kubectl for everything we can (which doesn't include "apply" operations) in the long run, but I didn't realize that none of the tasks that already use kubeclient had been tested in our production Shipit environment. This feature is high-priority, and switching to kubeclient is far from it. With the problems we've had, we may even want to reconsider this particular gem choice.

cc @Shopify/cloudplatform

### DIFF
--- a/lib/kubernetes-deploy/ejson_secret_provisioner.rb
+++ b/lib/kubernetes-deploy/ejson_secret_provisioner.rb
@@ -3,6 +3,7 @@ require 'json'
 require 'base64'
 require 'open3'
 require 'kubernetes-deploy/logger'
+require 'kubernetes-deploy/kubectl'
 
 module KubernetesDeploy
   class EjsonSecretError < FatalDeploymentError
@@ -17,10 +18,13 @@ module KubernetesDeploy
     EJSON_SECRETS_FILE = "secrets.ejson"
     EJSON_KEYS_SECRET = "ejson-keys"
 
-    def initialize(namespace:, template_dir:, client:)
+    def initialize(namespace:, context:, template_dir:)
       @namespace = namespace
+      @context = context
       @ejson_file = "#{template_dir}/#{EJSON_SECRETS_FILE}"
-      @kubeclient = client
+
+      raise FatalDeploymentError, "Cannot create secrets without a namespace" if @namespace.blank?
+      raise FatalDeploymentError, "Cannot create secrets without a context" if @context.blank?
     end
 
     def secret_changes_required?
@@ -52,25 +56,27 @@ module KubernetesDeploy
 
     def prune_managed_secrets
       ejson_secret_names = encrypted_ejson.fetch(MANAGED_SECRET_EJSON_KEY, {}).keys
-      live_secrets = @kubeclient.get_secrets(namespace: @namespace)
+      live_secrets = run_kubectl_json("get", "secrets")
 
       live_secrets.each do |secret|
-        secret_name = secret.metadata.name
+        secret_name = secret["metadata"]["name"]
         next unless secret_managed?(secret)
         next if ejson_secret_names.include?(secret_name)
 
         KubernetesDeploy.logger.info("Pruning secret #{secret_name}")
-        @kubeclient.delete_secret(secret_name, @namespace)
+        out, err, st = run_kubectl("delete", "secret", secret_name)
+        KubernetesDeploy.logger.debug(out)
+        raise EjsonSecretError, err unless st.success?
       end
     end
 
     def managed_secrets_exist?
-      all_secrets = @kubeclient.get_secrets(namespace: @namespace)
+      all_secrets = run_kubectl_json("get", "secrets")
       all_secrets.any? { |secret| secret_managed?(secret) }
     end
 
     def secret_managed?(secret)
-      secret.metadata.annotations.to_h.stringify_keys.key?(MANAGEMENT_ANNOTATION)
+      secret["metadata"].fetch("annotations", {}).key?(MANAGEMENT_ANNOTATION)
     end
 
     def encrypted_ejson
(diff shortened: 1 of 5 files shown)


lib/kubernetes-deploy/ejson_secret_provisioner.rb
```diff
@@ -96,31 +99,45 @@ def validate_secret_spec(secret_name, spec)
...
-        KubernetesDeploy.logger.info("Updating secret #{secret_name}")
-        @kubeclient.update_secret(secret)
-      else
-        KubernetesDeploy.logger.info("Creating secret #{secret_name}")
-        @kubeclient.create_secret(secret)
+      secret_yaml = generate_secret_yaml(secret_name, secret_type, data)
+      msg = secret_exists?(secret_name) ? "Updating secret #{secret_name}" : "Creating secret #{secret_name}"
+      KubernetesDeploy.logger.info(msg)
+
+      file = Tempfile.new(secret_name)
```
KnVerey: Alternatively, we could construct a really long command with `--from-literal=k=v --from-literal=k=v ...`

lib/kubernetes-deploy/ejson_secret_provisioner.rb
```diff
@@ -152,17 +169,28 @@ def decrypt_ejson(key_dir)
...
+    def run_kubectl_json(*args)
+      args += ["--output=json"]
+      out, err, st = run_kubectl(*args)
+      raise EjsonSecretError, err unless st.success?
+      result = JSON.parse(out)
+      result.fetch('items', result)
+    end
+
+    def run_kubectl(*args)
+      raise FatalDeploymentError, "Namespace missing for namespaced command" if @namespace.blank?
```
camilo: Why not check this on the constructor?

lib/kubernetes-deploy/ejson_secret_provisioner.rb
```diff
@@ -152,17 +169,28 @@ def decrypt_ejson(key_dir)
...
+      raise KubectlError, "Explicit context is required to run this command" if @context.blank?
```
camilo: Same here

lib/kubernetes-deploy/ejson_secret_provisioner.rb
```diff
@@ -96,31 +99,45 @@ def validate_secret_spec(secret_name, spec)
...
-        @kubeclient.create_secret(secret)
+      secret_yaml = generate_secret_yaml(secret_name, secret_type, data)
+      msg = secret_exists?(secret_name) ? "Updating secret #{secret_name}" : "Creating secret #{secret_name}"
+      KubernetesDeploy.logger.info(msg)
+
+      file = Tempfile.new(secret_name)
+      file.write(secret_yaml)
+      file.close
+
+      out, err, st = run_kubectl("apply", "--filename=#{file.path}")
```
mkobetic: We should make sure the file is deleted too, would `Tempfile.open` work for this?

lib/kubernetes-deploy/ejson_secret_provisioner.rb
```diff
@@ -96,31 +99,45 @@ def validate_secret_spec(secret_name, spec)
...
-      else
-        KubernetesDeploy.logger.info("Creating secret #{secret_name}")
-        @kubeclient.create_secret(secret)
+      secret_yaml = generate_secret_yaml(secret_name, secret_type, data)
+      msg = secret_exists?(secret_name) ? "Updating secret #{secret_name}" : "Creating secret #{secret_name}"
+      KubernetesDeploy.logger.info(msg)
+
+      file = Tempfile.new(secret_name)
+      file.write(secret_yaml)
+      file.close
```
KnVerey: To address both this and @mkobetic comment below, kubectl won't be able to read the file unless we close it. Tempfiles are automatically GC'd when the process exits, if not sooner. `Tempfile.open { |f| ... }` doesn't return the file. I don't think it would be possible for the file to get GC'd when we need it in this case, since we only need it for the duration of this method, which retrains a reference to it. However, I can switch to a regular file and manually unlink it in an `ensure` to be extra super safe.

mkobetic: I would expect that if the file was flushed after write the shell command should be able to read it then, but I haven't tried.

KnVerey: TBH I think this is the simplest solution. The GC should take care of unlinking at the appropriate time, but I'll add `file.unlink if file` in an `ensure` just in case.

lib/kubernetes-deploy/ejson_secret_provisioner.rb
KnVerey: Because this is copied from another place where it wouldn't make sense in the constructor. :P

Really, it's a super-paranoid check to make sure we don't run commands without a context/namespace. In reality, it should be impossible to hit in production (several other things would have raised before we get here). But I don't feel strongly about where it goes and can move it.

KnVerey:
Updated

lib/kubernetes-deploy/ejson_secret_provisioner.rb
```diff
@@ -99,17 +102,19 @@ def validate_secret_spec(secret_name, spec)
...
+      secret_yaml = generate_secret_yaml(secret_name, secret_type, data)
       file = Tempfile.new(secret_name)
       file.write(secret_yaml)
       file.close
 
       out, err, st = run_kubectl("apply", "--filename=#{file.path}")
       KubernetesDeploy.logger.debug(out)
       raise EjsonSecretError, err unless st.success?
+    ensure
+      file.unlink if file
```
mkobetic: I wonder if this can actually error out in some circumstances, I'd be fine to rely on Tempfile to take care of this. But I'm fine either way

PR #92, opened 2017-05-09
Don't log command failures in EjsonSecretProvisioner

KnVerey:
In all cases other than `secret_exists?`, we already raise an error with the stderr text if the command fails, making the failure logging redundant. And for `secret_exists?` we don't want to log an error on failure, because the failure is very likely 404. Here's an example of the useless warning this gets rid of:

image


ibawt:
I guess there's no way to just filter out the 404? 

code lgtm

KnVerey:
Other than scanning the err text for `NotFound`, not so much. 😞

PR #96, opened 2017-05-23
Add Support for PodDisruptionBudget Resources

psycotica0-shopify:
This appears to be the first resource that we have to do a
replace --force to update.

So, that took a little bit of reworking.

Also, in trying to get the tests to fail properly I realized we weren't
checking the return value of the `create` when a replace failed.

Now if I don't return `:replace_force` for the `#deploy_method`, the
test fail.

### DIFF
--- a/lib/kubernetes-deploy/kubernetes_resource/pod_disruption_budget.rb
+++ b/lib/kubernetes-deploy/kubernetes_resource/pod_disruption_budget.rb
@@ -1,0 +1,32 @@
+# frozen_string_literal: true
+module KubernetesDeploy
+  class PodDisruptionBudget < KubernetesResource
+    TIMEOUT = 30.seconds
+
+    def initialize(name, namespace, context, file)
+      @name = name
+      @namespace = namespace
+      @context = context
+      @file = file
+    end
+
+    def sync
+      _, _err, st = run_kubectl("get", type, @name)
+      @found = st.success?
+      @status = @found ? "Available" : "Unknown"
+      log_status
+    end
+
+    def deploy_succeeded?
+      exists?
+    end
+
+    def deploy_method
+      :replace_force
+    end
+
+    def exists?
+      @found
+    end
+  end
+end
--- a/lib/kubernetes-deploy/kubernetes_resource/pod_disruption_budget.rb
+++ b/lib/kubernetes-deploy/kubernetes_resource/pod_disruption_budget.rb
@@ -1,7 +1,7 @@
 # frozen_string_literal: true
 module KubernetesDeploy
   class PodDisruptionBudget < KubernetesResource
-    TIMEOUT = 30.seconds
+    TIMEOUT = 10.seconds
 
     def initialize(name, namespace, context, file)
       @name = name
@@ -22,6 +22,7 @@ module KubernetesDeploy
     end
 
     def deploy_method
+      # Required until https://github.com/kubernetes/kubernetes/issues/45398 changes
       :replace_force
     end
 
--- a/lib/kubernetes-deploy/runner.rb
+++ b/lib/kubernetes-deploy/runner.rb
@@ -310,13 +310,16 @@ MSG
           _, _, st = run_kubectl("replace", "-f", r.file.path)
         when :replace_force
           _, _, st = run_kubectl("replace", "--force", "-f", r.file.path)
+        else
+          # Fail Fast! This is a programmer mistake.
+          raise "Unexpected deploy method! (#{r.deploy_method.inspect})"
         end
 
         unless st.success?
           # it doesn't exist so we can't replace it
           _, err, st = run_kubectl("create", "-f", r.file.path)
           unless st.success?
-            raise FatalDeploymentError, <<~MSG
+            raise FatalDeploymentError, <<-MSG.strip_heredoc
               Failed to replace or create resource: #{r.id}
               #{err}
             MSG
--- a/lib/kubernetes-deploy/runner.rb
+++ b/lib/kubernetes-deploy/runner.rb
@@ -17,6 +17,7 @@ require 'kubernetes-deploy/kubernetes_resource'
   service
   pod_template
   bugsnag
+  pod_disruption_budget
 ).each do |subresource|
   require "kubernetes-deploy/kubernetes_resource/#{subresource}"
 end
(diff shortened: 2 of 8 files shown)


lib/kubernetes-deploy/runner.rb
```diff
@@ -295,25 +296,38 @@ def validate_configuration
...
+        when :replace_force
+          _, _, st = run_kubectl("replace", "--force", "-f", r.file.path)
+        end
 
         unless st.success?
           # it doesn't exist so we can't replace it
-          run_kubectl("create", "-f", r.file.path)
+          _, err, st = run_kubectl("create", "-f", r.file.path)
+          unless st.success?
+            raise FatalDeploymentError, <<~MSG
```
KnVerey: This gem is for use in Shipit, which afaik we're keeping 2.2 compatible, so I've been avoiding 2.3-specific features, as nice as they are. You have `.strip_heredoc` from ActiveSupport available though.

lib/kubernetes-deploy/runner.rb
```diff
@@ -295,25 +296,38 @@ def validate_configuration
...
+      individuals.each do |r|
         KubernetesDeploy.logger.info("- #{r.id}")
         r.deploy_started = Time.now.utc
-        _, _, st = run_kubectl("replace", "-f", r.file.path)
+        case r.deploy_method
+        when :replace
+          _, _, st = run_kubectl("replace", "-f", r.file.path)
+        when :replace_force
+          _, _, st = run_kubectl("replace", "--force", "-f", r.file.path)
+        end
```
KnVerey: Can you add an `else` case that raises? Shouldn't ever get hit, but could be helpful for debugging a future stupid mistake.

lib/kubernetes-deploy/runner.rb
```diff
@@ -295,25 +296,38 @@ def validate_configuration
...
         end
       end
-    end
 
-    def deploy_resources(resources, prune: false)
-      KubernetesDeploy.logger.info("Deploying resources:")
+      apply_all(applyables, prune)
+    end
 
-      # TPRs must use update for now: https://github.com/kubernetes/kubernetes/issues/39906
```
KnVerey: Can you move this note to above the `tpr? ? :replace : :apply` line? We will want to change TPRs to `apply` once all our clusters are on 1.6 because these resources otherwise can't get pruned.

psycotica0-shopify:
Hmm. I can't find any issue regarding PodDisruptionBudget and apply...

I just assumed it was a NoNo somehow.

wfarr:
> Hmm. I can't find any issue regarding PodDisruptionBudget and apply...
> 
> I just assumed it was a NoNo somehow.

All official upstream resources support `apply`. TPRs have weird caveats with apply and were broken for all of 1.5 with `apply`, which is what led to the use of `create` and `replace`.

psycotica0-shopify:
@wfarr These are what I tried, if I've done something wrong, I'd be all ears. The `--force` struck me as a little bold as well.
```
$ kubectl get poddisruptionbudgets kafka -o json | kubectl apply -f - 
Warning: kubectl apply should be used on resource created by either kubectl create --save-config or kubectl apply
The PodDisruptionBudget "kafka" is invalid: spec: Forbidden: updates to poddisruptionbudget spec are forbidden.
```

```
$ kubectl get poddisruptionbudgets kafka --context kafka-staging -o json | kubectl replace -f - --context kafka-staging
poddisruptionbudget "kafka" replaced
```

```
$ kubectl get poddisruptionbudgets kafka --context kafka-staging -o json | jq '.spec.minAvailable |= 4'  | kubectl --context  kafka-staging replace -f -                                                                                       
The PodDisruptionBudget "kafka" is invalid: spec: Forbidden: updates to poddisruptionbudget spec are forbidden.
```

```
$ kubectl patch poddisruptionbudgets kafka --context kafka-staging -p '{"spec":{"minAvailable": 4}}'
The PodDisruptionBudget "kafka" is invalid: spec: Forbidden: updates to poddisruptionbudget spec are forbidden.
```

```
$ kubectl version --context kafka-staging                                                                                                  
Client Version: version.Info{Major:"1", Minor:"6", GitVersion:"v1.6.0", GitCommit:"fff5156092b56e6bd60fff75aad4dc9de6b6ef37", GitTreeState:"clean", BuildDate:"2017-03-28T16:36:33Z", GoVersion:"go1.7.5", Compiler:"gc", Platform:"darwin/amd64"}                                                          
Server Version: version.Info{Major:"1", Minor:"6", GitVersion:"v1.6.2", GitCommit:"477efc3cbe6a7effca06bd1452fa356e2201e1ee", GitTreeState:"clean", BuildDate:"2017-04-19T20:22:08Z", GoVersion:"go1.7.5", Compiler:"gc", Platform:"linux/amd64"}
```

(This is the deploy that lead to this PR)
https://shipit.shopify.io/shopify/kafka-infrastructure/staging/deploys/447361

KnVerey:
Here's an issue discussing the problem: https://github.com/kubernetes/kubernetes/issues/45398
Let's minimally add a comment in the code like we have for TPR.

lib/kubernetes-deploy/runner.rb
```diff
@@ -295,25 +296,41 @@ def validate_configuration
...
         r.deploy_started = Time.now.utc
-        _, _, st = run_kubectl("replace", "-f", r.file.path)
+        case r.deploy_method
+        when :replace
+          _, _, st = run_kubectl("replace", "-f", r.file.path)
+        when :replace_force
+          _, _, st = run_kubectl("replace", "--force", "-f", r.file.path)
+        else
+          # Fail Fast! This is a programmer mistake.
+          raise "Unexpected deploy method! (#{r.deploy_method.inspect})"
```
kirs: Nit: there's a best practice to specify at least some error rather than using `raise "message"`. Can we make it `raise ArgumentError, "message"`?

lib/kubernetes-deploy/kubernetes_resource/pod_disruption_budget.rb
```diff
@@ -0,0 +1,32 @@
+# frozen_string_literal: true
+module KubernetesDeploy
+  class PodDisruptionBudget < KubernetesResource
+    TIMEOUT = 30.seconds
```
kirs: I think it takes way less time to apply PodDisruptionBudget? Does it have to be 30 seconds? I'm asking because it sounds inefficient to sleep for 30 seconds if `PodDisruptionBudget` failed to apply.

lib/kubernetes-deploy/runner.rb
psycotica0-shopify: In this case I didn't want the exception to be caught, since I considered it more like an assertion.

I could make it an ArgumentError if that's preferred though.

lib/kubernetes-deploy/kubernetes_resource/pod_disruption_budget.rb
psycotica0-shopify: I copied the 30 seconds from the ConfigMap which I considered to be very similar.

Maybe both should be changed?

KnVerey: It's true that 30s is overly generous for both. In practice configmaps appear pretty much instantly (unless of course `apply` fails, in which case we'd already fail the deploy right away), but that's not guaranteed. How about 10s, which will lead us to look for it 3 times?

Issue #52, opened 2017-02-27
Use --kubeconfig instead of env var in kubectl invocations


KnVerey:
Make `Kubectl`'s commands use `--kubeconfig` instead of relying on the env var. This will enable us to stop modifying the env var in tests. We should still derive the value of the flag from the environment, as is standard in other tools, so this change will be transparent to end users.

rendhalver:
Why don't we follow the conventions used by kubectl itself?
If you pass --kubeconfig it will ignore the `KUBECONFIG` env var.
Or is that what you are proposing?
This will keep the current behaviour (which is useful in some cases) and extend it.

KnVerey:
This issue is actually about an internal detail of our `KubernetesDeploy::Kubeclient` class, not the CLI args for our commands.

dturn:
Hi @rendhalver. I believe @KnVerey is suggesting: 

>If you pass --kubeconfig it will ignore the `KUBECONFIG` env var.

PR #97, opened 2017-05-24
Groundwork for output overhaul

KnVerey:
@Shopify/cloudplatform 

## Background

This PR is groundwork for an overhaul of deploy output. See [these screenshots](https://gist.github.com/KnVerey/fed879e4da297f6e7bbd21fcfbb671b0) for a preview of the goal (this PR is extracted from wip branch [improve_output](https://github.com/Shopify/kubernetes-deploy/compare/improve_output)). Notable goals of the overhaul for the purposes of this PR:
- make log lines more consistent (same prefixes, consistent left margin, consistent formatting)
- get rid of redundant logging
- get rid of `KUBESTATUS` in favour of targeted debug information in a new summary section. We may yet build a visualization, but it won't be all that soon, and ideally it wouldn't be done at the expense of deploy log clarity.

## Purpose of this PR

These changes make the gem use a single logger instance with access to deploy context for all logging, whether it's done by the main command (e.g. KubernetesDeploy::Runner), the kubectl executor, a KubernetesResource, or something else. 
* That logger's tags (i.e. the `[INFO][2017-05-24][context][namespace]` prefix) can _optionally_ include the context and namespace; this is now off by default, since it's noise for everything but Core.**
* The kubestatus logging's tag is now additional to those output by the unified logger. However, this logging will be removed entirely in the follow-up.
* `ENV[PRINT_LOGS]` is available to make the test logger print to stderr, so you can easily use tests for tophatting output.

**Fundamentally, the reason this PR changes the logger and Kubectl from class-level to instance-level constructs is the need for them to access the current context/namespace and an aversion to making these globally available at the class level. If these tags weren't important to core (without them, you have no idea what log line came from what deploy process), this PR could be much simpler. :-/

Incidentally fixes 2/3 items in https://github.com/Shopify/kubernetes-deploy/issues/52.

## Rollout

This doesn't improve much on its own, so I don't currently intend to release a version for it. However, it also shouldn't make the output any worse, so if we need to do a release between this and my main PR merging, that's fine.

🎩  before merging:
- [x] eyeball output of all three commands via tests
- [x] eyeball output of a few success/failure deploys via tests
- [x] run the exes to make sure I didn't break them

### DIFF
--- a/exe/kubernetes-deploy
+++ b/exe/kubernetes-deploy
@@ -10,7 +10,7 @@ template_dir = nil
 allow_protected_ns = false
 prune = true
 bindings = {}
-verbose_log_tags = false
+verbose_log_prefix = false
 
 ARGV.options do |opts|
   opts.on("--bindings=BINDINGS", Array, "k1=v1,k2=v2") do |binds|
@@ -26,7 +26,7 @@ ARGV.options do |opts|
   opts.on("--allow-protected-ns") { allow_protected_ns = true }
   opts.on("--no-prune") { prune = false }
   opts.on("--template-dir=DIR") { |v| template_dir = v }
-  opts.on("--verbose-log-tags") { verbose_log_tags = true }
+  opts.on("--verbose-log-prefix") { verbose_log_prefix = true }
   opts.parse!
 end
 
@@ -47,7 +47,7 @@ end
 
 namespace = ARGV[0]
 context = ARGV[1]
-logger = KubernetesDeploy::Logger.build(namespace, context, verbose_tags: verbose_log_tags)
+logger = KubernetesDeploy::FormattedLogger.build(namespace, context, verbose_prefix: verbose_log_prefix)
 
   runner = KubernetesDeploy::Runner.new(
     namespace: namespace,
@@ -60,7 +60,7 @@ logger = KubernetesDeploy::Logger.build(namespace, context, verbose_tags: verbos
 
 KubernetesDeploy::Runner.with_friendly_errors do
   runner.run(
-    skip_wait: skip_wait,
+    verify_result: !skip_wait,
     allow_protected_ns: allow_protected_ns,
     prune: prune
   )
--- a/exe/kubernetes-deploy
+++ b/exe/kubernetes-deploy
@@ -49,19 +49,18 @@ namespace = ARGV[0]
 context = ARGV[1]
 logger = KubernetesDeploy::FormattedLogger.build(namespace, context, verbose_prefix: verbose_log_prefix)
 
-  runner = KubernetesDeploy::Runner.new(
-    namespace: namespace,
-    context: context,
-    current_sha: revision,
-    template_dir: template_dir,
-    bindings: bindings,
-    logger: logger
-  )
+runner = KubernetesDeploy::Runner.new(
+  namespace: namespace,
+  context: context,
+  current_sha: revision,
+  template_dir: template_dir,
+  bindings: bindings,
+  logger: logger
+)
 
-KubernetesDeploy::Runner.with_friendly_errors do
-  runner.run(
-    verify_result: !skip_wait,
-    allow_protected_ns: allow_protected_ns,
-    prune: prune
-  )
-end
+success = runner.run(
+  verify_result: !skip_wait,
+  allow_protected_ns: allow_protected_ns,
+  prune: prune
+)
+exit 1 unless success
--- a/exe/kubernetes-deploy
+++ b/exe/kubernetes-deploy
@@ -10,6 +10,7 @@ template_dir = nil
 allow_protected_ns = false
 prune = true
 bindings = {}
+verbose_log_tags = false
 
 ARGV.options do |opts|
   opts.on("--bindings=BINDINGS", Array, "k1=v1,k2=v2") do |binds|
@@ -25,6 +26,7 @@ ARGV.options do |opts|
   opts.on("--allow-protected-ns") { allow_protected_ns = true }
   opts.on("--no-prune") { prune = false }
   opts.on("--template-dir=DIR") { |v| template_dir = v }
+  opts.on("--verbose-log-tags") { verbose_log_tags = true }
   opts.parse!
 end
 
@@ -43,16 +45,23 @@ revision = ENV.fetch('REVISION') do
   exit 1
 end
 
-KubernetesDeploy::Runner.with_friendly_errors do
+namespace = ARGV[0]
+context = ARGV[1]
+logger = KubernetesDeploy::Logger.build(namespace, context, verbose_tags: verbose_log_tags)
+
   runner = KubernetesDeploy::Runner.new(
-    namespace: ARGV[0],
-    context: ARGV[1],
+    namespace: namespace,
+    context: context,
     current_sha: revision,
     template_dir: template_dir,
-    wait_for_completion: !skip_wait,
-    allow_protected_ns: allow_protected_ns,
-    prune: prune,
-    bindings: bindings
+    bindings: bindings,
+    logger: logger
   )
-  runner.run
+
+KubernetesDeploy::Runner.with_friendly_errors do
+  runner.run(
+    skip_wait: skip_wait,
+    allow_protected_ns: allow_protected_ns,
+    prune: prune
+  )
 end
--- a/lib/kubernetes-deploy/ejson_secret_provisioner.rb
+++ b/lib/kubernetes-deploy/ejson_secret_provisioner.rb
@@ -22,6 +22,7 @@ module KubernetesDeploy
       @context = context
       @ejson_file = "#{template_dir}/#{EJSON_SECRETS_FILE}"
       @logger = logger
+      @kubectl = Kubectl.new(namespace: @namespace, context: @context, logger: @logger, log_failure_by_default: false)
     end
 
     def secret_changes_required?
@@ -61,7 +62,7 @@ module KubernetesDeploy
         next if ejson_secret_names.include?(secret_name)
 
         @logger.info("Pruning secret #{secret_name}")
-        out, err, st = kubectl.run("delete", "secret", secret_name)
+        out, err, st = @kubectl.run("delete", "secret", secret_name)
         @logger.debug(out)
         raise EjsonSecretError, err unless st.success?
       end
@@ -107,7 +108,7 @@ module KubernetesDeploy
       file.write(secret_yaml)
       file.close
 
-      out, err, st = kubectl.run("apply", "--filename=#{file.path}")
+      out, err, st = @kubectl.run("apply", "--filename=#{file.path}")
       @logger.debug(out)
       raise EjsonSecretError, err unless st.success?
     ensure
@@ -141,7 +142,7 @@ module KubernetesDeploy
     end
 
     def secret_exists?(secret_name)
-      _out, _err, st = kubectl.run("get", "secret", secret_name)
+      _out, _err, st = @kubectl.run("get", "secret", secret_name)
       st.success?
     end
 
@@ -186,14 +187,10 @@ module KubernetesDeploy
 
     def run_kubectl_json(*args)
       args += ["--output=json"]
-      out, err, st = kubectl.run(*args)
+      out, err, st = @kubectl.run(*args)
       raise EjsonSecretError, err unless st.success?
       result = JSON.parse(out)
       result.fetch('items', result)
-    end
-
-    def kubectl
-      @kubectl ||= Kubectl.new(namespace: @namespace, context: @context, logger: @logger, log_failure_by_default: false)
     end
   end
 end
--- a/lib/kubernetes-deploy/ejson_secret_provisioner.rb
+++ b/lib/kubernetes-deploy/ejson_secret_provisioner.rb
@@ -2,7 +2,6 @@
 require 'json'
 require 'base64'
 require 'open3'
-require 'kubernetes-deploy/logger'
 require 'kubernetes-deploy/kubectl'
 
 module KubernetesDeploy
@@ -18,13 +17,11 @@ module KubernetesDeploy
     EJSON_SECRETS_FILE = "secrets.ejson"
     EJSON_KEYS_SECRET = "ejson-keys"
 
-    def initialize(namespace:, context:, template_dir:)
+    def initialize(namespace:, context:, template_dir:, logger:)
       @namespace = namespace
       @context = context
       @ejson_file = "#{template_dir}/#{EJSON_SECRETS_FILE}"
-
-      raise FatalDeploymentError, "Cannot create secrets without a namespace" if @namespace.blank?
-      raise FatalDeploymentError, "Cannot create secrets without a context" if @context.blank?
+      @logger = logger
     end
 
     def secret_changes_required?
@@ -42,7 +39,7 @@ module KubernetesDeploy
       with_decrypted_ejson do |decrypted|
         secrets = decrypted[MANAGED_SECRET_EJSON_KEY]
         unless secrets.present?
-          KubernetesDeploy.logger.warn("#{EJSON_SECRETS_FILE} does not have key #{MANAGED_SECRET_EJSON_KEY}."\
+          @logger.warn("#{EJSON_SECRETS_FILE} does not have key #{MANAGED_SECRET_EJSON_KEY}."\
             "No secrets will be created.")
           return
         end
@@ -63,9 +60,9 @@ module KubernetesDeploy
         next unless secret_managed?(secret)
         next if ejson_secret_names.include?(secret_name)
 
-        KubernetesDeploy.logger.info("Pruning secret #{secret_name}")
-        out, err, st = run_kubectl("delete", "secret", secret_name)
-        KubernetesDeploy.logger.debug(out)
+        @logger.info("Pruning secret #{secret_name}")
+        out, err, st = kubectl.run("delete", "secret", secret_name)
+        @logger.debug(out)
         raise EjsonSecretError, err unless st.success?
       end
     end
@@ -103,15 +100,15 @@ module KubernetesDeploy
 
     def create_or_update_secret(secret_name, secret_type, data)
       msg = secret_exists?(secret_name) ? "Updating secret #{secret_name}" : "Creating secret #{secret_name}"
-      KubernetesDeploy.logger.info(msg)
+      @logger.info(msg)
 
       secret_yaml = generate_secret_yaml(secret_name, secret_type, data)
       file = Tempfile.new(secret_name)
       file.write(secret_yaml)
       file.close
 
-      out, err, st = run_kubectl("apply", "--filename=#{file.path}")
-      KubernetesDeploy.logger.debug(out)
+      out, err, st = kubectl.run("apply", "--filename=#{file.path}")
+      @logger.debug(out)
       raise EjsonSecretError, err unless st.success?
     ensure
       file.unlink if file
@@ -144,7 +141,7 @@ module KubernetesDeploy
     end
 
     def secret_exists?(secret_name)
-      _out, _err, st = run_kubectl("get", "secret", secret_name)
+      _out, _err, st = kubectl.run("get", "secret", secret_name)
       st.success?
     end
 
@@ -166,7 +163,7 @@ module KubernetesDeploy
     end
 
     def decrypt_ejson(key_dir)
-      KubernetesDeploy.logger.info("Decrypting #{EJSON_SECRETS_FILE}")
+      @logger.info("Decrypting #{EJSON_SECRETS_FILE}")
       # ejson seems to dump both errors and output to STDOUT
       out_err, st = Open3.capture2e("EJSON_KEYDIR=#{key_dir} ejson decrypt #{@ejson_file}")
       raise EjsonSecretError, out_err unless st.success?
@@ -176,7 +173,7 @@ module KubernetesDeploy
     end
 
     def fetch_private_key_from_secret
-      KubernetesDeploy.logger.info("Fetching ejson private key from secret #{EJSON_KEYS_SECRET}")
+      @logger.info("Fetching ejson private key from secret #{EJSON_KEYS_SECRET}")
 
       secret = run_kubectl_json("get", "secret", EJSON_KEYS_SECRET)
       encoded_private_key = secret["data"][public_key]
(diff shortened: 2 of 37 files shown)


lib/kubernetes-deploy/ejson_secret_provisioner.rb
```diff
@@ -18,13 +17,11 @@ class EjsonSecretProvisioner
     EJSON_SECRETS_FILE = "secrets.ejson"
     EJSON_KEYS_SECRET = "ejson-keys"
 
-    def initialize(namespace:, context:, template_dir:)
+    def initialize(namespace:, context:, template_dir:, logger:)
       @namespace = namespace
       @context = context
       @ejson_file = "#{template_dir}/#{EJSON_SECRETS_FILE}"
-
-      raise FatalDeploymentError, "Cannot create secrets without a namespace" if @namespace.blank?
```
KnVerey: `Kubectl.new` will take care of raising if these are blank

exe/kubernetes-deploy
```diff
@@ -43,16 +45,23 @@ revision = ENV.fetch('REVISION') do
...
+    logger: logger
+  )
+
+KubernetesDeploy::Runner.with_friendly_errors do
+  runner.run(
+    skip_wait: skip_wait,
     allow_protected_ns: allow_protected_ns,
-    prune: prune,
-    bindings: bindings
+    prune: prune
```
KnVerey: I split these out simply because the number of args to the initializer was getting crazy, and these seem like switches on the run itself.

lib/kubernetes-deploy/logger.rb
```diff
@@ -3,42 +3,42 @@
...
-            end
-          end
-          l
+        case severity
+        when "FATAL"
+          colorized_line.red
+        when "ERROR", "WARN"
+          colorized_line.yellow
+        when "INFO"
+          msg.match(%r{^\[(KUBESTATUS|Pod)}) ? colorized_line : colorized_line.blue
```
KnVerey: This matching is a temporary hack to keep from over-coloring the output if this is deployed without the output revamp PR, which will remove the hack.

lib/kubernetes-deploy/runner.rb
```diff
@@ -69,39 +69,24 @@ class Runner
 
     def self.with_friendly_errors
       yield
-    rescue FatalDeploymentError => error
-      KubernetesDeploy.logger.fatal <<-MSG
```
KnVerey: I moved this logging into a rescue on `run` (that re-raises, so we still land here) so that the output if you're tophatting via a test (which does not use `with_friendly_errors`) is consistent with what you'd see by running the executable (which does).

lib/kubernetes-deploy/runner.rb
```diff
@@ -69,39 +69,24 @@ class Runner
...
-#{error.class}: #{error.message}
-  #{error.backtrace && error.backtrace.join("\n  ")}
```
KnVerey: I removed the backtrace because in our experience seeing it leads people to assume the deploy itself blew up and the failure is not their own fault.

test/unit/kubernetes-deploy/resource_watcher_test.rb
```diff
@@ -33,38 +33,67 @@ def test_failure_with_mock_resource
...
-    watcher = KubernetesDeploy::ResourceWatcher.new([resource])
+    watcher = KubernetesDeploy::ResourceWatcher.new([resource], logger: test_logger)
     watcher.run(delay_sync: 0.1)
 
     assert_logs_match(/Waiting for web-pod with 1s timeout/)
     assert_logs_match(/web-pod failed to deploy with status 'timeout'/)
     assert_logs_match(/Spent (\S+)s waiting for web-pod/)
   end
 
+  def test_wait_logging_when_resources_do_not_finish_together
```
KnVerey: The test class refactors are for the purposes of this test, which will become more meaningful with the upcoming logging changes.

exe/kubernetes-deploy
```diff
@@ -25,6 +26,7 @@ ARGV.options do |opts|
   opts.on("--allow-protected-ns") { allow_protected_ns = true }
   opts.on("--no-prune") { prune = false }
   opts.on("--template-dir=DIR") { |v| template_dir = v }
+  opts.on("--verbose-log-tags") { verbose_log_tags = true }
```
KnVerey: This is for core to use to turn on the inclusion of `[context][namespace]` on all log lines. Suggestions for better flag name appreciated.

lib/kubernetes-deploy/logger.rb
```diff
@@ -3,43 +3,43 @@
...
-            end
-          end
-          l
+        case severity
+        when "FATAL"
+          colorized_line.red
+        when "ERROR", "WARN"
+          colorized_line.yellow
+        when "INFO"
+          msg =~ /^\[(KUBESTATUS|Pod)/ ? colorized_line : colorized_line.blue
```
KnVerey: This matching is a temporary hack to keep from over-coloring the output if this is deployed without the output revamp PR, which will remove the hack.

exe/kubernetes-deploy
camilo: > Suggestions for better flag name appreciated

`--verbose-log-prefix`? 

I think the name is fine tho.

lib/kubernetes-deploy/runner.rb
```diff
@@ -339,32 +338,24 @@ def deploy_resources(resources, prune: false)
...
-      if namespaced
-        raise KubectlError, "Namespace missing for namespaced command" if @namespace.blank?
-      end
-
-      if with_context
-        raise KubectlError, "Explicit context is required to run this command" if @context.blank?
-      end
-
-      Kubectl.run_kubectl(*args, namespace: @namespace, context: @context)
+    def kubectl
```
camilo: why the lazy creation here? can't `@kubectl` be assigned at initailization time?

lib/kubernetes-deploy/runner.rb
```diff
@@ -123,15 +113,18 @@ def run
...
-      deploy_resources(resources, prune: @prune)
+      deploy_resources(resources, prune: prune)
 
-      return unless wait_for_completion?
+      return if skip_wait
       wait_for_completion(resources)
       report_final_status(resources)
+    rescue FatalDeploymentError => error
+      @logger.fatal "#{error.class}: #{error.message}"
+      raise error
```
kirs: `raise` without argument from inside a rescue block re-raises the original exception. It's up to you but `raise` and `raise error` would have the same result here.

test/helpers/fixture_deploy_helper.rb
```diff
@@ -51,12 +51,14 @@ def deploy_dir(dir, wait: true, allow_protected_ns: false, prune: true, bindings
...
       template_dir: dir,
-      wait_for_completion: wait,
-      allow_protected_ns: allow_protected_ns,
-      prune: prune,
+      logger: test_logger,
       bindings: bindings
     )
-    runner.run
+    runner.run(
+      skip_wait: !wait,
```
kirs: I preferred `wait_for_completion` instead of `skip_wait` here because I usually try to avoid double negation in flag names (`disabled = true` vs `enabled = false`). What do you think?

test/test_helper.rb
```diff
@@ -20,18 +20,29 @@
...
+      end
     end
 
     def teardown
+      ColorizedString.disable_colorization = false
       @logger_stream.close
     end
 
     def assert_logs_match(regexp, times = nil)
+      if ENV["PRINT_LOGS"]
```
kirs: Do you want to keep `PRINT_LOGS` in the test suite for debugging? I'm not against it, actually I used something like this myself too.

test/unit/kubernetes-deploy/kubectl_test.rb
```diff
@@ -0,0 +1,87 @@
...
+  end
+
+  private
+
+  def build_kubectl(log_failure_by_default: true)
+    KubernetesDeploy::Kubectl.new(namespace: 'testn', context: 'testc', logger: test_logger,
+      log_failure_by_default: log_failure_by_default)
+  end
+
+  def stub_open3(command, resp:, err: "", success: true)
```
kirs: 👍 for the helper

test/unit/kubernetes-deploy/kubectl_test.rb
```diff
@@ -0,0 +1,87 @@
...
+
+  def test_run_does_not_log_failures_when_log_failure_by_default_is_false_and_override_is_unspecified
+    stub_open3(%w(kubectl get pods --namespace=testn --context=testc), resp: "", err: "oops", success: false)
+    test_logger.expects(:warn).never
+    build_kubectl(log_failure_by_default: false).run("get", "pods")
+  end
+
+  def test_run_does_not_log_failures_when_log_failure_by_default_is_false_and_override_is_also_false
+    stub_open3(%w(kubectl get pods --namespace=testn --context=testc), resp: "", err: "oops", success: false)
+    test_logger.expects(:warn).never
```
kirs: I see why we'd use `expects` for open3 but for the logger? I'd just assert that the resulted log doesn't have `WARN` keywords, instead of `expects(:warn).never`

test/unit/kubernetes-deploy/logger_test.rb
```diff
@@ -2,39 +2,65 @@
...
+    # [WARN][2017-05-19 20:07:31 -0400]\tWarning
+    # [WARN][2017-05-19 20:07:31 -0400]\t
+    # [ERROR][2017-05-19 20:07:31 -0400]\tError
+    # [FATAL][2017-05-19 20:07:31 -0400]\t
+    # [FATAL][2017-05-19 20:07:31 -0400]\tFatal
+
+    @logger_stream.rewind
+    scanner = StringScanner.new(@logger_stream.read)
+
+    assert scanner.scan_until(/^\[INFO\].*\]\tFYI$/)
```
kirs: `scan_until` 😍

lib/kubernetes-deploy/ejson_secret_provisioner.rb
```diff
@@ -189,14 +186,14 @@ def fetch_private_key_from_secret
...
-      out, err, st = run_kubectl(*args)
+      out, err, st = kubectl.run(*args)
       raise EjsonSecretError, err unless st.success?
       result = JSON.parse(out)
       result.fetch('items', result)
     end
 
-    def run_kubectl(*args)
-      Kubectl.run_kubectl(*args, namespace: @namespace, context: @context, log_failure: false)
+    def kubectl
```
KnVerey: Do you mean vs. passing it in like I'm doing for the logger? Mostly because this one (like the resources one) uses `log_failure_by_default: false` whereas the runner wants `log_failure_by_default: true`.

lib/kubernetes-deploy/logger.rb
```diff
@@ -3,43 +3,43 @@
...
+        else
+          colorized_line
         end
       end
 
-      private
+      def l.blank_line(level = :info)
```
KnVerey: Yeah, that's much better. Thanks for calling out the hack. 😄

lib/kubernetes-deploy/runner.rb
KnVerey: No particular reason. I can change it.

test/helpers/fixture_deploy_helper.rb
KnVerey: The thing that triggered this change was that this option becomes a local variable in `run` that had the same name as the actual `wait_for_completion` method (that does the waiting), which was confusing. I reached for this name to match the CLI flag (`opts.on("--skip-wait") { skip_wait = true }`), but I could change it to just `wait` or maybe `verify_result` instead.

test/test_helper.rb
KnVerey: Yes, both for debugging and for developing features that affect the log output. The integration tests are a really nice way to run through various scenarios!

test/unit/kubernetes-deploy/kubectl_test.rb
KnVerey: Fair. I'll add a `refute_logs_match` for this.

lib/kubernetes-deploy/runner.rb
KnVerey: OTOH for the Runner, the laziness lets us raise friendlier errors (from `validate_configuration` inside `run` instead of from the initializer building the kubectl instance), and for KubernetesResource, I think it makes more sense for the abstract superclass to provide a ready-to-use method than an ivar that hopefully subclass initializers won't forget to populate.

test/helpers/fixture_deploy_helper.rb
kirs: I think even `opts.on("--skip-wait") { wait_for_completion = false }` would be better than `skip_wait` in the runner

lib/kubernetes-deploy/ejson_secret_provisioner.rb
kirs: I asked because there's `kubectl` method defined on the resource level and I didn't see why `EjsonSecretProvisioner` would redefine this method. `log_failure_by_default: false` is a good reason 👍

KnVerey:
I'll rebase/fix policial in the morning, but please take a look at the commits I just pushed, especially the second:
* e474a4f **Code Review** -- what it sounds like
* ddc7ac2 **All core run/perform methods should handle output and return boolean** -- This corrects an oversight relating to `with_friendly_errors`. Previously, I changed it to just rescue the error and exit, assuming the error messages had already been logged. I decided this is dangerous/misleading. Instead, this commit gets rid of that method entirely, requiring each of the run/perform methods to return a boolean the executables can use to determine their exit code. The change itself is quite small/simple, but a bunch of tests needed to be updated. (this is exactly what we discussed IRL @camilo )
* e5c7cbe  **Don't make the test logger accessor an actual test** -- Rename method to fix a funny problem
* 9219b0f **Disable net connect in unit tests** -- I noticed in preparing ddc7ac2 that some of the "unit" tests required connecting to minikube. I moved the tests in question in that commit, but this commit actually makes the minor changes required to prevent such behaviour.

test/unit/kubernetes-deploy/runner_task_test.rb
```diff
@@ -3,52 +3,20 @@
 require 'kubernetes-deploy/runner_task'
 
 class RunnerTaskUnitTest < KubernetesDeploy::TestCase
-  def test_missing_namespace
```
KnVerey: This and the last test in the file are the two unit tests I mentioned in my comment that require connecting to minikube. I moved them to the integration test file.

kirs:
@KnVerey what advantage do you see in return value flow vs the previous `raise + with_friendly_errors`? I'm not against it, just trying to get the reasoning.

exe/kubernetes-deploy
```diff
@@ -43,16 +45,22 @@ revision = ENV.fetch('REVISION') do
...
+runner = KubernetesDeploy::Runner.new(
+  namespace: namespace,
+  context: context,
+  current_sha: revision,
+  template_dir: template_dir,
+  bindings: bindings,
+  logger: logger
+)
+
+success = runner.run(
```
kirs: I like the distinguish between initializer flags and run flags. Good design 👍

exe/kubernetes-deploy
```diff
@@ -43,16 +45,22 @@ revision = ENV.fetch('REVISION') do
...
+  verify_result: !skip_wait,
```
kirs: What was the reason to change `wait_for_completion` to `verify_result`? What this flag enables is waiting for resources to become available. I think it's still more _waiting_ than _verifying_

test/unit/kubernetes-deploy/runner_task_test.rb
```diff
@@ -1,50 +1,27 @@
...
-      )
-
-      task_runner.run(
-        task_template: 'hello-cloud-template-runner',
-        entrypoint: nil,
-        args: nil
-      )
-    end
+  def setup
+    Kubeclient::Client.any_instance.stubs(:discover)
```
kirs: If you want unit tests to never hit kubeclient, does it make sense to use `Kubeclient::Client.any_instance.expects(:discover).never`?

test/unit/kubernetes-deploy/resource_watcher_test.rb
```diff
@@ -33,38 +33,67 @@ def test_failure_with_mock_resource
...
+    assert_logs_match(/Waiting for first, second, third, fourth with 4s timeout/)
+    assert_logs_match(/second failed to deploy with status 'timeout'/)
+    assert_logs_match(/third failed to deploy with status 'failed'/)
+    assert_logs_match(/Spent (\S+)s waiting for first, second, third, fourth/)
+  end
+
   private
 
-  MockResource = Struct.new(:id, :timeout, :status) do
+  MockResource = Struct.new(:id, :hits_to_complete, :status) do
```
kirs: 👍 for `hits`

exe/kubernetes-deploy
KnVerey: The actual method in `Runner` that does the waiting is called `wait_for_completion`, and now that this is an arg to `Runner.run`, it ends up being a local var. So you have both a local var and a method called `wait_for_completion` being used on adjacent lines. I don't feel as strongly as you about this naming, so I'll use whatever you want. But I need two names: one for the flag and one for the method.

test/unit/kubernetes-deploy/runner_task_test.rb
KnVerey: That would seem better, but the problem is that our `build_kubeclient` method eagerly call `discover`, which means that if the kubeclient instance so much as gets created, discover will get called. Tbh the errors that happen if the test does end up calling a kubeclient method is super ugly; since discover was stubbed out, none of the `get_` (etc.) methods are defined, so it's a NoMethodError. I'm not sure what to do about that though.

kirs: We could probably record the discovery as a VCR cassette to make unit tests work. The downside that I see is that those would have to be rebuilt with new k8s versions.

KnVerey:
> what advantage do you see in return value flow vs the previous raise + with_friendly_errors? I'm not against it, just trying to get the reasoning.

Ok, I'll try to explain the whole thought process.  😄 

The main problem with the exception-based flow we had before this PR was that it caused tests to behave differently than production with respect to logging. To address that, this PR originally just gutted `with_friendly_errors`, making the run/perform methods in charge of the logging (so that it always happened, even in tests) and then re-raise. `with_friendly_errors` was  then just catching the error a second time and calling `exit 1`. 

I realized yesterday that that new behaviour sucks for a few reasons. First, the name `with_friendly_errors` is super misleading... it basically suppresses errors now, haha. This entire method would be better replaced with a begin/rescue in the executables themselves--I also prefer keeping the exit codes in the executables in general, so that seemed fine. However, looking at that solution I felt the need to write a comment to the effect of "This exception has already been logged, so just exit silently". That's true right now, but feels like a sketchy assumption to introduce. So I then changed the run/perform methods to raise a new `JustExit1Error` instead of the original. And that's when I realized that we're effectively using this exception to communicate a boolean result to the executable, from a place that could easily just return. Furthermore, logging notwithstanding, that exception itself was a discrepancy between the behaviour our users see and the integration tests, which is something I want to minimize. 

Make sense?

test/unit/kubernetes-deploy/runner_task_test.rb
KnVerey: That's a good idea. Are you ok with this for now though? This PR is getting enormous. 😄

kirs: Absolutely, jus thoughts for the future.

kirs:
Makes sense. One downside of communicating with boolean is that it would't work that well with `TaskRunner` which can be executed by other Ruby process from tools like failover that @xldenis have been building.

Previously `TaskRunner` would raise with an error and you could use that error in failover script to determine what exactly went wrong. Now there's only boolean flag and a log in own format that I don't think failover should parse.

KnVerey:
> Previously TaskRunner would raise with an error and you could use that error in failover script to determine what exactly went wrong. Now there's only boolean flag and a log in own format that I don't think failover should parse.

Oh, so the failover is not using the executable? I did not realize that. How about we make a `run!` twin that does not catch the exception for RunnerTask specifically? @xldenis

kirs:
Yep it's not using executable:
https://github.com/Shopify/activefailover/blob/eb700f6c41997adddfe20b340c61500b6bb9c797/test/unit/task/transfer_jobs_test.rb#L62

KnVerey:
Pushed another commit adding `run!` (and some test for it) to `RunnerTask` specifically. So ActiveFailover will simply need to add `!` to preserve the behaviour when it updates the gem.

Final 👀  please @kirs @camilo.

kirs:
LGTM! Great idea about about `run!`.

xldenis:
> Pushed another commit adding run! (and some test for it) to RunnerTask specifically. So ActiveFailover will simply need to add ! to preserve the behaviour when it updates the gem.

SGTM. I prefer not shelling out to just call ruby again :P also plays nicer with controlling log output and handling errors.

lib/kubernetes-deploy/runner.rb
```diff
@@ -68,41 +68,20 @@ class Runner
     PRUNE_WHITELIST_V_1_5 = %w(extensions/v1beta1/HorizontalPodAutoscaler).freeze
     PRUNE_WHITELIST_V_1_6 = %w(autoscaling/v1/HorizontalPodAutoscaler).freeze
 
-    def self.with_friendly_errors
```
camilo: 🔥 🔥 🔥 🔥 🔥 🔥

lib/kubernetes-deploy/runner.rb
```diff
@@ -293,37 +283,36 @@ def validate_configuration
...
+          _, _, st = kubectl.run("replace", "--force", "-f", r.file.path, log_failure: false)
         else
           # Fail Fast! This is a programmer mistake.
           raise ArgumentError, "Unexpected deploy method! (#{r.deploy_method.inspect})"
         end
 
+        next if st.success?
+        # it doesn't exist so we can't replace it
+        _, err, st = kubectl.run("create", "-f", r.file.path, log_failure: false)
         unless st.success?
```
camilo: Isn't this redundant give the fact that you are bailing here: `next if st.success?`

KnVerey: This is a new `st` from line 310: the status of the create command, rather than the original replace command.

camilo: ah derp, yes.

PR #106, opened 2017-06-03
More tweaks to deploy output

KnVerey:
@Shopify/cloudplatform 

## Purpose
This PR makes a few more small improvements to the deploy output:

1) Makes the `Successfully deployed in X.Xs` statements more accurate by making them compare to the actual time that step started rather than the time the watch started (which led to a lot of things supposedly deploying in 0.0s).

2) Changes the "Deploying resources:" logging to be a single line when only one resource is being deployed.

Before:

image

After: 

image

3) Puts the headings back to being _cyan_. When I switched to using the colorize gem, I used `:blue`, but they looked super dark in Shipit, so I switched them to light blue. Should have just checked the original code: `when "INFO" then "\033[0;36m#{log_text}\x1b[0m\n" # blue` : the comment is lies, it is cyan. Right now the headings are white because apparently light blue doesn't show so well on shipit. 😞 

4) Limits the max number of log lines displayed from a managed pod. See inline comment for more.

### Rubocop / Policial

The `Rubocop` commit fixes all outstanding rubocop violations in the project. Policial does not like that I fixed `kubernetes-deploy.rb` by exempting the filename, but I do not see the value in changing the main file's name, affecting `require`s everywhere. After this commit, you can run a quick `bundle exec rubocop -a` on this repo to fix stupid spacing mistakes like the one I originally made on this PR.

image


lib/kubernetes-deploy/kubernetes_resource/pod.rb
```diff
@@ -75,12 +75,14 @@ def fetch_logs
...
-        out, _err, _st = kubectl.run(
+        cmd = [
           "logs",
           @name,
           "--container=#{container_name}",
-          "--since-time=#{@deploy_started.to_datetime.rfc3339}"
-        )
+          "--since-time=#{@deploy_started.to_datetime.rfc3339}",
+        ]
+        cmd << "--tail=50" unless unmanaged?
```
KnVerey: The assumption is that unmanaged pods (i.e. where you have a pod template in your deploy directory) are performing a deploy-time operation and you want to see all their output inline. When we dump logs from a managed pod however (e.g. on deployment failure), the line count could be absurd, making the summary section unusable.

In researching deploy failures that happened since I made the output changes, I found a couple examples of super long log dumps:
- [this one](https://shipit.shopify.io/shopify/plus-account-centre/production/deploys/451844) has 128 lines of stack trace
- [this one](https://shipit.shopify.io/shopify/facebook-commerce/staging/deploys/451871) has 126 lines of output that looks completely irrelevant to the deploy failure

Arguably, the stack trace case won't be helpful truncated to the last 50 lines. But then again, maybe the takeaway in such cases should be "I need to check Bugsnag"? I'm really not sure what the limit should be, but we should have _something_, even if it is really high. What do you think?

karanthukral:
code LGTM. Not sure what a sane limit on log truncation would be though

lib/kubernetes-deploy/runner_task.rb
```diff
@@ -68,11 +68,8 @@ def validate_configuration(task_template, args)
       begin
         @kubeclient.get_namespace(@namespace) if @namespace.present?
       rescue KubeException => e
-        errors << if e.error_code == 404
-          "Namespace was not found"
-        else
-          "Could not connect to kubernetes cluster"
-        end
+        msg = e.error_code == 404 ? "Namespace was not found" : "Could not connect to kubernetes cluster"
```
stefanmb: What was wrong with the old formulation? I find it easier to read than the ternary operator.

KnVerey: Rubocop was asking for crazy non-standard indentation on the old version, seemingly because of a problem recognizing the left margin when the return value of the `if` is being concatenated. As much as I dislike changing valid code to please robots, I wasn't able to quickly find the relevant rubocop setting and didn't want to waste time on it. Ultimately, this conditional will be replaced with rescuing two different exceptions once we have a version of kubeclient with [kirs's patch](https://github.com/abonas/kubeclient/pull/233) included.

Issue #42, opened 2017-02-27
Zero-replicas deployments never succeed


KnVerey:
[Example](https://shipit.shopify.io/shopify/services-db/production/deploys/424256)

The monitoring for both the deployment (which currently has a guard clause on at least one replica being available) and the service (which waits to have endpoints to succeed) is causing this.

Issue #36, opened 2017-02-27
Deployment rollout monitoring revamp


KnVerey:
The fact that deployment rollout monitoring currently looks at all pods associated with that deployment instead of pods in the new ReplicaSet has caused several different bugs:
- ~~Deploys never succeed when there are evicted pods associated with the deployment, even though those pods are old~~ (fixed another way)
- Pod warnings get shown for pods that are being shut down, in the case where the last deploy was bad and the current one is actually succeeding (very confusing output)
- Deploy success is delayed by waiting for all old pods to fully disappear
- ~~[This](https://shipit.shopify.io/shopify/stock-photos/production/deploys/419524) false-positive deploy result is likely caused by this.~~ ~~Last poll before it "succeeded" was:~~ (I now think it probably briefly became available before failing a probe, or something like that)
```
[KUBESTATUS] {"group":"Deployments","name":"jobs","status_string":"Waiting for rollout to finish: 0 of 1 updated replicas are available...","exists":true,"succeeded":true,"failed":false,"timed_out":false,"replicas":{"updatedReplicas":1,"replicas":1,"availableReplicas":1},"num_pods":1}
```

Related technical notes:
- We are selecting related pods using an assumption that they are labelled with the deployment name. I believe this is a bad assumption that has flown under the radar so far because all our templates are labelled this way by convention. The new ReplicaSet version should not do this.
- [Here's how kubectl gets old/new rs](https://github.com/kubernetes/kubernetes/blob/a9f5065833d782ea93a6f462674fc7d3fb3affe7/pkg/controller/deployment/util/deployment_util.go#L490-L511)

KnVerey:
- We should base the timeout on `progressDeadlineSeconds` when present.
- With k8s 1.6, we should select pod based on the new `ownerRef` field.

PR #107, opened 2017-06-05
Replicaset awareness

KnVerey:
@Shopify/cloudplatform 

This PR makes `KubernetesResource::Deployment` aware of replica sets. This should fix a number of outstanding issues:
- Pod warnings get shown for pods that are being shut down, in the case where the last deploy was bad and the current one is actually succeeding (very confusing output)
- The way we try to identify a representative pod for the purpose of outputting related events and logs on failed deploy is sketchy (https://github.com/Shopify/kubernetes-deploy/pull/98#discussion_r119244951) and could theoretically still select a pod from the wrong RS.
- Rolling updates currently cannot fail, and will never time out on the basis of their pod states ([notes in code](https://github.com/Shopify/kubernetes-deploy/blob/master/lib/kubernetes-deploy/kubernetes_resource/deployment.rb#L62-L67))

Fixes https://github.com/Shopify/kubernetes-deploy/issues/36
Fixes https://github.com/Shopify/kubernetes-deploy/issues/42 (with small amount of extra code + test)

## Problem

We currently make the link between a deployment and its pods by looking for a "name" label. This is super fragile, since we're not using the full selector set, plus the presence of that label is a Shopify convention, not a k8s requirement. Furthermore, even using the full selector set does not guarantee you're looking at the right pods if you've gotten yourself into a really bad state (selector overlap). Ideally we'd connect Deployments to the pods we care about in a robust way; this is much easier to do as of Kubernetes 1.6 (see "Additional context" below).

## Solution

This PR uses the full set of label selectors to grab all possibly relevant RS. It then uses the new `ownerReferences` information to select the RS it owns, and the `deployment.kubernetes.io/revision` to select the current RS. That RS in turn uses label selectors and `ownerReferences` to list its pods.

## Additional context

- kubectl itself identifies the new replicaset by calculating and comparing pod template hashes: https://github.com/kubernetes/kubernetes/blob/a9f5065833d782ea93a6f462674fc7d3fb3affe7/pkg/controller/deployment/util/deployment_util.go#L611-L622. I don' know whether they plan to switch to the new attributes I'm using here, but either way, that approach is not feasible for us.
- `ownerReferences` were added to Deployment in 1.6. The [PR](https://github.com/kubernetes/kubernetes/pull/35676) that introduced them mentions they will be used for RS garbage collection. However, the [issue it resolved](https://github.com/kubernetes/kubernetes/issues/33845) mentions use cases like ours here.
- The `deployment.kubernetes.io/revision` annotation is much older (circa 1.2). According to the [PR](https://github.com/kubernetes/kubernetes/pull/19581) introducing it, the rollout history and rollback features leverage it.

### DIFF
--- a/lib/kubernetes-deploy/kubernetes_resource.rb
+++ b/lib/kubernetes-deploy/kubernetes_resource.rb
@@ -136,10 +136,11 @@ module KubernetesDeploy
         if container_logs.blank? || container_logs.values.all?(&:blank?)
           helpful_info << "  - Logs: #{DEBUG_RESOURCE_NOT_FOUND_MESSAGE}"
         else
-          helpful_info << "  - Logs (last #{LOG_LINE_COUNT} lines shown):"
-          container_logs.each do |identifier, logs|
-            logs.split("\n").each do |line|
-              helpful_info << "      [#{identifier}]\t#{line}"
+          sorted_logs = container_logs.sort_by { |_, log_lines| log_lines.length }
+          sorted_logs.each do |identifier, log_lines|
+            helpful_info << "  - Logs from container '#{identifier}' (last #{LOG_LINE_COUNT} lines shown):"
+            log_lines.each do |line|
+              helpful_info << "      #{line}"
             end
           end
         end
--- a/lib/kubernetes-deploy/kubernetes_resource/deployment.rb
+++ b/lib/kubernetes-deploy/kubernetes_resource/deployment.rb
@@ -4,78 +4,84 @@ module KubernetesDeploy
     TIMEOUT = 5.minutes
 
     def sync
-      json_data, _err, st = kubectl.run("get", type, @name, "--output=json")
+      raw_json, _err, st = kubectl.run("get", type, @name, "--output=json")
       @found = st.success?
-      @rollout_data = {}
-      @status = nil
-      @representative_pod = nil
-      @pods = []
 
       if @found
-        @rollout_data = JSON.parse(json_data)["status"]
-          .slice("updatedReplicas", "replicas", "availableReplicas", "unavailableReplicas")
-        @status = @rollout_data.map { |replica_state, num| "#{num} #{replica_state}" }.join(", ")
-
-        pod_list, _err, st = kubectl.run("get", "pods", "-a", "-l", "name=#{name}", "--output=json")
-        if st.success?
-          pods_json = JSON.parse(pod_list)["items"]
-          pods_json.each do |pod_json|
-            pod = Pod.new(
-              namespace: namespace,
-              context: context,
-              definition: pod_json,
-              logger: @logger,
-              parent: "#{@name.capitalize} deployment",
-              deploy_started: @deploy_started
-            )
-            pod.sync(pod_json)
-
-            if !@representative_pod && pod_probably_new?(pod_json)
-              @representative_pod = pod
-            end
-            @pods << pod
-          end
-        end
+        deployment_data = JSON.parse(raw_json)
+        @latest_rs = find_latest_rs(deployment_data)
+        @rollout_data = { "replicas" => 0 }.merge(deployment_data["status"]
+          .slice("replicas", "updatedReplicas", "availableReplicas", "unavailableReplicas"))
+        @status = @rollout_data.map { |state_replicas, num| "#{num} #{state_replicas.chop.pluralize(num)}" }.join(", ")
+      else # reset
+        @latest_rs = nil
+        @rollout_data = { "replicas" => 0 }
+        @status = nil
       end
-    end
-
-    def fetch_logs
-      @representative_pod ? @representative_pod.fetch_logs : {}
     end
 
     def fetch_events
       own_events = super
-      return own_events unless @representative_pod
-      own_events.merge(@representative_pod.fetch_events)
+      return own_events unless @latest_rs.present?
+      own_events.merge(@latest_rs.fetch_events)
+    end
+
+    def fetch_logs
+      return {} unless @latest_rs.present?
+      @latest_rs.fetch_logs
     end
 
     def deploy_succeeded?
-      return false unless @rollout_data.key?("availableReplicas")
-      # TODO: this should look at the current replica set's pods too
+      return false unless @latest_rs
+
+      @latest_rs.deploy_succeeded? &&
+      @latest_rs.desired_replicas == desired_replicas && # latest RS fully scaled up
       @rollout_data["updatedReplicas"].to_i == @rollout_data["replicas"].to_i &&
       @rollout_data["updatedReplicas"].to_i == @rollout_data["availableReplicas"].to_i
     end
 
     def deploy_failed?
-      # TODO: this should look at the current replica set's pods only or it'll never be true for rolling updates
-      @pods.present? && @pods.all?(&:deploy_failed?)
+      @latest_rs && @latest_rs.deploy_failed?
     end
 
     def deploy_timed_out?
-      # TODO: this should look at the current replica set's pods only or it'll never be true for rolling updates
-      super || @pods.present? && @pods.all?(&:deploy_timed_out?)
+      super || @latest_rs && @latest_rs.deploy_timed_out?
     end
 
     def exists?
       @found
     end
 
+    def desired_replicas
+      @definition["spec"]["replicas"].to_i
+    end
+
     private
 
-    def pod_probably_new?(pod_json)
-      return false unless @deploy_started
-      # Shitty, brittle workaround to identify a pod from the new ReplicaSet before implementing ReplicaSet awareness
-      Time.parse(pod_json["metadata"]["creationTimestamp"]) >= (@deploy_started - 30.seconds)
+    def find_latest_rs(deployment_data)
+      label_string = deployment_data["spec"]["selector"]["matchLabels"].map { |k, v| "#{k}=#{v}" }.join(",")
+      raw_json, _err, st = kubectl.run("get", "replicasets", "--output=json", "--selector=#{label_string}")
+      return unless st.success?
+
+      all_rs_data = JSON.parse(raw_json)["items"]
+      current_revision = deployment_data["metadata"]["annotations"]["deployment.kubernetes.io/revision"]
+
+      latest_rs_data = all_rs_data.find do |rs|
+        rs["metadata"]["ownerReferences"].any? { |ref| ref["uid"] == deployment_data["metadata"]["uid"] } &&
+        rs["metadata"]["annotations"]["deployment.kubernetes.io/revision"] == current_revision
+      end
+      return unless latest_rs_data.present?
+
+      rs = ReplicaSet.new(
+        namespace: namespace,
+        context: context,
+        definition: latest_rs_data,
+        logger: @logger,
+        parent: "#{@name.capitalize} deployment",
+        deploy_started: @deploy_started
+      )
+      rs.sync(latest_rs_data)
+      rs
     end
   end
 end
--- a/lib/kubernetes-deploy/kubernetes_resource/pod.rb
+++ b/lib/kubernetes-deploy/kubernetes_resource/pod.rb
@@ -53,12 +53,11 @@ module KubernetesDeploy
 
     # Returns a hash in the following format:
     # {
-    #   "pod/web-1/app-container" => "giant blob of logs\nas a single string"
-    #   "pod/web-1/nginx-container" => "another giant blob of logs\nas a single string"
+    #   "app" => ["array of log lines", "received from app container"],
+    #   "nginx" => ["array of log lines", "received from nginx container"]
     # }
     def fetch_logs
       return {} unless exists? && @containers.present?
-
       @containers.each_with_object({}) do |container_name, container_logs|
         cmd = [
           "logs",
@@ -68,7 +67,7 @@ module KubernetesDeploy
         ]
         cmd << "--tail=#{LOG_LINE_COUNT}" unless unmanaged?
         out, _err, _st = kubectl.run(*cmd)
-        container_logs["#{id}/#{container_name}"] = out
+        container_logs[container_name] = out.split("\n")
       end
     end
 
@@ -110,13 +109,11 @@ module KubernetesDeploy
 
       container_logs.each do |container_identifier, logs|
         if logs.blank?
-          @logger.warn("No logs found for #{container_identifier}")
+          @logger.warn("No logs found for container '#{container_identifier}'")
         else
           @logger.blank_line
-          @logger.info("Logs from #{container_identifier}:")
-          logs.split("\n").each do |line|
-            @logger.info("[#{container_identifier}]\t#{line}")
-          end
+          @logger.info("Logs from #{id} container '#{container_identifier}':")
+          logs.each { |line| @logger.info("\t#{line}") }
           @logger.blank_line
         end
       end
--- a/lib/kubernetes-deploy/kubernetes_resource/replica_set.rb
+++ b/lib/kubernetes-deploy/kubernetes_resource/replica_set.rb
@@ -1,0 +1,105 @@
+# frozen_string_literal: true
+module KubernetesDeploy
+  class ReplicaSet < KubernetesResource
+    TIMEOUT = 5.minutes
+
+    def initialize(namespace:, context:, definition:, logger:, parent: nil, deploy_started: nil)
+      @parent = parent
+      @deploy_started = deploy_started
+      @rollout_data = { "replicas" => 0 }
+      @pods = []
+      super(namespace: namespace, context: context, definition: definition, logger: logger)
+    end
+
+    def sync(rs_data = nil)
+      if rs_data.blank?
+        raw_json, _err, st = kubectl.run("get", type, @name, "--output=json")
+        rs_data = JSON.parse(raw_json) if st.success?
+      end
+
+      if rs_data.present?
+        @found = true
+        @rollout_data = { "replicas" => 0 }.merge(rs_data["status"]
+          .slice("replicas", "availableReplicas", "readyReplicas"))
+        @status = @rollout_data.map { |state_replicas, num| "#{num} #{state_replicas.chop.pluralize(num)}" }.join(", ")
+        @pods = find_pods(rs_data)
+      else # reset
+        @found = false
+        @rollout_data = { "replicas" => 0 }
+        @status = nil
+        @pods = []
+      end
+    end
+
+    def deploy_succeeded?
+      @rollout_data["replicas"].to_i == @rollout_data["availableReplicas"].to_i &&
+      @rollout_data["replicas"].to_i == @rollout_data["readyReplicas"].to_i
+    end
+
+    def deploy_failed?
+      @pods.present? && @pods.all?(&:deploy_failed?)
+    end
+
+    def deploy_timed_out?
+      super || @pods.present? && @pods.all?(&:deploy_timed_out?)
+    end
+
+    def exists?
+      @found
+    end
+
+    def desired_replicas
+      @definition["spec"]["replicas"].to_i
+    end
+
+    def fetch_events
+      own_events = super
+      return own_events unless @pods.present?
+      own_events.merge(@pods.first.fetch_events)
+    end
+
+    def fetch_logs
+      container_names.each_with_object({}) do |container_name, container_logs|
+        out, _err, _st = kubectl.run(
+          "logs",
+          id,
+          "--container=#{container_name}",
+          "--since-time=#{@deploy_started.to_datetime.rfc3339}",
+          "--tail=#{LOG_LINE_COUNT}"
+        )
+        container_logs[container_name] = out.split("\n")
+      end
+    end
+
+    private
+
+    def unmanaged?
+      @parent.blank?
+    end
+
+    def container_names
+      @definition["spec"]["template"]["spec"]["containers"].map { |c| c["name"] }
+    end
+
+    def find_pods(rs_data)
+      label_string = rs_data["spec"]["selector"]["matchLabels"].map { |k, v| "#{k}=#{v}" }.join(",")
+      raw_json, _err, st = kubectl.run("get", "pods", "-a", "--output=json", "--selector=#{label_string}")
+      return [] unless st.success?
+
+      all_pods = JSON.parse(raw_json)["items"]
+      all_pods.each_with_object([]) do |pod_data, relevant_pods|
+        next unless pod_data["metadata"]["ownerReferences"].any? { |ref| ref["uid"] == rs_data["metadata"]["uid"] }
+        pod = Pod.new(
+          namespace: namespace,
+          context: context,
+          definition: pod_data,
+          logger: @logger,
+          parent: "#{@name.capitalize} replica set",
+          deploy_started: @deploy_started
+        )
+        pod.sync(pod_data)
+        relevant_pods << pod
+      end
+    end
+  end
+end
--- a/lib/kubernetes-deploy/kubernetes_resource/service.rb
+++ b/lib/kubernetes-deploy/kubernetes_resource/service.rb
@@ -16,7 +16,11 @@ module KubernetesDeploy
     end
 
     def deploy_succeeded?
-      @num_endpoints > 0
+      if exposes_zero_replica_deployment?
+        @num_endpoints == 0
+      else
+        @num_endpoints > 0
+      end
     end
 
     def deploy_failed?
@@ -32,6 +36,24 @@ module KubernetesDeploy
 
     def exists?
       @found
+    end
+
+    private
+
+    def exposes_zero_replica_deployment?
+      related_deployment_replicas && related_deployment_replicas == 0
+    end
+
+    def related_deployment_replicas
+      @related_deployment_replicas ||= begin
+        selector = @definition["spec"]["selector"].map { |k, v| "#{k}=#{v}" }.join(",")
+        raw_json, _err, st = kubectl.run("get", "deployments", "--selector=#{selector}", "--output=json")
+        return unless st.success?
+
+        deployments = JSON.parse(raw_json)["items"]
+        return unless deployments.length == 1
+        deployments.first["spec"]["replicas"].to_i
+      end
     end
   end
 end
--- a/test/fixtures/hello-cloud/bare_replica_set.yml
+++ b/test/fixtures/hello-cloud/bare_replica_set.yml
@@ -1,0 +1,27 @@
+---
+apiVersion: extensions/v1beta1
+kind: ReplicaSet
+metadata:
+  labels:
+    app: hello-cloud
+    name: bare-replica-set
+  name: bare-replica-set
+spec:
+  replicas: 1
+  selector:
+    matchLabels:
+      app: hello-cloud
+      name: bare-replica-set
+  template:
+    metadata:
+      labels:
+        app: hello-cloud
+        name: bare-replica-set
+    spec:
+      containers:
+      - image: nginx:alpine
+        name: app
+        ports:
+        - containerPort: 80
+          name: http
+          protocol: TCP
--- a/test/integration/kubernetes_deploy_test.rb
+++ b/test/integration/kubernetes_deploy_test.rb
@@ -6,6 +6,18 @@ class KubernetesDeployTest < KubernetesDeploy::IntegrationTest
     assert deploy_fixtures("hello-cloud")
     hello_cloud = FixtureSetAssertions::HelloCloud.new(@namespace)
     hello_cloud.assert_all_up
+
+    assert_logs_match_all([
+      "Deploying ConfigMap/hello-cloud-configmap-data (timeout: 30s)",
+      "Hello from Docker!", # unmanaged pod logs
+      "Result: SUCCESS",
+      "Successfully deployed 12 resources"
+    ], in_order: true)
+
+    assert_logs_match_all([
+      %r{ReplicaSet/bare-replica-set\s+1 replica, 1 availableReplica, 1 readyReplica},
+      %r{Deployment/web\s+1 replica, 1 updatedReplica, 1 availableReplica}
+    ])
   end
 
   def test_partial_deploy_followed_by_full_deploy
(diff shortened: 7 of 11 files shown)


lib/kubernetes-deploy/kubernetes_resource/replica_set.rb
```diff
@@ -0,0 +1,116 @@
...
+      @status = @rollout_data.map { |st, num| "#{num} #{st.chop.pluralize(num)}" }.join(", ")
+    end
+
+    def deploy_succeeded?
+      @rollout_data["replicas"].to_i == @rollout_data["availableReplicas"].to_i &&
+      @rollout_data["replicas"].to_i == @rollout_data["readyReplicas"].to_i
+    end
+
+    def deploy_failed?
+      @pods.present? && @pods.all?(&:deploy_failed?)
```
KnVerey: This (and `deploy_timed_out?`) still will basically never fail on account of the pods, since the Pod timeout is higher and the failure condition is super narrow (`@phase == "Failed"`). However, my next task is to make pod failures smarter / more aggressive, and at that point this should make a huge difference in how often our users hit the Deployment timeout.

lib/kubernetes-deploy/kubernetes_resource/deployment.rb
```diff
@@ -4,68 +4,54 @@ class Deployment < KubernetesResource
...
+      @latest_rs.fetch_events
     end
 
-    def fetch_events
-      own_events = super
-      return own_events unless @representative_pod
-      own_events.merge(@representative_pod.fetch_events)
+    def fetch_logs
+      return {} unless @latest_rs.present?
+      @latest_rs.fetch_logs
```
KnVerey: `kubectl get logs deployment/web -c nginx` (or whatever) is a valid command. However, looking at the source, I  think it might favour pods from the old RS under some circumstances, and certainly won't exclude them. https://github.com/kubernetes/kubernetes/blob/dbd1503b6546b0b6048787bad0e63515f3281506/pkg/controller/controller_utils.go#L669-L698

lib/kubernetes-deploy/kubernetes_resource/replica_set.rb
```diff
@@ -0,0 +1,116 @@
...
+      unmanaged? ? @found : true
+    end
+
+    def fetch_events
+      own_events = super
+      return own_events unless @pods.present?
+      own_events.merge(@pods.first.fetch_events)
+    end
+
+    def fetch_logs
```
KnVerey: See [the source](https://github.com/kubernetes/kubernetes/blob/dbd1503b6546b0b6048787bad0e63515f3281506/pkg/controller/controller_utils.go#L669-L698) for how kubectl selects which pod to dump logs from. The gist is that it is trying to select a pod that is more likely to actually have logs.

lib/kubernetes-deploy/kubernetes_resource/deployment.rb
```diff
@@ -4,68 +4,54 @@ class Deployment < KubernetesResource
...
+          .slice("replicas", "updatedReplicas", "availableReplicas", "unavailableReplicas"))
+        @status = @rollout_data.map { |state_replicas, num| "#{num} #{state_replicas.chop.pluralize(num)}" }.join(", ")
       end
     end
 
-    def fetch_logs
-      @representative_pod ? @representative_pod.fetch_logs : {}
+    def fetch_events
+      return {} unless @latest_rs.present?
+      @latest_rs.fetch_events
```
KnVerey: We could include deployment events in here too, but I don't think they'd generally be useful (typically, they're about "Scaled replicaSet foo"), and there can be a ton of them since `fetch_events` doesn't filter by time right now (though I'm considering adding that).

lib/kubernetes-deploy/kubernetes_resource/replica_set.rb
```diff
@@ -0,0 +1,116 @@
...
+      return {} unless container_names.present?
+      container_names.each_with_object({}) do |container_name, container_logs|
+        out, _err, _st = kubectl.run(
+          "logs",
+          id,
+          "--container=#{container_name}",
+          "--since-time=#{@deploy_started.to_datetime.rfc3339}",
+          "--tail=#{LOG_LINE_COUNT}"
+        )
+        container_logs["#{id}/#{container_name}"] = out
```
mkobetic: Given that we're displaying logs for all the containers in the pod, might make sense to display shorter ones first to avoid having a chatty container drown out a less chatty and possibly more relevant one. Visually separating individual container logs would probably be helpful as well.

lib/kubernetes-deploy/kubernetes_resource/deployment.rb
```diff
@@ -4,68 +4,54 @@ class Deployment < KubernetesResource
     TIMEOUT = 5.minutes
 
     def sync
-      json_data, _err, st = kubectl.run("get", type, @name, "--output=json")
+      raw_json, _err, st = kubectl.run("get", type, @name, "--output=json")
       @found = st.success?
-      @rollout_data = {}
+      @rollout_data = { "replicas" => 0 }
```
camilo: Do we expect many calls to `sycn` on the same instance?  or should this be set in the constructor?

lib/kubernetes-deploy/kubernetes_resource/replica_set.rb
```diff
@@ -0,0 +1,116 @@
...
+
+    def initialize(name:, namespace:, context:, file:, parent: nil, logger:)
+      @name = name
+      @namespace = namespace
+      @context = context
+      @file = file
+      @parent = parent
+      @logger = logger
+      @pods = []
+      @rollout_data = { "replicas" => 0 }
```
camilo: Oh interesting, this one does do the intialization here, there must be a reason I'm not seeing.

lib/kubernetes-deploy/kubernetes_resource/replica_set.rb
```diff
@@ -0,0 +1,116 @@
...
+        rs_data = JSON.parse(raw_json)
+        interpret_json_data(rs_data)
+      else # reset
+        @rollout_data = { "replicas" => 0 }
+        @status = nil
+        @pods = []
+      end
+    end
+
+    def interpret_json_data(rs_data)
```
camilo: Does this method need to be public?

lib/kubernetes-deploy/kubernetes_resource/replica_set.rb
```diff
@@ -0,0 +1,116 @@
...
+      pods_data = get_pods(rs_data)
```
camilo: I might be argued out of this one, but same as in the RS class, this should probably just return the objects.

lib/kubernetes-deploy/kubernetes_resource.rb
```diff
@@ -192,5 +193,10 @@ def pretty_status
     def kubectl
       @kubectl ||= Kubectl.new(namespace: @namespace, context: @context, logger: @logger, log_failure_by_default: false)
     end
+
+    def template
```
camilo: Why is this needed, and why is it okay to return `nil` if the file is not present?

lib/kubernetes-deploy/kubernetes_resource/deployment.rb
```diff
@@ -4,68 +4,54 @@ class Deployment < KubernetesResource
...
+        latest_rs_data = get_latest_rs(deployment_data)
+        @latest_rs = ReplicaSet.new(
+          name: latest_rs_data["metadata"]["name"],
+          namespace: namespace,
+          context: context,
+          file: nil,
+          parent: "#{@name.capitalize} deployment",
+          logger: @logger
+        )
+        @latest_rs.interpret_json_data(latest_rs_data)
```
camilo: I even think this should be an optional(?) param to the constructor:

`ReplicaSet.new(blah blah... data: json)`

lib/kubernetes-deploy/kubernetes_resource/replica_set.rb
camilo: I get it now, because the data is set into the objects after intialization, as I mentioned in the other comment, sounds like it could be an optional param to the constructor.

lib/kubernetes-deploy/kubernetes_resource/deployment.rb
```diff
@@ -4,68 +4,54 @@ class Deployment < KubernetesResource
...
+      raw_json, _err, st = kubectl.run("get", type, @name, "--output=json")
       @found = st.success?
-      @rollout_data = {}
+      @rollout_data = { "replicas" => 0 }
       @status = nil
-      @representative_pod = nil
-      @pods = []
+      @latest_rs = nil
 
       if @found
```
camilo: I know is not a change here, but why is there no defined behaviour for if not found?

lib/kubernetes-deploy/kubernetes_resource/deployment.rb
kirs: That ^ or `@latest_rs = ReplicaSet.from_json(...)`

test/fixtures/hello-cloud/bare_replica_set.yml
```diff
@@ -0,0 +1,27 @@
+---
+apiVersion: extensions/v1beta1
+kind: ReplicaSet
```
kirs: Is there a use case when we would have a manifest for ReplicaSet? I thought it's created automatically by Deployment

lib/kubernetes-deploy/kubernetes_resource/deployment.rb
```diff
@@ -4,68 +4,54 @@ class Deployment < KubernetesResource
...
+        @rollout_data.merge!(deployment_data["status"]
+          .slice("replicas", "updatedReplicas", "availableReplicas", "unavailableReplicas"))
+        @status = @rollout_data.map { |state_replicas, num| "#{num} #{state_replicas.chop.pluralize(num)}" }.join(", ")
       end
     end
 
-    def fetch_logs
-      @representative_pod ? @representative_pod.fetch_logs : {}
+    def fetch_events
+      return {} unless @latest_rs.present?
```
kirs: I suppose that the deployment would always have a replica set link. Should we raise if it doesn't (catching the possible inconsistent state)?

lib/kubernetes-deploy/kubernetes_resource/deployment.rb
kirs: AFAIK there's no `kubectl get logs`, you probably meant `kubectl logs`. Though I agree that `kubectl logs deployment/web -c web` can be ambiguous.

How did it work before for a deployment with 100 replicas (== pods)? Would `representative_pod` be one of a hundred?

Do we fetch logs for the developers to tell them what went wrong?

lib/kubernetes-deploy/kubernetes_resource/deployment.rb
```diff
@@ -74,10 +60,18 @@ def exists?
...
-      return false unless @deploy_started
-      # Shitty, brittle workaround to identify a pod from the new ReplicaSet before implementing ReplicaSet awareness
-      Time.parse(pod_json["metadata"]["creationTimestamp"]) >= (@deploy_started - 30.seconds)
+    def get_latest_rs(deployment_data)
+      label_string = deployment_data["spec"]["selector"]["matchLabels"].map { |k, v| "#{k}=#{v}" }.join(",")
+      raw_json, _err, st = kubectl.run("get", "replicasets", "--output=json", "--selector=#{label_string}")
+      return unless st.success?
+
+      all_rs_data = JSON.parse(raw_json)["items"]
+      current_revision = deployment_data["metadata"]["annotations"]["deployment.kubernetes.io/revision"]
```
kirs: It may be handy to use `deployment_data.dig` here (up to you)

lib/kubernetes-deploy/kubernetes_resource/deployment.rb
```diff
@@ -4,68 +4,54 @@ class Deployment < KubernetesResource
...
-      @representative_pod = nil
-      @pods = []
+      @latest_rs = nil
 
       if @found
-        @rollout_data = JSON.parse(json_data)["status"]
-          .slice("updatedReplicas", "replicas", "availableReplicas", "unavailableReplicas")
-        @status = @rollout_data.map { |replica_state, num| "#{num} #{replica_state}" }.join(", ")
+        deployment_data = JSON.parse(raw_json)
+        latest_rs_data = get_latest_rs(deployment_data)
```
kirs: Or `find_latest_rs` since this method actually gets multiple RS and finds the right one.

lib/kubernetes-deploy/kubernetes_resource/pod.rb
```diff
@@ -14,13 +14,13 @@ def initialize(name:, namespace:, context:, file:, parent: nil, logger:)
...
-      out, _err, st = kubectl.run("get", type, @name, "-a", "--output=json")
+      raw_json, _err, st = kubectl.run("get", type, @name, "-a", "--output=json")
       if @found = st.success?
-        pod_data = JSON.parse(out)
+        pod_data = JSON.parse(raw_json)
         interpret_json_data(pod_data)
       else # reset
         @status = @phase = nil
-        @ready = false
+        @ready = @found = false
```
kirs: Why do we need the `@found` flag?

lib/kubernetes-deploy/kubernetes_resource/replica_set.rb
kirs: ❤️ for checking it and for the reference to k8s source that was not easy to find

test/integration/kubernetes_deploy_test.rb
```diff
@@ -381,8 +418,27 @@ def test_output_when_switching_labels_incorrectly
...
-    assert_logs_match("one of your templates is invalid")
-    assert_logs_match(/The Deployment "web" is invalid.*`selector` does not match template `labels`/)
+
+    assert_logs_match_all([
+      "one of your templates is invalid",
+      /The Deployment "web" is invalid.*`selector` does not match template `labels`/
+    ])
+  end
+
+  def test_can_deploy_deployment_with_zero_replicas
```
kirs: I was going to mention that it's extremely important to make sure that it works with multiple long-running revisions / pods that take more time to shutdown / custom terminationGracePeriod, but it's already tested by `test_long_running_deployment` and if the test passes I don't have much to add.

lib/kubernetes-deploy/kubernetes_resource/deployment.rb
KnVerey: Yes, I meant `kubectl logs`. And yes, we're only fetching logs as debug information for these resource types when they fail (unlike unmanaged pods, for which we always dump logs). `representative_pod` was a hack to be able to introduce the debug output without requiring this PR to be ready in addition to the extensive changes that did go in at the time. It attempted to select a pod from the new replicaSet by looking at pod creationTimestamps.

lib/kubernetes-deploy/kubernetes_resource/deployment.rb
KnVerey: My other PR adds filtering by time, so we can actually keep these events after all.

lib/kubernetes-deploy/kubernetes_resource/deployment.rb
KnVerey: Yes, `sync` will be called many, many times on the same instance--about once every three seconds from the beginning of the deploy until the time the resource succeeds/fails. The exception to this is resources that were created to model subresources of the thing that was actually in the template set, e.g. managed replica sets or pods. We'll get new instances of those on every sync.

lib/kubernetes-deploy/kubernetes_resource.rb
KnVerey: I put it here because multiple subclasses in this PR use it:
- `ReplicaSet` uses it to grab the container names in the case where it was not derived from a deployment (managed ReplicaSets will not have a template file)
- `Service` uses it to find the related deployment so it can be smarter about how many endpoints to expect (notably so that it can expect zero when appropriate)

Your comment make me think that perhaps I should move the fallback right into this method: if we have a file, load it from file; if we don't, fetch it from the API.

lib/kubernetes-deploy/kubernetes_resource/deployment.rb
KnVerey: The behaviour in that case is a reset, which is accomplished by the code above. Basically, we always reset, and then if we find new data, we repopulate using it.

lib/kubernetes-deploy/kubernetes_resource/deployment.rb
KnVerey: 👍

lib/kubernetes-deploy/kubernetes_resource/deployment.rb
KnVerey: I remember trying the `ReplicaSet.from_json` model at one point in this work and running into difficulties... but I don't remember what they were, so I'll try again. :P

lib/kubernetes-deploy/kubernetes_resource/deployment.rb
KnVerey: No, I don't think we should raise. For one, the purpose of `fetch_events` is collection of debug information, so raising during that would throw a wrench in our attempt to be helpful. For another, since the RS creation happens asynchronously, it is theoretically possible for the RS not to exist (yet) at a given point.

lib/kubernetes-deploy/kubernetes_resource/deployment.rb
KnVerey: That's a Ruby 2.3 feature, so I can't (unless we decide to break with the version supported by shipit-engine, or that gem bumped its requirement and I didn't notice)

lib/kubernetes-deploy/kubernetes_resource/pod.rb
KnVerey: It's used by `exists?`... but I'm not sure why I moved it here if that's what you mean. It should already be set, so I'll change it back.

lib/kubernetes-deploy/kubernetes_resource/replica_set.rb
KnVerey: The reason has to do with the possible origins of these resource types:
- Deployments _always_ come from a file in your deploy directory and get their observed state via `sync`.
- ReplicaSets _usually_ show up via deployments, but it is possible to deploy them separately, so I've supported that here (and added a test).
- Pods are very commonly created both ways: unmanaged from a file in your deploy directory, or via a deployment/replicaSet.

For deployments, we have it easy: we see the file, create the instance, and then call sync on it repeatedly to reset and update the observed state. For the other two, there is the additional possibility that they are being created when their parent syncs, and that the parent already has the data on their observed state. In that case, we create the instance and then call `interpret_json_data` with that data; `sync` is never called. Effectively, `interpret_json_data` is the common core between the two flows. I'll investigate making it private / clarifying these two flows.

test/fixtures/hello-cloud/bare_replica_set.yml
KnVerey: I haven't encountered a use case, but it is definitely possible to create them separately.

lib/kubernetes-deploy/kubernetes_resource/replica_set.rb
KnVerey: How about this?

image

- each gets its own section, with a heading instead of tags on every line
- sorted by line count
- identifier format `container 'foo'` instead of the full resource path (`Deployment/web/foo`)

lib/kubernetes-deploy/kubernetes_resource.rb
```diff
@@ -139,10 +140,11 @@ def debug_message
...
         else
-          helpful_info << "  - Logs (last #{LOG_LINE_COUNT} lines shown):"
-          container_logs.each do |identifier, logs|
-            logs.split("\n").each do |line|
-              helpful_info << "      [#{identifier}]\t#{line}"
+          sorted_logs = container_logs.sort_by { |_, log_lines| log_lines.length }
+          sorted_logs.each do |identifier, log_lines|
+            helpful_info << "  - Logs from container '#{identifier}' (last #{LOG_LINE_COUNT} lines shown):"
+            log_lines.each do |line|
+              helpful_info << "      #{line}"
```
KnVerey: Now looks like this:
image
[...event noise addressed by other PR...]
image

lib/kubernetes-deploy/kubernetes_resource/pod.rb
```diff
@@ -4,50 +4,38 @@ class Pod < KubernetesResource
...
+        if @deploy_started
+          log_suspicious_states(pod_data["status"].fetch("containerStatuses", []))
+        end
+      else # reset
+        @found = false
+        @containers = []
+        @phase = @status = nil
+        @ready = false
       end
+      display_logs if unmanaged? && deploy_succeeded?
```
KnVerey: Example of this output:

image

KnVerey:
I made some architectural changes to this PR based on your questions and feedback. Now, whether the resource was created from a deploy template or during a parent resource's `sync`, its observed state is recorded through its own `sync` method.`interpret_json_data` is gone. This way, we have a more unified code path for the resources that can be created both ways (i.e. Pods, ReplicaSets). I'm hoping this will be a lot clearer--lmk what you think!

lib/kubernetes-deploy/kubernetes_resource/deployment.rb
```diff
@@ -74,10 +51,27 @@ def exists?
...
+      end
+
+      ReplicaSet.new(
+        name: latest_rs_data["metadata"]["name"],
+        namespace: namespace,
+        context: context,
+        parent: "#{@name.capitalize} deployment",
+        logger: @logger,
+        deploy_started: @deploy_started
+      ).tap { |rs| rs.sync(latest_rs_data) }
```
KnVerey: I can also make a method like this if you still think it is better:
```ruby
    def self.from_json(**args)
      json_data = args.delete(:json_data)
      inst = new(**args)
      inst.sync(json_data)
      inst
    end
```
I'm not sure it adds much value, since most of the initializer params aren't from the json, but I'm not against it.

lib/kubernetes-deploy/kubernetes_resource/replica_set.rb
mkobetic: ❤️ 👍

lib/kubernetes-deploy/kubernetes_resource/service.rb
```diff
@@ -16,7 +16,11 @@ def sync
     end
 
     def deploy_succeeded?
-      @num_endpoints > 0
+      if related_deployment_replicas
```
KnVerey: The purpose of this change is to be able to handle zero-replica deployments (that have been temporarily scaled down for some reason). I'm second-guessing the what I've done here though--I think it might be too aggressive. If the deployment has rolled out and the service has _some_ endpoints, it's not really valuable to hold up the deploy, is it?

WDTY think about this logic instead:

```ruby
if exposes_zero_replica_deployment?
 @num_endpoints == 0
else
  @num_endpoints > 0
end
```

lib/kubernetes-deploy/kubernetes_resource/deployment.rb
camilo: I think is fine as is, what I don't like is `tap` I find that method to be kinda not very useful sintactic sugar; WDYJ:

```
rs = RS.new(..)
rs.sync
```

lib/kubernetes-deploy/kubernetes_resource/pod.rb
```diff
@@ -4,50 +4,38 @@ class Pod < KubernetesResource
...
     SUSPICIOUS_CONTAINER_STATES = %w(ImagePullBackOff RunContainerError ErrImagePull).freeze
 
-    def initialize(name:, namespace:, context:, file:, parent: nil, logger:)
+    def initialize(name:, namespace:, context:, file: nil, parent: nil, logger:, deploy_started: nil)
       @name = name
       @namespace = namespace
       @context = context
       @file = file
       @parent = parent
+      @deploy_started = deploy_started
```
camilo: This is an interesting semantic change here; why would a `KubernetesResource` be aware of when a deploy started?

lib/kubernetes-deploy/kubernetes_resource/service.rb
camilo: I like the explicitness of `exposes_zero_replica_deployment?` over the current more implicit implementation

lib/kubernetes-deploy/kubernetes_resource.rb
KnVerey: Memoizing _should_ be safe... From file, the result obviously won't change. Although resources do change cluster-side, the ones that don't have a file are child resources for which that shouldn't be true. For example, if the initial status check (before the deploy makes any changes) on a deployment resource memoized a template from the cluster, that'd be bad, since that template would change when the deploy proceeds. But the child resource objects are replaced by their parents every time the parent syncs.

lib/kubernetes-deploy/kubernetes_resource/deployment.rb
```diff
@@ -4,68 +4,45 @@ class Deployment < KubernetesResource
...
-
-            if !@representative_pod && pod_probably_new?(pod_json)
-              @representative_pod = pod
-            end
-            @pods << pod
-          end
-        end
+        deployment_data = JSON.parse(raw_json)
+        @latest_rs = find_latest_rs(deployment_data)
+        @rollout_data = { "replicas" => 0 }.merge!(deployment_data["status"]
```
KnVerey: no good reason. I refactored this section a bunch of times, so maybe there was some intermediate version that didn't use the return value. I'll fix it.

lib/kubernetes-deploy/kubernetes_resource/deployment.rb
KnVerey: Because we need to return the rs, so it'd be _three_ lines instead of one 😉 
Though our style guide doesn't mention it, I've heard other Shopifolk hating on `tap` in the past. I can change it.

lib/kubernetes-deploy/kubernetes_resource/pod.rb
KnVerey: It's not actually new. I added it as a parameter to save the parent resource from having to call the setter immediately in these cases. I'll change it back if you think this introduces confusion. See [here in the diff](https://github.com/Shopify/kubernetes-deploy/pull/107/files#diff-4709b5257970978173e422904663525fL32) for the setter being used instead.

As for why it is needed, it is primarily so that they know when to timeout, although we also use it to filter logs and events. It's not the global deploy start time; it is the time that specific resource was `apply`'d. For resources in your yamls, the instance is created when we first read in the templates, and this attribute is it is set later,  [here](https://github.com/Shopify/kubernetes-deploy/blob/b61fe310824a17ce82f57964bc44d641ffc6c169/lib/kubernetes-deploy/runner.rb#L376).

KnVerey:
@camilo @kirs This has been rebased on the refactor I extracted out of it based on your feedback. What's left should look pretty familiar, but could you please take a final look? 😄

KnVerey:
@ibawt can you please review this one?

lib/kubernetes-deploy/kubernetes_resource/replica_set.rb
```diff
@@ -0,0 +1,105 @@
...
+    end
+
+    def desired_replicas
+      @definition["spec"]["replicas"].to_i
+    end
+
+    def fetch_events
+      own_events = super
+      return own_events unless @pods.present?
+      own_events.merge(@pods.first.fetch_events)
```
ibawt: pods.first: does that mean we'll lose some data or is it the same for the entire thing?

KnVerey: My thinking was that the events will be very similar across the entire set of pods. It is possible some pods will be more interesting than others (e.g. the case you saw where a pod landed on a bad node), but at this stage I don't think we have a nice way of figuring out which one. After my next PR, we could deliberately try to find a failed pod (if any) and use its events... 🤔  Did you have any ideas around this?

test/integration/kubernetes_deploy_test.rb
```diff
@@ -384,8 +422,27 @@ def test_output_when_switching_labels_incorrectly
...
+    success = deploy_fixtures("hello-cloud", subset: ["configmap-data.yml", "web.yml.erb"]) do |fixtures|
+      web = fixtures["web.yml.erb"]["Deployment"].first
+      web["spec"]["replicas"] = 0
```
kirs: A use case of zero replicas that I remember is:

1. Create a deployment with unspecified `replicas` field
2. The replicas number is scaled dynamically (can sometimes be zero)
3. At the moment of the deploy it's zero

I see the test works with YAML that explicitly sets `replicas` to zero, but would it also support the case of dynamic `replicas` value set to zero?

lib/kubernetes-deploy/kubernetes_resource/deployment.rb
```diff
@@ -4,78 +4,84 @@ class Deployment < KubernetesResource
...
+      label_string = deployment_data["spec"]["selector"]["matchLabels"].map { |k, v| "#{k}=#{v}" }.join(",")
+      raw_json, _err, st = kubectl.run("get", "replicasets", "--output=json", "--selector=#{label_string}")
+      return unless st.success?
+
+      all_rs_data = JSON.parse(raw_json)["items"]
+      current_revision = deployment_data["metadata"]["annotations"]["deployment.kubernetes.io/revision"]
+
+      latest_rs_data = all_rs_data.find do |rs|
+        rs["metadata"]["ownerReferences"].any? { |ref| ref["uid"] == deployment_data["metadata"]["uid"] } &&
+        rs["metadata"]["annotations"]["deployment.kubernetes.io/revision"] == current_revision
```
KnVerey: I'm using `ownerReferences` to be completely certain that the RS belongs to the deployment, and `revision` to select the new one, as opposed to the `n` other RS that could exists for this deployment. This seemed like the best available option based one the k8s source and issues I read, but I could be missing something. Are you aware of a better way?

test/integration/kubernetes_deploy_test.rb
KnVerey: Yes, that should work fine. [The implementation](https://github.com/Shopify/kubernetes-deploy/pull/107/files#diff-bfab9900d04f0a61190aa04f896aed1fR50) uses selectors to grab a live copy of the related deployment's template, so it should work even if the deployment wasn't managed by `kubernetes-deploy`.

kirs: :+1: great work!

lib/kubernetes-deploy/kubernetes_resource/deployment.rb
kirs: Ah, I somehow thought it's the k8s revision

lib/kubernetes-deploy/kubernetes_resource/replica_set.rb
ibawt: no ideas just wanted to make sure I knew the reason.

Issue #108, opened 2017-06-06
pod.rb:82:in `block in fetch_logs': undefined method `to_datetime' for nil:NilClass


airhorns:
Got this exception while trying to deploy `wedge-viewer` for the first time, looks like the `@deploy_started` variable hasn't been set by the time we try to use it there :(

Full trace:
```

[INFO][2017-06-06 23:07:21 +0000]	----------------------------Phase 2: Checking initial resource statuses-----------------------------
[INFO][2017-06-06 23:07:23 +0000]	
[INFO][2017-06-06 23:07:23 +0000]	------------------------------------------Result: FAILURE-------------------------------------------
[FATAL][2017-06-06 23:07:23 +0000]	No actions taken
/usr/lib/ruby-shopify/ruby-shopify-2.3.3/lib/ruby/gems/2.3.0/gems/kubernetes-deploy-0.7.4/lib/kubernetes-deploy/kubernetes_resource/pod.rb:82:in `block in fetch_logs': undefined method `to_datetime' for nil:NilClass (NoMethodError)
	from /usr/lib/ruby-shopify/ruby-shopify-2.3.3/lib/ruby/gems/2.3.0/gems/kubernetes-deploy-0.7.4/lib/kubernetes-deploy/kubernetes_resource/pod.rb:77:in `each'
	from /usr/lib/ruby-shopify/ruby-shopify-2.3.3/lib/ruby/gems/2.3.0/gems/kubernetes-deploy-0.7.4/lib/kubernetes-deploy/kubernetes_resource/pod.rb:77:in `each_with_object'
	from /usr/lib/ruby-shopify/ruby-shopify-2.3.3/lib/ruby/gems/2.3.0/gems/kubernetes-deploy-0.7.4/lib/kubernetes-deploy/kubernetes_resource/pod.rb:77:in `fetch_logs'
	from /usr/lib/ruby-shopify/ruby-shopify-2.3.3/lib/ruby/gems/2.3.0/gems/kubernetes-deploy-0.7.4/lib/kubernetes-deploy/kubernetes_resource/pod.rb:98:in `display_logs'
	from /usr/lib/ruby-shopify/ruby-shopify-2.3.3/lib/ruby/gems/2.3.0/gems/kubernetes-deploy-0.7.4/lib/kubernetes-deploy/kubernetes_resource/pod.rb:26:in `sync'
[...]

PR #109, opened 2017-06-07
Fail the deploy when unmanaged pod predates it

KnVerey:
Fixes https://github.com/Shopify/kubernetes-deploy/issues/108

The problem was that the unmanaged pod name was hardcoded on that repo. The first deploy (the one where it spent a while trying to pull the container, then ultimately failed because of a missing rake task) successfully ran that pod. Subsequent deploys then tried to deploy the same pod, which therefore looked like it succeeded before the deploy even started. 

If it weren't for the `nil` bug, it would have dumped its logs during `Checking initial resource statuses`, which also would have been bad. IMO the deploy should fail immediately in that case, because if you `apply` a pod that has already run, it won't do anything, and these pods are expected to run during each deploy.

Here's what that deploy would have shown if this patch were in:

image

### DIFF
--- a/lib/kubernetes-deploy/kubernetes_resource/pod.rb
+++ b/lib/kubernetes-deploy/kubernetes_resource/pod.rb
@@ -15,7 +15,19 @@ module KubernetesDeploy
 
     def sync
       out, _err, st = kubectl.run("get", type, @name, "-a", "--output=json")
-      if @found = st.success?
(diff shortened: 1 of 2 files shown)


KnVerey:
Out of date. Replaced by https://github.com/Shopify/kubernetes-deploy/pull/125.

PR #110, opened 2017-06-07
Improve event fetching

KnVerey:
@Shopify/cloudplatform @kirs 

Makes a few improvements to the event fetching code:
* Fixes event parsing when the event message includes a newline (note: in a rather large sample of events, I didn't see newlines in any other relevant fields, and I didn't see any tabs)
* Excludes events that occurred before the deploy started (relevant for controllers, such as Deployment, that span multiple deploys)
* Cleans up the code around event fetching and adds some unit test cases

### DIFF
--- a/lib/kubernetes-deploy/kubernetes_resource.rb
+++ b/lib/kubernetes-deploy/kubernetes_resource.rb
@@ -154,25 +154,18 @@ module KubernetesDeploy
     # Returns a hash in the following format:
     # {
     #   "pod/web-1" => [
-    #     {"subject_kind" => "Pod", "some" => "stuff"}, # event 1
-    #     {"subject_kind" => "Pod", "other" => "stuff"}, # event 2
+    #     "Pulling: pulling image "hello-world:latest" (1 events)",
+    #     "Pulled: Successfully pulled image "hello-world:latest" (1 events)"
     #   ]
     # }
     def fetch_events
       return {} unless exists?
-      fields = "{.involvedObject.kind}\t{.count}\t{.message}\t{.reason}"
-      out, _err, st = kubectl.run("get", "events",
-        %(--output=jsonpath={range .items[?(@.involvedObject.name=="#{name}")]}#{fields}\n{end}))
+      out, _err, st = kubectl.run("get", "events", "--output=jsonpath=#{Event.jsonpath_for(name)}")
       return {} unless st.success?
 
       event_collector = Hash.new { |hash, key| hash[key] = [] }
-      out.split("\n").each_with_object(event_collector) do |event_blob, events|
-        pieces = event_blob.split("\t")
-        subject_kind = pieces[0]
-        # jsonpath can only filter by one thing at a time, and we chose involvedObject.name
-        # This means we still need to filter by involvedObject.kind here to make sure we only retain relevant events
-        next unless subject_kind.downcase == type.downcase
-        events[id] << "#{pieces[3] || 'Unknown'}: #{pieces[2]} (#{pieces[1]} events)" # Reason: Message (X events)
+      Event.from_jsonpath_blob(out).each_with_object(event_collector) do |candidate, events|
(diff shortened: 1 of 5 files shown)


lib/kubernetes-deploy/kubernetes_resource.rb
```diff
@@ -192,5 +185,51 @@ def pretty_status
...
       @kubectl ||= Kubectl.new(namespace: @namespace, context: @context, logger: @logger, log_failure_by_default: false)
     end
+
+    class Event
+      JSONPATH_FIELDS =
+        "{.involvedObject.kind}\t{.involvedObject.name}\t{.count}\t{.lastTimestamp}\t{.reason}\t{.message}"
+      JSONPATH_SEPARATOR = "ENDEVENT\nBEGINEVENT"
+
+      def self.jsonpath_for(name)
+        %({range .items[?(@.involvedObject.name=="#{name}")]}#{JSONPATH_FIELDS}#{JSONPATH_SEPARATOR}{end})
```
KnVerey: Anyone with jsonpath experience know a way to compose this less sketchily? (or any other reasonably efficient way to get relevant events)

Relevant:
* https://kubernetes.io/docs/user-guide/jsonpath/
* [issue: jsonpath template output should be json](https://github.com/kubernetes/kubernetes/issues/16707)
* [issue: kubectl: no way to list events about a specific object](https://github.com/kubernetes/kubernetes/issues/11994)
* [issue: Unable to use kubectl get with a fieldSelector](https://github.com/kubernetes/kubernetes/issues/14129)

mkobetic: Looks like you can emit \n with jsonpath if you make it part of jsonpath string literal, i.e. instead of just `\n` make it `{"\n"}`

mkobetic: e.g. this is almost legit json array output
```tier3 get --all-namespaces pods -ojsonpath='{"["}{range .items[?(@.status.phase!="Running")]}{"{"}node:"{.spec.nodeName}",namespace:"{.metadata.namespace}",name:"{.metadata.name}",reason:"{.status.reason}"{"},\n"}{end}{"]"}'```

**Comment by KnVerey** [comment]
I just pushed a new commit that switches this to use go templating as discussed with @mkobetic yesterday. I think it is much better! It also now screens out "Created" and "Started" events, as I realized they can be very, very noisy if your container is in a crash backoff loop (the integration test output showed this).

File: lib/kubernetes-deploy/kubernetes_resource.rb [hunk 2]
```diff
@@ -192,5 +185,61 @@ def pretty_status
...
+          %[(eq .involvedObject.name "#{name}")],
+          '(ne .reason "Started")',
+          '(ne .reason "Created")'
+        ]
+        condition_start = "{{if and #{and_conditions.join(' ')}}}"
+        field_part = FIELDS.map { |f| "{{#{f}}}" }.join(%({{print "#{FIELD_SEPARATOR}"}}))
+        %({{range .items}}#{condition_start}#{field_part}{{print "#{EVENT_SEPARATOR}"}}{{end}}{{end}})
+      end
+
+      def self.from_go_template_blob(blob)
```
camilo: naming nitpick: this name does not reflect the fact that you'll get back a collection of events not just one

lib/kubernetes-deploy/kubernetes_resource.rb
```diff
@@ -192,5 +185,61 @@ def pretty_status
     def kubectl
       @kubectl ||= Kubectl.new(namespace: @namespace, context: @context, logger: @logger, log_failure_by_default: false)
     end
+
+    class Event
```
camilo: good use of an intenal class 👍

lib/kubernetes-deploy/kubernetes_resource.rb
KnVerey: True. Something like `extract_all_from_go_template_blob`?

PR #111, opened 2017-06-14
Log reminders every 30s if nothing new happens

KnVerey:
@Shopify/cloudplatform 

This fixes a problem some people have run into (including because of Cloudsql taking forever to start the first time) where Shipit aborts the deploy because nothing has been logged in the past 300 seconds. We could instead change that timeout on the Shipit side, but I think logging something every now and then to show we are not stuck would be generally better UX.

Example from the test output:

image

Note: this also pulls over some test helpers from https://github.com/Shopify/kubernetes-deploy/pull/107 so that I can use them now.

### DIFF
--- a/lib/kubernetes-deploy/resource_watcher.rb
+++ b/lib/kubernetes-deploy/resource_watcher.rb
@@ -12,38 +12,57 @@ module KubernetesDeploy
       @deploy_started_at = deploy_started_at
     end
 
-    def run(delay_sync: 3.seconds)
-      delay_sync_until = Time.now.utc
+    def run(delay_sync: 3.seconds, reminder_interval: 30.seconds)
+      delay_sync_until = last_message_logged_at = Time.now.utc
 
       while @resources.present?
         if Time.now.utc < delay_sync_until
           sleep(delay_sync_until - Time.now.utc)
         end
-        watch_time = (Time.now.utc - @deploy_started_at).round(1)
         delay_sync_until = Time.now.utc + delay_sync # don't pummel the API if the sync is fast
+
         @resources.each(&:sync)
         newly_finished_resources, @resources = @resources.partition(&:deploy_finished?)
 
-        new_success_list = []
-        newly_finished_resources.each do |resource|
-          if resource.deploy_failed?
-            @logger.error("#{resource.id} failed to deploy after #{watch_time}s")
-          elsif resource.deploy_timed_out?
-            @logger.error("#{resource.id} deployment timed out")
-          else
-            new_success_list << resource.id
-          end
-        end
-
-        unless new_success_list.empty?
-          success_string = ColorizedString.new("Successfully deployed in #{watch_time}s:").green
-          @logger.info("#{success_string} #{new_success_list.join(', ')}")
-        end
-
-        if newly_finished_resources.present? && @resources.present? # something happened this cycle, more to go
-          @logger.info("Continuing to wait for: #{@resources.map(&:id).join(', ')}")
+        if newly_finished_resources.present?
+          watch_time = (Time.now.utc - @deploy_started_at).round(1)
+          report_what_just_happened(newly_finished_resources, watch_time)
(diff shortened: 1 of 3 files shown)


lib/kubernetes-deploy/resource_watcher.rb
```diff
@@ -12,38 +12,57 @@ def initialize(resources, logger:, deploy_started_at: Time.now.utc)
...
+
+    def report_what_is_left(reminder:)
+      return unless @resources.present?
+      resource_list = @resources.map(&:id).join(', ')
+      msg = reminder ? "Still waiting for: #{resource_list}" : "Continuing to wait for: #{resource_list}"
+      @logger.info(msg)
+    end
+
+    def due_for_reminder?(last_message_logged_at, reminder_interval)
+      (last_message_logged_at.to_f + reminder_interval.to_f) <= Time.now.utc.to_f
```
ibawt: amusingly with floating point is unlikely to ever be equal :)

test/test_helper.rb
```diff
@@ -48,31 +48,37 @@ def teardown
...
-        assertion = "\033[0;35mrefute_logs_match(#{regexp.inspect})\033[0;33m"
-        $stderr.puts("\033[0;33mWARNING: Skipping #{assertion} while logs are redirected to stderr\033[0m")
-        return
+    def assert_logs_match_all(entry_list, in_order: false)
+      logging_assertion do |logs|
+        scanner = StringScanner.new(logs)
+        entry_list.each do |entry|
+          regex = entry.is_a?(Regexp) ? entry : Regexp.new(Regexp.escape(entry))
+          if in_order
+            failure_msg = "'#{entry}' not found in the expected sequence in the following logs:\n#{logs}"
```
ibawt: does the scanner keeps it's place or something?

ibawt: I enjoy this solution as well if that's the case 👍

lib/kubernetes-deploy/resource_watcher.rb
```diff
@@ -12,38 +12,57 @@ def initialize(resources, logger:, deploy_started_at: Time.now.utc)
...
-          elsif resource.deploy_timed_out?
-            @logger.error("#{resource.id} deployment timed out")
-          else
-            new_success_list << resource.id
-          end
+        if newly_finished_resources.present?
+          watch_time = (Time.now.utc - @deploy_started_at).round(1)
+          report_what_just_happened(newly_finished_resources, watch_time)
+          report_what_is_left(reminder: false)
+          last_message_logged_at = Time.now.utc
```
camilo: I like this bit of code a lot more, much better level of abstraction.

lib/kubernetes-deploy/resource_watcher.rb
```diff
@@ -12,38 +12,57 @@ def initialize(resources, logger:, deploy_started_at: Time.now.utc)
...
+        elsif due_for_reminder?(last_message_logged_at, reminder_interval)
+          report_what_is_left(reminder: true)
+          last_message_logged_at = Time.now.utc
```
camilo: you do this in both branches of the if, why not just do it once after?

KnVerey: It is possible to for both things to be false (i.e. nothing happened this cycle, and we weren't due for a reminder), so that'd have to be wrapped in `if newly_finished_resources.present? || due_for_reminder?(...)`

test/test_helper.rb
KnVerey: Indeed!

> Scans the string until the pattern is matched. Returns the substring up to and including the end of the match, advancing the scan pointer to that location. If there is no match, nil is returned.

image

lib/kubernetes-deploy/resource_watcher.rb
camilo: right see, nm I'm dumb

PR #112, opened 2017-06-16
Fix passing env vars into the task runneR

xldenis:
@jordanwheeler noticed that the code to pass ENV vars into the task runner was broken. This fixes it and adds a test case.

### DIFF
--- a/test/integration/runner_task_test.rb
+++ b/test/integration/runner_task_test.rb
@@ -65,6 +65,23 @@ class RunnerTaskTest < KubernetesDeploy::IntegrationTest
     end
   end
 
+  def test_run_with_env_vars
+    deploy_fixtures("hello-cloud", subset: ["template-runner.yml"])
+
+    task_runner = build_task_runner
+    assert task_runner.run(
+      task_template: 'hello-cloud-template-runner',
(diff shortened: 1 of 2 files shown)


jordanwheeler:
thanks 👍

test/integration/runner_task_test.rb
```diff
@@ -65,6 +65,23 @@ def test_run_bang_with_template_missing_raises_exception
...
+      task_template: 'hello-cloud-template-runner',
+      entrypoint: nil,
+      args: %w(rake some_task),
+      env_vars: ['ENV=VAR1']
+    )
+
+    pods = kubeclient.get_pods(namespace: @namespace)
+    assert_equal 1, pods.length, "Expected 1 pod to exist, found #{pods.length}"
+    assert_equal %w(rake some_task), pods.first.spec.containers.first.args
+    assert_equal 'ENV', pods.first.spec.containers.env.first.name
```
KnVerey: Let's add an assertion on the value as well.

xldenis:
I've fixed the test and run BK.. I'm going to merge this unless there are any objections.

KnVerey:
:shipit:

PR #113, opened 2017-06-21
Improve handling of KubernetesResource tempfiles/templates

KnVerey:
This is a refactor split out of https://github.com/Shopify/kubernetes-deploy/pull/107 to make it more reviewable. It includes the change to simplify `Pod`'s public API (aka get rid of `Pod#interpret_json_data`) discussed on that PR, plus the bigger change described below.

### Before

Two possible paths:
- If we start with a file in the deploy directory, we split it into one tempfile per resource in `runner.rb`, then initialize the instance with `file: tempfile`. If we later need the parsed template (which we do as of the linked PR), we need to either re-parse the tempfile or fetch it from the API.
- If we start with a child resource we discovered during a sync (e.g. a pod that is a child of a deployment), we initialize the instance with `file: nil`. To make the template available later (which again is necessary as of the linked PR), we need to add another parameter to the initializer to pass in the template data (which we already got from the API when we discovered the resource).

### After

One path: Always pass a parsed template into the initializer. It turns out we already _always_ have this when we create the instance. In the child resource case, this is obvious; we got it from the API. As it turns out, in the deploy directory case, we are generating and discarding it in `Runner#split_templates`. The trick is to use that copy to initialize the instance, and have the instance itself lazily create the tempfile if we need it.

### DIFF
--- a/lib/kubernetes-deploy/kubernetes_resource.rb
+++ b/lib/kubernetes-deploy/kubernetes_resource.rb
@@ -22,14 +22,15 @@ module KubernetesDeploy
       If you have reason to believe it will succeed, retry the deploy to continue to monitor the rollout.
       MSG
 
-    def self.build(namespace:, context:, template:, logger:)
-      opts = { namespace: namespace, context: context, template: template, logger: logger }
-      if KubernetesDeploy.const_defined?(template["kind"])
-        klass = KubernetesDeploy.const_get(template["kind"])
+    def self.build(namespace:, context:, definition:, logger:)
+      opts = { namespace: namespace, context: context, definition: definition, logger: logger }
+      if KubernetesDeploy.const_defined?(definition["kind"])
+        klass = KubernetesDeploy.const_get(definition["kind"])
         klass.new(**opts)
       else
         inst = new(**opts)
-        inst.tap { |r| r.type = template["kind"] }
+        inst.type = definition["kind"]
+        inst
       end
     end
 
@@ -41,26 +42,25 @@ module KubernetesDeploy
       self.class.timeout
     end
 
-    def initialize(namespace:, context:, template:, logger:)
+    def initialize(namespace:, context:, definition:, logger:)
       # subclasses must also set these if they define their own initializer
-      @name = template.fetch("metadata", {})["name"]
+      unless @name = definition.fetch("metadata", {})["name"]
+        logger.summary.add_paragraph("Rendered template content:\n#{definition.to_yaml}")
+        raise FatalDeploymentError, "Template is missing required field metadata.name"
+      end
+
       @namespace = namespace
       @context = context
       @logger = logger
-      @template = template
+      @definition = definition
     end
 
     def id
       "#{type}/#{name}"
     end
 
-    def file
-      @file ||= begin
-        f = Tempfile.new(["#{type}-#{name}", ".yml"])
-        f.write(YAML.dump(@template))
-        f.close
-        f
-      end
+    def file_path
+      file.path
     end
 
     def sync
@@ -236,6 +236,20 @@ module KubernetesDeploy
       def to_s
         "#{@reason}: #{@message} (#{@count} events)"
       end
+    end
+
+    private
+
+    def file
+      @file ||= create_definition_tempfile
+    end
+
+    def create_definition_tempfile
+      file = Tempfile.new(["#{type}-#{name}", ".yml"])
+      file.write(YAML.dump(@definition))
+      file
+    ensure
+      file.close if file
     end
   end
 end
--- a/lib/kubernetes-deploy/kubernetes_resource.rb
+++ b/lib/kubernetes-deploy/kubernetes_resource.rb
@@ -44,7 +44,8 @@ module KubernetesDeploy
 
     def initialize(namespace:, context:, definition:, logger:)
       # subclasses must also set these if they define their own initializer
-      unless @name = definition.fetch("metadata", {})["name"]
+      @name = definition.fetch("metadata", {})["name"]
+      unless @name.present?
         logger.summary.add_paragraph("Rendered template content:\n#{definition.to_yaml}")
         raise FatalDeploymentError, "Template is missing required field metadata.name"
       end
--- a/lib/kubernetes-deploy/kubernetes_resource.rb
+++ b/lib/kubernetes-deploy/kubernetes_resource.rb
@@ -22,27 +22,14 @@ module KubernetesDeploy
       If you have reason to believe it will succeed, retry the deploy to continue to monitor the rollout.
       MSG
 
-    def self.for_type(type:, name:, namespace:, context:, file:, logger:)
-      subclass = case type
-      when 'cloudsql' then Cloudsql
-      when 'configmap' then ConfigMap
-      when 'deployment' then Deployment
-      when 'pod' then Pod
-      when 'redis' then Redis
-      when 'bugsnag' then Bugsnag
-      when 'ingress' then Ingress
-      when 'persistentvolumeclaim' then PersistentVolumeClaim
-      when 'service' then Service
-      when 'podtemplate' then PodTemplate
-      when 'poddisruptionbudget' then PodDisruptionBudget
-      end
-
-      opts = { name: name, namespace: namespace, context: context, file: file, logger: logger }
-      if subclass
-        subclass.new(**opts)
+    def self.build(namespace:, context:, template:, logger:)
+      opts = { namespace: namespace, context: context, template: template, logger: logger }
+      if KubernetesDeploy.const_defined?(template["kind"])
+        klass = KubernetesDeploy.const_get(template["kind"])
+        klass.new(**opts)
       else
         inst = new(**opts)
-        inst.tap { |r| r.type = type }
+        inst.tap { |r| r.type = template["kind"] }
       end
     end
 
@@ -54,17 +41,26 @@ module KubernetesDeploy
       self.class.timeout
     end
 
-    def initialize(name:, namespace:, context:, file:, logger:)
+    def initialize(namespace:, context:, template:, logger:)
       # subclasses must also set these if they define their own initializer
-      @name = name
+      @name = template.fetch("metadata", {})["name"]
       @namespace = namespace
       @context = context
-      @file = file
       @logger = logger
+      @template = template
     end
 
     def id
       "#{type}/#{name}"
+    end
+
+    def file
+      @file ||= begin
+        f = Tempfile.new(["#{type}-#{name}", ".yml"])
+        f.write(YAML.dump(@template))
+        f.close
+        f
+      end
     end
 
     def sync
--- a/lib/kubernetes-deploy/kubernetes_resource/deployment.rb
+++ b/lib/kubernetes-deploy/kubernetes_resource/deployment.rb
@@ -23,7 +23,7 @@ module KubernetesDeploy
             pod = Pod.new(
               namespace: namespace,
               context: context,
-              template: pod_json,
+              definition: pod_json,
               logger: @logger,
               parent: "#{@name.capitalize} deployment",
               deploy_started: @deploy_started
--- a/lib/kubernetes-deploy/kubernetes_resource/deployment.rb
+++ b/lib/kubernetes-deploy/kubernetes_resource/deployment.rb
@@ -20,17 +20,15 @@ module KubernetesDeploy
         if st.success?
           pods_json = JSON.parse(pod_list)["items"]
           pods_json.each do |pod_json|
-            pod_name = pod_json["metadata"]["name"]
             pod = Pod.new(
-              name: pod_name,
               namespace: namespace,
               context: context,
-              file: nil,
+              template: pod_json,
+              logger: @logger,
               parent: "#{@name.capitalize} deployment",
-              logger: @logger
+              deploy_started: @deploy_started
             )
-            pod.deploy_started = @deploy_started
-            pod.interpret_json_data(pod_json)
+            pod.sync(pod_json)
 
             if !@representative_pod && pod_probably_new?(pod_json)
               @representative_pod = pod
--- a/lib/kubernetes-deploy/kubernetes_resource/pod.rb
+++ b/lib/kubernetes-deploy/kubernetes_resource/pod.rb
@@ -4,50 +4,31 @@ module KubernetesDeploy
     TIMEOUT = 10.minutes
     SUSPICIOUS_CONTAINER_STATES = %w(ImagePullBackOff RunContainerError ErrImagePull).freeze
 
-    def initialize(name:, namespace:, context:, file:, parent: nil, logger:)
-      @name = name
-      @namespace = namespace
-      @context = context
-      @file = file
+    def initialize(namespace:, context:, template:, logger:, parent: nil, deploy_started: nil)
       @parent = parent
-      @logger = logger
+      @deploy_started = deploy_started
+      @containers = template["spec"]["containers"].map { |c| c["name"] }
+      super(namespace: namespace, context: context, template: template, logger: logger)
     end
 
-    def sync
-      out, _err, st = kubectl.run("get", type, @name, "-a", "--output=json")
-      if @found = st.success?
-        pod_data = JSON.parse(out)
-        interpret_json_data(pod_data)
+    def sync(pod_data = nil)
+      if pod_data.blank?
+        raw_json, _err, st = kubectl.run("get", type, @name, "-a", "--output=json")
+        pod_data = JSON.parse(raw_json) if st.success?
+      end
+
+      if pod_data.present?
+        @found = true
+        interpret_pod_status_data(pod_data["status"], pod_data["metadata"]) # sets @phase, @status and @ready
+        if @deploy_started
+          log_suspicious_states(pod_data["status"].fetch("containerStatuses", []))
+        end
       else # reset
-        @status = @phase = nil
+        @found = false
+        @phase = @status = nil
         @ready = false
-        @containers = []
       end
       display_logs if unmanaged? && deploy_succeeded?
-    end
-
-    def interpret_json_data(pod_data)
-      @phase = (pod_data["metadata"]["deletionTimestamp"] ? "Terminating" : pod_data["status"]["phase"])
-      @containers = pod_data["spec"]["containers"].map { |c| c["name"] }
-
-      if @deploy_started && pod_data["status"]["containerStatuses"]
-        pod_data["status"]["containerStatuses"].each do |status|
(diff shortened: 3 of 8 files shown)


lib/kubernetes-deploy/kubernetes_resource.rb
```diff
@@ -22,27 +22,14 @@ class KubernetesResource
...
-      when 'podtemplate' then PodTemplate
-      when 'poddisruptionbudget' then PodDisruptionBudget
-      end
-
-      opts = { name: name, namespace: namespace, context: context, file: file, logger: logger }
-      if subclass
-        subclass.new(**opts)
+    def self.build(namespace:, context:, template:, logger:)
+      opts = { namespace: namespace, context: context, template: template, logger: logger }
+      if KubernetesDeploy.const_defined?(template["kind"])
```
kirs: ❤️ instead of case / when

lib/kubernetes-deploy/kubernetes_resource.rb
```diff
@@ -54,19 +41,28 @@ def timeout
       self.class.timeout
     end
 
-    def initialize(name:, namespace:, context:, file:, logger:)
+    def initialize(namespace:, context:, template:, logger:)
       # subclasses must also set these if they define their own initializer
-      @name = name
+      @name = template.fetch("metadata", {})["name"]
```
kirs: In what case the metadata will be blank and the name won't be present?

lib/kubernetes-deploy/kubernetes_resource.rb
```diff
@@ -54,19 +41,28 @@ def timeout
...
 
     def id
       "#{type}/#{name}"
     end
 
+    def file
+      @file ||= begin
+        f = Tempfile.new(["#{type}-#{name}", ".yml"])
+        f.write(YAML.dump(@template))
+        f.close
```
kirs: Do we need to put `f.close` into `ensure` block?

lib/kubernetes-deploy/kubernetes_resource.rb
```diff
@@ -54,19 +41,28 @@ def timeout
...
-      @file = file
       @logger = logger
+      @template = template
     end
 
     def id
       "#{type}/#{name}"
     end
 
+    def file
```
kirs: This method would return an instance of closed `Tempfile`:

```
#<Tempfile:/var/folders/9k/zxhywlyx2nx_3z3qnrzpkjd00000gn/T/abc20170623-92852-1180pr4 (closed)>
```

That we'd be unable to read (because read of closed file raises with `IOError: closed stream`)

I don't have full context on how you're using the return value of this method, just letting you know that it may not work as expected.

lib/kubernetes-deploy/kubernetes_resource.rb
```diff
@@ -54,19 +41,28 @@ def timeout
...
-    def initialize(name:, namespace:, context:, file:, logger:)
+    def initialize(namespace:, context:, template:, logger:)
       # subclasses must also set these if they define their own initializer
-      @name = name
+      @name = template.fetch("metadata", {})["name"]
       @namespace = namespace
       @context = context
-      @file = file
       @logger = logger
+      @template = template
```
kirs: This is more of a representation of the resource in JSON, not a template. WDYT? The name `template` may confuse someone who reads the code.

lib/kubernetes-deploy/kubernetes_resource/pod.rb
```diff
@@ -89,6 +70,27 @@ def fetch_logs
 
     private
 
+    def interpret_pod_status_data(status_data, metadata)
+      @status = @phase = (metadata["deletionTimestamp"] ? "Terminating" : status_data["phase"])
```
kirs: It always confuses me that "Terminating" is not a state and that you always have to check `deletionTimestamp` 🙈

lib/kubernetes-deploy/kubernetes_resource.rb
camilo: https://en.wikipedia.org/wiki/Composition_over_inheritance indeed

lib/kubernetes-deploy/kubernetes_resource.rb
```diff
@@ -22,27 +22,14 @@ class KubernetesResource
...
-        subclass.new(**opts)
+    def self.build(namespace:, context:, template:, logger:)
+      opts = { namespace: namespace, context: context, template: template, logger: logger }
+      if KubernetesDeploy.const_defined?(template["kind"])
+        klass = KubernetesDeploy.const_get(template["kind"])
+        klass.new(**opts)
       else
         inst = new(**opts)
-        inst.tap { |r| r.type = type }
+        inst.tap { |r| r.type = template["kind"] }
```
camilo: why do you need to `tap` ?

camilo: I meant to say, I rather the setter function be exposed than using tap; I rarely find `tap` to be necessary.

camilo:
Overall LGTM I had a couple nits but Kirs got the big ones. 

A larger design question still in my mind is the need to pass deploy started time so much; I feel that there could be another object tracking that but that is absolutely NOT a blocker.

lib/kubernetes-deploy/kubernetes_resource.rb
KnVerey: I did this because we now have not yet done the dry run (which validates the template) at the time we instantiate these classes. So although the template would not be valid if this field were missing, it is not guaranteed to be valid at this point.

kirs: There're few methods where we use `name` - is there a chance they'll access nil or empty string and break?

lib/kubernetes-deploy/kubernetes_resource.rb
KnVerey: True, it's the entire resource definition, which actually contains things called "template". I'm hesitant to call in JSON since it is parsed. Kubernetes docs refer to this as [Kubernetes Objects](https://kubernetes.io/docs/concepts/overview/working-with-objects/kubernetes-objects/), but I don't think that would really be clearer (especially given the significance of "object" in Ruby). How do you feel about `resource_definition`?

lib/kubernetes-deploy/kubernetes_resource.rb
KnVerey: This code is copy-pasted from what we're using today. It is primarily used for [passing its path to kubectl](https://github.com/Shopify/kubernetes-deploy/pull/113/files#diff-ae1efd633a69ca93e2ad9f56702555ebR230) when we do the dry run and when we actually deploy the resource. However, we may in fact read it [here](https://github.com/Shopify/kubernetes-deploy/pull/113/files#diff-ae1efd633a69ca93e2ad9f56702555ebR189) for debugging purposes. `file.read` would raise that error, but `File.read(file.path)` works fine, and that's what we're doing.

lib/kubernetes-deploy/kubernetes_resource.rb
KnVerey: Probably a good idea, yes

lib/kubernetes-deploy/kubernetes_resource.rb
kirs: 👍 for `resource_definition` or just `definition`

lib/kubernetes-deploy/kubernetes_resource.rb
kirs: This can also result to tempfiles not being garbage collected after the process is done...
If we're already using it this way today then it doesn't have to be addressed in this PR - but may be something that we want to revisit in future. WDYT?

kirs: IMO when a method returns an instance of `File` (or any `IO`) you'd expect to read it directly without the need for `File.read(file.path)`. Though it's only used for debugging and it's not critical.

KnVerey:
> A larger design question still in my mind is the need to pass deploy started time so much; I feel that there could be another object tracking that but that is absolutely NOT a blocker.

Maybe I haven't explained this properly. Normally, a resource's `deploy_started` time is set with an accessor at the time that specific resource is deployed. Each resource needs to know its own unique start time for things like implementing the timeout. What perhaps seems odd here is that parents pass their own start time along to their children, because the existence of that child was set in motion not by us observing it (instantiation time) but by the parent being deployed. For Pods, an alternative would be to use the pod's own creationTimestamp as its deploy start time. For ReplicaSets, I don't think that would work, as the RS could predate the deploy (e.g. when you're on version 2 of your deployment and you roll back to version 1, it will scale up the existing version 1 RS). Does that make sense? Do you think we should switch pods to deploy_started = creationTimestamp?

lib/kubernetes-deploy/kubernetes_resource.rb
KnVerey: > This can also result to tempfiles not being garbage collected after the process is done...

What can result in that? Tempfiles are automatically GC'd when the process exits.

> IMO when a method returns an instance of File (or any IO) you'd expect to read it directly without the need for File.read(file.path). Though it's only used for debugging and it's not critical.

Hm. The thing is that for kubectl to be able to read it, it needs to be closed. It seems to me having something external close it would be more confusing overall. Idea: We could make this method private, and only expose one called `file_path`. Then callers clearly need to do `File.read`. WDYT?

lib/kubernetes-deploy/kubernetes_resource.rb
KnVerey: The validation happens _immediately_ after this (unless this is a child resource, in which case it is inherently valid), so it shouldn't be a problem. Making it safer is a good idea, but we shouldn't be filling in missing names. I guess empty string would be slightly better than the current `nil` from that perspective?

kirs: At least `nil` would fail and we'd know when and where that happened. With an empty string we may never know if it happens. Also IMO `nil` if better from the semantics perspective because NULL value kind of unknown value.

lib/kubernetes-deploy/kubernetes_resource.rb
kirs: > Tempfiles are automatically GC'd when the process exits.

You're right, I forgot about it.

> We could make this method private, and only expose one called file_path. Then callers clearly need to do File.read. WDYT?

Sounds good 👍

lib/kubernetes-deploy/kubernetes_resource.rb
KnVerey: The user will definitely know, because the template is not valid if this is missing so the deploy can't possibly succeed. Another option is to raise a FatalDeploymentError right here to immediately abort the deploy if it is missing.

KnVerey: I implemented the FatalDeploymentError option with a test. LMK what you think.

KnVerey:
@kirs @camilo I rebased this and updated it based on your feedback. Can you please take another look? I'd love to get it in tomorrow.

lib/kubernetes-deploy/kubernetes_resource.rb
```diff
@@ -54,19 +42,27 @@ def timeout
...
+      unless @name = definition.fetch("metadata", {})["name"]
+        logger.summary.add_paragraph("Rendered template content:\n#{definition.to_yaml}")
+        raise FatalDeploymentError, "Template is missing required field metadata.name"
+      end
+
       @namespace = namespace
       @context = context
-      @file = file
       @logger = logger
+      @definition = definition
```
kirs: 😍

lib/kubernetes-deploy/kubernetes_resource/deployment.rb
```diff
@@ -20,17 +20,15 @@ def sync
...
-              file: nil,
+              definition: pod_json,
+              logger: @logger,
               parent: "#{@name.capitalize} deployment",
-              logger: @logger
+              deploy_started: @deploy_started
             )
-            pod.deploy_started = @deploy_started
-            pod.interpret_json_data(pod_json)
+            pod.sync(pod_json)
```
kirs: Just an idea for future refactorings: if `pod.sync` doesn't actually sync and just assigns the definition from `pod_json`, could this be as simple as:

```ruby
pod.definition = pod_json
```

(with some setter logic like `log_suspicious_states`)

lib/kubernetes-deploy/kubernetes_resource/pod.rb
```diff
@@ -4,50 +4,31 @@ class Pod < KubernetesResource
...
-    def initialize(name:, namespace:, context:, file:, parent: nil, logger:)
-      @name = name
-      @namespace = namespace
-      @context = context
-      @file = file
+    def initialize(namespace:, context:, definition:, logger:, parent: nil, deploy_started: nil)
       @parent = parent
-      @logger = logger
+      @deploy_started = deploy_started
+      @containers = definition["spec"]["containers"].map { |c| c["name"] }
```
kirs: Nit: using `definition.fetch("spec")` instead of `definition["spec"]` would automatically raise if this key is not present (fail earlier), which is good because now when `spec` key is missing it will lead to `NoMethodError`

lib/kubernetes-deploy/kubernetes_resource/pod.rb
```diff
@@ -89,6 +70,27 @@ def fetch_logs
...
 
+    def interpret_pod_status_data(status_data, metadata)
+      @status = @phase = (metadata["deletionTimestamp"] ? "Terminating" : status_data["phase"])
+
+      if @phase == "Failed" && status_data['reason'].present?
+        @status += " (Reason: #{status_data['reason']})"
+      elsif @phase != "Terminating"
+        ready_condition = status_data.fetch("conditions", []).find { |condition| condition["type"] == "Ready" }
+        @ready = ready_condition.present? && (ready_condition["status"] == "True")
+        @status += " (Ready: #{@ready})"
```
kirs: I couldn't find the test coverage for the pod status. Something that could be added as a follow-up in the future

lib/kubernetes-deploy/kubernetes_resource/pod.rb
```diff
@@ -89,6 +70,27 @@ def fetch_logs
...
+      end
+    end
+
+    def log_suspicious_states(container_statuses)
+      container_statuses.each do |status|
+        waiting_state = status["state"]["waiting"] if status["state"]
+        reason = waiting_state["reason"] if waiting_state
+        next unless SUSPICIOUS_CONTAINER_STATES.include?(reason)
+        @logger.warn("#{id} has container in state #{reason} (#{waiting_state['message']})")
```
kirs: Will a test fail if we remove this line?

lib/kubernetes-deploy/kubernetes_resource/pod.rb
```diff
@@ -89,6 +70,27 @@ def fetch_logs
...
+        ready_condition = status_data.fetch("conditions", []).find { |condition| condition["type"] == "Ready" }
+        @ready = ready_condition.present? && (ready_condition["status"] == "True")
+        @status += " (Ready: #{@ready})"
+      end
+    end
+
+    def log_suspicious_states(container_statuses)
+      container_statuses.each do |status|
+        waiting_state = status["state"]["waiting"] if status["state"]
+        reason = waiting_state["reason"] if waiting_state
```
kirs: You could probably avoid those `if`s by using try:

```
reason = status.try!(:[], "state").try!(:[], "waiting").try!(:[], "reason")
```

Dunno if this chain makes the code better.

test/integration/kubernetes_deploy_test.rb
```diff
@@ -388,6 +388,21 @@ def test_output_when_switching_labels_incorrectly
...
+  def test_deploy_aborts_immediately_if_metadata_name_missing
+    success = deploy_fixtures("hello-cloud", subset: ["configmap-data.yml"]) do |fixtures|
+      definition = fixtures["configmap-data.yml"]["ConfigMap"].first
+      definition["metadata"].delete("name")
+    end
+    assert_equal false, success, "Deploy succeeded when it was expected to fail"
+
+    assert_logs_match_all([
+      "Result: FAILURE",
+      "Template is missing required field metadata.name",
```
kirs: 👍

lib/kubernetes-deploy/kubernetes_resource.rb
```diff
@@ -54,19 +42,27 @@ def timeout
       self.class.timeout
     end
 
-    def initialize(name:, namespace:, context:, file:, logger:)
+    def initialize(namespace:, context:, definition:, logger:)
       # subclasses must also set these if they define their own initializer
-      @name = name
+      unless @name = definition.fetch("metadata", {})["name"]
```
kirs: It looks like this doesn't handle the case when `name` is an empty string? Could probably use `present?` from ActiveSupport

lib/kubernetes-deploy/kubernetes_resource/deployment.rb
KnVerey: `Pod` is going to change a lot again in a follow-up PR that helps us fail doomed deploys faster, but I will keep this suggestion in mind.

lib/kubernetes-deploy/kubernetes_resource/pod.rb
KnVerey: Yeah, my third branch in this stack (the one you haven't seen yet) should cover this. 👍

lib/kubernetes-deploy/kubernetes_resource/pod.rb
KnVerey: I don't think so, but this is being replaced by fail-fast logic in the third PR I've been referencing. TBH people largely ignore it right now.

lib/kubernetes-deploy/kubernetes_resource/pod.rb
KnVerey: Hm, maybe I should raise here like I'm now doing in the base class if metadata.name is missing. This is similarly a user error that should be exposed as a template problem, not a stacktrace.

lib/kubernetes-deploy/kubernetes_resource/pod.rb
KnVerey: That's a decent suggestion, but I'm going to leave it as is because my other branch deletes this code.

lib/kubernetes-deploy/kubernetes_resource/pod.rb
KnVerey: I tophatted this by removing the failure stubbing in the `test_deploy_result_logging_for_mixed_result_deploy` test case (and adding `CrashLoopBackOff` to the suspicious states list--it should have been there). Result: 

screenshot

It still works, and it is still noisy. I can't wait to replace this!

PR #114, opened 2017-06-22
Add `context` and `namespace` template variables

DazWorrall:
I have a need to know the region a pod is in at runtime, but curling the GCP metadata is turning out to be flaky (sometimes it's empty), and trying to expose it from a node label doesn't appear to be possible.

I _can_ infer it from the kube context though, so exposing it to the template solves a problem for me. I added the namespace for good measure while I was in there.

🎩 'ed locally against some test CI agents.

### DIFF
--- a/lib/kubernetes-deploy/runner.rb
+++ b/lib/kubernetes-deploy/runner.rb
@@ -137,6 +137,8 @@ module KubernetesDeploy
       {
         'current_sha' => @current_sha,
         'deployment_id' => @id,
+        'context' => @context,
+        'namespace' => @namespace,
       }.merge(@bindings)
     end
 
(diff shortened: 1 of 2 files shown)


KnVerey:
I don't think we need this. Here's why:

- If your pod needs to know what namespace it is running in, the correct way to get this info is the downward API. 
- The context name reflects the deployer's kubeconfig file, not necessarily an inherent property of the cluster. Therefore I don't think it is something people should be relying on in their templates.
- We already have a mechanism for adding arbitrary bindings required by particular templates: [--bindings](https://github.com/Shopify/kubernetes-deploy/blob/master/exe/kubernetes-deploy#L16). We definitely need to document this though. And this gem in general. 😄 

Do these options meet your needs @DazWorrall?

DazWorrall:
> Do these options meet your needs @DazWorrall?

The last one in particular, thanks!

PR #116, opened 2017-07-04
Make deployments fail faster when they are likely doomed

KnVerey:
@Shopify/cloudplatform @kirs 

## What?

Minimize the frequency with which doomed deploys result in a timeout. If we can detect that it is doomed, we should fail fast!

## Why?

Timeouts are currently super common, especially when apps are onboarding. There's nothing worse than having to repeatedly wait 5 minutes to see what the problem is, and in some cases not getting an obvious failure reason even then.

## How?

Implements a `Container` class inside `Pod` that knows how to detect several of the most common doom conditions:
* container mounting secret that doesn't exist
* crashing init container
* crashing regular container
* container command doesn't exist
* bad container image (e.g. path is wrong, or image build not finished)

When any of these conditions is present, `pod.deploy_failed?` will be true. If all of the current replicaSet's pods have failed, the `deployment.deploy_failed?` will be true as well.

It also leverages this new class to improve the timeout message when the readiness probe is failing (but does not make the deploy fail, since who knows how long an arbitrary pod is supposed to take to become ready). Sadly it does not improve anything in the cases of mounting a non-existent volume, as I couldn't find any way to deduce that fact from the pod status (events already do tell you this today though).

## Review

Please give special attention to the kubernetes logic here -- could this fail deploys too eagerly? are the checks too fragile? does it make bad assumptions about potential mixes of pod/container states? etc.

Check out the comments on the test titles for examples of the resulting deploy summary sections.

### DIFF
--- a/lib/kubernetes-deploy/kubernetes_resource.rb
+++ b/lib/kubernetes-deploy/kubernetes_resource.rb
@@ -124,7 +124,7 @@ module KubernetesDeploy
 
       events = fetch_events
       if events.present?
-        helpful_info << "  - Events:"
+        helpful_info << "  - Events (common success events excluded):"
         events.each do |identifier, event_hashes|
           event_hashes.each { |event| helpful_info << "      [#{identifier}]\t#{event}" }
         end
--- a/lib/kubernetes-deploy/kubernetes_resource.rb
+++ b/lib/kubernetes-deploy/kubernetes_resource.rb
@@ -139,6 +139,11 @@ module KubernetesDeploy
         else
           sorted_logs = container_logs.sort_by { |_, log_lines| log_lines.length }
           sorted_logs.each do |identifier, log_lines|
+            if log_lines.empty?
+              helpful_info << "  - Logs from container '#{identifier}': #{DEBUG_RESOURCE_NOT_FOUND_MESSAGE}"
+              next
+            end
+
             helpful_info << "  - Logs from container '#{identifier}' (last #{LOG_LINE_COUNT} lines shown):"
             log_lines.each do |line|
               helpful_info << "      #{line}"
@@ -220,7 +225,8 @@ module KubernetesDeploy
           %[(eq .involvedObject.kind "#{kind}")],
           %[(eq .involvedObject.name "#{name}")],
           '(ne .reason "Started")',
-          '(ne .reason "Created")'
+          '(ne .reason "Created")',
+          '(ne .reason "SuccessfulCreate")'
         ]
         condition_start = "{{if and #{and_conditions.join(' ')}}}"
         field_part = FIELDS.map { |f| "{{#{f}}}" }.join(%({{print "#{FIELD_SEPARATOR}"}}))
--- a/lib/kubernetes-deploy/kubernetes_resource.rb
+++ b/lib/kubernetes-deploy/kubernetes_resource.rb
@@ -226,7 +226,10 @@ module KubernetesDeploy
           %[(eq .involvedObject.name "#{name}")],
           '(ne .reason "Started")',
           '(ne .reason "Created")',
-          '(ne .reason "SuccessfulCreate")'
+          '(ne .reason "SuccessfulCreate")',
+          '(ne .reason "Scheduled")',
+          '(ne .reason "Pulling")',
+          '(ne .reason "Pulled")'
         ]
         condition_start = "{{if and #{and_conditions.join(' ')}}}"
         field_part = FIELDS.map { |f| "{{#{f}}}" }.join(%({{print "#{FIELD_SEPARATOR}"}}))
--- a/lib/kubernetes-deploy/kubernetes_resource/pod.rb
+++ b/lib/kubernetes-deploy/kubernetes_resource/pod.rb
@@ -2,16 +2,16 @@
 module KubernetesDeploy
   class Pod < KubernetesResource
     TIMEOUT = 10.minutes
-    SUSPICIOUS_CONTAINER_STATES = %w(ImagePullBackOff RunContainerError ErrImagePull CrashLoopBackOff).freeze
 
     def initialize(namespace:, context:, definition:, logger:, parent: nil, deploy_started: nil)
       @parent = parent
       @deploy_started = deploy_started
-      @containers = definition.fetch("spec", {}).fetch("containers", {}).map { |c| c["name"] }
+      @containers = definition.fetch("spec", {}).fetch("containers", []).map { |c| Container.new(c) }
       unless @containers.present?
         logger.summary.add_paragraph("Rendered template content:\n#{definition.to_yaml}")
         raise FatalDeploymentError, "Template is missing required field spec.containers"
       end
+      @containers += definition["spec"].fetch("initContainers", []).map { |c| Container.new(c, init: true) }
       super(namespace: namespace, context: context, definition: definition, logger: logger)
     end
 
@@ -23,15 +23,16 @@ module KubernetesDeploy
 
       if pod_data.present?
         @found = true
-        interpret_pod_status_data(pod_data["status"], pod_data["metadata"]) # sets @phase, @status and @ready
-        if @deploy_started
-          log_suspicious_states(pod_data["status"].fetch("containerStatuses", []))
-        end
+        @phase = @status = pod_data["status"]["phase"]
+        @status += " (Reason: #{pod_data['status']['reason']})" if pod_data['status']['reason'].present?
+        @ready = ready?(pod_data["status"])
+        update_container_statuses(pod_data["status"])
       else # reset
-        @found = false
-        @phase = @status = nil
-        @ready = false
+        @found = @ready = false
+        @status = @phase = 'Unknown'
+        @containers.each(&:reset_status)
       end
+
       display_logs if unmanaged? && deploy_succeeded?
     end
 
@@ -44,11 +45,34 @@ module KubernetesDeploy
     end
 
     def deploy_failed?
-      @phase == "Failed"
+      return true if @phase == "Failed"
+      @containers.any?(&:doomed?)
     end
 
     def exists?
       @found
+    end
+
+    def timeout_message
+      return unless @phase == "Running" && !@ready
+      pieces = ["Your pods are running, but the following containers seem to be failing their readiness probes:"]
+      @containers.each do |c|
+        next if c.init_container? || c.ready?
+        yellow_name = ColorizedString.new(c.name).yellow
+        pieces << "> #{yellow_name} must respond with a good status code at '#{c.probe_location}'"
+      end
+      pieces.join("\n") + "\n"
+    end
+
+    def failure_message
+      doomed_containers = @containers.select(&:doomed?)
+      return unless doomed_containers.present?
+      container_messages = doomed_containers.map do |c|
+        red_name = ColorizedString.new(c.name).red
+        "> #{red_name}: #{c.doom_reason}"
+      end
+      intro = "The following containers are in a state that is unlikely to be recoverable:\n"
+      intro + container_messages.join("\n") + "\n"
     end
 
     # Returns a hash in the following format:
@@ -58,39 +82,35 @@ module KubernetesDeploy
     # }
     def fetch_logs
       return {} unless exists? && @containers.present?
-      @containers.each_with_object({}) do |container_name, container_logs|
+      @containers.each_with_object({}) do |container, container_logs|
         cmd = [
           "logs",
           @name,
-          "--container=#{container_name}",
+          "--container=#{container.name}",
           "--since-time=#{@deploy_started.to_datetime.rfc3339}",
         ]
         cmd << "--tail=#{LOG_LINE_COUNT}" unless unmanaged?
         out, _err, _st = kubectl.run(*cmd)
-        container_logs[container_name] = out.split("\n")
+        container_logs[container.name] = out.split("\n")
       end
     end
 
     private
 
-    def interpret_pod_status_data(status_data, metadata)
-      @status = @phase = (metadata["deletionTimestamp"] ? "Terminating" : status_data["phase"])
-
-      if @phase == "Failed" && status_data['reason'].present?
-        @status += " (Reason: #{status_data['reason']})"
-      elsif @phase != "Terminating"
-        ready_condition = status_data.fetch("conditions", []).find { |condition| condition["type"] == "Ready" }
-        @ready = ready_condition.present? && (ready_condition["status"] == "True")
-        @status += " (Ready: #{@ready})"
-      end
+    def ready?(status_data)
+      ready_condition = status_data.fetch("conditions", []).find { |condition| condition["type"] == "Ready" }
+      ready_condition.present? && (ready_condition["status"] == "True")
     end
 
-    def log_suspicious_states(container_statuses)
-      container_statuses.each do |status|
-        waiting_state = status["state"]["waiting"] if status["state"]
-        reason = waiting_state["reason"] if waiting_state
-        next unless SUSPICIOUS_CONTAINER_STATES.include?(reason)
-        @logger.warn("#{id} has container in state #{reason} (#{waiting_state['message']})")
+    def update_container_statuses(status_data)
+      @containers.each do |c|
+        key = c.init_container? ? "initContainerStatuses" : "containerStatuses"
+        if status_data.key?(key)
+          data = status_data[key].find { |st| st["name"] == c.name }
+          c.update_status(data)
+        else
+          c.reset_status
+        end
       end
     end
 
@@ -119,6 +139,68 @@ module KubernetesDeploy
       end
 
       @already_displayed = true
+    end
+
+    class Container
+      STATUS_SCAFFOLD = {
+        "state" => {
+          "running" => {},
+          "waiting" => {},
+          "terminated" => {},
+        },
+        "lastState" => {
+          "terminated" => {}
+        }
+      }.freeze
+
+      attr_reader :name, :probe_location
+
+      def initialize(definition, init: false)
+        @init = init
+        @name = definition["name"]
+        @image = definition["image"]
+        @probe_location = definition.fetch("readinessProbe", {}).fetch("httpGet", {})["path"]
+        @status = STATUS_SCAFFOLD.dup
+      end
+
+      def doomed?
+        doom_reason.present?
+      end
+
+      def doom_reason
+        exit_code = @status['lastState']['terminated']['exitCode']
+        last_terminated_reason = @status["lastState"]["terminated"]["reason"]
+        limbo_reason = @status["state"]["waiting"]["reason"]
+
+        if last_terminated_reason == "ContainerCannotRun"
+          # ref: https://github.com/kubernetes/kubernetes/blob/562e721ece8a16e05c7e7d6bdd6334c910733ab2/pkg/kubelet/dockershim/docker_container.go#L353
+          "Failing to start (exit #{exit_code}): #{@status['lastState']['terminated']['message']}"
+        elsif limbo_reason == "CrashLoopBackOff"
+          "Crashing repeatedly (exit #{exit_code}). See logs for more information."
+        elsif %w(ImagePullBackOff ErrImagePull).include?(limbo_reason)
+          "Failing to pull image #{@image}. "\
+          "Did you wait for it to be built and pushed to the registry before deploying?"
+        elsif @status["state"]["waiting"]["message"] == "Generate Container Config Failed"
+          # reason/message seem to be backwards for this case -- reason is the free-form part
+          "Failing to generate container configuration: #{limbo_reason}"
+        end
+      end
+
+      def ready?
+        @status['ready'] == "true"
+      end
+
+      def init_container?
+        @init
+      end
+
+      def update_status(data)
+        @status = STATUS_SCAFFOLD.deep_merge(data || {})
+      end
+
+      def reset_status
+        @status = STATUS_SCAFFOLD.dup
+      end
     end
   end
 end
--- a/lib/kubernetes-deploy/kubernetes_resource/pod.rb
+++ b/lib/kubernetes-deploy/kubernetes_resource/pod.rb
@@ -11,7 +11,7 @@ module KubernetesDeploy
         logger.summary.add_paragraph("Rendered template content:\n#{definition.to_yaml}")
         raise FatalDeploymentError, "Template is missing required field spec.containers"
       end
-      @containers += definition["spec"].fetch("initContainers", []).map { |c| Container.new(c, init: true) }
+      @containers += definition["spec"].fetch("initContainers", []).map { |c| Container.new(c, init_container: true) }
       super(namespace: namespace, context: context, definition: definition, logger: logger)
     end
 
@@ -54,6 +54,7 @@ module KubernetesDeploy
     end
 
     def timeout_message
+      return if unmanaged?
       return unless @phase == "Running" && !@ready
       pieces = ["Your pods are running, but the following containers seem to be failing their readiness probes:"]
       @containers.each do |c|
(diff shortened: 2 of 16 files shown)


test/integration/kubernetes_deploy_test.rb
```diff
@@ -211,6 +212,69 @@ def test_bad_container_image_on_run_once_halts_and_fails_deploy
     hello_cloud.refute_web_resources_exist
   end
 
+  def test_deployment_container_mounting_secret_that_does_not_exist_as_env_var_fails_quickly
```
KnVerey: image

test/integration/kubernetes_deploy_test.rb
```diff
@@ -211,6 +212,69 @@ def test_bad_container_image_on_run_once_halts_and_fails_deploy
...
+    assert_equal false, success, "Deploy succeeded when it was expected to fail"
+
+    assert_logs_match_all([
+      "Deployment/web: FAILED",
+      "The following containers are in a state that is unlikely to be recoverable:",
+      "app: Failing to generate container configuration: secrets \"monitoring-token\" not found"
+    ], in_order: true)
+  end
+
+  def test_bad_container_image_on_deployment_pod_fails_quickly
```
KnVerey: image

test/integration/kubernetes_deploy_test.rb
```diff
@@ -211,6 +212,69 @@ def test_bad_container_image_on_run_once_halts_and_fails_deploy
...
+
+    assert_logs_match_all([
+      "Deployment/cannot-run: FAILED",
+      "The following containers are in a state that is unlikely to be recoverable:",
+      "container-cannot-run: Failing to pull image some-invalid-image:badtag.",
+      "Did you wait for it to be built and pushed to the registry before deploying?"
+    ], in_order: true)
+  end
+
+  def test_bad_init_container_on_deployment_fails_quickly
```
KnVerey: image

test/integration/kubernetes_deploy_test.rb
```diff
@@ -211,6 +212,69 @@ def test_bad_container_image_on_run_once_halts_and_fails_deploy
...
+    refute deploy_fixtures("invalid", subset: ["init_crash.yml"])
+    assert_logs_match_all([
+      "Deployment/init-crash: FAILED",
+      "The following containers are in a state that is unlikely to be recoverable:",
+      "init-crash-loop-back-off: Crashing repeatedly (exit 1). See logs for more information.",
+      "ls: /not-a-dir: No such file or directory" # logs
+    ], in_order: true)
+  end
+
+  def test_crashing_container_on_deployment_fails_quickly
```
KnVerey: image

test/integration/kubernetes_deploy_test.rb
```diff
@@ -211,6 +212,69 @@ def test_bad_container_image_on_run_once_halts_and_fails_deploy
...
+
+    assert_logs_match_all([
+      "Deployment/crash-loop: FAILED",
+      "The following containers are in a state that is unlikely to be recoverable:",
+      "crash-loop-back-off: Crashing repeatedly (exit 1). See logs for more information.",
+      'nginx: [error] open() "/var/run/nginx.pid" failed (2: No such file or directory)' # Logs
+    ], in_order: true)
+  end
+
+  def test_unrunnable_container_on_deployment_pod_fails_quickly
```
KnVerey: image

test/integration/kubernetes_deploy_test.rb
```diff
@@ -255,13 +319,15 @@ def test_deploy_fails_if_required_binding_not_present
...
       assert deploy_fixtures('long-running')
+      assert_logs_match(%r{Service/multi-replica\s+Selects 2 pods})
     end
 
     pods = kubeclient.get_pods(namespace: @namespace, label_selector: 'name=jobs,app=fixtures')
     assert_equal 4, pods.size
 
     count = count_by_revisions(pods)
     assert_equal [2, 2], count.values
+    assert_logs_match(%r{Service/multi-replica\s+Selects 4 pods})
```
KnVerey: These new assertions (and the new service in this test case) are regression tests for https://github.com/Shopify/kubernetes-deploy/pull/116/files#diff-ae1efd633a69ca93e2ad9f56702555ebR153 (displaying stale info for the "final" status of resources that succeeded quickly, e.g. because the service was pulled out of the poll loop as soon as 1 replica existed, but it had 4 by the end of the deploy).

test/integration/kubernetes_deploy_test.rb
```diff
@@ -340,45 +406,40 @@ def test_pruning_of_existing_managed_secrets_when_ejson_file_has_been_deleted
   end
 
   def test_deploy_result_logging_for_mixed_result_deploy
```
KnVerey: image

lib/kubernetes-deploy/kubernetes_resource.rb
```diff
@@ -201,7 +206,8 @@ def self.go_template_for(kind, name)
           %[(eq .involvedObject.kind "#{kind}")],
           %[(eq .involvedObject.name "#{name}")],
           '(ne .reason "Started")',
-          '(ne .reason "Created")'
+          '(ne .reason "Created")',
+          '(ne .reason "SuccessfulCreate")'
```
KnVerey: This filters out these event types. Looking at the test output (see the screenshots), I'm thinking perhaps we should filter out `Scheduled`, `Pulling` and `Pulled` as well. WDYT? Now that pull errors should be reported, could those events be useful?

lib/kubernetes-deploy/kubernetes_resource.rb
```diff
@@ -138,6 +138,11 @@ def debug_message
         else
           sorted_logs = container_logs.sort_by { |_, log_lines| log_lines.length }
           sorted_logs.each do |identifier, log_lines|
+            if log_lines.empty?
+              helpful_info << "  - Logs from container '#{identifier}': #{DEBUG_RESOURCE_NOT_FOUND_MESSAGE}"
```
KnVerey: This is for consistency. However, in the interests of reducing noise, I'm wondering if we should go the other way and get rid of these messages entirely (don't output anything if logs/events can't be fetched. WDYT?

lib/kubernetes-deploy/kubernetes_resource/replica_set.rb
```diff
@@ -40,6 +40,14 @@ def deploy_failed?
       @pods.present? && @pods.all?(&:deploy_failed?)
     end
 
+    def failure_message
+      @pods.map(&:failure_message).compact.uniq.join("\n")
```
KnVerey: The `failure_message`s are fairly arbitrary strings, so this could be a bit intensive if there are a ton of failed pods in the set, and in the vast majority of cases, it will only yield one failure message anyway. Wdyt, worth doing over `@pods.find(&:deploy_failed?).failure_message`?

test/integration/kubernetes_deploy_test.rb
```diff
@@ -211,17 +212,85 @@ def test_bad_container_image_on_run_once_halts_and_fails_deploy
...
+      "Deployment/web: FAILED",
+      "The following containers are in a state that is unlikely to be recoverable:",
+      "app: Failing to generate container configuration: secrets \"monitoring-token\" not found",
+      "Final status: 3 replicas, 3 updatedReplicas, 3 unavailableReplicas"
+    ], in_order: true)
+
+    assert_logs_match("The following containers are in a state that is unlikely to be recoverable", 1) # no duplicates
+  end
+
+  def test_bad_container_image_on_deployment_pod_fails_quickly
```
KnVerey: image

lib/kubernetes-deploy/kubernetes_resource.rb
dalehamel: I think that telling the user there are no logs is information that is useful for debugging.

From an ideal perspective, i like the unix "no news is good news" though. I could go either way here - I guess it depends how often this might come up in the logs.

lib/kubernetes-deploy/kubernetes_resource/pod.rb
```diff
@@ -44,53 +45,72 @@ def deploy_succeeded?
     end
 
     def deploy_failed?
-      @phase == "Failed"
+      return true if @phase == "Failed"
+      @containers.any?(&:doomed?)
```
dalehamel: elegant!

dalehamel:
@KnVerey i'm afraid I don't yet have my sea-legs yet for this project in order to give a critical / useful review for you here, but I can say that moving the logic and state out to a container class as you have done here seems to make sense, and the failure conditions you identified are sensible too. I might try taking another pass at the specific logic around those cases, to try and poke holes it in, when I get some free time.

lib/kubernetes-deploy/kubernetes_resource/pod.rb
```diff
@@ -120,5 +140,67 @@ def display_logs
...
+        if last_terminated_reason == "ContainerCannotRun"
+          # ref: https://github.com/kubernetes/kubernetes/blob/562e721ece8a16e05c7e7d6bdd6334c910733ab2/pkg/kubelet/dockershim/docker_container.go#L353
+          "Failing to start (exit #{exit_code}): #{@status['lastState']['terminated']['message']}"
+        elsif limbo_reason == "CrashLoopBackOff"
+          "Crashing repeatedly (exit #{exit_code}). See logs for more information."
+        elsif %w(ImagePullBackOff ErrImagePull).include?(limbo_reason)
+          "Failing to pull image #{@image}. "\
+          "Did you wait for it to be built and pushed to the registry before deploying?"
+        elsif @status["state"]["waiting"]["message"] == "Generate Container Config Failed"
+          # reason/message seem to be backwards for this case -- reason is the free-form part
```
stefanmb: This seems to have been fixed upstream a few days ago: https://github.com/kubernetes/kubernetes/commit/df41787b1a3f51b73fb6db8a2203f0a7c7c92931

Maybe check for both the legacy way and the correct way since we'll likely forget about this fix later?

lib/kubernetes-deploy/kubernetes_resource.rb
stefanmb: `Scheduled` is often useful in determining progress when new nodes have to be allocated, I would keep the messages.

lib/kubernetes-deploy/kubernetes_resource.rb
stefanmb: Let's keep it.

lib/kubernetes-deploy/kubernetes_resource/service.rb
```diff
@@ -41,19 +31,26 @@ def exists?
...
+
+    def fetch_related_pod_count
+      raw_json, _err, st = kubectl.run("get", "pods", "--selector=#{selector}", "--output=json")
+      return 0 unless st.success?
+      JSON.parse(raw_json)["items"].length
+    end
+
+    def fetch_related_replica_count
+      raw_json, _err, st = kubectl.run("get", "deployments", "--selector=#{selector}", "--output=json")
+      return 1 unless st.success?
```
stefanmb: If I understand correctly the point of this is to cause the replica count check to fail if we're unable to retrieve the actual replica count? Should we just use `nil` for that case as opposed to an arbitrary integer not equal to 0?

lib/kubernetes-deploy/kubernetes_resource/replica_set.rb
stefanmb: Do we have any reason to believe this will cause a perf issue?

lib/kubernetes-deploy/kubernetes_resource/pod.rb
```diff
@@ -120,5 +140,67 @@ def display_logs
...
+        exit_code = @status['lastState']['terminated']['exitCode']
+        last_terminated_reason = @status["lastState"]["terminated"]["reason"]
+        limbo_reason = @status["state"]["waiting"]["reason"]
+
+        if last_terminated_reason == "ContainerCannotRun"
+          # ref: https://github.com/kubernetes/kubernetes/blob/562e721ece8a16e05c7e7d6bdd6334c910733ab2/pkg/kubelet/dockershim/docker_container.go#L353
+          "Failing to start (exit #{exit_code}): #{@status['lastState']['terminated']['message']}"
+        elsif limbo_reason == "CrashLoopBackOff"
+          "Crashing repeatedly (exit #{exit_code}). See logs for more information."
+        elsif %w(ImagePullBackOff ErrImagePull).include?(limbo_reason)
```
stefanmb: One concern I have here is that on large deployments a single container going into `ImagePullBackOff` will fail the entire deploy, when it could be a recoverable condition. Can we check for "not found" as part of the message before failing immediately for this reason?

e.g `rpc error: code = 2 desc = net/http: request canceled"` warrants a retry, where as `"Error: image foo:bar not found` may not.

lib/kubernetes-deploy/kubernetes_resource.rb
KnVerey: You'd keep Scheduled, SuccessfulCreate, Pulling and Pulled, or just Scheduled? My concern here is generating so much noise that people stop reading the output. Keep in mind that the other events are always available in Splunk for more advanced debugging.

stefanmb: Personally I like seeing those, but maybe you're right that we should reduce noise. I won't object to filtering all 4.

lib/kubernetes-deploy/kubernetes_resource/replica_set.rb
KnVerey: Not really. It's true that even for a large deployment that somehow manages to fail in multiple ways, in most cases (except maxUnavail/Surge 100%) the number of failed pods would be limited by the rollout strategy. I don't have any evidence that the number of comparisons likely to be involved would actually be slow.

lib/kubernetes-deploy/kubernetes_resource/pod.rb
KnVerey: > One concern I have here is that on large deployments a single container going into ImagePullBackOff will fail the entire deploy, when it could be a recoverable condition. 

Since the condition is `@pods.present? && @pods.all?(&:deploy_failed?)` (all?, not any?), it would only fail the deploy if they _all_ hit that same transient failure. You make a very good point though. It is easy enough to look for "not found". Could be a little fragile, but a test will fail if a future k8s version changes the message, so I'll do it 👍 .

lib/kubernetes-deploy/kubernetes_resource/pod.rb
KnVerey: Good find! I'll change the comment to reference that commit. The test `test_deployment_container_mounting_secret_that_does_not_exist_as_env_var_fails_quickly` will fail when we upgrade to a version that includes that commit, so I'm thinking we can flip the condition at that time as part of the version support upgrade.

lib/kubernetes-deploy/kubernetes_resource/pod.rb
ibawt: that would be wonderful!

lib/kubernetes-deploy/kubernetes_resource/service.rb
KnVerey: > If I understand correctly the point of this is to cause the replica count check to fail if we're unable to retrieve the actual replica count?

Yeah, my thinking was that if we encounter a blip on this poll that makes us unable to retrieve the correct count (we poll every 3 sec currently), we should assume it has not succeeded. So by default assume that we require at least one pod (here) and that we see 0 (default for fetch_related_pod_count). I was trying to avoid making the consumers handle nil, but it isn't that many places, and lying about the result is indeed sketchy. I'll change it.

KnVerey:
Updated

lib/kubernetes-deploy/kubernetes_resource/replica_set.rb
kirs: Does it mean that if Shopify/shopify with 500 pods fails to start, it would print logs for all of them?

test/fixtures/long-running/jobs.yml.erb
```diff
@@ -1,3 +1,16 @@
+apiVersion: v1
+kind: Service
+metadata:
+  name: multi-replica
+spec:
+  ports:
+  - port: 80
+    name: http
```
kirs: Why does something that emulates a long-running background queue jobs need http spec?

lib/kubernetes-deploy/kubernetes_resource/pod.rb
```diff
@@ -120,5 +140,69 @@ def display_logs
...
+        },
+        "lastState" => {
+          "terminated" => {}
+        }
+      }.freeze
+
+      attr_reader :name, :probe_location
+
+      def initialize(definition, init: false)
+        @init = init
```
kirs: When I was reading this I thought it's a hack to bypass some initializer code - but then I realized it's a flag for k8s `initContainer`. Can we rename the kwarg and the instance variable to `init_container`?

lib/kubernetes-deploy/kubernetes_resource/pod.rb
```diff
@@ -120,5 +140,69 @@ def display_logs
 
       @already_displayed = true
     end
+
+    class Container
```
kirs: Should we add some basic unit tests for this class? Or do you consider it very internal?

test/fixtures/invalid/init_crash.yml
```diff
@@ -0,0 +1,19 @@
...
+spec:
+  replicas: 2
+  template:
+    metadata:
+      labels:
+        name: web
+        app: crash-app
+    spec:
+      initContainers:
+      - name: init-crash-loop-back-off
```
kirs: should we add a test case with successful `initContainer` and unsuccessful `container`?

lib/kubernetes-deploy/kubernetes_resource/pod.rb
```diff
@@ -120,5 +140,69 @@ def display_logs
...
+      STATUS_SCAFFOLD = {
```
kirs: Is this for "unknown" status?

ibawt:
this is the error we shouldn't bail on:
```Failed to pull image "gcr.io/shopify-docker-images/apps/production/shopify:bd8e25c82c9a56d6f5f0d7dfc7a4d71a24150c42": rpc error: code = 2 desc = net/http: request canceled	```

File: lib/kubernetes-deploy/kubernetes_resource/pod.rb (hunk 18 above)
Comment (KnVerey) [inline lib/kubernetes-deploy/kubernetes_resource/pod.rb]: I consider it very internal. I'm going to try to get coverage of the case Ian mentioned somehow though, and it can't really be an integration test since the intended behaviour is a potentially long timeout. I'll see what I can do.

File: lib/kubernetes-deploy/kubernetes_resource/pod.rb (hunk 20 above)
Comment (KnVerey) [inline lib/kubernetes-deploy/kubernetes_resource/pod.rb]: Sorta. It's used to provide a reliable structure so that I don't have to be nil checking everywhere.

File: lib/kubernetes-deploy/kubernetes_resource/pod.rb (hunk 17 above)
Comment (KnVerey) [inline lib/kubernetes-deploy/kubernetes_resource/pod.rb]: Sure

File: lib/kubernetes-deploy/kubernetes_resource/replica_set.rb (hunk 10 above)
Comment (KnVerey) [inline lib/kubernetes-deploy/kubernetes_resource/replica_set.rb]: No, this is the custom part of the message that optionally gets printed before the logs/events/etc. If 500 pods fail to start and the `Container#doom_reasons` aren't all the same, it would print however many variations there were. I'd expect that there would be very little variation though, given that the dynamic info in doom reasons is just exit status code, image name and container config problem (e.g. missing secret name).

File: lib/kubernetes-deploy/kubernetes_resource/pod.rb (hunk 20 above)
Comment (kirs) [inline lib/kubernetes-deploy/kubernetes_resource/pod.rb]: Then maybe it's a `DEFAULT_STATUS`?

File: test/fixtures/invalid/init_crash.yml (hunk 19 above)
Comment (KnVerey) [inline test/fixtures/invalid/init_crash.yml]: That's a good point; there isn't currently any test with a successful init container.

File: test/fixtures/long-running/jobs.yml.erb (hunk 16 above)
Comment (KnVerey) [inline test/fixtures/long-running/jobs.yml.erb]: It's because I'm shoehorning that existing test into also being able to assert proper logging for a service whose number of endpoints changes. The test is doing exactly what I'd want to do for this new test (see https://github.com/Shopify/kubernetes-deploy/pull/116/files#r125535926) and it inherently is longer-running than most. My thinking was that the ideological mismatch of adding a service to the template was a lesser evil than adding an extra 15s to every test run just for a logging assertion. Is that terrible?

File: lib/kubernetes-deploy/kubernetes_resource/pod.rb (hunk 20 above)
Comment (KnVerey) [inline lib/kubernetes-deploy/kubernetes_resource/pod.rb]: Not really, since it is empty? It's not there to provide fallback data, but rather to make it easier not to 💥  if the json we got doesn't contain all the keys we use.

File: lib/kubernetes-deploy/kubernetes_resource/pod.rb [hunk 21]
```diff
@@ -120,5 +140,69 @@ def display_logs
...
+        @probe_location = definition.fetch("readinessProbe", {}).fetch("httpGet", {})["path"]
+        @status = STATUS_SCAFFOLD.dup
+      end
+
+      def doomed?
+        doom_reason.present?
+      end
+
+      def doom_reason
+        exit_code = @status['lastState']['terminated']['exitCode']
```
dalehamel: from your scaffold it looks like these are always safe to dereference, but might that change?

lib/kubernetes-deploy/kubernetes_resource/replica_set.rb
```diff
@@ -53,7 +61,8 @@ def exists?
     def fetch_events
       own_events = super
       return own_events unless @pods.present?
-      own_events.merge(@pods.first.fetch_events)
+      most_useful_pod = @pods.find(&:deploy_failed?) || @pods.find(&:deploy_timed_out?) || @pods.first
```
dalehamel: so you are trying to find a pod that has explicitly been identified as the one that cause the deploy to fail, falling back to one that explicitly caused a timeout, falling back to just the first pod (arbitrarily)?

KnVerey: Yes, exactly. If there aren't any failed/timed out pods, all pods should be pretty much equal for debugging purposes. The case where we'll definitely hit the `pods.first` fallback is where the timeout was triggered by the parent, not a pod.

lib/kubernetes-deploy/kubernetes_resource/pod.rb
KnVerey: The idea of the scaffold is to make sure the keys will be there by merging the scaffold with the data we actually receive. I guess that falls apart if we explicitly received something like `state: { running: nil}`, but I don't think that is a possibility. Is this too sketchy?

KnVerey:
Updated again. Please take a final look @kirs.

lib/kubernetes-deploy/kubernetes_resource/pod.rb
```diff
@@ -71,8 +72,13 @@ def failure_message
         red_name = ColorizedString.new(c.name).red
         "> #{red_name}: #{c.doom_reason}"
       end
-      intro = "The following containers are in a state that is unlikely to be recoverable:\n"
-      intro + container_messages.join("\n") + "\n"
+
+      intro = if unmanaged?
+        "The following containers encountered errors:"
```
KnVerey: The "unlikely to be recoverable" message doesn't really make sense for a dead run-once pod.

test/fixtures/long-running/jobs.yml.erb
kirs: It's not terrible. Maybe we could adjust the name of the fixture?

lib/kubernetes-deploy/kubernetes_resource/pod.rb
kirs: Don't spend too much time on it since it has a good integration coverage anyway 👌

KnVerey:
I'm going to ship this tomorrow.

dalehamel:
:clap:

PR #117, opened 2017-07-05
Set desired_replicas from API definition, not yaml

KnVerey:
@ibawt @n1koo @stefanmb This should fix the issue we ran into in v0.7.9 for some of our less cookie-cutter deployments.
cc @kirs 

`spec.replicas` can inherit a default value or otherwise be set out of band, so the value in the yaml file is not necessarily the actual desired replica count. The `.spec.replicas` value retrieved from the API is correct (I checked this on a couple of the deployments that were failing), and it is what HPAs update when they scale a deployment.

### DIFF
--- a/lib/kubernetes-deploy/kubernetes_resource/deployment.rb
+++ b/lib/kubernetes-deploy/kubernetes_resource/deployment.rb
@@ -9,6 +9,7 @@ module KubernetesDeploy
 
       if @found
         deployment_data = JSON.parse(raw_json)
+        @desired_replicas = deployment_data["spec"]["replicas"].to_i
         @latest_rs = find_latest_rs(deployment_data)
         @rollout_data = { "replicas" => 0 }.merge(deployment_data["status"]
           .slice("replicas", "updatedReplicas", "availableReplicas", "unavailableReplicas"))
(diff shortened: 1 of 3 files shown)


lib/kubernetes-deploy/kubernetes_resource/deployment.rb
```diff
@@ -35,8 +36,8 @@ def deploy_succeeded?
       return false unless @latest_rs
 
       @latest_rs.deploy_succeeded? &&
-      @latest_rs.desired_replicas == desired_replicas && # latest RS fully scaled up
-      @rollout_data["updatedReplicas"].to_i == @rollout_data["replicas"].to_i &&
```
KnVerey: In practice it seems like `.status.replicas` (aka `@rollout_data["replicas"].to_i`) pretty much always matches `.spec.replicas`. However, `.spec.replicas` seems to be more strictly the source of truth for desired replica count, so I think we should switch to it for this comparison as well (and the equivalent in the RS class below). A divergence in these values could explain how I still got a 1/200ish false positive rate for a certain previously-more-flakey test on the branch that became v1.7.9.

kirs:
👍

PR #118, opened 2017-07-06
Add StatsD monitoring

n1koo:
Barebones monitoring for various durations. Use `#{namespace}.deploy` as prefix so we end up with eg. `services-db.deploy.duration.succes`

- Monitor full duration of the deploy run
- Monitor various successes and tag this with resources related to it
- Report success/failure/timeout
- Add `sha` tag  

**Sucky**
Due to how we batch run (and monitor in intervals) its hard to get duration per resource without major refactoring.


dwradcliffe:
I think using the namespace as a tag rather than in the metric name might be better.

n1koo:
Sure, I was pretty 50/50 about it

lib/kubernetes-deploy/resource_watcher.rb
```diff
@@ -41,14 +41,19 @@ def run(delay_sync: 3.seconds, reminder_interval: 30.seconds)
     def report_what_just_happened(resources, watch_time)
       new_successes, new_failures = resources.partition(&:deploy_succeeded?)
       new_failures.each do |resource|
+        tags = %W(namespace:#{@namespace}, sha:#{ENV['REVISION']}, resource:#{resource.id})
```
dwradcliffe: Is this syntax right?

n1koo: Yeah it is, the resources get tangled up (eg. `#resources:ConfigMap/kube-shellService/memcachedDeployment/memcachedIngress/nginx`) because `,` is not supported though. I guess I could join without any delimiter since its not supported up anyways

n1koo: eg. ```[StatsD] measure kubernetes_deploy.duration.success:2200.0 #namespace: #sha:6795b97b6423a33d571c2359efa8399f4e6fa0de #resources:ConfigMap/kube-shellService/memcachedDeployment/memcachedIngress/nginx```

Though I'm somewhat pondering if this should be dropped altogether since due to the interval of checks whats going on we are not really super correct about duration

dwradcliffe: This has extra commas though: `["namespace:mynamespace,", "sha:876sd87f6asd78f6asd78f6ads78f6ads7f,", "resource:asdf"]`

n1koo: 🤦‍♂️  ah true, missed it because the statsd part strips it. Good catch

lib/kubernetes-deploy/resource_watcher.rb
```diff
@@ -41,14 +41,19 @@ def run(delay_sync: 3.seconds, reminder_interval: 30.seconds)
     def report_what_just_happened(resources, watch_time)
       new_successes, new_failures = resources.partition(&:deploy_succeeded?)
       new_failures.each do |resource|
+        tags = %W(namespace:#{@namespace} sha:#{ENV['REVISION']} resource:#{resource.id})
```
KnVerey: What are you hoping to get out of the resource tag? Could it be more informative to split it into `resource_name` and `resource_type` tags?

lib/kubernetes-deploy/runner.rb
```diff
@@ -131,6 +138,8 @@ def run(verify_result: true, allow_protected_ns: false, prune: true)
       success = false
     ensure
       @logger.print_summary(success)
+      tags = %W(namespace:#{@namespace} sha:#{ENV['REVISION']})
```
KnVerey: Actually, what is the value of the SHA tag? Wouldn't the idea of emitting metrics like this be to look over several deploys? (the duration info is already printed in the deploy output if you're talking about a single one).

lib/kubernetes-deploy/resource_watcher.rb
```diff
@@ -41,14 +41,19 @@ def run(delay_sync: 3.seconds, reminder_interval: 30.seconds)
...
         if resource.deploy_failed?
+          ::StatsD.increment('failure', 1, tags: tags)
```
KnVerey: I think we should make it more obvious which metrics deal with a whole deploy task and which are about subresources, while keeping what we emit about each somewhat parallel. Maybe `KubernetesDeploy.deploy.status` vs `KubernetesDeploy.deploy.resource.status` and have success/failure/timeout as tags? And then similarly `KubernetesDeploy.deploy.duration` etc. (The "deploy" part is because this gem can also do restarts and task runs) Wdyt?

lib/kubernetes-deploy/resource_watcher.rb
kirs: I'm not sure how REVISION/sha would be useful in the DD dashboard - did you include it for any specific reason?

lib/kubernetes-deploy/runner.rb
```diff
@@ -104,7 +105,9 @@ def run(verify_result: true, allow_protected_ns: false, prune: true)
 
       if deploy_has_priority_resources?(resources)
         @logger.phase_heading("Predeploying priority resources")
-        predeploy_priority_resources(resources)
+        ::StatsD.measure('duration.priority_resources', tags: %W(namespace:#{@namespace} sha:#{ENV['REVISION']})) do
```
kirs: I see that it measures the duration of predeploy resources and the duration of verification separately, but from the dashboard perspective I think that the most valuable metric would be the total deploy time. Should we measure it as well?

lib/kubernetes-deploy/runner.rb
```diff
@@ -113,11 +116,15 @@ def run(verify_result: true, allow_protected_ns: false, prune: true)
...
       if verify_result
-        deploy_resources(resources, prune: prune, verify: true)
+        ::StatsD.measure('duration.normal_resources', tags: %W(namespace:#{@namespace} sha:#{ENV['REVISION']})) do
+          deploy_resources(resources, prune: prune, verify: true)
+        end
         record_statuses(resources)
         success = resources.all?(&:deploy_succeeded?)
       else
-        deploy_resources(resources, prune: prune, verify: false)
+        ::StatsD.measure('duration.normal_resources', tags: %W(namespace:#{@namespace} sha:#{ENV['REVISION']})) do
```
kirs: When we start deploying shopify core to multiple clusters (contexts), all of them would be deployed to the same namespace ("shopify-core" or whatever). From what I see now that would result to mix up of metrics.
How do you think we should solve it? We could also include `context` into the tags, however I'm not sure if it's the best idea.

lib/kubernetes-deploy/statsd.rb
```diff
@@ -0,0 +1,20 @@
+# frozen_string_literal: true
+require 'statsd-instrument'
+
+module KubernetesDeploy
+  class StatsD
+    def self.build(logger)
+      ::StatsD.default_sample_rate = 1.0
+      ::StatsD.prefix = "kubernetes_deploy"
+
+      if !ENV['STATSD_DEV'].empty?
```
n1koo: 👍

exe/kubernetes-deploy
```diff
@@ -48,6 +47,7 @@ end
 namespace = ARGV[0]
 context = ARGV[1]
 logger = KubernetesDeploy::FormattedLogger.build(namespace, context, verbose_prefix: verbose_log_prefix)
+KubernetesDeploy::StatsD.build(logger)
```
n1koo: Usually logs and metrics go hand in hand so this seemed like a natural place. Dont feel strongly about it so i'll move to `kubernetes-deploy.rb`

lib/kubernetes-deploy/resource_watcher.rb
n1koo: Its useful of tracking down which deploy specifically was slow. We do the same in core so seems like a good pattern to follow

lib/kubernetes-deploy/resource_watcher.rb
```diff
@@ -41,14 +41,19 @@ def run(delay_sync: 3.seconds, reminder_interval: 30.seconds)
...
           @logger.error("#{resource.id} failed to deploy after #{watch_time}s")
         else
+          ::StatsD.increment('timeout', 1, tags: tags)
           @logger.error("#{resource.id} deployment timed out")
         end
       end
 
       if new_successes.present?
+        tags = %W(namespace:#{@namespace} sha:#{ENV['REVISION']} resources:#{new_successes.map(&:id).join(';')})
+        ::StatsD.measure('duration.success', watch_time * 1000, tags: tags)
```
n1koo: Yeah, the higher interval is more of a hack than solution. We would need to have real watcher for this to work ideally.

lib/kubernetes-deploy/runner.rb
n1koo: Explained in https://github.com/Shopify/kubernetes-deploy/pull/118#discussion_r126109263

n1koo: Also this was liberate since `@current_sha` was set in `runner`and I wanted to keep things identical in all calling places

lib/kubernetes-deploy/runner.rb
n1koo: Absolutely, its done in https://github.com/Shopify/kubernetes-deploy/pull/118/files#diff-ae1efd633a69ca93e2ad9f56702555ebR142

lib/kubernetes-deploy/runner.rb
n1koo: Yup, I thought about this but decided to leave it out of here since we already have `location` etc tags to make that difference

lib/kubernetes-deploy/statsd.rb
```diff
@@ -0,0 +1,20 @@
+# frozen_string_literal: true
+require 'statsd-instrument'
+
+module KubernetesDeploy
+  class StatsD
+    def self.build(logger)
+      ::StatsD.default_sample_rate = 1.0
+      ::StatsD.prefix = "kubernetes_deploy"
```
n1koo: 👍

n1koo:
> Would it work to centralize the instrumentation itself, following Shipit's example?

I love that example but unlike shipit we are drilling a little more deeper here. Eg. for just the global duration we could wrap, but things like specific task durations where we iterate over list not really. 

Doing both approaches (eg. some inline, some via middleware) would be confusing (?)

Gemfile
```diff
@@ -8,3 +8,4 @@ gem 'pry-byebug'
 gem 'rubocop'
 gem 'timecop'
 gem 'byebug'
+gem 'statsd-instrument'
```
KnVerey: Since this is a production dependency, it needs to go in `kubernetes-deploy.gemspec` instead

lib/kubernetes-deploy/resource_watcher.rb
KnVerey: > We can add the type if we want, though just having the failed name is should be enough

`id` is in the form `type/name`, so this is emitting both, but tacked together. My thinking is that if you emit those separately, you'd also be able to do things like see things like one type in general taking longer than others--e.g. Pods, which always have unique ids between deploys. Otoh names aren't necessarily unique, so perhaps the useful combination would be `id` and `type` as you said. I'm just speculating though.

lib/kubernetes-deploy/resource_watcher.rb
```diff
@@ -41,14 +41,23 @@ def run(delay_sync: 3.seconds, reminder_interval: 30.seconds)
     def report_what_just_happened(resources, watch_time)
       new_successes, new_failures = resources.partition(&:deploy_succeeded?)
       new_failures.each do |resource|
+        tags = %W(namespace:#{@namespace} sha:#{ENV['REVISION']} resource:#{resource.id})
         if resource.deploy_failed?
+          ::StatsD.increment('deploy.resource.status', 1, tags: tags << "status:failed")
```
KnVerey: Sorry, I just realized we're reporting this from the watcher, which is used by the other task types (restart and run) as well. Maybe the metric should just be `resource.status` then... wdyt @n1koo @kirs , if you're looking at that metric, is it a good or a confusing thing that you'd see data from how quickly it started during a restart task as well?

lib/kubernetes-deploy/resource_watcher.rb
```diff
@@ -41,14 +41,23 @@ def run(delay_sync: 3.seconds, reminder_interval: 30.seconds)
...
+          ::StatsD.increment('deploy.resource.status', 1, tags: tags << "status:timeout")
           @logger.error("#{resource.id} deployment timed out")
         end
       end
 
       if new_successes.present?
+        new_successes.each do |success|
+          tags = %W(namespace:#{@namespace} sha:#{ENV['REVISION']} resource:#{success.id})
+          ::StatsD.increment('deploy.resource.status', 1, tags: tags << "status:success")
+          ::StatsD.measure('deploy.resouce.duration', watch_time * 1000, tags: tags)
```
KnVerey: Are you intentionally reporting duration for successes only?

lib/kubernetes-deploy/runner.rb
```diff
@@ -113,11 +116,15 @@ def run(verify_result: true, allow_protected_ns: false, prune: true)
...
       if verify_result
-        deploy_resources(resources, prune: prune, verify: true)
+        ::StatsD.measure('duration.normal_resources', tags: %W(namespace:#{@namespace} sha:#{@current_sha})) do
+          deploy_resources(resources, prune: prune, verify: true)
+        end
         record_statuses(resources)
         success = resources.all?(&:deploy_succeeded?)
       else
-        deploy_resources(resources, prune: prune, verify: false)
+        ::StatsD.measure('duration.normal_resources', tags: %W(namespace:#{@namespace} sha:#{@current_sha})) do
```
KnVerey: I don't think we should measure in this case. If `verify_result` is false, we're skipping the polling step entirely, so it wouldn't be measuring any real deploy time (just the time of `kubectl apply` itself, basically).

lib/kubernetes-deploy/runner.rb
```diff
@@ -131,6 +138,8 @@ def run(verify_result: true, allow_protected_ns: false, prune: true)
       success = false
     ensure
       @logger.print_summary(success)
+      tags = %W(namespace:#{@namespace} sha:#{@current_sha})
```
KnVerey: This method repeatedly constructs the same tag set. Maybe make a method `statsd_tags` that returns this?

lib/kubernetes-deploy/statsd.rb
```diff
@@ -0,0 +1,22 @@
...
+require 'statsd-instrument'
+require 'logger'
+
+module KubernetesDeploy
+  class StatsD
+    def self.build
+      ::StatsD.default_sample_rate = 1.0
+      ::StatsD.prefix = "KubernetesDeploy"
+
+      if !ENV['STATSD_DEV'].empty?
```
KnVerey: We have pieces of ActiveSupport in this gem, including `.present?` FYI

KnVerey:
Can you please [run CI](buildkite.com/shopify/kubernetes-deploy-gem) after your next update?

> I love that example but unlike shipit we are drilling a little more deeper here. Eg. for just the global duration we could wrap, but things like specific task durations where we iterate over list not really.

Couldn't we instrument `report_what_just_happened` for that, like:
```ruby
def report_what_just_happened(resources, watch_time)
  resources.each do |r| 
    status = if r.deploy_succeeded?
     'success'
    ...
    # report metric
  end
  super
end
```

In any case, I won't insist. 😄

lib/kubernetes-deploy/resource_watcher.rb
n1koo: Collisions on names like `web` (ingress vs pod) are pound to happen but fair comment about type. That seems useful, just pondering the cost vs value at this point

lib/kubernetes-deploy/resource_watcher.rb
n1koo: I'd say changing the name back makes sense. I don't think having that data is confusing considering the tools name, usage pattern etc.

lib/kubernetes-deploy/resource_watcher.rb
n1koo: Yes, but maybe its more inline with the reporting in the logs if we also report it on failures

lib/kubernetes-deploy/runner.rb
n1koo: 👍

n1koo:
@KnVerey we definitely could. I was trying to say that we would need bunch of these wrappers for different things since `report_what_just_happened(resources, watch_time)` isn't the only place where we use it. So unlike shipit where it seems to be just wrapper around a single thing here we would end up with either bunch of or mixed bag. Also not a huge fan of latter solution because it would make any later additions in other places more complicated.

n1koo:
I think I covered all the feedback in latest commit

lib/kubernetes-deploy/resource_watcher.rb
```diff
@@ -42,13 +42,19 @@ def report_what_just_happened(resources, watch_time)
       new_successes, new_failures = resources.partition(&:deploy_succeeded?)
       new_failures.each do |resource|
         if resource.deploy_failed?
+          status = "failure"
```
KnVerey: Now that the classes are in charge of the reporting themselves, we could make this even simpler, i.e. right at the beginning of this method, `resources.each { |r| r.report_status_to_statsd(watch_time) }`. The resources themselves can easily figure out the status tag text.

PR #122, opened 2017-07-11
up the deployment/service timeout

ibawt:
## Wat
Until the progressDeadline thing is done, I'd like to up the timeout to prevent core from failing the deploy.  

At the moment we get a few failures a day where it's only a few replicas of the total set that are missing like 1 or 2.  

I think if we bump this timeout it should alleviate the errors and allow for some variance on 5minutes.

@dwradcliffe

### DIFF
--- a/lib/kubernetes-deploy/kubernetes_resource/deployment.rb
+++ b/lib/kubernetes-deploy/kubernetes_resource/deployment.rb
@@ -1,7 +1,7 @@
 # frozen_string_literal: true
 module KubernetesDeploy
   class Deployment < KubernetesResource
-    TIMEOUT = 5.minutes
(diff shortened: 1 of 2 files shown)


ibawt:
running tests locally now

ibawt:
man I can't get these tests to pass, and I'm not sure what's going on :S .

KnVerey:
> man I can't get these tests to pass, and I'm not sure what's going on :S .

Locally? They look good on CI. Locally the problem is typically that your minikube clock is out of sync.

ibawt:
hrm I did destroy and delete a bunch of times. I'll merge tomorrow and roll it out.

PR #123, opened 2017-07-11
Various StatsD improvements

n1koo:
Followup to: https://github.com/Shopify/kubernetes-deploy/pull/118

- Move from milliseconds to seconds on durations
- Report `context` on everything (eg. which cluster was deployed)
- Report status for the full deploy ( `status.total` )
- Report success/failure on the total metrics ( `"status:#{status}"`

### DIFF
--- a/lib/kubernetes-deploy/kubernetes_resource.rb
+++ b/lib/kubernetes-deploy/kubernetes_resource.rb
@@ -187,18 +187,8 @@ module KubernetesDeploy
 
     def report_status_to_statsd(watch_time)
       unless @statsd_report_done
-        status = if deploy_failed?
-          "failure"
-        elsif deploy_timed_out?
-          "timeout"
-        elsif deploy_succeeded?
-          "success"
-        else
-          "unknown"
-        end
-        tags = %W(namespace:#{namespace} resource:#{id} type:#{type} sha:#{ENV['REVISION']} status:#{status})
-        ::StatsD.measure('resource.duration', watch_time * 1000, tags: tags)
-        ::StatsD.increment('resource.status', 1, tags: tags)
+        ::StatsD.measure('resource.duration', watch_time, tags: statsd_tags)
+        ::StatsD.increment('resource.status', 1, tags: statsd_tags)
         @statsd_report_done = true
       end
     end
@@ -271,6 +261,19 @@ module KubernetesDeploy
       file
     ensure
       file.close if file
+    end
+
+    def statsd_tags
+      status = if deploy_failed?
+        "failure"
+      elsif deploy_timed_out?
+        "timeout"
+      elsif deploy_succeeded?
+        "success"
+      else
+        "unknown"
+      end
+      %W(context:#{context} namespace:#{namespace} resource:#{id} type:#{type} sha:#{ENV['REVISION']} status:#{status})
     end
   end
 end
--- a/lib/kubernetes-deploy/kubernetes_resource.rb
+++ b/lib/kubernetes-deploy/kubernetes_resource.rb
@@ -188,7 +188,6 @@ module KubernetesDeploy
     def report_status_to_statsd(watch_time)
       unless @statsd_report_done
         ::StatsD.measure('resource.duration', watch_time, tags: statsd_tags)
-        ::StatsD.increment('resource.status', 1, tags: statsd_tags)
         @statsd_report_done = true
       end
     end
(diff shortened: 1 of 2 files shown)


lib/kubernetes-deploy/kubernetes_resource.rb
```diff
@@ -185,6 +185,10 @@ def kubectl
       @kubectl ||= Kubectl.new(namespace: @namespace, context: @context, logger: @logger, log_failure_by_default: false)
     end
 
+    def statsd_tags
+      %W(context:#{context} namespace:#{namespace} resource:#{id} type:#{type} sha:#{ENV['REVISION']} status:#{status})
```
KnVerey: The "status" available here isn't the one you want; you need to extract the conditional too. 

Nit: can you move this to the `private` section?

lib/kubernetes-deploy/runner.rb
```diff
@@ -136,7 +136,9 @@ def run(verify_result: true, allow_protected_ns: false, prune: true)
       success = false
     ensure
       @logger.print_summary(success)
-      ::StatsD.measure('duration.total', (Time.now.utc - start) * 1000.0, tags: statsd_tags)
+      status = success ? "success" : "failed"
+      ::StatsD.increment('status.total', 1, tags: statsd_tags << "status:#{status}")
+      ::StatsD.measure('duration.total', (Time.now.utc - start), tags: statsd_tags << "status:#{status}")
```
KnVerey: When exploring this metric in practice I got momentarily confused by `total`--it sounds like it is summing something. Wdyt about changing it to `full_deploy` or `all_resources` or `overall` or something like that?

KnVerey:
Another idea I just had while playing in Datadog: what about sending a single `context/namespace` tag, like how we're sending `resource:type/name`? What we're looking for is an identifier, right? If so, I think that'd work better.

n1koo:
Single is probably fine... I'm slightly worried it being hard to combine to app level dashes where they are separate dash variables. WDYT?

lib/kubernetes-deploy/kubernetes_resource.rb
n1koo: 🤦‍♂️

lib/kubernetes-deploy/runner.rb
n1koo: `all_resources` seems good 👍

lib/kubernetes-deploy/runner.rb
```diff
@@ -105,9 +105,9 @@ def run(verify_result: true, allow_protected_ns: false, prune: true)
 
       if deploy_has_priority_resources?(resources)
         @logger.phase_heading("Predeploying priority resources")
-        ::StatsD.measure('duration.priority_resources', tags: statsd_tags) do
-          predeploy_priority_resources(resources)
-        end
+        start_priority_resource = Time.now.utc
+        predeploy_priority_resources(resources)
+        ::StatsD.measure('duration.priority_resources', (Time.now.utc - start_priority_resource), tags: statsd_tags)
```
KnVerey: Should we round this number (and the other two)? If we don't we have the slight weirdness that the individual resource times can appear to be longer than this time because of the extra resolution here. 

```
I, [2017-07-12T14:09:52.499086 #10373]  INFO -- : [StatsD] measure KubernetesDeploy.resource.duration:0.3 #context:minikube #namespace:k8sdeploy-test-deploy-result-logging-for-mixed-dcae97ad348d6836 #resource:ConfigMap/hello-cloud-configmap-data #type:ConfigMap #sha: #status:success
I, [2017-07-12T14:09:52.499495 #10373]  INFO -- : [StatsD] measure KubernetesDeploy.duration.priority_resources:0.270041 #namespace:k8sdeploy-test-deploy-result-logging-for-mixed-dcae97ad348d6836 #sha:5698f34b7826 #context:minikube
```

KnVerey:
>  I'm slightly worried it being hard to combine to app level dashes where they are separate dash variables. WDYT?

Oh, I just understood what you meant. That's a fair point at least for how we're generating those dashes today. I'm fine either way.

lib/kubernetes-deploy/runner.rb
n1koo: Sure

KnVerey:
Since @n1koo is done for the day and I'd love to get this shipped, I made the changes we just agreed to:
- rounded all the `measures` to the same resolution
- removed the `increment` since they're just duplicating data at this point
- fixed up the `measures` to have consistent names (`{resource_group}.duration`--previously we also had `duration.{resource_group}`).

@dwradcliffe could you take a look?

Here's a 🎩  screenshot:

image

PR #124, opened 2017-07-12
Fail fast on image pull back off too

KnVerey:
Apparently if we miss the state where one of the "not found" errors is printed, the message will start saying "Back-off pulling image" and the deploy will still hit the 5 min timeout. Here's the pod status from a deploy that reached the timeout after my doom detection PR shipped:

image

@ibawt it is possible we'd enter this state after getting a few of the transient errors that you pointed out we shouldn't bail on. IMO it this change would alleviate more pain than it would cause though. WDYT?


kirs:
Does it mean that the k8s changes the state from "image not found" to "Back-off pulling image" after some time?

ibawt:
fwiw the frequency of the errors is drastically less now.  So I think you should merge and we can revert it's a problem for core or figure a better solution.

KnVerey:
> Does it mean that the k8s changes the state from "image not found" to "Back-off pulling image" after some time?

It seems that the container status message initially references the specific reason the image could not be pulled (the unit tests capture the variations I've seen), but once it has happened multiple times, it simply says "Back-off pulling image".

PR #126, opened 2017-07-12
Update tagging instructions. Document STATSD_DEV var.

KnVerey:
@ibawt @kirs I tried this for the 1.9.1 release and it worked great.

### DIFF
--- a/README.md
+++ b/README.md
@@ -257,6 +257,8 @@ To install this gem onto your local machine, run `bundle exec rake install`.
 
 To see the full-color output of a specific integration test, you can use `PRINT_LOGS=1 bundle exec ruby -I test test/integration/kubernetes_deploy_test.rb -n/test_name/`.
 
+To make StatsD log what it would have emitted, run a test with `STATSD_DEV=1`.
+
 
 
 ![test-output](screenshots/test-output.png)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -265,11 +267,12 @@ To see the full-color output of a specific integration test, you can use `PRINT_
 
 ## Releasing a new version (Shopify employees)
 
-1. Update the version number in `version.rb` and commit that change alone to master.
-2. Tag the version with `git tag vx.y.z && git push --tags`
-3. Use the [Shipit Stack](https://shipit.shopify.io/shopify/kubernetes-deploy/rubygems) to build the `.gem` file and upload to [rubygems.org](https://rubygems.org/gems/kubernetes-deploy).
+1. Update the version number in `version.rb` and commit that change with message "Version x.y.z". Don't push yet or you'll confuse Shipit.
+2. Tag the version with `git tag vx.y.z -a -m "Version x.y.z"`
```
kirs: I should have mentioned that earlier, but working on gems most of developers use `rake release`, which does everything: creates and pushes the tag and uploads new version to rubygems. You only need to bump `version.rb`.

I'm not sure that with `rake release` describing these steps is necessary.

KnVerey: We're using Shipit though. Nobody should do `rake release` locally anymore. Only you and I still have the permissions to do so.

kirs: Sorry, I somehow missed the point when we stopped doing `rake release` and started using Shipit.

Looking at how Shipit [releases the gem](https://shipit.shopify.io/shopify/kubernetes-deploy/rubygems/deploys/464645), I see that it takes care of creating and pushing the tag. Why it it necessary to do it by hands as suggested in step 2 of the README?

ibawt: oh hrm, we should do the minimal amount of things if shipit does it.

KnVerey: It doesn't create it for you. Check out this deploy for an example of what happens when you don't create it first (or push the commit before the tag, which messes up Shipit's cache): https://shipit.shopify.io/shopify/kubernetes-deploy/rubygems/deploys/464644

kirs: I see, so for some good reason Shipit has this check for exiting tag.

kirs: Good catch, Katrina!

PR #129, opened 2017-07-13
Handle rare unstable deploy result

KnVerey:
Since shipping the doom detection PR, I've seen two really strange deploy results, where the deploy says it failed but has no failure reason and lists all the resources as successes. In [the clearest case](https://shipit.shopify.io/shopify/services-db/production/deploys/463168) it seems the new sync added to make sure we print the latest info in the summary failed (my theory: related to the master upgrade we did around then), resetting many resource statuses to Unknown. In the other, two of the supposedly "successful" resources list unavailable pods.

It is unavoidably possible for the resources to go from fully up to not fully up between last poll and status output. This is a consequence of how Kubernetes works, and not necessarily indicative of a problem. Instead of restarting the polling in this case, IMO we should stick to our original conclusion re:success and pass the deploy. As a rule, the "status" we print should reflect the condition used to determine that success, without being overly specific.

This PR removes the additional summary-time sync and adjusts `Service` to print a less specific status.

@Shopify/cloudplatform @n1koo @dwradcliffe  - The issue this fixes could affect core deploy stability. I'm on vacation tomorrow. Feel free to review and merge without me.

### DIFF
--- a/lib/kubernetes-deploy/deferred_summary_logging.rb
+++ b/lib/kubernetes-deploy/deferred_summary_logging.rb
@@ -71,7 +71,7 @@ module KubernetesDeploy
       # Example:
       # # The resulting summary will begin with "Created 3 secrets and failed to deploy 2 resources"
       # @logger.summary.add_action("created 3 secrets")
-      # @logger.summary.add_cation("failed to deploy 2 resources")
+      # @logger.summary.add_action("failed to deploy 2 resources")
       def add_action(sentence_fragment)
         @actions_taken << sentence_fragment
       end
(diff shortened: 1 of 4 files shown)


kirs:
I'm not sure where `Selects at least 1 pod` would pop up. In the example that you linked the status is "not found":

screen shot 2017-07-14 at 14 03 19

KnVerey:
> If I get it right, removing successful_resources.map(&:sync) should help to avoid the weird state, and the rest of changes are just cosmetic.

Correct. The reason that sync was introduced was to make things like Service that finish way before others display a more accurate result in the summary. But it's better to make them just display something more vague that is tied to the success condition itself.

Issue #63, opened 2017-03-09
Enable per-resource timeout overrides via an annotation


KnVerey:
Originally suggested in https://github.com/Shopify/kubernetes-deploy/pull/60. See also https://github.com/Shopify/kubernetes-deploy/pull/62. If this pattern works nicely, it could come in handy for other types of overrides (e.g. priority resource identification) as well.

cc @kirs @sirupsen @wvanbergen @DazWorrall

sirupsen:
✨

KnVerey:
Actually, for Deployments specifically (probably what people want to override most of the time?) a more sophisticated version of this already exists as `progressDeadlineSeconds`: 

<image src="https://screenshot.click/2017-03-15--193323_6936z-hjx63.png" width="600px">

The future automatic rollback aspect scares me, as a rollback of a single resource within a Shipit deploy would be really bizarre UX. Hopefully that will be configurable at a level we can control with the deployment script.

In any case, when this parameter is present, we should probably time out based only on the presence of the related status condition.

karanthukral:
Looking into implementing override for deployments using `Progress Deadline Seconds`

PR #130, opened 2017-07-14
Add support for progress conditions on deployments

karanthukral:
## What?
- Resolves https://github.com/Shopify/kubernetes-deploy/issues/63
- Adds support for `progressDeadlineSeconds` for deployments
- If a deployment template has `spec.progressDeadlineSeconds` defined, the deployment class will use the `Progressing` condition to evaluate timeout instead of the one defined for the class

## What needs to be improved?
- ~~The test is still a work in progress~~
- ~~In order to replicate the case of overriding the timeout and causing a deploy to fail, I tried to use an invalid image, or an image with an invalid entrypoint. In either case, kubernetes-deploy fails the deploy due to detecting unrecoverable states.~~
- ~~The current test attempts to deploy 10 nginx containers with a 2 second `progressDeadlineSeconds` with a `minReadySeconds` set to 1 second. This causes the deploy to time out.~~

- Fixed the test by creating a simple busy box container that simply sleeps and the test deployment adds a readiness probe to run `ls` with a `progressDeadlineSeconds` of 2 seconds which causes the deployment to timeout.

- Open to suggestions if there is a better way to write this test


karanthukral:
cc/ @Shopify/cloudplatform

karanthukral:
The following events keep the deployment "progressing"
```
Kubernetes marks a Deployment as progressing when one of the following tasks is performed:
- The Deployment creates a new ReplicaSet.
- The Deployment is scaling up its newest ReplicaSet.
- The Deployment is scaling down its older ReplicaSet(s).
- New Pods become ready or available (ready for at least MinReadySeconds).
```

https://kubernetes.io/docs/concepts/workloads/controllers/deployment/

lib/kubernetes-deploy/kubernetes_resource/deployment.rb
```diff
@@ -14,6 +14,9 @@ def sync
         @rollout_data = { "replicas" => 0 }.merge(deployment_data["status"]
           .slice("replicas", "updatedReplicas", "availableReplicas", "unavailableReplicas"))
         @status = @rollout_data.map { |state_replicas, num| "#{num} #{state_replicas.chop.pluralize(num)}" }.join(", ")
+        deployment_data["status"]["conditions"].each do |condition|
+          @progress = condition["type"] == 'Progressing' ? condition : nil
```
KnVerey: This will set `@progress` to `nil` if the Progressing condition isn't last.

lib/kubernetes-deploy/kubernetes_resource/deployment.rb
```diff
@@ -54,13 +57,33 @@ def timeout_message
...
+        super || @latest_rs && @latest_rs.deploy_timed_out?
+      end
     end
 
     def exists?
       @found
     end
 
+    def timeout
+      raw_json, _err, st = kubectl.run("get", type, @name, "--output=json")
```
KnVerey: Since there isn't a default value for this (i.e. if it isn't set in the template, it won't be set ever), we can look at `@definition["spec"]["progressDeadlineSeconds"]` for this instead of querying the API. As a rule, I try to keep all API queries inside the `sync` method. This reduces the number of overall calls (since e.g. `deploy_succeeded?` might get called a bunch of times in a row), and also makes sure we keep a consistent view of the cluster within a given polling cycle.

EDIT: In the end I don't think we should override this anyway (see my other comment).

lib/kubernetes-deploy/kubernetes_resource/deployment.rb
```diff
@@ -54,13 +57,33 @@ def timeout_message
     end
 
     def deploy_timed_out?
-      super || @latest_rs && @latest_rs.deploy_timed_out?
+      if @progress
+        @progress["status"] == 'False'
+      else
+        super || @latest_rs && @latest_rs.deploy_timed_out?
```
KnVerey: If they had `progressDeadlineSeconds` in their spec but the status condition is missing, we'd fail the deploy with a very short timeout. That's probably not what we want. I'm thinking what you have right here is correct--if the status condition failed, time out without bothering to look at the actual progressDeadlineSeconds number. However, IMO the class-level hard timeout that kicks in if the condition is missing should retain its normal definition.

test/fixtures/misc-templates/deployment-w-progress.yml
```diff
@@ -0,0 +1,29 @@
+apiVersion: extensions/v1beta1
```
KnVerey: Would the test/fixtures/long-running/undying-deployment.yml.erb fixture work for this? It just sleeps, and you can modify it to add the `progressDeadlineSeconds` by using the block form of `deploy_fixtures`.

lib/kubernetes-deploy/kubernetes_resource/deployment.rb
karanthukral: I am not entirely sure what you mean here? The current logic falls back to the default logic if the progress condition does not exist. Isn't that what we want?

test/fixtures/misc-templates/deployment-w-progress.yml
karanthukral: So I tried this and was able to add the `progressDeadlineSeconds` into the spec but adding a `readinessProbe` is not as easy. We need a `readinessProbe` so that the container fails the ready check causing the timeout

karanthukral:
@KnVerey made the changes you recommended except the one question. The tests started failing with:
```
KubernetesDeployTest#test_deploy_result_logging_for_mixed_result_deploy:
  | RuntimeError: can't modify frozen String
```
Have you ever seen that before or have any ideas why this could be happening after I wrote the `timeout_message` method in `deployment.rb`

lib/kubernetes-deploy/kubernetes_resource/deployment.rb
```diff
@@ -14,6 +14,11 @@ def sync
         @rollout_data = { "replicas" => 0 }.merge(deployment_data["status"]
           .slice("replicas", "updatedReplicas", "availableReplicas", "unavailableReplicas"))
         @status = @rollout_data.map { |state_replicas, num| "#{num} #{state_replicas.chop.pluralize(num)}" }.join(", ")
+        deployment_data["status"]["conditions"].each do |condition|
```
KnVerey: I think this could be shortened to:
`@progress = deployment_data["status"]["conditions"].find { |condition| condition['type'] == 'Progressing' }`

lib/kubernetes-deploy/kubernetes_resource/deployment.rb
KnVerey: Sorry, I probably should have commented on `def timeout` instead. I was trying to say we should remove that override so that it always uses `TIMEOUT`. Here's an example of the scenario I'm worried about:
- spec.progressDeadlineSeconds is set to a low value, e.g. 30s
- the deployment has a ton of pods and a conservative rollout strategy, so it takes, say, 5min to finish
- partway through, for whatever reason, we don't observe the status condition (so `@progress` becomes nil)
- `deploy_timed_out?` sees `@progress == nil` and uses `super`, which thinks the timeout is 30s
- deploy times out even though it was still progressing, which is what we're trying to avoid with this feature

Of course landing in the `super` when progressDeadlineSeconds was set would be unexpected. But what I'm trying to say is if that happens for some reason, I think the sane behaviour would be to use the global timeout.

test/fixtures/misc-templates/deployment-w-progress.yml
KnVerey: Oh true, I probably should have suggested `test/fixtures/invalid/bad_probe.yml` instead.

lib/kubernetes-deploy/kubernetes_resource/deployment.rb
karanthukral: I'll remove the `def timeout` override 👍

karanthukral:
cc/ @ibawt

ibawt:
code LGTM, I'll just have to whip this on core and see what happens I think.

ibawt:
cc @dwradcliffe this is how we can avoid bumping the global timeout for deployments/services

karanthukral:
@KnVerey can I get a 👍  on the final changes?

PR #132, opened 2017-07-17
Add Service Account resource

karanthukral:
## What?
- Add a simple resource to track deploying a service account
- The resource follows a similar functionality as the [ingress resource](https://github.com/Shopify/kubernetes-deploy/blob/master/lib/kubernetes-deploy/kubernetes_resource/ingress.rb)


lib/kubernetes-deploy/kubernetes_resource/service_account.rb
```diff
@@ -0,0 +1,23 @@
...
+      @found = st.success?
+    end
+
+    def deploy_succeeded?
+      exists?
+    end
+
+    def deploy_failed?
+      false
+    end
```
KnVerey: We should probably use 
```ruby
    def timeout_message
      UNUSUAL_FAILURE_MESSAGE
    end
```
here (like for configmap) since this shouldn't ever time out

test/integration/kubernetes_deploy_test.rb
```diff
@@ -11,7 +11,7 @@ def test_full_hello_cloud_set_deploy_succeeds
       "Deploying ConfigMap/hello-cloud-configmap-data (timeout: 30s)",
       "Hello from Docker!", # unmanaged pod logs
       "Result: SUCCESS",
-      "Successfully deployed 12 resources"
+      "Successfully deployed 13 resources"
```
KnVerey: Can you please also add an assertion about this resource to [HelloCloud#assert_all_up](https://github.com/Shopify/kubernetes-deploy/blob/master/test/helpers/fixture_sets/hello_cloud.rb#L9-L18)?

test/helpers/fixture_set.rb
```diff
@@ -120,6 +120,15 @@ def assert_secret_present(secret_name, expected_data = nil, type: 'Opaque', mana
       assert_equal expected_data, secret_data
     end
 
+    def assert_service_account_present(name)
+      service_accounts = kubeclient.get_service_accounts(namespace: namespace)
+      found = false
+      service_accounts.each do |sa|
+        found = true if sa.metadata.name == name
+      end
+      assert found
```
KnVerey: Suggestion: 
```ruby
desired = service_accounts.find { |sa| sa.metadata.name == name }
assert desired.present?, "Service account #{name} does not exist"
```

lib/kubernetes-deploy/kubernetes_resource/service_account.rb
```diff
@@ -0,0 +1,27 @@
+# frozen_string_literal: true
+module KubernetesDeploy
+  class ServiceAccount < KubernetesResource
+    TIMEOUT = 30.seconds
+
+    def sync
+      _, _err, st = kubectl.run("get", type, @name, "--output=json")
+      @found = st.success?
```
KnVerey: Sorry I didn't notice last time, but we should set `@status` here too. I guess what Ingress has here would work fine for this case too: `@status = st.success? ? "Created" : "Unknown"`

That'll fix the summary:

image

PR #133, opened 2017-07-18
Add Daemon Set resource

karanthukral:
## What?
- Adds daemon set as a kubernetes resource
- Replicates logic built into the Replica Set resource [here](https://github.com/Shopify/kubernetes-deploy/blob/master/lib/kubernetes-deploy/kubernetes_resource/replica_set.rb)
- Adds a test for successful and timed out deploy for a daemon set
- The timeout currently is the same as a replica set

cc/ @Shopify/cloudplatform

### DIFF
--- a/lib/kubernetes-deploy/kubernetes_resource/daemon_set.rb
+++ b/lib/kubernetes-deploy/kubernetes_resource/daemon_set.rb
@@ -1,0 +1,95 @@
+# frozen_string_literal: true
+module KubernetesDeploy
+  class DaemonSet < KubernetesResource
+    TIMEOUT = 5.minutes
+
+    def sync
+      raw_json, _err, st = kubectl.run("get", type, @name, "--output=json")
+      @found = st.success?
+
+      if @found
+        daemonset_data = JSON.parse(raw_json)
+        @current_generation = daemonset_data["metadata"]["generation"]
+        @observed_generation = daemonset_data["status"]["observedGeneration"]
+        @rollout_data = daemonset_data["status"]
+          .slice("currentNumberScheduled", "desiredNumberScheduled", "numberReady", "numberAvailable")
+        @status = @rollout_data.map { |state_replicas, num| "#{num} #{state_replicas}" }.join(", ")
+        @pods = find_pods(daemonset_data)
+      else # reset
+        @rollout_data = { "currentNumberScheduled" => 0 }
+        @current_generation = 1 # to make sure the current and observed generations are different
+        @observed_generation = 0
+        @status = nil
+        @pods = []
+      end
+    end
+
+    def deploy_succeeded?
+      @rollout_data["desiredNumberScheduled"].to_i == @rollout_data["currentNumberScheduled"].to_i &&
+      @rollout_data["desiredNumberScheduled"].to_i == @rollout_data["numberAvailable"].to_i &&
+      @current_generation == @observed_generation
+    end
+
+    def deploy_failed?
+      @pods.present? && @pods.any?(&:deploy_failed?)
+    end
+
+    def failure_message
+      @pods.map(&:failure_message).compact.uniq.join("\n")
+    end
+
+    def timeout_message
+      @pods.map(&:timeout_message).compact.uniq.join("\n")
+    end
+
+    def deploy_timed_out?
+      super || @pods.present? && @pods.any?(&:deploy_timed_out?)
+    end
+
+    def exists?
+      @found
+    end
+
+    def fetch_events
+      own_events = super
+      return own_events unless @pods.present?
+      most_useful_pod = @pods.find(&:deploy_failed?) || @pods.find(&:deploy_timed_out?) || @pods.first
+      own_events.merge(most_useful_pod.fetch_events)
+    end
+
+    def fetch_logs
+      most_useful_pod = @pods.find(&:deploy_failed?) || @pods.find(&:deploy_timed_out?) || @pods.first
+      most_useful_pod.fetch_logs
+    end
+
+    private
+
+    def find_pods(ds_data)
+      label_string = ds_data["spec"]["selector"]["matchLabels"].map { |k, v| "#{k}=#{v}" }.join(",")
+      raw_json, _err, st = kubectl.run("get", "pods", "-a", "--output=json", "--selector=#{label_string}")
+      return [] unless st.success?
+
+      all_pods = JSON.parse(raw_json)["items"]
+      current_generation = ds_data["metadata"]["generation"]
+
+      latest_pods = all_pods.find_all do |pods|
+        pods["metadata"]["ownerReferences"].any? { |ref| ref["uid"] == ds_data["metadata"]["uid"] } &&
+        pods["metadata"]["labels"]["pod-template-generation"].to_i == current_generation.to_i
+      end
+      return unless latest_pods.present?
+
+      latest_pods.each_with_object([]) do |pod_data, relevant_pods|
+        pod = Pod.new(
+          namespace: namespace,
+          context: context,
+          definition: pod_data,
+          logger: @logger,
+          parent: "#{@name.capitalize} daemon set",
+          deploy_started: @deploy_started
+        )
+        pod.sync(pod_data)
+        relevant_pods << pod
+      end
+    end
+  end
+end
--- a/test/fixtures/hello-cloud/daemon_set.yml
+++ b/test/fixtures/hello-cloud/daemon_set.yml
@@ -1,0 +1,16 @@
+apiVersion: extensions/v1beta1
+kind: DaemonSet
+metadata:
+  name: nginx
+spec:
+  template:
+    metadata:
+      labels:
+        app: hello-cloud
+        name: nginx
+    spec:
+      containers:
+      - name: nginx
+        image: nginx
+        ports:
+        - containerPort: 80
--- a/test/helpers/fixture_set.rb
+++ b/test/helpers/fixture_set.rb
@@ -126,6 +126,13 @@ module FixtureSetAssertions
       assert desired.present?, "Service account #{name} does not exist"
     end
 
+    def assert_daemon_set_present(name)
+      labels = "name=#{name},app=#{app_name}"
+      daemon_sets = v1beta1_kubeclient.get_daemon_sets(namespace: namespace, label_selector: labels)
+      desired = daemon_sets.find { |ds| ds.metadata.name == name }
+      assert desired.present?, "Daemon set #{name} does not exist"
+    end
+
     def assert_annotated(obj, annotation)
       annotations = obj.metadata.annotations.to_h.stringify_keys
       assert annotations.key?(annotation), "Expected secret to have annotation #{annotation}, but it did not"
--- a/test/integration/kubernetes_deploy_test.rb
+++ b/test/integration/kubernetes_deploy_test.rb
@@ -11,13 +11,14 @@ class KubernetesDeployTest < KubernetesDeploy::IntegrationTest
       "Deploying ConfigMap/hello-cloud-configmap-data (timeout: 30s)",
       "Hello from Docker!", # unmanaged pod logs
       "Result: SUCCESS",
-      "Successfully deployed 13 resources"
+      "Successfully deployed 14 resources"
     ], in_order: true)
 
     assert_logs_match_all([
       %r{ReplicaSet/bare-replica-set\s+1 replica, 1 availableReplica, 1 readyReplica},
       %r{Deployment/web\s+1 replica, 1 updatedReplica, 1 availableReplica},
-      %r{Service/web\s+Selects at least 1 pod}
+      %r{Service/web\s+Selects at least 1 pod},
+      %r{DaemonSet/nginx\s+1 currentNumberScheduled, 1 desiredNumberScheduled, 1 numberReady, 1 numberAvailable}
     ])
   end
 
@@ -49,9 +50,10 @@ class KubernetesDeployTest < KubernetesDeploy::IntegrationTest
       'pod "unmanaged-pod-',
       'service "web"',
       'deployment "web"',
-      'ingress "web"'
+      'ingress "web"',
+      'daemonset "nginx"'
     ] # not necessarily listed in this order
-    expected_msgs = [/Pruned 5 resources and successfully deployed 3 resources/]
+    expected_msgs = [/Pruned 6 resources and successfully deployed 3 resources/]
     expected_pruned.map do |resource|
       expected_msgs << /The following resources were pruned:.*#{resource}/
     end
@@ -630,6 +632,26 @@ class KubernetesDeployTest < KubernetesDeploy::IntegrationTest
     end
     assert_equal false, success, "Deploy succeeded when it was expected to fail"
     assert_logs_match("Unmanaged pods like Pod/oops-it-is-static must have unique names on every deploy")
+  end
+
+  def test_bad_container_on_daemon_sets_fails
+    success = deploy_fixtures("hello-cloud", subset: ["daemon_set.yml"]) do |fixtures|
+      daemon_set = fixtures['daemon_set.yml']['DaemonSet'].first
+      container = daemon_set['spec']['template']['spec']['containers'].first
+      container["image"] = "busybox"
+      container["command"] = ["ls", "/not-a-dir"]
+    end
+
+    refute success
+    assert_logs_match_all([
+      "DaemonSet/nginx: FAILED",
+      "nginx: Crashing repeatedly (exit 1). See logs for more information.",
+      "Final status: 1 currentNumberScheduled, 1 desiredNumberScheduled, 0 numberReady",
+      "Events (common success events excluded):",
+      "BackOff: Back-off restarting failed container",
+      "Logs from container 'nginx' (last 250 lines shown):",
+      "ls: /not-a-dir: No such file or directory"
+    ], in_order: true)
   end
 
   private
--- a/lib/kubernetes-deploy/runner.rb
+++ b/lib/kubernetes-deploy/runner.rb
@@ -20,6 +20,7 @@ require 'kubernetes-deploy/kubernetes_resource'
   pod_disruption_budget
   replica_set
   service_account
+  daemon_set
 ).each do |subresource|
   require "kubernetes-deploy/kubernetes_resource/#{subresource}"
 end
--- a/test/helpers/fixture_sets/hello_cloud.rb
+++ b/test/helpers/fixture_sets/hello_cloud.rb
@@ -15,6 +15,7 @@ module FixtureSetAssertions
       assert_poddisruptionbudget
       assert_bare_replicaset_up
       assert_all_service_accounts_up
+      assert_daemon_set_up
     end
 
     def assert_unmanaged_pod_statuses(status, count = 1)
@@ -83,6 +84,10 @@ module FixtureSetAssertions
 
     def assert_all_service_accounts_up
       assert_service_account_present("build-robot")
+    end
+
+    def assert_daemon_set_up
+      assert_daemon_set_present("nginx")
     end
   end
 end


karanthukral:
I have another branch ready for stateful sets but I'll wait for review on this before opening that one since they might have similar comments

lib/kubernetes-deploy/kubernetes_resource/daemon_set.rb
```diff
@@ -0,0 +1,99 @@
...
+      return own_events unless @pods.present?
+      most_useful_pod = @pods.find(&:deploy_failed?) || @pods.find(&:deploy_timed_out?) || @pods.first
+      own_events.merge(most_useful_pod.fetch_events)
+    end
+
+    def fetch_logs
+      container_names.each_with_object({}) do |container_name, container_logs|
+        out, _err, _st = kubectl.run(
+          "logs",
+          id,
```
KnVerey: This command isn't implemented for daemonsets unfortunately.

```
༶ kubectl logs ds/dd-agent                                                                                                                                                                                                                                                                      
error: cannot get the logs from extensions/__internal, Kind=DaemonSet
```

Ref  https://github.com/kubernetes/kubernetes/pull/40927, which implemented it for deployments, jobs and statefulsets. We'll need to pick a pod ourselves and use its logs. The `most_useful_pod` logic from `fetch_events` is probably good enough for now, though it isn't as smart as what kubectl does. Note that there are tradeoffs to prioritizing failing pods, which I discuss in https://github.com/Shopify/kubernetes-deploy/issues/138.

lib/kubernetes-deploy/kubernetes_resource/daemon_set.rb
```diff
@@ -0,0 +1,99 @@
...
+          "--since-time=#{@deploy_started.to_datetime.rfc3339}",
+          "--tail=#{LOG_LINE_COUNT}"
+        )
+        container_logs[container_name] = out.split("\n")
+      end
+    end
+
+    private
+
+    def unmanaged?
```
KnVerey: This concept doesn't apply to DaemonSets--they aren't generated by other resources.

test/fixtures/hello-cloud/daemon_set.yml
```diff
@@ -0,0 +1,16 @@
...
+metadata:
+  name: busybox
+spec:
+  template:
+    metadata:
+      labels:
+        app: busybox
+    spec:
+      containers:
+      - name: busybox
```
KnVerey: nit: Doesn't make a difference since this won't come close to hitting 40s, but I'd use nginx here like we do to emulate web, since Daemonsets aren't expected to ever complete.

test/helpers/fixture_set.rb
```diff
@@ -126,6 +126,15 @@ def assert_service_account_present(name)
       assert desired.present?, "Service account #{name} does not exist"
     end
 
+    def assert_daemon_set_present(name)
+      found = false
+      daemon_sets = v1beta1_kubeclient.get_daemon_sets(namespace: namespace, label_selector: "app=#{name}")
+      daemon_sets.each do |ds|
```
KnVerey: I'd suggest the same thing as we went with above for service accounts (`.find` + message on the assertion)

test/fixtures/hello-cloud/daemon_set.yml
```diff
@@ -0,0 +1,16 @@
+apiVersion: extensions/v1beta1
+kind: DaemonSet
+metadata:
+  name: busybox
+spec:
+  template:
+    metadata:
+      labels:
+        app: busybox
```
KnVerey: Nit: the "app" label for this fixture set is usually set to "hello-cloud"

test/helpers/fixture_set.rb
```diff
@@ -126,6 +126,15 @@ def assert_service_account_present(name)
...
+        found = true if ds.metadata.name == name
+      end
+      assert found
```
KnVerey: We should also have an availability assertion like we do for deployments and replicasets.

test/integration/kubernetes_deploy_test.rb
```diff
@@ -550,6 +550,32 @@ def test_success_detection_tolerates_out_of_band_deployment_scaling
     assert_equal true, success, "Failed to deploy deployment with dynamic replica count"
   end
 
+  def test_successful_daemon_set_deploy
+    assert deploy_fixtures("hello-cloud", subset: ["daemon_set.yml"])
+
+    assert_logs_match_all([
+      'Successfully deployed 1 resource',
+      '1 currentNumberSchedule, 1 desiredNumberSchedule, 1 numberRead'
```
KnVerey: Since these tests are expensive, I'd recommend adding this assertion to the group at L17 (including a resource type prefix like the others there have) instead of having a separate test.

test/integration/kubernetes_deploy_test.rb
```diff
@@ -550,6 +550,32 @@ def test_success_detection_tolerates_out_of_band_deployment_scaling
...
+    ])
+  end
+
+  def test_timed_out_daemon_set_deploy
```
KnVerey: Since the timeout has not special logic for this resource type, I'd rather swap this out for a test of a failure scenario. There are a few deployment tests you can use as a starting point (no need to duplicate all of them, since they're ultimately using the same pod logic). One thing it is important to show is that we're successfully fetching events from both the DS and a pod, as well as logs.

lib/kubernetes-deploy/kubernetes_resource/daemon_set.rb
```diff
@@ -0,0 +1,99 @@
...
+    end
+
+    def find_pods(ds_data)
+      label_string = ds_data["spec"]["selector"]["matchLabels"].map { |k, v| "#{k}=#{v}" }.join(",")
+      raw_json, _err, st = kubectl.run("get", "pods", "-a", "--output=json", "--selector=#{label_string}")
+      return [] unless st.success?
+
+      all_pods = JSON.parse(raw_json)["items"]
+      all_pods.each_with_object([]) do |pod_data, relevant_pods|
+        next unless pod_data["metadata"]["ownerReferences"].any? { |ref| ref["uid"] == ds_data["metadata"]["uid"] }
```
KnVerey: Seemingly this will select all pods belonging to the DaemonSet, both old and updated. Is there a way to select only the ones in the updated generation? It's the intermediate ReplicaSet layer that provides this guarantee for Deployments. If it isn't possible to filter down to the relevant pods, then we'll need to change our logic for success/failure.

test/fixtures/hello-cloud/daemon_set.yml
karanthukral: I fixed that but still need to push it :P

lib/kubernetes-deploy/kubernetes_resource/daemon_set.rb
```diff
@@ -0,0 +1,99 @@
...
+        @pods = find_pods(daemonset_data)
+      else # reset
+        @rollout_data = { "currentNumberScheduled" => 0 }
+        @status = nil
+        @pods = []
+      end
+    end
+
+    def deploy_succeeded?
+      @desired_number == @rollout_data["desiredNumberScheduled"].to_i &&
```
karanthukral: 🤦‍♂️

lib/kubernetes-deploy/kubernetes_resource/daemon_set.rb
karanthukral: Thought I added the revision check too. I'll add that

test/helpers/fixture_set.rb
kirs: I'd also add the assertion message

karanthukral:
Made all the requested changes. Should be good for another set of 👀 . Will rebase the commits before merging

lib/kubernetes-deploy/kubernetes_resource/daemon_set.rb
```diff
@@ -0,0 +1,103 @@
...
+      else # reset
+        @rollout_data = { "currentNumberScheduled" => 0 }
+        @status = nil
+        @pods = []
+      end
+    end
+
+    def deploy_succeeded?
+      @rollout_data["desiredNumberScheduled"] == @rollout_data["currentNumberScheduled"].to_i &&
+      @rollout_data["desiredNumberScheduled"] == @rollout_data["numberAvailable"].to_i
```
KnVerey: I believe this could generate a false positive if we look right before the rollout of the new pods begins. I.e. at that point desired == current == available, and all of the pods in question are from the old generation. The status includes a `updatedNumberScheduled` number, doesn't it? Do you think we should also look at  `@pods` at all?

lib/kubernetes-deploy/kubernetes_resource/daemon_set.rb
```diff
@@ -0,0 +1,103 @@
...
+          "--tail=#{LOG_LINE_COUNT}"
+        )
+        container_logs[container_name] = out.split("\n")
+      end
+    end
+
+    private
+
+    def container_names
+      @definition["spec"]["template"]["spec"]["containers"].map { |c| c["name"] }
```
KnVerey: Can DaemonSets have init containers too?

lib/kubernetes-deploy/kubernetes_resource/daemon_set.rb
```diff
@@ -0,0 +1,103 @@
...
+      label_string = ds_data["spec"]["selector"]["matchLabels"].map { |k, v| "#{k}=#{v}" }.join(",")
+      raw_json, _err, st = kubectl.run("get", "pods", "-a", "--output=json", "--selector=#{label_string}")
+      return [] unless st.success?
+
+      all_pods = JSON.parse(raw_json)["items"]
+      current_generation = ds_data["metadata"]["generation"].to_s
+
+      latest_pods = all_pods.find_all do |pods|
+        pods["metadata"]["ownerReferences"].any? { |ref| ref["uid"] == ds_data["metadata"]["uid"] } &&
+        pods["metadata"]["labels"]["pod-template-generation"] == current_generation
```
KnVerey: This sounds like the right thing to look at, but for my education (since this is different from Deployment) do you have a doc or PR link that describes Generation?

test/integration/kubernetes_deploy_test.rb
```diff
@@ -567,6 +568,24 @@ def test_success_detection_tolerates_out_of_band_deployment_scaling
...
+      container["image"] = "busybox"
+      container["command"] = ["ls", "/not-a-dir"]
+    end
+
+    refute success
+
+    assert_logs_match_all([
+      "DaemonSet/nginx: FAILED",
+      "The following containers are in a state that is unlikely to be recoverable:",
+      "Logs from container 'nginx' (last 250 lines shown):",
```
KnVerey: Add an assertion on an event too please

test/integration/kubernetes_deploy_test.rb
```diff
@@ -11,13 +11,14 @@ def test_full_hello_cloud_set_deploy_succeeds
...
-      "Successfully deployed 13 resources"
+      "Successfully deployed 14 resources"
     ], in_order: true)
 
     assert_logs_match_all([
       %r{ReplicaSet/bare-replica-set\s+1 replica, 1 availableReplica, 1 readyReplica},
       %r{Deployment/web\s+1 replica, 1 updatedReplica, 1 availableReplica},
-      %r{Service/web\s+Selects at least 1 pod}
+      %r{Service/web\s+Selects at least 1 pod},
+      %r{DaemonSet/nginx\s+1 currentNumberSchedule, 1 desiredNumberSchedule, 1 numberRead, 1 numberAvailabl}
```
KnVerey: Why are "numberRead" and "numberAvailabl" truncated?

lib/kubernetes-deploy/kubernetes_resource/daemon_set.rb
karanthukral: I mostly figured this out by testing it and seeing what we had access too. 👀 at the k8s repo I did find [isPodUpdated](https://github.com/kubernetes/kubernetes/blob/50ec4384ff0ac482beb164767e562765ad82bde7/pkg/controller/daemon/util/daemonset_util.go#L72) and [rollout status ref](https://github.com/kubernetes/kubernetes/blob/7560142e27ba9c0c535695356a14ac169acd7476/pkg/kubectl/rollout_status.go#L97). Hopefully that helps. If not I can investigate more tomorrow

lib/kubernetes-deploy/kubernetes_resource/daemon_set.rb
karanthukral: It can. Sorry forgot to add it back while jumping between too many branches last week

lib/kubernetes-deploy/kubernetes_resource/daemon_set.rb
karanthukral: I'll look into the `updatedNumberScheduled`. I skipped checking the pods since `numberAvailable` should handle the case when a pod deploy fails. If you see a reason for it to be there, I can look into adding it.

KnVerey: > I skipped checking the pods since numberAvailable should handle the case when a pod deploy fails.  If you see a reason for it to be there, I can look into adding it.

Theoretically you should be right. However, in debugging a test failure that happened about 1% of the time locally and more often on CI, I discovered that the following is possible _for deployments_:
- Deployment template is up to date
- New replica set exists and is scaled to zero (so its desired/current/available are equal)
- Deployment's status still describes the old replica set, which is fully scaled still (so the overall desired/current/available are also equal)

That's why Deployment's success condition looks at the latest replicaSet too, not just the status fields. I don't know that DS have the same issue, but it seems plausible. An in-between option would be to include `@pods.length == desiredNumberScheduled`. WDYT?

KnVerey: One more note on this: don't forget to convert all of them to integers (you've got a mix right now)

KnVerey: Looking at the code you linked in your other comment, we should be safe without looking at pods as long as we also check `status.observedGeneration == metadata.Generation`. Looks like that strategy would be available for Deployment too actually.

lib/kubernetes-deploy/kubernetes_resource/daemon_set.rb
KnVerey: I started from those links and poked around a bit. tl;dr I think what you have here is correct for 1.6, but it'll change to `controller-revision-hash` in 1.7 (deprecation)/1.8(final). Some of the stuff I just read:
- [Daemonset Update Proposal](https://github.com/kubernetes/community/blob/master/contributors/design-proposals/daemonset-update.md#implementation) with some relevant details about the field changes (looks like the old one will be kept and the new one added when a DS rolls in 1.7)
- [pod-template-generation is deprecated](https://github.com/kubernetes/kubernetes/blob/ebf24c14a916adebc37023be69f112d2c97759a9/staging/src/k8s.io/api/extensions/v1beta1/types.go#L552-L563) 
- [Issue mentioning the DS/deploy/SS labels for 1.7](https://github.com/kubernetes/kubernetes/issues/47554) (it was apparently going to be `daemonset-controller-hash` for DS but was changed before release for consistency)

test/integration/kubernetes_deploy_test.rb
```diff
@@ -51,7 +52,7 @@ def test_pruning_works
       'deployment "web"',
       'ingress "web"'
     ] # not necessarily listed in this order
-    expected_msgs = [/Pruned 5 resources and successfully deployed 3 resources/]
+    expected_msgs = [/Pruned 6 resources and successfully deployed 3 resources/]
```
KnVerey: let's also add a message to the `expected_prune` list above

lib/kubernetes-deploy/kubernetes_resource/daemon_set.rb
karanthukral: Since 1.6 we don't have access to the revision, I'll make an issue to update this when we look at migrating to 1.7

test/integration/kubernetes_deploy_test.rb
karanthukral: Fixing now :)

test/integration/kubernetes_deploy_test.rb
```diff
@@ -567,6 +569,25 @@ def test_success_detection_tolerates_out_of_band_deployment_scaling
...
+      container["image"] = "busybox"
+      container["command"] = ["ls", "/not-a-dir"]
+    end
+
+    refute success
+
+    assert_logs_match_all([
+      "DaemonSet/nginx: FAILED",
+      "The following containers are in a state that is unlikely to be recoverable:",
+      "Events: None found. Please check your usual logging service (e.g. Splunk).",
```
KnVerey: Hm, really? I'd expect there to be some events here. Indeed, I see two when I run this test locally:
```
[Pod/nginx-cnkzq]	BackOff: Back-off restarting failed container (2 events)
[Pod/nginx-cnkzq]	FailedSync: Error syncing pod, skipping: failed to "StartContainer" for "nginx" with CrashLoopBackOff: "Back-off 10s restarting failed container=nginx pod=nginx-cnkzq_k8sdeploy-test-bad-container-on-daemon-sets-fa-be1043b5b97dfe2b(969dd56c-7152-11e7-be12-76b0ec54df39)" (2 events)
```

test/unit/kubernetes-deploy/google_friendly_config_test.rb
```diff
@@ -14,7 +14,7 @@ def teardown
   def test_auth_use_default_gcp_success
     config = KubernetesDeploy::KubeclientBuilder::GoogleFriendlyConfig.new(kubeconfig, "")
 
-    stub_request(:post, 'https://www.googleapis.com/oauth2/v3/token')
+    stub_request(:post, 'https://www.googleapis.com/oauth2/v4/token')
```
KnVerey: This is on master if you rebase FYI

test/integration/kubernetes_deploy_test.rb
karanthukral: My local tests for some reason is not printing any logs or events so I wanted to see if buildkite would do it

lib/kubernetes-deploy/kubernetes_resource/daemon_set.rb
```diff
@@ -0,0 +1,95 @@
...
+    end
+
+    def deploy_succeeded?
+      @rollout_data["desiredNumberScheduled"].to_i == @rollout_data["currentNumberScheduled"].to_i &&
+      @rollout_data["desiredNumberScheduled"].to_i == @rollout_data["numberAvailable"].to_i &&
+      @current_generation == @observed_generation
+    end
+
+    def deploy_failed?
+      @pods.present? && @pods.all?(&:deploy_failed?)
```
kirs: Why `.all?` and not `any?`? What if daemon set has one crashing and one starting pod?

karanthukral: Good point. I'll fix that

karanthukral:
I'll merge this first thing tomorrow since I don't want to ship it EOD

lib/kubernetes-deploy/kubernetes_resource/daemon_set.rb
KnVerey: I think `all?` is correct actually. Since these are clones, we don't want to fail the deploy if a large deployment has a single bad pod, which will automatically be rescheduled. We're trying to catch the case where something is systemically wrong. ReplicaSets also use `all?` for this reason.

karanthukral: I feel like `all?` makes sense for replica_sets since they are redundant whereas a daemon set is meant to make sure there is a pod on each node. I don't have strong feelings either way though

KnVerey: Very true. And when a single pod is bad, it is often because of a problem with the node, so in the case of a DS pod it is pretty doomed. I'm especially nervous that we'll wrongly fail deploys due to transient registry errors that unfortunately can surface as 404s, but I definitely see the logic. We can always try `any?` if we feel it's the strictly correct solution as long as we keep an eye on our DaemonSet success metrics and switch to `all?` if it is an actual problem.

karanthukral: Sounds good 👍

kirs: 👍

PR #140, opened 2017-07-19
Add Stateful Set Resoruce

karanthukral:
## What?
- Adds stateful set as a kubernetes resource
- Replicates logic built into the Replica Set resource [here](https://github.com/Shopify/kubernetes-deploy/blob/master/lib/kubernetes-deploy/kubernetes_resource/replica_set.rb)
- The timeout currently is the same as a replica set
- As mentioned in a comment below, stateful sets pods currently (1.6) only support update `onDelete` which means simply using `apply` or `replace` won't actually update the pods. Kubernetes 1.7 is adding support for `RollingUpdate`.
  - Since we want to move to 1.7 in the short term, it might not be worth implementing custom deploy logic to delete each pod as 1.7 would allow us to simply use `apply`
  - If we feel like this is needed before we move to 1.6 then we can look into implementing custom logic for deploys

### DIFF
--- a/lib/kubernetes-deploy/kubernetes_resource/stateful_set.rb
+++ b/lib/kubernetes-deploy/kubernetes_resource/stateful_set.rb
@@ -1,0 +1,66 @@
+# frozen_string_literal: true
+require 'kubernetes-deploy/kubernetes_resource/pod_set_base'
+module KubernetesDeploy
+  class StatefulSet < PodSetBase
+    TIMEOUT = 10.minutes
+    ONDELETE = 'OnDelete'
+    attr_reader :pods
+
+    def sync
+      raw_json, _err, st = kubectl.run("get", type, @name, "--output=json")
+      @found = st.success?
+
+      if @found
+        stateful_data = JSON.parse(raw_json)
+        @desired_replicas = stateful_data["spec"]["replicas"].to_i
+        @status_data = stateful_data["status"]
+        rollout_data = stateful_data["status"].slice("replicas", "readyReplicas", "currentReplicas")
+        @update_strategy = if kubectl.server_version < Gem::Version.new("1.7.0")
+          ONDELETE
+        else
+          stateful_data['spec']['updateStrategy']['type']
+        end
+        @status = rollout_data.map { |state_replicas, num| "#{num} #{state_replicas.chop.pluralize(num)}" }.join(", ")
+        @pods = find_pods(stateful_data)
+      else # reset
+        @status_data = { 'readyReplicas' => '-1', 'currentReplicas' => '-2' }
+        @status = nil
+        @pods = []
+      end
+    end
+
+    def deploy_succeeded?
+      if @update_strategy == ONDELETE
+        # Gem cannot monitor update since it doesn't occur until delete
+        unless @success_assumption_warning_shown
+          @logger.warn("WARNING: Your StatefulSet's updateStrategy is set to OnDelete, "\
+                       "which means updates will not be applied until its pods are deleted. "\
+                       "If you are using k8s 1.7+, consider switching to rollingUpdate.")
+          @success_assumption_warning_shown = true
+        end
+        true
+      else
+        @status_data['currentRevision'] == @status_data['updateRevision'] &&
+        @desired_replicas == @status_data['readyReplicas'].to_i &&
+        @desired_replicas == @status_data['currentReplicas'].to_i
+      end
+    end
+
+    def deploy_failed?
+      return false if @update_strategy == ONDELETE
+      pods.present? && pods.any?(&:deploy_failed?)
+    end
+
+    def exists?
+      @found
+    end
+
+    private
+
+    def parent_of_pod?(set_data, pod_data)
+      return false unless pod_data.dig("metadata", "ownerReferences")
+      pod_data["metadata"]["ownerReferences"].any? { |ref| ref["uid"] == set_data["metadata"]["uid"] } &&
+      set_data["status"]["currentRevision"] == pod_data["metadata"]["labels"]["controller-revision-hash"]
+    end
+  end
+end
--- a/test/fixtures/hello-cloud/stateful_set.yml
+++ b/test/fixtures/hello-cloud/stateful_set.yml
@@ -1,0 +1,19 @@
+apiVersion: apps/v1beta1
+kind: StatefulSet
+metadata:
+  name: nginx-ss
+spec:
+  serviceName: "nginx-ss"
+  replicas: 2
+  template:
+    metadata:
+      labels:
+        app: hello-cloud
+        name: nginx-ss
+    spec:
+      containers:
+      - name: nginx
+        image: gcr.io/google_containers/nginx-slim:0.8
+        command: ["sleep", "40"]
+        ports:
+        - containerPort: 80
--- a/test/integration/kubernetes_deploy_test.rb
+++ b/test/integration/kubernetes_deploy_test.rb
@@ -11,14 +11,15 @@ class KubernetesDeployTest < KubernetesDeploy::IntegrationTest
       "Deploying ConfigMap/hello-cloud-configmap-data (timeout: 30s)",
       "Hello from Docker!", # unmanaged pod logs
       "Result: SUCCESS",
-      "Successfully deployed 13 resources"
+      "Successfully deployed 14 resources"
     ], in_order: true)
 
     assert_logs_match_all([
       %r{ReplicaSet/bare-replica-set\s+1 replica, 1 availableReplica, 1 readyReplica},
       %r{Deployment/web\s+1 replica, 1 updatedReplica, 1 availableReplica},
       %r{Service/web\s+Selects at least 1 pod},
-      %r{DaemonSet/nginx\s+1 currentNumberScheduled, 1 desiredNumberScheduled, 1 numberReady}
+      %r{DaemonSet/nginx\s+1 currentNumberScheduled, 1 desiredNumberScheduled, 1 numberReady},
+      %r{StatefulSet/nginx-ss}
     ])
 
     # Verify that success section isn't duplicated for predeployed resources
@@ -55,9 +56,10 @@ class KubernetesDeployTest < KubernetesDeploy::IntegrationTest
       'service "web"',
       'deployment "web"',
       'ingress "web"',
-      'daemonset "nginx"'
+      'daemonset "nginx"',
+      'statefulset "nginx-ss"'
     ] # not necessarily listed in this order
-    expected_msgs = [/Pruned 6 resources and successfully deployed 3 resources/]
+    expected_msgs = [/Pruned 7 resources and successfully deployed 3 resources/]
     expected_pruned.map do |resource|
       expected_msgs << /The following resources were pruned:.*#{resource}/
     end
@@ -669,6 +671,53 @@ invalid type for io.k8s.apimachinery.pkg.apis.meta.v1.ObjectMeta.labels:",
       "BackOff: Back-off restarting failed container",
       "Logs from container 'nginx' (last 250 lines shown):",
       "ls: /not-a-dir: No such file or directory"
+    ], in_order: true)
+  end
+
+  def test_bad_container_on_stateful_sets_fails_with_rolling_update
+    skip if KUBE_CLIENT_VERSION < Gem::Version.new("1.7.0")
+    result = deploy_fixtures("hello-cloud", subset: ["stateful_set.yml"]) do |fixtures|
+      stateful_set = fixtures['stateful_set.yml']['StatefulSet'].first
+      stateful_set['spec']['updateStrategy'] = { 'type' => 'RollingUpdate' }
+      container = stateful_set['spec']['template']['spec']['containers'].first
+      container["image"] = "busybox"
+      container["command"] = ["ls", "/not-a-dir"]
+    end
+
+    assert_deploy_failure(result)
+    assert_logs_match_all([
+      "StatefulSet/nginx-ss: FAILED",
+      "nginx: Crashing repeatedly (exit 1). See logs for more information.",
+      "Events (common success events excluded):",
+      "[Pod/nginx-ss-0]	FailedSync: Error syncing pod",
+      "Logs from container 'nginx' (last 250 lines shown):",
+      "ls: /not-a-dir: No such file or directory"
+    ], in_order: true)
+  end
+
+  def test_on_delete_stateful_sets_are_not_monitored
+    result = deploy_fixtures("hello-cloud", subset: ["stateful_set.yml"])
+
+    assert_deploy_success(result)
+    assert_logs_match_all([
+      "WARNING: Your StatefulSet's updateStrategy is set to OnDelete",
+      "Successful resources",
+      "StatefulSet/nginx-ss"
+    ], in_order: true)
+  end
+
+  def test_rolling_update_stateful_sets_are_monitored
+    skip if KUBE_CLIENT_VERSION < Gem::Version.new("1.7.0")
+    result = deploy_fixtures("hello-cloud", subset: ["stateful_set.yml"]) do |fixtures|
+      stateful_set = fixtures['stateful_set.yml']['StatefulSet'].first
+      stateful_set['spec']['updateStrategy'] = { 'type' => 'RollingUpdate' }
+    end
+
+    assert_deploy_success(result)
+    assert_logs_match_all([
+      "Successfully deployed",
+      "Successful resources",
+      %r{StatefulSet/nginx-ss\s+2 replicas, 2 readyReplicas, 2 currentReplicas}
     ], in_order: true)
   end
 
--- a/lib/kubernetes-deploy/kubeclient_builder.rb
+++ b/lib/kubernetes-deploy/kubeclient_builder.rb
@@ -35,6 +35,14 @@ module KubernetesDeploy
       )
     end
 
+    def build_apps_v1beta1_kubeclient(context)
+      _build_kubeclient(
+        api_version: "v1beta1",
+        context: context,
+        endpoint_path: "/apis/apps"
+      )
+    end
+
     def _build_kubeclient(api_version:, context:, endpoint_path: nil)
       config = GoogleFriendlyConfig.read(ENV.fetch("KUBECONFIG"))
       unless config.contexts.include?(context)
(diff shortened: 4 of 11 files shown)


airhorns:
This would be super awesome to get out <3

lib/kubernetes-deploy/kubernetes_resource/stateful_set.rb
```diff
@@ -0,0 +1,96 @@
...
+    private
+
+    def container_names
+      regular_containers = @definition["spec"]["template"]["spec"]["containers"].map { |c| c["name"] }
+      init_containers = @definition["spec"]["template"]["spec"].fetch("initContainers", {}).map { |c| c["name"] }
+      regular_containers + init_containers
+    end
+
+    def find_pods(ss_data)
+      label_string = ss_data["spec"]["selector"]["matchLabels"].map { |k, v| "#{k}=#{v}" }.join(",")
```
karanthukral: @KnVerey Currently (in k8s 1.6) the updates to the stateful set pods are only applied `onDelete` which means `kubectl apply` or `kubectl replace` only updates the stateful set definition. In k8s 1.7, they are adding `RollingUpdates` for them and will add the revision tag to match the pods. 

Do you think its worth implementing a custom apply logic to delete each pod during the deploy or do we want to wait for 1.7 where `apply` would be sufficient?

KnVerey: > Do you think its worth implementing a custom apply logic to delete each pod during the deploy or do we want to wait for 1.7 where apply would be sufficient?

Definitely not worth spending time on custom logic. Arguably making the deploy time out when the rollout doesn't happen automatically under 1.6 still would be more correct though. Let's chat tomorrow about the 1.7 timeline and what would be best for our (very few) users of StatefulSets if it isn't imminent.

KnVerey:
@karanthukral let's pick this up again now that we're supporting 1.7

karanthukral:
@KnVerey Tests are running to make sure nothing broke. Added a new test for failure to make sure pods are selected. Should be good for review 👀

karanthukral:
@stefanmb Does 1.6.4 and 1.7.5 ci's use different kubectl versions? Trying to fix the tests, they are broken on master as well

stefanmb:
Yes, each one uses the stated version of `kubectl`.

karanthukral:
> Yes, each one uses the stated version of kubectl.

I might be wrong but it looks like `1.7.5` is using kubectl 1.8 since I can replicate the failure locally by using 1.8 but not with 1.7.5

stefanmb:
Hmm, seems you're right. The minikube installation doesn't seem to install `kubectl`.

stefanmb:
So the CI situation right now is that each cluster has the correct server version, but the client used is `1.8.1` everywhere. cc: @KnVerey

karanthukral:
@KnVerey whenever you get a chance

lib/kubernetes-deploy/kubernetes_resource/stateful_set.rb
```diff
@@ -0,0 +1,98 @@
...
+        @pods = []
+      end
+    end
+
+    def deploy_succeeded?
+      @current_generation == @observed_generation && @desired_replicas == @rollout_data["replicas"].to_i
+    end
+
+    def deploy_failed?
+      @pods.present? && @pods.all?(&:deploy_failed?)
```
stefanmb: `all?` or `any?`

I realized this is the same logic as `ReplicaSet` but I'd like to understand why.

karanthukral: It should be `any?` that was my fault. Will fix

stefanmb: But the same logic error exists in `ReplicaSet` then?

karanthukral: > But the same logic error exists in ReplicaSet then?

Not sure why 😕 . I use `any?` in `DaemonSets`. @KnVerey might know more about `ReplicaSets`

karanthukral:
> could we extract some of it into a base class?

I'll give that a shot :)

karanthukral:
@stefanmb Hopefully this cleans up the repeated code

karanthukral:
This should be good for 👀  too

lib/kubernetes-deploy/kubernetes_resource/base_set.rb
```diff
@@ -0,0 +1,73 @@
+# frozen_string_literal: true
+module KubernetesDeploy
+  class BaseSet < KubernetesResource
+    def deploy_failed?
+      @pods.present? && @pods.any?(&:deploy_failed?)
```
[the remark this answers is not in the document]
KnVerey: To make a point related to this specific method: I'd rather sick with  `.all?` for ReplicaSet. For DaemonSet, `.any?` makes sense because the pods are more unique in that a bad node can cause a single one to fail, and that one will not be able to be rescheduled. One bad pod on a RS, where they're truly clones, is much more likely to be recovered automatically. For StatefulSet I could go either way but would lean toward `any?` because the nature of SS suggests the pods could fail for independent, unrecoverable reasons. IMO this decision needs to be made class-by-class, balancing deploy fragility/speed concerns, and the base class shouldn't define a default as a result.

lib/kubernetes-deploy/kubernetes_resource/base_set.rb
```diff
@@ -0,0 +1,73 @@
...
+
+    def fetch_events
+      own_events = super
+      return own_events unless @pods.present?
+      most_useful_pod = @pods.find(&:deploy_failed?) || @pods.find(&:deploy_timed_out?) || @pods.first
+      own_events.merge(most_useful_pod.fetch_events)
+    end
+
+    def fetch_logs
+      container_names.each_with_object({}) do |container_name, container_logs|
```
KnVerey: Add the `return {} unless @pods.present? # the kubectl command times out if no pods exist` guard from RS

lib/kubernetes-deploy/kubernetes_resource/stateful_set.rb
```diff
@@ -0,0 +1,30 @@
...
+    def sync
+      raw_json, _err, st = kubectl.run("get", type, @name, "--output=json")
+      @found = st.success?
+
+      if @found
+        stateful_data = JSON.parse(raw_json)
+        @current_generation = stateful_data["metadata"]["generation"]
+        @observed_generation = stateful_data["status"]["observedGeneration"]
+        @desired_replicas = stateful_data["spec"]["replicas"].to_i
+        @rollout_data = stateful_data["status"].slice("replicas")
```
KnVerey: This is slicing out the exact same thing you're using for `desired_replicas`, so if I'm not mistaken, your current success condition will always pass as soon as the first new replica is created. I see in [the docs](https://v1-7.docs.kubernetes.io/docs/api-reference/v1.7/#statefulsetstatus-v1beta1-apps) that SS have a distinctive set of fields, two of which reference `Revision` (that's probably referring to [controllerRevision](https://github.com/kubernetes/kubernetes/pull/45867)?). Please look into and document which combinations of these fields we should look at to most reliably detect rollout completion.

karanthukral:
Skipped monitoring onDelete statefulsets

lib/kubernetes-deploy/kubernetes_resource/stateful_set.rb
```diff
@@ -0,0 +1,61 @@
...
+    def sync
+      raw_json, _err, st = kubectl.run("get", type, @name, "--output=json")
+      @found = st.success?
+
+      if @found
+        stateful_data = JSON.parse(raw_json)
+        @status_data = stateful_data["status"]
+        @rollout_data = stateful_data["status"].slice("replicas")
+        @update_strategy = stateful_data["spec"]["updateStrategy"]["type"]
+        @status = @rollout_data.map { |state_replicas, num| "#{num} #{state_replicas.chop.pluralize(num)}" }.join(", ")
```
KnVerey: This should minimally include `readyReplicas` too... maybe even the other `xReplicas` fields as well, wdyt?

lib/kubernetes-deploy/kubernetes_resource/stateful_set.rb
```diff
@@ -0,0 +1,61 @@
...
+        @pods = find_pods(stateful_data)
+      else # reset
+        @rollout_data = { "replicas" => 0 }
```
KnVerey: This should reset `@status_data` instead in a way that will not pass `deploy_succeeded?`

lib/kubernetes-deploy/kubernetes_resource/stateful_set.rb
```diff
@@ -0,0 +1,61 @@
...
+        @status = nil
+        @pods = []
+      end
+    end
+
+    def deploy_succeeded?
+      if @update_strategy == ONDELETE
+        # Gem cannot monitor update since it doesn't occur until delete
+        unless @success_assumption_warning_shown
+          @logger.warn("WARNING: Your SS's updateStrategy is set to OnDelete, "\
```
KnVerey: - Spell out "StatefulSet"
- "its deleted" -> "its pods are deleted" 
- `RollingUpdates` -> `rollingUpdate` (nit: actual field name)

lib/kubernetes-deploy/kubernetes_resource/stateful_set.rb
```diff
@@ -0,0 +1,61 @@
...
+                       "which means updates will not be applied until its deleted. "\
+                       "If you are using k8s 1.7+, consider switching to RollingUpdates.")
+          @success_assumption_warning_shown = true
+        end
+        true
+      else
+        @status_data['current_generation'] == @status_data['observed_generation'] &&
```
KnVerey: these field names seem wrong... I'm guessing the test passes because they are both `nil`. 

The real field I see is `.status.observedGeneration`. The current generation is in  `.metadata.generation`, not the status.

KnVerey: The fact that SS has so many different status fields is a bit confusing... I went and found an example of a SS that is having trouble rolling out an update to help clarify what we need. The following status belongs to a SS that is in a pretty messed up state:

```yaml
status:
  currentReplicas: 1
  currentRevision: statefulset-3627988389 # there is one pod in this version
  observedGeneration: 4 # this is equal to .metadata.generation... useless?
  readyReplicas: 2 # these are from a revision not mentioned in the status
  replicas: 3
  updateRevision: statefulset-1264314452 # there are no pods in this version
```

And of a healthy SS:

```yaml
status:
  currentReplicas: 2
  currentRevision: other-statefulset-335438845
  observedGeneration: 64
  readyReplicas: 2 
  replicas: 2
  updateRevision: other-statefulset-335438845
```

Based on that, I think we need:
- `.status.currentRevision == .status.updateRevision`
- `.spec.replicas == .status.currentReplicas == .status.readyReplicas`

lib/kubernetes-deploy/kubernetes_resource/stateful_set.rb
```diff
@@ -0,0 +1,61 @@
...
+        @status_data['currentRevision'] == @status_data['updateRevision'] &&
+        @status_data['replicas'].to_i == @status_data['readyReplicas'].to_i
```
KnVerey: According to the docs, `replicas` is the number created by the controller, which doesn't sound like the same thing as the number we ultimately want. I think we need to compare to `.spec.replicas` instead. (see my other comment though)

lib/kubernetes-deploy/kubernetes_resource/stateful_set.rb
```diff
@@ -0,0 +1,61 @@
+# frozen_string_literal: true
+require 'kubernetes-deploy/kubernetes_resource/pod_set_base'
+module KubernetesDeploy
+  class StatefulSet < PodSetBase
+    TIMEOUT = 5.minutes
```
KnVerey: Do you think this is long enough for statefulsets? Since they roll out one by one, I'd tend to think they'd need something more generous. Maybe check with DataDelivery?

karanthukral: Good point. I'll check with them

lib/kubernetes-deploy/kubernetes_resource/stateful_set.rb
karanthukral: I'll add `currentReplicas` and `readyReplicas`

lib/kubernetes-deploy/kubernetes_resource/stateful_set.rb
karanthukral: Good catch. The ss fields are confusing in how they are layed out

lib/kubernetes-deploy/kubernetes_resource/stateful_set.rb
KnVerey: What did they say?

lib/kubernetes-deploy/kubernetes_resource/stateful_set.rb
```diff
@@ -0,0 +1,62 @@
...
+    def sync
+      raw_json, _err, st = kubectl.run("get", type, @name, "--output=json")
+      @found = st.success?
+
+      if @found
+        stateful_data = JSON.parse(raw_json)
+        @desired_replicas = stateful_data["spec"]["replicas"].to_i
+        @status_data = stateful_data["status"]
+        rollout_data = stateful_data["status"].slice("replicas", "readyReplicas", "currentReplicas")
+        @update_strategy = stateful_data.dig('spec', 'updateStrategy', 'type') || ONDELETE
```
KnVerey: Hm, why did you have to add a default here? I would have expected the server to always fill this in if it isn't present. The actual default will vary by groupversion; in `apps/v1beta1` it is indeed OnDelete, but it is switching to RollingUpdate in v1beta2: https://kubernetes.io/docs/api-reference/v1.8/#statefulsetupdatestrategy-v1beta2-apps

lib/kubernetes-deploy/kubernetes_resource/stateful_set.rb
```diff
@@ -0,0 +1,62 @@
...
+        @pods = []
+      end
+    end
+
+    def deploy_succeeded?
+      if @update_strategy == ONDELETE
+        # Gem cannot monitor update since it doesn't occur until delete
+        unless @success_assumption_warning_shown
+          @logger.warn("WARNING: Your StatefulSet's updateStrategy is set to OnDelete, "\
+                       "which means updates will not be applied until its pods deleted. "\
```
KnVerey: the word "are" is missing in "until its pods deleted"

lib/kubernetes-deploy/kubernetes_resource/stateful_set.rb
```diff
@@ -0,0 +1,62 @@
...
+      if @found
+        stateful_data = JSON.parse(raw_json)
+        @desired_replicas = stateful_data["spec"]["replicas"].to_i
+        @status_data = stateful_data["status"]
+        rollout_data = stateful_data["status"].slice("replicas", "readyReplicas", "currentReplicas")
+        @update_strategy = stateful_data.dig('spec', 'updateStrategy', 'type') || ONDELETE
+        @status = rollout_data.map { |state_replicas, num| "#{num} #{state_replicas.chop.pluralize(num)}" }.join(", ")
+        @pods = find_pods(stateful_data)
+      else # reset
+        @status_data = { "replicas" => 0 }
```
KnVerey: We don't even look at `replicas`, so resetting it isn't particularly useful. The important part is to make sure that `deploy_succeeded?` won't pass based on this reset. One way to do this is to set mismatched defaults (e.g. current -1 / ready -2); another is to add a guard to `deploy_succeeded?` checking that the keys we need are actually present.

test/fixtures/hello-cloud/stateful_set.yml
```diff
@@ -0,0 +1,36 @@
+---
+apiVersion: v1
+kind: Service
```
KnVerey: What's the use of having a service in this fixture?

test/integration/kubernetes_deploy_test.rb
```diff
@@ -672,6 +676,57 @@ def test_bad_container_on_daemon_sets_fails
     ], in_order: true)
   end
 
+  def test_bad_container_on_stateful_sets_fails_with_rolling_update
+    skip if KUBE_CLIENT_VERSION < Gem::Version.new("1.7.0")
+    result = deploy_fixtures("hello-cloud", subset: ["stateful_set.yml"]) do |fixtures|
+      stateful_set = fixtures['stateful_set.yml']['StatefulSet'].first
+      stateful_set['spec']['updateStrategy'] = {}
```
KnVerey: nit: set `{ 'type' => 'RollingUpdate' }` and save a line

test/integration/kubernetes_deploy_test.rb
```diff
@@ -672,6 +676,57 @@ def test_bad_container_on_daemon_sets_fails
...
+      "Successful resources",
+      "StatefulSet/nginx-ss"
+    ], in_order: true)
+  end
+
+  def test_rolling_update_stateful_sets_are_monitored
+    skip if KUBE_CLIENT_VERSION < Gem::Version.new("1.7.0")
+    result = deploy_fixtures("hello-cloud", subset: ["stateful_set.yml"]) do |fixtures|
+      stateful_set = fixtures['stateful_set.yml']['StatefulSet'].first
+      stateful_set['spec']['updateStrategy'] = {}
```
KnVerey: same nit here

test/integration/kubernetes_deploy_test.rb
```diff
@@ -672,6 +676,57 @@ def test_bad_container_on_daemon_sets_fails
...
+      stateful_set['spec']['updateStrategy']['type'] = 'RollingUpdate'
+    end
+
+    assert_deploy_success(result)
+    assert_logs_match_all([
+      "Deploying resources:",
+      "Successfully deployed",
+      "Service/nginx-ss",
+      "Successful resources",
+      "StatefulSet/nginx-ss"
```
KnVerey: Let's assert on the `2 replicas, 2 readyReplicas, 2 currentReplicas` output here, since we'd normally do that in the "hello cloud" test but can't because it uses OnDelete.

test/fixtures/hello-cloud/stateful_set.yml
karanthukral: Good point. I'll remove it

Issue #134, opened 2017-07-18
Revisit supported ruby version


KnVerey:
We nominally support ruby 2.1 or greater, and have Rubocop set up accordingly. However, it turns out this has not been true since we added the ActiveSupport dependency:

```ruby
activesupport-5.1.1 requires ruby version >= 2.2.2, which is incompatible with the current version, ruby 2.1.8p440
```

Should we:

A) Change the dependency to fall in line with shipit-engine's minimum version as originally intended; or

B) Set an independent, higher requirement.

Given the extent to which we work with variable parsed JSON in this gem, the safe navigation operator introduced in  2.3 would be tremendously useful. And although we certainly designed this gem around Shipit's use case, it isn't all that functionally tied to Shipit. For those reasons, I vote for (B), and making that requirement ruby 2.3.

@Shopify/cloudplatform @kirs

kirs:
👍 for 2.3 as long it's available on Shopify's Shipit setup

n1koo:
👍 for 2.3, I don't see any reason for supporting older versions than what Shipit is using as long as theres no other major users

ibawt:
2.3 sounds fine to me.

PR #143, opened 2017-07-20
Add a dev.yml which uses dev's minikube support to get Shopify employees set up quickly

airhorns:
Getting `kubectl` and `minikube` isn't the hardest thing in the world but it turns out `dev` is actually pretty good at it. Dev isn't open source but since the majority of contributors are likely to be Shopifiers I figure we may as well add the file that makes it real easy for us to get set up with the normal `dev up` command. I don't think the `dev.yml` will ever have anything sensitive in it so I think this is safe from a security and privacy standpoint.

Thoughts? I 🎩 'd and the test suite passes locally after `dev up`ing.

### DIFF
--- a/dev.yml
+++ b/dev.yml
@@ -1,0 +1,18 @@
+---
+name: kubernetes-deploy
+up:
+  - ruby: 2.3.2
+  - bundler
+  - minikube
+
+minikube: {}
+
+commands:
+  test:
+    run: bundle exec rake test
+  tophat:
+    run: PRINT_LOGS=1 bundle exec ruby -I test test/integration/kubernetes_deploy_test.rb -n/${1}/
+    desc: Tophat a change by running a test scenario with logging output enabled.
+    syntax:
+      optional:
+        argument: TEST_REGEX
--- a/dev.yml
+++ b/dev.yml
@@ -1,13 +1,17 @@
 ---
 name: kubernetes-deploy
 up:
-  - ruby: 2.3.2
+  - ruby: 2.3.3
   - bundler
-  - minikube
-
-minikube: {}
-
+  - homebrew:
+    - Caskroom/cask/minikube
+  - custom:
+      name: Minikube Cluster
+      met?: test $(minikube status | grep Running | wc -l) -eq 2 && $(minikube status | grep -q 'Correctly Configured')
+      meet: minikube start --vm-driver=xhyve --kubernetes-version=v1.7.5
+      down: minikube stop
 commands:
+  reset-minikube: minikube delete && rm -rf ~/.minikube
   test:
     run: bundle exec rake test
   tophat:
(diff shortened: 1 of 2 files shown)


dev.yml
```diff
@@ -0,0 +1,12 @@
+---
+name: kubernetes-deploy
+up:
+  - ruby: 2.4.1
```
KnVerey: Let's use 2.3.3 instead to match Shipit (see also https://github.com/Shopify/kubernetes-deploy/issues/134)

dev.yml
```diff
@@ -0,0 +1,12 @@
...
+  - bundler
+  - minikube
+
+minikube: {}
+
+commands:
+  test:
```
KnVerey: Another command that could be really handy (I have a personal alias for it I use all the time) is something like:
```yaml
  tophat:
    run: PRINT_LOGS=1 bundle exec ruby -I test test/integration/kubernetes_deploy_test.rb -n/${1}/
    desc: Tophat a change by running a test scenario with logging output enabled.
    syntax:
      optional:
        argument: TEST_REGEX
```

image

KnVerey:
Btw the fix for your CI failures is on master now 😄

stefanmb:
@KnVerey @airhorns It seems `dev` minikube support is deprecated: https://dev-accel.shopify.io/dev-acceleration/Developer%20Environments/minikube_is_dead

KnVerey:
oh no, that's too bad. It does say:

> The system is being left in place for use in infrastructure related experiments.

I wonder for how long.

dev.yml
```diff
@@ -0,0 +1,18 @@
+---
+name: kubernetes-deploy
+up:
+  - ruby: 2.3.2
+  - bundler
+  - minikube
+
+minikube: {}
```
stefanmb: We can revive this PR with Minikube from brew - dev supports it.

KnVerey:
I've updated this to use minikube from brew as suggested, based on our setup for cloudbuddies. @stefanmb @karanthukral can you please 👀 and try it out?

dev.yml
```diff
@@ -0,0 +1,22 @@
...
+up:
+  - ruby: 2.3.3
+  - bundler
+  - homebrew:
+    - Caskroom/cask/minikube
+  - custom:
+      name: Minikube Cluster
+      met?: test $(minikube status | grep Running | wc -l) -eq 2 && $(minikube status | grep -q 'Correctly Configured')
+      meet: minikube start --vm-driver=xhyve --kubernetes-version=v1.7.5
+      down: minikube stop
```
stefanmb: FWIW the reason I used `minikube delete` elsewhere was that with `stop` I had noticed on start `minikube` would not switch namespaces - but I think it was a red herring, it appears to work properly here.

PR #141, opened 2017-07-19
Improve output re: invalid templates

KnVerey:
I set out this morning to add a quick fix for a spacing issue I noticed on a couple production deploys:

image

And then I went down a bit of rabbit hole. This PR:
- Fixes that spacing issue.
- Unifies recording of invalid templates (i.e. same code path for failed dry run and failed apply)
- Loosens file path identification to work with the new error examples
- Handles multiple invalid templates in a single apply failure
- Adds a warning on apply failures that some resources were likely created 

I'll post screenshots of the new test output inline.

PR #144, opened 2017-07-21
Ruby 2.3

KnVerey:
Upgrades our minimum requirement to Ruby 2.3. Also makes a few easy-win changes to 2.3-style squiggly heredocs, `dig` and `&.`.

The new `# frozen_string_literal: true` came from running `bx rubocop -a`. Dunno why it didn't care about those files until now.

Closes https://github.com/Shopify/kubernetes-deploy/issues/134

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -2,7 +2,4 @@ inherit_from:
   - http://shopify.github.io/ruby-style-guide/rubocop.yml
 
 AllCops:
-  TargetRubyVersion: 2.1
-
-Layout/IndentHeredoc:
-  EnforcedStyle: active_support
+  TargetRubyVersion: 2.3
--- a/kubernetes-deploy.gemspec
+++ b/kubernetes-deploy.gemspec
@@ -1,4 +1,5 @@
 # coding: utf-8
+# frozen_string_literal: true
 lib = File.expand_path('../lib', __FILE__)
 $LOAD_PATH.unshift(lib) unless $LOAD_PATH.include?(lib)
 require 'kubernetes-deploy/version'
@@ -20,6 +21,8 @@ Gem::Specification.new do |spec|
   spec.bindir        = "exe"
   spec.executables   = spec.files.grep(%r{^exe/}) { |f| File.basename(f) }
   spec.require_paths = ["lib"]
+
+  spec.required_ruby_version = '>= 2.3.0'
   spec.add_dependency "activesupport", ">= 4.2"
   spec.add_dependency "kubeclient", "~> 2.3"
   spec.add_dependency "googleauth", ">= 0.5"
--- a/lib/kubernetes-deploy/kubernetes_resource/deployment.rb
+++ b/lib/kubernetes-deploy/kubernetes_resource/deployment.rb
@@ -47,15 +47,15 @@ module KubernetesDeploy
     end
 
     def failure_message
-      @latest_rs.failure_message
+      @latest_rs&.failure_message
     end
 
     def timeout_message
       progress_seconds = @definition['spec']['progressDeadlineSeconds']
       if progress_seconds
-        "Deploy timed out due to progressDeadlineSeconds of #{progress_seconds} seconds. #{@latest_rs.timeout_message}"
+        "Deploy timed out due to progressDeadlineSeconds of #{progress_seconds} seconds. #{@latest_rs&.timeout_message}"
       else
-        @latest_rs.timeout_message
+        @latest_rs&.timeout_message
       end
     end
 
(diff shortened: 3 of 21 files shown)


lib/kubernetes-deploy/kubernetes_resource/replica_set.rb
```diff
@@ -86,7 +86,7 @@ def unmanaged?
 
     def container_names
       regular_containers = @definition["spec"]["template"]["spec"]["containers"].map { |c| c["name"] }
-      init_containers = @definition["spec"]["template"]["spec"].fetch("initContainers", {}).map { |c| c["name"] }
+      init_containers = @definition["spec"]["template"]["spec"].fetch("initContainers", []).map { |c| c["name"] }
```
stefanmb: Will this still work if the default value is an array instead of a dictionary?

.rubocop.yml
```diff
@@ -2,7 +2,4 @@ inherit_from:
   - http://shopify.github.io/ruby-style-guide/rubocop.yml
 
 AllCops:
-  TargetRubyVersion: 2.1
-
-Layout/IndentHeredoc:
-  EnforcedStyle: active_support
+  TargetRubyVersion: 2.3
```
kirs: > The new # frozen_string_literal: true came from running bx rubocop -a. Dunno why it didn't care about those files until now.

Because it has been validating code for 2.1, and in that version there was no such thing as frozen string literal

kubernetes-deploy.gemspec
```diff
@@ -20,6 +21,8 @@ Gem::Specification.new do |spec|
   spec.bindir        = "exe"
   spec.executables   = spec.files.grep(%r{^exe/}) { |f| File.basename(f) }
   spec.require_paths = ["lib"]
+
+  spec.required_ruby_version = '>= 2.3.0'
```
kirs: 👍

lib/kubernetes-deploy/kubernetes_resource/deployment.rb
```diff
@@ -47,15 +47,15 @@ def deploy_failed?
     end
 
     def failure_message
-      @latest_rs.failure_message
+      @latest_rs&.failure_message
```
kirs: Why this change if it worked before?

test/unit/kubernetes-deploy/google_friendly_config_test.rb
```diff
@@ -33,7 +33,7 @@ def test_auth_use_default_gcp_success
   def test_auth_use_default_gcp_failure
     config = KubernetesDeploy::KubeclientBuilder::GoogleFriendlyConfig.new(kubeconfig, "")
 
-    stub_request(:post, 'https://www.googleapis.com/oauth2/v3/token')
+    stub_request(:post, 'https://www.googleapis.com/oauth2/v4/token')
```
kirs: Curious why this is necessary if dependencies weren't updated.

.rubocop.yml
KnVerey: Ah, of course! I could have sworn it had bugged me to add some of those in the past, but I must be misremembering. Or maybe I ran it with the wrong config somehow. ¯\\\_(ツ)_/¯

lib/kubernetes-deploy/kubernetes_resource/deployment.rb
KnVerey: It is unlikely but possible that no replicaset will be found. That's why we have `return false unless @latest_rs` in other spots. I'm not sure what made me notice these guys now, but we shouldn't assume we have an RS here either.

lib/kubernetes-deploy/kubernetes_resource/replica_set.rb
KnVerey: yeah, `{}.map` and `[].map` both have the same result--do nothing and return an array.

image

I changed it because `initContainers` would be an array if present

test/unit/kubernetes-deploy/google_friendly_config_test.rb
KnVerey: Stefan [asked the same thing in the other PR](https://github.com/Shopify/kubernetes-deploy/pull/141#discussion_r128847918), which has a copy of this commit (CI won't pass without it or an equivalent fix):

> Since this is a gem, the Gemfile.lock isn't committed to the repo. Our CI bundle installs from scratch every time, so it'll upgrade as far as the Gemfile allows it to when new versions become available. Because of this, CI just started using v0.4.0 of kubeclient, which is what caused this difference. Alternatively, we could pin kubeclient more strictly in the gemspec.

LMK if you think we should change the gemspec instead.

PR #145, opened 2017-07-25
Add concurrency to slowest parts

KnVerey:
## Problem

Our largest deploys (core) are spending a ton of time in template discovery and in the early polling cycles that involve the full (massive) set of templates. The following sanitized example is typical of a few deploys I checked:

```
[INFO][2017-07-20 01:23:13 +0000]    Discovering templates:
...
[INFO][2017-07-20 01:23:37 +0000]    ----------------------------Phase 2: Checking initial resource statuses 
# that was 24s of discovering things -- not great!
[INFO][2017-07-20 01:26:45 +0000]    Service/[thing]
# that was nearly 3 MINUTES for the initial sync. WTF!
```

I did some profiling using tests (see [this branch](https://github.com/Shopify/kubernetes-deploy/compare/threads...profile)--lmk if you think that'd be useful to commit to the repo), which supports this slowness. I would have expected that with a template set of the "hello-cloud" size, most of the time would be spent sleeping, but that isn't the case:

Before (main thread):
k8s-d-flame-before

## Solution

Arguably it would be a good idea in the long run to batch requests, i.e. request everything of a given type in the namespace at once and distribute the JSON to the appropriate instances' `sync`. However, since we're spending all of this time in IO wait, concurrency can help us here. Unlike batching, such concurrency is almost trivial to add. I think we should try it first.

After (main thread):
image

cc @Shopify/cloudplatform

### DIFF
--- a/lib/kubernetes-deploy/concurrency.rb
+++ b/lib/kubernetes-deploy/concurrency.rb
@@ -1,0 +1,18 @@
+# frozen_string_literal: true
+module KubernetesDeploy
+  module Concurrency
+    MAX_THREADS = 8
+
+    def self.split_across_threads(all_work, &block)
+      return if all_work.empty?
+      raise ArgumentError, "Block of work is required" unless block_given?
+
+      slice_size = ((all_work.length + MAX_THREADS - 1) / MAX_THREADS)
+      threads = []
+      all_work.each_slice(slice_size) do |work_group|
+        threads << Thread.new { work_group.each(&block) }
+      end
+      threads.each(&:join)
+    end
+  end
+end
--- a/test/unit/kubernetes-deploy/concurrency_test.rb
+++ b/test/unit/kubernetes-deploy/concurrency_test.rb
@@ -1,0 +1,66 @@
+# frozen_string_literal: true
+require 'test_helper'
+
+class ConcurrencyTest < KubernetesDeploy::TestCase
+  class TestWork
+    attr_accessor :worked_by_threads
+    def initialize
+      @worked_by_threads = []
+    end
+  end
+
+  def test_split_across_threads_raises_without_a_block
+    assert_raises_message(ArgumentError, "Block of work is required") do
+      KubernetesDeploy::Concurrency.split_across_threads([TestWork.new])
+    end
+  end
+
+  def test_split_across_threads_works_with_zero_work
+    KubernetesDeploy::Concurrency.split_across_threads([]) do |individual|
+      individual.worked_by_threads << Thread.current.object_id
+    end
+    # nothing raised
+  end
+
+  def test_split_across_threads_works_with_one_work
+    all_work = [TestWork.new]
+    KubernetesDeploy::Concurrency.split_across_threads(all_work) do |individual|
+      individual.worked_by_threads << Thread.current.object_id
+    end
+    assert_work_distribution(all_work, [1])
+  end
+
+  def test_split_across_threads_splits_evenly_with_small_work
+    all_work = 2.times.with_object([]) { |_, all| all << TestWork.new }
+    KubernetesDeploy::Concurrency.split_across_threads(all_work) do |individual|
+      individual.worked_by_threads << Thread.current.object_id
+    end
+    assert_work_distribution(all_work, [1, 1])
+  end
+
+  def test_split_across_threads_splits_evenly_with_equal_work_and_threads
+    all_work = KubernetesDeploy::Concurrency::MAX_THREADS.times.with_object([]) { |_, all| all << TestWork.new }
+    KubernetesDeploy::Concurrency.split_across_threads(all_work) do |individual|
+      individual.worked_by_threads << Thread.current.object_id
+    end
+    assert_work_distribution(all_work, [1, 1, 1, 1, 1, 1, 1, 1])
+  end
+
+  def test_split_across_threads_splits_evenly_with_large_work
+    all_work = 31.times.with_object([]) { |_, all| all << TestWork.new }
+    KubernetesDeploy::Concurrency.split_across_threads(all_work) do |individual|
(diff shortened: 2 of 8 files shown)


lib/kubernetes-deploy/concurrency.rb
```diff
@@ -0,0 +1,18 @@
...
+  module Concurrency
+    MAX_THREADS = 8
+
+    def self.split_across_threads(all_work, &block)
+      return if all_work.empty?
+      raise ArgumentError, "Block of work is required" unless block_given?
+
+      slice_size = ((all_work.length + MAX_THREADS - 1) / MAX_THREADS)
+      threads = []
+      all_work.each_slice(slice_size) do |work_group|
```
kirs: We should use `Thread.abort_on_exception = true`, otherwise those threads will never report exceptions

lib/kubernetes-deploy/concurrency.rb
```diff
@@ -0,0 +1,18 @@
+# frozen_string_literal: true
+module KubernetesDeploy
+  module Concurrency
+    MAX_THREADS = 8
+
+    def self.split_across_threads(all_work, &block)
+      return if all_work.empty?
+      raise ArgumentError, "Block of work is required" unless block_given?
+
+      slice_size = ((all_work.length + MAX_THREADS - 1) / MAX_THREADS)
```
kirs: Smart trick slicing the work into N pieces!

Another common pattern would be to create 8 threads and make each of them `pop` from the `Queue` object (which is a thread safe primitive). But I think your approach is even simpler and we should stick with it.

lib/kubernetes-deploy/concurrency.rb
kirs: (it should be set in the parent thread, as per example in http://ruby-doc.org/core-2.4.1/Thread.html#method-c-abort_on_exception-3D )

test/unit/kubernetes-deploy/concurrency_test.rb
```diff
@@ -0,0 +1,66 @@
...
+    assert_raises_message(ArgumentError, "Block of work is required") do
+      KubernetesDeploy::Concurrency.split_across_threads([TestWork.new])
+    end
+  end
+
+  def test_split_across_threads_works_with_zero_work
+    KubernetesDeploy::Concurrency.split_across_threads([]) do |individual|
+      individual.worked_by_threads << Thread.current.object_id
+    end
+    assert_work_distribution([], [])
```
kirs: Do we need this assertion at all?

lib/kubernetes-deploy/concurrency.rb
KnVerey: Yeah, I considered `Queue` as well and didn't use it for exactly that reason--it wouldn't be as simple, and the expected benefit would be low given that these pieces of work should be roughly equal.

lib/kubernetes-deploy/concurrency.rb
ibawt: 👍

lib/kubernetes-deploy/concurrency.rb
ibawt: and all with integer math :D

lib/kubernetes-deploy/concurrency.rb
KnVerey: > otherwise those threads will never report exceptions

That's what I thought that option meant too when I was reading the `Thread` docs, but when I played with it that's not what I saw. It seems like it makes the main thread abort immediately when any other thread aborts, as opposed to when you join the other threads. Even more unexpectedly, it doesn't seem to abort the other child threads immediately. Here's the experiment I did:
```ruby
      threads = []
      Thread.abort_on_exception = true # comment out this line for the other version of experiment
      all_work.each_slice(slice_size) do |work_group|
        thread = Thread.new do
          puts ColorizedString.new("This thing started: #{work_group.map(&:id)}").yellow
          work_group.each(&block)
          puts ColorizedString.new("This thing ran: #{work_group.map(&:id)}").green
          raise FatalDeploymentError, "this thing raised: #{work_group.map(&:id)}" if work_group.any? { |work| work.id.include?("Service/redis") }
        end
        threads << thread
      end
      puts "SLEEPING"
      sleep 10
      puts "JOINING THREADS"
      threads.each(&:join)
```

`Thread.abort_on_exception = false`
image


`Thread.abort_on_exception = true`
image

The former behaviour looks nicer to me. Am I missing something?

KnVerey:
> To confirm does this PR introduce concurrency in three spots: dry-run, Runner, ResourceWatcher

Yes, specifically the initial `sync` in Runner, the regular `sync` in the loop in ResourceWatcher, and the dry run we do for template validation

> Also, have you tried running it on core templates after the change to see how it improves the total time?

No, I haven't. I could do this for the dry run part by commenting out a bunch of code, but the syncs are the bigger win, and to be realistic I'd need to actually run this against the real core cluster (since it is affected by what is observed). IMO the changes in the flamegraphs make it realistic to expect an improvement, and I'd like to ship this and get the real numbers on how much and then iterate (or revert if it is useless) than lock Shipit and yolo deploy core from my laptop on this branch.

test/unit/kubernetes-deploy/concurrency_test.rb
KnVerey: True, this assertion couldn't possibly do anything. I can wrap this in an `assert_nothing_raised` instead I suppose.

kirs: `assert_nothing_raised` is just `yield` and some developers try to avoid it. Without `assert_nothing_raised ` if something raises, the test fails anyway.

ibawt: I dislike nothing_raised as well but it does show the intention.  ( this is the usual argument )

PR #146, opened 2017-07-26
Daemon Set Fix

karanthukral:
## What?
- In order to check for a successful deploy we were using `numberAvailable` from the ds status field
- `numberAvailable` is the number of pods that have been ready for at least `minReadySeconds`
- I have been seeing flakiness in the `numberAvailable` status being upgraded causing deploys to fail [here](https://shipit.shopify.io/shopify/k8s-cluster-services/production/deploys/466283)
  - Successive deploys have failed on different clusters, different portions of the deploy
  - In all failures I checked the status field of the daemon sets and `numberReady` was always up to date whereas `numberAvailable` was often off by one. Sometimes it would resolve itself whereas sometimes it never resolved or only resolved when I force killed a pod

## Other issues?
- tier1 has 180 nodes which makes the 5 min timeout too low due to the `minReadyRequirements`
- Also ran into a failure where `@pods` were empty for the timeout message and caused a failure

Not sure if this is the best way to fix it, open to ideas. In the meantime I'll keep 👀 

cc/ @KnVerey @kirs @Shopify/cloudplatform

### DIFF
--- a/lib/kubernetes-deploy/kubernetes_resource/daemon_set.rb
+++ b/lib/kubernetes-deploy/kubernetes_resource/daemon_set.rb
@@ -39,6 +39,7 @@ module KubernetesDeploy
     end
 
     def timeout_message
+      STANDARD_TIMEOUT_MESSAGE unless @pods.present?
       @pods.map(&:timeout_message).compact.uniq.join("\n")
     end
 
--- a/lib/kubernetes-deploy/kubernetes_resource/daemon_set.rb
+++ b/lib/kubernetes-deploy/kubernetes_resource/daemon_set.rb
@@ -1,7 +1,7 @@
 # frozen_string_literal: true
 module KubernetesDeploy
   class DaemonSet < KubernetesResource
-    TIMEOUT = 7.minutes
+    TIMEOUT = 5.minutes
 
     def sync
       raw_json, _err, st = kubectl.run("get", type, @name, "--output=json")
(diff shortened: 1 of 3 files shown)


karanthukral:
check
Even in checking the initial status of resources before the deploy, the `numberAvailable` was at 36 and `numberReady` at 37

lib/kubernetes-deploy/kubernetes_resource/daemon_set.rb
```diff
@@ -39,6 +39,7 @@ def failure_message
     end
 
     def timeout_message
+      STANDARD_TIMEOUT_MESSAGE unless @pods.present?
```
KnVerey: I think you meant to `return` this

Deployment, ReplicaSet and Pod have very similar issues, where they should be returning `STANDARD_TIMEOUT_MESSAGE` and instead are returning `nil` or exactly what we were doing here. Mind fixing them too?

KnVerey: Also, we should probably add a timeout test after all...

KnVerey:
Notes/todos from investigating the failed deploys:

- [ ] We need to change the RollingUpdate strategy on a bunch of our DS before shipping this again. `maxUnavailable: 2` is unreasonable for huge numbers of nodes; it should be a percentage.
- [ ] Some of our datadog DS seem to have actual stability issues. This is going to rightly interfere with deploys until we fix them. We have some ideas.
- [ ] I think we're looking at the wrong field when selecting pods. `DS#spec.Generation` matches up with `DS#status.observedGeneration` for making sure the status we're looking at is up to date, so that's fine. However, it seems to be `DS#spec.templateGeneration` that matches up with `Pod#metadata.labels.pod-template-generation`. We have a DS where Generation and templateGeneration are really different.
- [ ] We have one case where `numberAvailable` is down by one even though all the pods appear to have been ready for minutes (where min ready seconds is 5). I still think that `numberAvailable` is the most correct thing to look, but you're right that we'll have to switch if we can't figure out why this is / it seems unreliable.

karanthukral:
closing for https://github.com/Shopify/kubernetes-deploy/pull/148

PR #148, opened 2017-07-27
Re-Add Daemon Sets

karanthukral:
## What?
- Re-adds daemon set resource
- Notes/todos from investigating the failed deploys after previous attempt:

  - [x] We need to change the RollingUpdate strategy on a bunch of our DS before shipping this again. `maxUnavailable: 2` is unreasonable for huge numbers of nodes; it should be a percentage.
    - https://github.com/Shopify/k8s-cluster-services/pull/148
  - [ ] Some of our datadog DS seem to have actual stability issues. This is going to rightly interfere with deploys until we fix them. We have some ideas.
  - [x] I think we're looking at the wrong field when selecting pods. `DS#spec.Generation` matches up with `DS#status.observedGeneration` for making sure the status we're looking at is up to date, so that's fine. However, it seems to be `DS#spec.templateGeneration` that matches up with `Pod#metadata.labels.pod-template-generation`. We have a DS where Generation and templateGeneration are really different.
  - [x] We have one case where `numberAvailable` is down by one even though all the pods appear to have been ready for minutes (where min ready seconds is 5). I still think that `numberAvailable` is the most correct thing to look, but you're right that we'll have to switch if we can't figure out why this is / it seems unreliable.

### DIFF
--- a/lib/kubernetes-deploy/kubernetes_resource/daemon_set.rb
+++ b/lib/kubernetes-deploy/kubernetes_resource/daemon_set.rb
@@ -72,9 +72,10 @@ module KubernetesDeploy
       all_pods = JSON.parse(raw_json)["items"]
       template_generation = ds_data["spec"]["templateGeneration"]
 
-      latest_pods = all_pods.find_all do |pods|
-        pods["metadata"]["ownerReferences"].any? { |ref| ref["uid"] == ds_data["metadata"]["uid"] } &&
-        pods["metadata"]["labels"]["pod-template-generation"].to_i == template_generation.to_i
+      latest_pods = all_pods.find_all do |pod|
+        next unless owners = pod.dig("metadata", "ownerReferences")
+        owners.any? { |ref| ref["uid"] == ds_data["metadata"]["uid"] } &&
+        pod["metadata"]["labels"]["pod-template-generation"].to_i == template_generation.to_i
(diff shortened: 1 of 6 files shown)


lib/kubernetes-deploy/kubernetes_resource/daemon_set.rb
```diff
@@ -0,0 +1,95 @@
...
+    def find_pods(ds_data)
+      label_string = ds_data["spec"]["selector"]["matchLabels"].map { |k, v| "#{k}=#{v}" }.join(",")
+      raw_json, _err, st = kubectl.run("get", "pods", "-a", "--output=json", "--selector=#{label_string}")
+      return [] unless st.success?
+
+      all_pods = JSON.parse(raw_json)["items"]
+      template_generation = ds_data["spec"]["templateGeneration"]
+
+      latest_pods = all_pods.find_all do |pods|
+        pods["metadata"]["ownerReferences"].any? { |ref| ref["uid"] == ds_data["metadata"]["uid"] } &&
```
KnVerey: We just saw `ownerReferences` being nil on a pod that matched an RS's labels. Seems incredibly rare, but we should guard against that for DS too since we know it can happen: https://github.com/Shopify/kubernetes-deploy/pull/151.

Issue #139, opened 2017-07-19
Readiness probe message makes bad assumptions


KnVerey:
A test in @karanthukral PR points out that the `Pod` code is assuming the readinessProbes are HTTP:
![image](https://user-images.githubusercontent.com/4789493/28347391-c76b0b54-6c04-11e7-983e-711a1228482d.png)
^ This message makes no sense, oops! 😄 

I noticed today that that message can also get displayed when the probes are fine but the rollout was reeeallly slow so the container just happens to be starting. Maybe we should change the beginning to be less confident, e.g. "Your pods are running, but are not ready. Please make sure they're passing their readiness probes". And then either not push a probe-specific message if `probe_location` is blank, or adjust it to work / make sense for both types of probes. 

@karanthukral would you be interested in fixing this?

PR #149, opened 2017-07-27
Handle exec probes and fix container readiness check

KnVerey:
Fixes https://github.com/Shopify/kubernetes-deploy/issues/139

- Makes the timeout heading for readiness probe failures less confident, since it can surface when a huge deployment takes forever to roll out and one of its containers is still brand new (and could very well pass its probes)
- Makes the list of failed readiness probes support exec probes
- Fixes the readiness probe check to correctly detect success (was `if true == "true"` 😞 ), and adds a test assertion for this (I noticed in production that it was printing all the probes when I could see that one was passing)

cc @Shopify/cloudplatform


test/integration/kubernetes_deploy_test.rb
```diff
@@ -476,18 +476,19 @@ def test_pruning_of_existing_managed_secrets_when_ejson_file_has_been_deleted
   end
 
   def test_deploy_result_logging_for_mixed_result_deploy
```
KnVerey: image

Issue #83, opened 2017-04-21
Remove special handling of TPR creation/update


KnVerey:
As of v1.6, `kubectl apply` can handle TPRs properly.

PR #150, opened 2017-07-28
Remove support for Kubernetes v1.5

KnVerey:
Closes https://github.com/Shopify/kubernetes-deploy/issues/83

- Removes code we added to support 1.5 and 1.6 at the same time
- Removes special handling of TPR creation/update, which is no longer required as of 1.6. This will also make it possible to prune TPRs if we want to go that route.


KnVerey:
> Let's also make sure that we check k8s version in the script and require a minimum

We don't currently, and we presumably should for all three scripts. I'll make an issue.

PR #156, opened 2017-08-01
Profiling via integration tests

KnVerey:
This is the setup I used to create the flamegraphs for https://github.com/Shopify/kubernetes-deploy/pull/145. @ibawt @kirs wdyt about committing something like this to the repo?

cc @Shopify/cloudplatform

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -9,4 +9,6 @@ gem 'pry-byebug'
 gem 'rubocop'
 gem 'timecop'
 gem 'byebug'
-gem 'codecov', require: false, group: :test
+gem 'codecov', require: false
+gem 'ruby-prof', require: false
+gem 'ruby-prof-flamegraph', require: false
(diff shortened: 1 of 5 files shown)


Gemfile
```diff
@@ -10,3 +10,5 @@ gem 'rubocop'
 gem 'timecop'
 gem 'byebug'
 gem 'codecov', require: false, group: :test
+gem 'ruby-prof', require: false, group: :test
```
kirs: Gemfile in the gem repo is only used for dev and testing. It's totally separate from gemspec dependencies. I don't think using groups makes a lot of sense here.

dev/flamegraph-from-tests.sh
```diff
@@ -0,0 +1,35 @@
+#!/usr/bin/env bash
```
kirs: With the shebang, `.sh` in the file name is not necessary and can easily be ✂️

KnVerey:
Great points, thanks @kirs !

PR #165, opened 2017-09-11
Kubernetes 1.7 upgrade

KnVerey:
## Purpose 

This PR prepares kubernetes-deploy for version 1.7. I've run the tests against v1.7.4 locally, which only revealed the following very minor issues:
- `activeDeadlineSeconds` doesn't make sense on a deployment, and k8s now validates this, preventing a test's setup from working. It was a regression test trying to simulate dead pods hanging around as a result of evictions. Performing an actual eviction against minikube does not seem to leave the pods around. The value of this test is very low (we now ignore the old replicaset completely, and there is a test for the ability to run multiple revisions at once) and it is not obvious how to make it work without the activeDeadlineSeconds hack, so I'm removing it.
- Unmanaged pods that die immediately with `ContainerCannotRun` have their termination data under `state.terminated` instead of `lastState.terminated`, so we need to check both places. I'm not certain whether a managed pod can also end up with `ContainerCannotRun` in `state.terminated`, but  if it does, using that info so seems acceptable (even desirable) to me. Without this fix, the result is still correct and all of the information is still logged, but the termination reason isn't surfaced at the top of the output as it should be.

I plan to upgrade CI to 1.7 once we're actually running that version. If you think I should do that sooner, LMK. (Maybe we should set up the ability to run CI against two versions for a period...)

## Incidentals

- Add a helper to assert the return value of `deploy_fixtures` in a consistent and more helpful way (inspired by one of the test failures I debugged here originally being `expected true, actual false`)
- Make sure ImagePullPolicy is explicitly set to IfNotPresent where that is not the default (i.e. when tag is `:latest` or unspecified). I noticed in passing that it was set to `Always` in some cases, which could theoretically be slowing down CI.

cc @Shopify/cloudplatform

### DIFF
--- a/lib/kubernetes-deploy/kubernetes_resource/pod.rb
+++ b/lib/kubernetes-deploy/kubernetes_resource/pod.rb
@@ -179,15 +179,18 @@ module KubernetesDeploy
       end
 
       def doom_reason
-        exit_code = @status.dig('lastState', 'terminated', 'exitCode')
-        last_terminated_reason = @status.dig("lastState", "terminated", "reason")
         limbo_reason = @status.dig("state", "waiting", "reason")
         limbo_message = @status.dig("state", "waiting", "message")
 
-        if last_terminated_reason == "ContainerCannotRun"
+        if @status.dig("lastState", "terminated", "reason") == "ContainerCannotRun"
           # ref: https://github.com/kubernetes/kubernetes/blob/562e721ece8a16e05c7e7d6bdd6334c910733ab2/pkg/kubelet/dockershim/docker_container.go#L353
+          exit_code = @status.dig('lastState', 'terminated', 'exitCode')
           "Failed to start (exit #{exit_code}): #{@status.dig('lastState', 'terminated', 'message')}"
+        elsif @status.dig("state", "terminated", "reason") == "ContainerCannotRun"
+          exit_code = @status.dig('state', 'terminated', 'exitCode')
+          "Failed to start (exit #{exit_code}): #{@status.dig('state', 'terminated', 'message')}"
         elsif limbo_reason == "CrashLoopBackOff"
+          exit_code = @status.dig('lastState', 'terminated', 'exitCode')
           "Crashing repeatedly (exit #{exit_code}). See logs for more information."
         elsif %w(ImagePullBackOff ErrImagePull).include?(limbo_reason) &&
           limbo_message.match(/(?:not found)|(?:back-off)/i)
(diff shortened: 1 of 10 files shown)


lib/kubernetes-deploy/kubernetes_resource/pod.rb
```diff
@@ -179,15 +179,18 @@ def doomed?
...
-        last_terminated_reason = @status.dig("lastState", "terminated", "reason")
         limbo_reason = @status.dig("state", "waiting", "reason")
         limbo_message = @status.dig("state", "waiting", "message")
 
-        if last_terminated_reason == "ContainerCannotRun"
+        if @status.dig("lastState", "terminated", "reason") == "ContainerCannotRun"
           # ref: https://github.com/kubernetes/kubernetes/blob/562e721ece8a16e05c7e7d6bdd6334c910733ab2/pkg/kubelet/dockershim/docker_container.go#L353
+          exit_code = @status.dig('lastState', 'terminated', 'exitCode')
           "Failed to start (exit #{exit_code}): #{@status.dig('lastState', 'terminated', 'message')}"
+        elsif @status.dig("state", "terminated", "reason") == "ContainerCannotRun"
```
kirs: Can you add a comment that this is a 1.7 thing?

test/test_helper.rb
```diff
@@ -64,6 +64,20 @@ def teardown
       @logger_stream.close
     end
 
+    def assert_deploy_status(desired_status, result)
```
kirs: Since the result of deploy is boolean, there's not much value in introducing statuses like `failure` and `success`, especially if there are only two of them. WDYT about helpers like `assert_deploy_succeeds` and `assert_deploy_fails`?

lib/kubernetes-deploy/kubernetes_resource/pod.rb
KnVerey: I couldn't find the change by poking around in the source, but I wouldn't be surprised if it was a fix / not going to change in the future. Unmanaged pods aren't restarted, so having `lastState` populated doesn't really make sense.

KnVerey:
> I agree that it would be cool to have kubernetes-deploy tested on both 1.6 and 1.7. Do you think it will take a lot of time to implement on the CI side?

@stefanmb mentioned at standup today that he is working on a way to run minikube-based CI on Pipa. If that pans out, I imagine it'll be very easy to do. 😄

test/test_helper.rb
KnVerey: I like that suggestion. What I was trying to avoid was `assert_deploy_status(false, result)`, which I felt read poorly.

PR #168, opened 2017-09-19
Improve restart command

KnVerey:
Fixes https://github.com/Shopify/cloudplatform/issues/562
@Shopify/cloudplatform 

## Goals

- Make the restart command's exit status accurate (see [this deploy](https://shipit.shopify.io/shopify/facebook-commerce/staging/tasks/478249) for an example of the problem)
- Bring the restart command's output behaviour in line with the deploy command's
  - Task failures and user errors should generate friendly messages and exit cleanly
  - Final status/debug information should always be printed in a summary section
- Output statsd duration metrics for the restart task too

### Successful restart

screenshot

screenshot

### Bad configuration

screenshot

screenshot

### Failed restart

screenshot

### DIFF
--- a/kubernetes-deploy.gemspec
+++ b/kubernetes-deploy.gemspec
@@ -24,7 +24,7 @@ Gem::Specification.new do |spec|
 
   spec.required_ruby_version = '>= 2.3.0'
   spec.add_dependency "activesupport", ">= 4.2"
-  spec.add_dependency "kubeclient", "~> 2.3"
+  spec.add_dependency "kubeclient", "~> 2.4"
   spec.add_dependency "googleauth", ">= 0.5"
   spec.add_dependency "ejson", "1.0.1"
   spec.add_dependency "colorize", "~> 0.8"
--- a/lib/kubernetes-deploy/resource_watcher.rb
+++ b/lib/kubernetes-deploy/resource_watcher.rb
@@ -1,7 +1,7 @@
 # frozen_string_literal: true
 module KubernetesDeploy
   class ResourceWatcher
-    def initialize(resources, logger:, deploy_started_at: Time.now.utc)
+    def initialize(resources, logger:, deploy_started_at: Time.now.utc, operation_name: "deploy")
       unless resources.is_a?(Enumerable)
         raise ArgumentError, <<~MSG
           ResourceWatcher expects Enumerable collection, got `#{resources.class}` instead
@@ -10,30 +10,33 @@ module KubernetesDeploy
       @resources = resources
       @logger = logger
       @deploy_started_at = deploy_started_at
+      @operation_name = operation_name
     end
 
-    def run(delay_sync: 3.seconds, reminder_interval: 30.seconds)
+    def run(delay_sync: 3.seconds, reminder_interval: 30.seconds, record_summary: true)
       delay_sync_until = last_message_logged_at = Time.now.utc
+      remainder = @resources.dup
 
-      while @resources.present?
+      while remainder.present?
         if Time.now.utc < delay_sync_until
           sleep(delay_sync_until - Time.now.utc)
         end
         delay_sync_until = Time.now.utc + delay_sync # don't pummel the API if the sync is fast
 
-        KubernetesDeploy::Concurrency.split_across_threads(@resources, &:sync)
-        newly_finished_resources, @resources = @resources.partition(&:deploy_finished?)
+        KubernetesDeploy::Concurrency.split_across_threads(remainder, &:sync)
+        newly_finished_resources, remainder = remainder.partition(&:deploy_finished?)
 
         if newly_finished_resources.present?
           watch_time = (Time.now.utc - @deploy_started_at).round(1)
           report_what_just_happened(newly_finished_resources, watch_time)
-          report_what_is_left(reminder: false)
+          report_what_is_left(remainder, reminder: false)
           last_message_logged_at = Time.now.utc
         elsif due_for_reminder?(last_message_logged_at, reminder_interval)
-          report_what_is_left(reminder: true)
+          report_what_is_left(remainder, reminder: true)
           last_message_logged_at = Time.now.utc
         end
       end
+      record_statuses_for_summary(@resources) if record_summary
     end
 
     private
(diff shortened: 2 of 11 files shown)


lib/kubernetes-deploy/deferred_summary_logging.rb
```diff
@@ -33,6 +33,23 @@ def heading(text, secondary_msg = '', secondary_msg_color = :cyan)
       info(part1 + part2 + part3)
     end
 
+    def record_statuses(resources, operation_name: "deploy")
```
KnVerey: It feels kinda wrong to put this in the logger's summary module, since it knows about methods on KubernetesResources. I'm not sure where else to put it though; I'd like the restart and deploy task copies to stay in sync.

lib/kubernetes-deploy/restart_task.rb
```diff
@@ -27,58 +23,71 @@ def initialize(context:, namespace:, logger:)
       @context = context
       @namespace = namespace
       @logger = logger
-      @kubeclient = build_v1_kubeclient(context)
-      @v1beta1_kubeclient = build_v1beta1_kubeclient(context)
-      @policy_v1beta1_kubeclient = build_policy_v1beta1_kubeclient(context)
```
KnVerey: I moved to initializing these lazily because they can raise. If the kubeconfig is invalid, I want to print a pretty error about it, not an exception (like deploys do).

test/integration/restart_task_test.rb
```diff
@@ -117,8 +162,45 @@ def test_restart_not_existing_namespace
...
+    ],
+      in_order: true)
+  end
+
+  def test_restart_failure
+    success = deploy_fixtures("hello-cloud", subset: ["configmap-data.yml", "web.yml.erb"]) do |fixtures|
+      deployment = fixtures["web.yml.erb"]["Deployment"].first
+      deployment["spec"]["progressDeadlineSeconds"] = 12
+      container = deployment["spec"]["template"]["spec"]["containers"].first
+      container["readinessProbe"] = {
```
KnVerey: Alternative ideas for simulating a failed restart greatly appreciated!

lib/kubernetes-deploy/resource_watcher.rb
```diff
@@ -10,30 +10,33 @@ def initialize(resources, logger:, deploy_started_at: Time.now.utc)
       @resources = resources
       @logger = logger
       @deploy_started_at = deploy_started_at
+      @operation_name = operation_name
     end
 
-    def run(delay_sync: 3.seconds, reminder_interval: 30.seconds)
+    def run(delay_sync: 3.seconds, reminder_interval: 30.seconds, record_summary: true)
       delay_sync_until = last_message_logged_at = Time.now.utc
+      remainder = @resources.dup
```
stefanmb: Was using the original object a bug?

lib/kubernetes-deploy/restart_task.rb
```diff
@@ -27,58 +23,71 @@ def initialize(context:, namespace:, logger:)
...
+      ResourceWatcher.new(resources, logger: @logger, operation_name: "restart").run
+      success = resources.all?(&:deploy_succeeded?)
+    rescue FatalDeploymentError => error
+      @logger.summary.add_action(error.message)
+      success = false
+    ensure
+      @logger.print_summary(success)
+      status = success ? "success" : "failed"
+      tags = %W(namespace:#{@namespace} context:#{@context} status:#{status} deployments:#{deployments.to_a.length}})
+      ::StatsD.measure('restart.duration', StatsD.duration(start), tags: tags)
```
stefanmb: 👍  on the stats

lib/kubernetes-deploy/restart_task.rb
```diff
@@ -27,58 +23,71 @@ def initialize(context:, namespace:, logger:)
...
-      @kubeclient = build_v1_kubeclient(context)
-      @v1beta1_kubeclient = build_v1beta1_kubeclient(context)
-      @policy_v1beta1_kubeclient = build_policy_v1beta1_kubeclient(context)
     end
 
     def perform(deployments_names = nil)
+      start = Time.now.utc
       @logger.reset
+
+      @logger.phase_heading("Initializing retart")
```
mkobetic: `retart`?

test/integration/restart_task_test.rb
stefanmb: This seems reasonable to me.

lib/kubernetes-deploy/resource_watcher.rb
KnVerey: It would have been a bug if anything tried to reuse the same ResourceWatcher instance (but we don't). Without this change, `record_statuses_for_summary` wouldn't record anything, as `@resources` was always empty by the end of the method. Technically I don't need to `dup` here as long as the body uses a separate variable, since the variable is being replaced (not mutated) by `newly_finished_resources, remainder = remainder.partition(&:deploy_finished?)`.

test/integration/restart_task_test.rb
```diff
@@ -117,8 +162,45 @@ def test_restart_not_existing_namespace
...
+      "Result: FAILURE",
+      "Namespace `walrus` not found in context `minikube`"
+    ],
+      in_order: true)
+  end
+
+  def test_restart_failure
+    success = deploy_fixtures("hello-cloud", subset: ["configmap-data.yml", "web.yml.erb"]) do |fixtures|
+      deployment = fixtures["web.yml.erb"]["Deployment"].first
+      deployment["spec"]["progressDeadlineSeconds"] = 12
```
kirs: Why 12?

kubernetes-deploy.gemspec
```diff
@@ -24,7 +24,7 @@ Gem::Specification.new do |spec|
 
   spec.required_ruby_version = '>= 2.3.0'
   spec.add_dependency "activesupport", ">= 4.2"
-  spec.add_dependency "kubeclient", "~> 2.3"
+  spec.add_dependency "kubeclient", "~> 2.4"
```
kirs: Looks like not much changed: https://github.com/abonas/kubeclient/compare/v2.3.0...v2.4.0

lib/kubernetes-deploy/restart_task.rb
```diff
@@ -27,58 +23,71 @@ def initialize(context:, namespace:, logger:)
...
+      @logger.phase_heading("Triggering restart by touching ENV[RESTARTED_AT]")
+      patch_kubeclient_deployments(deployments)
+
+      @logger.phase_heading("Waiting for rollout")
+      resources = build_watchables(deployments, start)
+      ResourceWatcher.new(resources, logger: @logger, operation_name: "restart").run
+      success = resources.all?(&:deploy_succeeded?)
+    rescue FatalDeploymentError => error
+      @logger.summary.add_action(error.message)
+      success = false
```
kirs: The flow is what variable is returned is a bit opaque.
Do you think this version would be a bit better readable?

```ruby
def perform
  # do things that might raise
  statsd_report(start, status: "success")
  true
rescue =>
  statsd_report(start, status: "failed")
  false
ensure
  @logger.print_summary
end
```

kubernetes-deploy.gemspec
KnVerey: Yeah, I was hoping they might have released a version with your NotFound error class. That's not the case, but I figured we might as well update anyway.

test/integration/restart_task_test.rb
KnVerey: Trial and error... just trying to pick a number that allows a bit of time for the expected logs/events to be reported but doesn't make the test too slow.

lib/kubernetes-deploy/restart_task.rb
KnVerey: That won't work as is because `print_summary` needs to know the success status, and the condition that didn't raise might still have failed (that's why it does `success = resources.all?(&:deploy_succeeded?)`). So that'd end up as:

```ruby
      report_statsd(start, success: success, num_deployments: deployments.to_a.length)
      success
      @logger.summary.add_action(error.message)
      report_statsd(start, success: false, num_deployments: deployments.to_a.length)
      success = false
    ensure
    end
```

Which seems pretty much the same to me, except that the statsD no longer leverages `status`. What do you find unclear about it? I note that the return value of this is now heavily tested.

KnVerey:
> Would a test fail if some future code change makes kubernetes-restart exit with 0 again?

It wasn't the exit code that was broken, it was the return value of `perform`. All of the tests now assert that value, and there's new coverage of the scenario where the restart rollout fails for some reason.

KnVerey:
The CI pipeline seems to be broken and we don't want to spend time fixing it because we're in the middle of migrating to our new setup with Pipa, which will support multiple k8s versions out of the box. Unfortunately the new setup isn't stable enough to enable yet. Stefan and I are actively working on it, but I don't want to hold up this PR any longer though, as we have several apps waiting on it. Here's a local test run:

image

I'll do the same on master after merging before tagging and releasing.

KnVerey:
On master
image

image

PR #172, opened 2017-09-21
Add ResourceQuotas to predeploy sequence

karanthukral:
## What?
- Add `ResourceQuota` to the pre-deploy sequence to make sure its updated before any pods/deployments are deployed

### DIFF
--- a/lib/kubernetes-deploy/kubernetes_resource/resource_quota.rb
+++ b/lib/kubernetes-deploy/kubernetes_resource/resource_quota.rb
@@ -4,13 +4,18 @@ module KubernetesDeploy
     TIMEOUT = 30.seconds
 
     def sync
-      _, _err, st = kubectl.run("get", type, @name)
+      raw_json, _err, st = kubectl.run("get", type, @name, "--output=json")
       @status = st.success? ? "Available" : "Unknown"
       @found = st.success?
+      @rollout_data = if @found
+        JSON.parse(raw_json)
+      else
+        {}
+      end
     end
 
     def deploy_succeeded?
-      exists?
+      @rollout_data["spec"]["hard"] == @rollout_data["status"]["hard"]
     end
 
     def deploy_failed?
--- a/lib/kubernetes-deploy/kubernetes_resource/resource_quota.rb
+++ b/lib/kubernetes-deploy/kubernetes_resource/resource_quota.rb
@@ -1,0 +1,28 @@
+# frozen_string_literal: true
+module KubernetesDeploy
+  class ResourceQuota < KubernetesResource
+    TIMEOUT = 30.seconds
+
+    def sync
+      _, _err, st = kubectl.run("get", type, @name)
+      @status = st.success? ? "Available" : "Unknown"
+      @found = st.success?
+    end
+
+    def deploy_succeeded?
+      exists?
+    end
+
+    def deploy_failed?
+      false
+    end
+
+    def timeout_message
+      UNUSUAL_FAILURE_MESSAGE
+    end
+
+    def exists?
+      @found
+    end
+  end
+end
--- a/lib/kubernetes-deploy/kubernetes_resource/resource_quota.rb
+++ b/lib/kubernetes-deploy/kubernetes_resource/resource_quota.rb
@@ -15,7 +15,7 @@ module KubernetesDeploy
     end
 
     def deploy_succeeded?
-      @rollout_data["spec"]["hard"] == @rollout_data["status"]["hard"]
+      @rollout_data.dig("spec", "hard") == @rollout_data.dig("status", "hard")
     end
 
     def deploy_failed?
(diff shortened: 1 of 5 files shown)


karanthukral:
@KnVerey for 👀 . I ran the tests locally and the results are below

branch-test

test/integration/kubernetes_deploy_test.rb
```diff
@@ -635,6 +635,19 @@ def test_bad_container_on_daemon_sets_fails
...
 
+  def test_resource_quotas_are_deployed_first
+    forced_timeout = 10
+    KubernetesDeploy::Deployment.any_instance.stubs(:timeout).returns(forced_timeout)
+    result = deploy_fixtures("resource-quota")
+    assert_deploy_failure(result)
+    assert_logs_match_all([
+      "Predeploying priority resources",
+      "Deploying ResourceQuota/resource-quotas (timeout: 30s)",
+      "Deploying all resources",
```
KnVerey: I'd suggest replacing this assertion with one for `ResourceQuota/resource-quotas                     Available` (so this test also proves the quota part of the deploy succeeded), and adding an additional assertion at the end on the event to show that the web failure had to do with the quota. Otherwise, the test could pass because of the forced short timeout even if the web rollout were eventually going to succeed.

Sidenote: it sucks that this deploy times out, since we should be able to tell right away that it is doomed. Would you be willing to explore (as a follow-up, not in this PR) making the deployment itself fail fast on observing the related status condition?
```yaml
    - lastTransitionTime: 2017-09-25T17:04:07Z
      lastUpdateTime: 2017-09-25T17:04:07Z
      message: 'unable to create pods: pods "web-3393528609-" is forbidden: failed
        quota: resource-quotas: must specify limits.cpu,limits.memory,requests.cpu,requests.memory'
      reason: FailedCreate
      status: "True"
      type: ReplicaFailure
```

lib/kubernetes-deploy/kubernetes_resource/resource_quota.rb
```diff
@@ -0,0 +1,28 @@
...
+    TIMEOUT = 30.seconds
+
+    def sync
+      _, _err, st = kubectl.run("get", type, @name)
+      @status = st.success? ? "Available" : "Unknown"
+      @found = st.success?
+    end
+
+    def deploy_succeeded?
+      exists?
```
KnVerey: I just noticed that this resource type [has a status](https://kubernetes.io/docs/api-reference/v1.6/#resourcequota-v1-core): `Status defines the actual enforced quota and its current usage`. Based on your experience, do you think it would be reasonable to define success as `spec.hard == status.hard`?

test/integration/kubernetes_deploy_test.rb
```diff
@@ -635,6 +635,19 @@ def test_bad_container_on_daemon_sets_fails
...
+      "Deployment/web deployment timed out"
+    ], in_order: true)
+  end
```
KnVerey: Let's also fetch the quota and assert something about its spec and/or status since we don't have a separate test for that.

lib/kubernetes-deploy/kubernetes_resource/resource_quota.rb
karanthukral: I think that would work. I am not entirely sure if its guaranteed to be in the same units though

karanthukral:
Ran tests again with the changes and 👍

lib/kubernetes-deploy/kubernetes_resource/resource_quota.rb
```diff
@@ -0,0 +1,33 @@
...
+      @found = st.success?
+      @rollout_data = if @found
+        JSON.parse(raw_json)
+      else
+        {}
+      end
+    end
+
+    def deploy_succeeded?
+      @rollout_data["spec"]["hard"] == @rollout_data["status"]["hard"]
```
KnVerey: You'll need to `dig` here since `@rollout_data` can be an empty hash

test/integration/kubernetes_deploy_test.rb
```diff
@@ -635,6 +635,29 @@ def test_bad_container_on_daemon_sets_fails
...
+      "Deployment/web: TIMED OUT (limit: 10s)",
+      "failed quota: resource-quotas"
+    ], in_order: true)
+
+    rqs = kubeclient.get_resource_quotas(namespace: @namespace)
+    assert_equal 1, rqs.length
+
+    rq = rqs[0]
+    assert_equal "resource-quotas", rq["metadata"]["name"]
+    assert rq["spec"].present?
```
KnVerey: nbd but it'd be nice to verify the success condition a little more specifically here, e.g. compare spec.hard.pods to status.hard.pods or something.

karanthukral: That gets checked during the success deploy though. I can do that here too but I thought just making sure the spec was present would be useful enough to check for existence

lib/kubernetes-deploy/kubernetes_resource/resource_quota.rb
karanthukral: Fixed :)

PR #174, opened 2017-09-25
Fix logging assertion

karanthukral:
## What?
- Turns out I was not on latest master before
- Fixes logging assertion

cc/ @KnVerey


KnVerey:
image

:shipit:

PR #177, opened 2017-09-26
Better error message when $KUBECONFIG not set

dturn:
Currently if $KUBECONFIG is not get you get an error message like:
```
[localhost][err] [FATAL][2017-09-26 09:39:15 -0400][context][namespace]    Configuration invalid
[localhost][err] [FATAL][2017-09-26 09:39:15 -0400][context][namespace]    - Kube config not found at
```

This will change the second line to `"$KUBECONFIG not set"` which should be easier to understand.


KnVerey:
Change looks good, but could you please add test coverage for this in test/unit/kubernetes-deploy/runner_test.rb?

dturn:
Test Added

PR #180, opened 2017-09-27
Ensure ejson secrets respect the no-prune flag

karanthukral:
## What?
- Currently the ejson secret provisioner does not respect the no-prune flag
  - This can be a huge pain since the deploy can delete secrets from an apps namespace
- This PR adds a prune variable to the provisioner and respects the `--no-prune` flag passed to the runner

cc/ @Shopify/cloudplatform

### DIFF
--- a/lib/kubernetes-deploy/ejson_secret_provisioner.rb
+++ b/lib/kubernetes-deploy/ejson_secret_provisioner.rb
@@ -32,11 +32,7 @@ module KubernetesDeploy
 
     def run
       create_secrets
-      if @prune
-        prune_managed_secrets
-      else
-        @logger.summary.add_action("no secrets were pruned")
-      end
+      prune_managed_secrets if @prune
     end
 
     private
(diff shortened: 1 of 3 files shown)


lib/kubernetes-deploy/ejson_secret_provisioner.rb
```diff
@@ -31,7 +32,11 @@ def secret_changes_required?
 
     def run
       create_secrets
-      prune_managed_secrets
+      if @prune
+        prune_managed_secrets
+      else
+        @logger.summary.add_action("no secrets were pruned")
```
KnVerey: IMO this isn't necessary; we don't usually record things we _didn't_ do, and there's no similar log when we skip pruning of regular resources. Also, this is just reporting the setting; it's possible there was nothing to prune in any case. If you feel we should log something about pruning disabled (could be useful, but I don't feel strongly either way), I think it should be logged in the "initializing deploy" section.

karanthukral: I'll remove that and update the test

Issue #128, opened 2017-07-13
Generic Custom Resource Definition support


KnVerey:
Instead of hardcoding details about what Shopify's TPR controllers are expected to create, we should provide generic support based on a conventional status field. According to @wfarr investigation, such a status field will be easier to implement in the new `CustomResourceDefinition`s.

- We'd query the cluster for valid CRD types at the beginning of the deploy, before creating resource instances.
- CRD instances would be expected to expose status fields from which we can derive their success/failure
  - We should have a convention for this that lines up with the fields/values first-party controllers set
  - We can provide an override mechanism, likely an annotation, for specifying custom fields/values

wfarr:
I need to dig more in practice, but my belief on looking at these particular bits of code lead me to believe that CRDs offer a true status sub-resource that can be called directly, avoiding some of the pains of circular updates:

* https://github.com/kubernetes/kubernetes/blob/04477ff90ceb55b864a4ac96fcc8c8b510414ead/staging/src/k8s.io/apiextensions-apiserver/pkg/apiserver/apiserver.go#L125-L133
* https://github.com/kubernetes/kubernetes/blob/04477ff90ceb55b864a4ac96fcc8c8b510414ead/staging/src/k8s.io/apiextensions-apiserver/pkg/registry/customresourcedefinition/etcd.go#L150-L173
* https://github.com/kubernetes/apiextensions-apiserver/blob/master/examples/client-go/main.go#L78-L91
* https://github.com/kubernetes/apiextensions-apiserver/blob/master/examples/client-go/apis/cr/v1/types.go
* https://github.com/kubernetes/apiextensions-apiserver/blob/master/examples/client-go/client/cr.go#L97-L99
* https://github.com/kubernetes/apiextensions-apiserver/blob/master/examples/client-go/controller/controller.go#L96-L113

PR #181, opened 2017-09-28
Add memcached resource

devonboyer:
## What 

Add memcached resource

### DIFF
--- a/lib/kubernetes-deploy/kubernetes_resource/cloudsql.rb
+++ b/lib/kubernetes-deploy/kubernetes_resource/cloudsql.rb
@@ -62,9 +62,9 @@ module KubernetesDeploy
     def cloudsql_resource_uuid
       return @cloudsql_resource_uuid if defined?(@cloudsql_resource_uuid) && @cloudsql_resource_uuid
 
-      redis, _err, st = kubectl.run("get", "cloudsqls", @name, "-o=json")
+      cloudsql, _err, st = kubectl.run("get", "cloudsqls", @name, "-o=json")
       if st.success?
-        parsed = JSON.parse(redis)
+        parsed = JSON.parse(cloudsql)
 
         @cloudsql_resource_uuid = parsed.dig("metadata", "uid")
       end
--- a/lib/kubernetes-deploy/kubernetes_resource/memcached.rb
+++ b/lib/kubernetes-deploy/kubernetes_resource/memcached.rb
@@ -1,0 +1,78 @@
+# frozen_string_literal: true
+module KubernetesDeploy
+  class Memcached < KubernetesResource
+    TIMEOUT = 5.minutes
+    SECRET_NAME = "memcached-url"
+
+    def sync
+      _, _err, st = kubectl.run("get", type, @name)
+      @found = st.success?
+      @deployment_exists = memcached_deployment_exists?
+      @service_exists = memcached_service_exists?
+      @secret_exists = memcached_secret_exists?
+      
+      @status = if @deployment_exists && @service_exists && @secret_exists
+        "Provisioned"
+      else
+        "Unknown"
+      end
+    end
+
+    def deploy_succeeded?
+      @deployment_exists && @service_exists && @secret_exists
+    end
+
+    def deploy_failed?
+      false
+    end
+
+    def exists?
+      @found
+    end
+
+    private
+
+    def memcached_deployment_exists?
+      deployment, _err, st = kubectl.run("get", "deployments", "-l \"name=#{@name}\"", "-o=json")
+
+      if st.success?
+        parsed = JSON.parse(deployment)
+
+        if parsed.fetch("status", {}).fetch("availableReplicas", -1) == parsed.fetch("status", {}).fetch("replicas", 0)
+          # all memcached pods are running
+          return true
+        end
+      end
+
+      false
+    end
+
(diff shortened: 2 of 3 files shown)


lib/kubernetes-deploy/kubernetes_resource/cloudsql.rb
```diff
@@ -62,9 +62,9 @@ def mysql_service_exists?
     def cloudsql_resource_uuid
       return @cloudsql_resource_uuid if defined?(@cloudsql_resource_uuid) && @cloudsql_resource_uuid
 
-      redis, _err, st = kubectl.run("get", "cloudsqls", @name, "-o=json")
+      cloudsql, _err, st = kubectl.run("get", "cloudsqls", @name, "-o=json")
```
stefanmb: Thanks!

test/helpers/fixture_sets/hello_cloud.rb
```diff
@@ -10,6 +10,7 @@ def assert_all_up
       assert_unmanaged_pod_statuses("Succeeded")
       assert_all_web_resources_up
       assert_all_redis_resources_up
+      assert_all_memcached_resources_up
```
KnVerey: This doesn't actually test your code added at all, since the fixtures you added contain a regular deployment/service (which already have plenty of coverage). Our CI is not running Shopify's closed-source controllers, so there sadly isn't a way to add tests for TPR/CRD support. As a result, changes to TPR/CRD support need to be manually tophatted against tierstaging; if you need pointers for that, please ping me on slack. Our long-term solution for this (which we'll be able to get started on very soon) is to make the CRD support generic: https://github.com/Shopify/kubernetes-deploy/issues/128.

devonboyer: Okay good to know 👍

lib/kubernetes-deploy/kubernetes_resource/memcached.rb
```diff
@@ -0,0 +1,78 @@
...
+
+      if st.success?
+        parsed = JSON.parse(secret)
+
+        if parsed.dig("data", @name).present?
+          return true
+        end
+      end
+
+      false
```
fbogsany:
```ruby
secret, ... = ...
return false unless st.success?
parsed = JSON.parse(secret)
parsed.dig(“data”, @name).present?
```
Same pattern for the other methods.

stefanmb: Agreed, but if we're going to change it here it should also be changed in the `cloudsql` resource, `redis` resource and wherever else this pattern was used.

devonboyer:
@KnVerey 

🎩 with `tierstaging`:

screen shot 2017-10-02 at 2 53 26 pm

lib/kubernetes-deploy/kubernetes_resource/memcached.rb
```diff
@@ -0,0 +1,56 @@
+# frozen_string_literal: true
+module KubernetesDeploy
+  class Memcached < KubernetesResource
```
stefanmb: Can we follow up with a PR refactoring the existing resources to follow the same more idiomatic patterns? I think consistency in this codebase is important.

devonboyer: https://github.com/Shopify/kubernetes-deploy/issues/184

I will take care of doing that.

lib/kubernetes-deploy/kubernetes_resource/memcached.rb
```diff
@@ -0,0 +1,60 @@
...
+    end
+
+    def exists?
+      @found
+    end
+
+    private
+
+    def memcached_deployment_exists?
+      deployments, _err, st = kubectl.run("get", "deployments", "-l name=#{@name}", "-o=json")
```
devonboyer: Swapped to using a label selector to verify the deployment exists instead. This prevents us from being forced to name the `Service` and `Deployment` exactly what the template provides or have knowledge we shouldn't about how things are named.

@stefanmb

stefanmb: I think that makes sense! Could I ask you to add this to the refactor list for `redis` and `cloudsql` as well? Outside this PR, of course.

devonboyer: I will add that to the issue but we will need to make additional changes to make sure the `Redis` and  `Cloudsql` TPR controllers actually add this label. The `Memcached` controller **does** add this label.

lib/kubernetes-deploy/kubernetes_resource/memcached.rb
```diff
@@ -0,0 +1,61 @@
+# frozen_string_literal: true
+module KubernetesDeploy
+  class Memcached < KubernetesResource
+    TIMEOUT = 5.minutes
```
KnVerey: @devonboyer This is a very high timeout. Is it possible for this resource to take this long to come up under common circumstances?

devonboyer: No that is probably to high. I just did the same thing as `redis`/`cloudsql` which also use seemingly crazy timeouts. I have an issue to refactor these TPR resources so I will add this to the list.

PR #188, opened 2017-10-17
Runtime resource discovery

stefanmb:
**WARNING: Some of this description is now obsolete (Jan26/2018)**

**What**
This PR introduces automatic discovery of CustomResourceDefinitions (1.7+) and ThirdPartyResources (1.6) in a backwards compatible manner. Once this PR is implemented, the monitoring logic for these resources can be removed from `kubernetes-deploy`.

**Motivation**
Currently we are handling each resource type by adding a new subclass of `KubernetesResource` to handle its state ([as per the documentation](https://github.com/Shopify/kubernetes-deploy#contributing-a-new-resource-type)).

This approach makes sense for core resources, but is not scalable for `CustomResourceDefinitions` (and `ThirdPartyResources`) for multiple reasons:

- We are including logic to handle CRDs such as `cloudsql` and `redis` which are not open sourced.
- There is too tight a coupling between our buddy/controller framework (closed source) and `k8s-deploy`
- We cannot support 3rd party use cases unless they contribute a resource definition back to `k8s-deploy` - we cannot centralize all logic for all possible resources users may wish to implement.

The goals of the PR are:
- Automatically detect custom resources present in the cluster at deploy time
- Decouple the concern of determining a custom resource's status from `k8s-deploy` by enabling controllers to annotate resources with their respective statuses.
- Remain **fully backward compatible** so that controllers can be upgraded with annotation support incrementally, and the CloudSQL, Redis, etc subclasses of `KubernetesResource` can be removed one at a time.
- Support both TPR and CRD custom resources.
 
**How**
`TODO`

**TODO**
Update `README` file.

**Review**
cc: @Shopify/cloudplatform @KnVerey

### DIFF
--- a/bin/test
+++ b/bin/test
@@ -28,8 +28,13 @@ if [[ ${PARALLELISM:=0} -lt 1 ]]; then
 fi
 
 if [[ ${CI:="0"} == "1" ]]; then
+  echo "--- :linux: Install jq"
+  wget -qP /usr/local/bin http://stedolan.github.io/jq/download/linux64/jq && chmod a+x /usr/local/bin/jq
+fi
+
+if [[ ${CI:="0"} == "1" ]]; then
   echo "--- :ruby: Bundle Install"
-  bundle install --jobs 4
+  bundle install --jobs 4 || exit 1
 fi
 
 print_header "Run Unit Tests"
--- a/bin/test
+++ b/bin/test
@@ -28,11 +28,6 @@ if [[ ${PARALLELISM:=0} -lt 1 ]]; then
 fi
 
 if [[ ${CI:="0"} == "1" ]]; then
-  echo "--- :linux: Install jq"
-  wget -qP /usr/local/bin http://stedolan.github.io/jq/download/linux64/jq && chmod a+x /usr/local/bin/jq
-fi
-
-if [[ ${CI:="0"} == "1" ]]; then
   echo "--- :ruby: Bundle Install"
   bundle install --jobs 4 || exit 1
 fi
--- a/kubernetes-deploy.gemspec
+++ b/kubernetes-deploy.gemspec
@@ -30,6 +30,7 @@ Gem::Specification.new do |spec|
   spec.add_dependency "colorize", "~> 0.8"
   spec.add_dependency "statsd-instrument", "~> 2.1"
   spec.add_dependency "rgl", "0.5.3"
+  spec.add_dependency "jsonpath", "0.8.8"
 
   spec.add_development_dependency "bundler"
   spec.add_development_dependency "rake", "~> 10.0"
--- a/kubernetes-deploy.gemspec
+++ b/kubernetes-deploy.gemspec
@@ -29,6 +29,7 @@ Gem::Specification.new do |spec|
   spec.add_dependency "ejson", "1.0.1"
   spec.add_dependency "colorize", "~> 0.8"
   spec.add_dependency "statsd-instrument", "~> 2.1"
+  spec.add_dependency "rgl", "0.5.3"
 
   spec.add_development_dependency "bundler"
   spec.add_development_dependency "rake", "~> 10.0"
--- a/lib/kubernetes-deploy/deploy_task.rb
+++ b/lib/kubernetes-deploy/deploy_task.rb
@@ -4,7 +4,11 @@ require 'yaml'
 require 'shellwords'
 require 'tempfile'
 require 'fileutils'
+require 'rgl/adjacency'
+require 'rgl/topsort'
+require 'kubernetes-deploy/discoverable_resource'
 require 'kubernetes-deploy/kubernetes_resource'
+
 %w(
   cloudsql
   config_map
@@ -27,6 +31,7 @@ require 'kubernetes-deploy/kubernetes_resource'
   bucket
   stateful_set
   cron_job
+  customresourcedefinition
 ).each do |subresource|
   require "kubernetes-deploy/kubernetes_resource/#{subresource}"
 end
@@ -40,16 +45,6 @@ module KubernetesDeploy
   class DeployTask
     include KubeclientBuilder
 
-    PREDEPLOY_SEQUENCE = %w(
-      ResourceQuota
-      Cloudsql
-      Redis
-      Memcached
-      ConfigMap
-      PersistentVolumeClaim
-      ServiceAccount
-      Pod
-    )
     PROTECTED_NAMESPACES = %w(
       default
       kube-system
@@ -64,23 +59,48 @@ module KubernetesDeploy
     # extensions/v1beta1/ReplicaSet -- managed by deployments
     # core/v1/Secret -- should not committed / managed by shipit
     def prune_whitelist
-      wl = %w(
-        core/v1/ConfigMap
-        core/v1/Pod
-        core/v1/Service
-        core/v1/ResourceQuota
-        batch/v1/Job
-        extensions/v1beta1/DaemonSet
-        extensions/v1beta1/Deployment
-        apps/v1beta1/Deployment
-        extensions/v1beta1/Ingress
-        apps/v1beta1/StatefulSet
-        autoscaling/v1/HorizontalPodAutoscaler
-      )
-      if server_version >= Gem::Version.new('1.8.0')
-        wl << "batch/v1beta1/CronJob"
+      @prune_whitelist ||= _build_prune_whitelist
+    end
+
+    def _build_prune_whitelist
+      prunable_resources = all_resources.select(&:prunable?)
+      prunable_resources.map(&:qualified_kind)
+    end
+
+    def predeploy_sequence
+      @predeploy_sequence ||= _build_predeploy_sequence
+    end
+
+    def _build_predeploy_sequence
+      # Express dependencies as DAG
+      graph = RGL::DirectedAdjacencyGraph.new
+      graph.add_vertex(:ROOT_NODE)
+
+      # This is a partially ordered set, so we must make sure everything is reachable:
+      predeploy_resources = all_resources.select(&:predeploy?)
+      predeploy_resources.each { |res| graph.add_edge(:ROOT_NODE, res.kind) }
+
+      # Find resources that have explicit predeploy (inter-)dependencies:
+      predeploy_res_with_deps = all_resources.select(&:predeploy_dependencies)
+      predeploy_res_with_deps.each do |res|
+        # Edge [A,B] means B requires A to be deployed first
+        res.predeploy_dependencies.each { |dep| graph.add_edge(dep, res.kind) }
       end
-      wl
+
+      raise FatalDeploymentError, "Cyclic predeploy requirements: #{graph.cycles.flatten}" unless graph.cycles.empty?
+
+      # Topological sort is not unique, but will respect the requirements
+      predeploy_order = graph.topsort_iterator.to_a
+      predeploy_order.delete(:ROOT_NODE)
+      predeploy_order
+    end
+
+    def all_resources
+      resources = DiscoverableResource.all + KubernetesResource.all
+      # Omit unqualified kinds (they are unsupported on this cluster, or discovery hasn't been performed yet)
+      resources.select do |res|
+        res.constants.include?(:GROUP) && res.constants.include?(:VERSION)
+      end
     end
 
     def server_version
@@ -124,7 +144,8 @@ module KubernetesDeploy
       confirm_context_exists
       confirm_namespace_exists
       @namespace_tags |= tags_from_namespace_labels
-      resources = discover_resources
+      discover_resources
+      resources = load_resource_from_file
       validate_definitions(resources)
 
       @logger.phase_heading("Checking initial resource statuses")
@@ -190,12 +211,12 @@ module KubernetesDeploy
     private
 
     def deploy_has_priority_resources?(resources)
-      resources.any? { |r| PREDEPLOY_SEQUENCE.include?(r.type) }
+      resources.any? { |r| predeploy_sequence.include?(r.kind) }
     end
 
     def predeploy_priority_resources(resource_list)
-      PREDEPLOY_SEQUENCE.each do |resource_type|
-        matching_resources = resource_list.select { |r| r.type == resource_type }
+      predeploy_sequence.each do |resource_kind|
+        matching_resources = resource_list.select { |r| r.kind == resource_kind }
         next if matching_resources.empty?
         deploy_resources(matching_resources, verify: true, record_summary: false)
 
@@ -225,6 +246,13 @@ module KubernetesDeploy
     end
 
     def discover_resources
+      # (Lazily) rebuild these lists after discovery if they were present.
+      @predeploy_sequence = nil
+      @prune_whitelist = nil
+      DiscoverableResource.discover(context: @context, logger: @logger, server_version: server_version)
+    end
+
+    def load_resource_from_file
       resources = []
       @logger.info("Discovering templates:")
 
@@ -232,8 +260,8 @@ module KubernetesDeploy
         next unless filename.end_with?(".yml.erb", ".yml", ".yaml", ".yaml.erb")
 
         split_templates(filename) do |r_def|
-          r = KubernetesResource.build(namespace: @namespace, context: @context, logger: @logger,
-                                       definition: r_def, statsd_tags: @namespace_tags)
+          r = DiscoverableResource.build(namespace: @namespace, context: @context, logger: @logger,
+                                         definition: r_def, statsd_tags: @namespace_tags)
           resources << r
           @logger.info "  - #{r.id}"
         end
@@ -385,7 +413,7 @@ module KubernetesDeploy
 
         if prune
           command.push("--prune", "--all")
-          prune_whitelist.each { |type| command.push("--prune-whitelist=#{type}") }
+          prune_whitelist.each { |kind| command.push("--prune-whitelist=#{kind}") }
         end
 
         out, err, st = kubectl.run(*command, log_failure: false)
--- a/lib/kubernetes-deploy/discoverable_resource.rb
+++ b/lib/kubernetes-deploy/discoverable_resource.rb
@@ -1,0 +1,237 @@
+# frozen_string_literal: true
+require 'kubernetes-deploy/kubernetes_resource'
+require 'kubernetes-deploy/kubeclient_builder'
+require 'erb'
+require 'json'
+require 'open3'
+
+module KubernetesDeploy
+  class DiscoverableResource < KubernetesResource
+    extend KubernetesDeploy::KubeclientBuilder
+
+    TRUE_VALUES = [true, 1, '1', 't', 'T', 'true', 'TRUE'].to_set
+    DEPLOY_METADATA_ANNOTATION = 'kubernetes-deploy.shopify.io/metadata'
+
+    def self.inherited(child_class)
+      DiscoverableResource.child_classes.add(child_class)
+    end
+
+    def self.discover(context:, logger:, server_version:)
+      logger.info("Discovering custom resources:")
+      with_retries { discover_groups(context) }
+      if server_version >= Gem::Version.new('1.7.0')
+        kube_client = v1beta1_crd_kubeclient(context)
+        with_retries { discover_crd(logger: logger, client: kube_client) }
+      end
+    end
+
+    def self.discover_groups(context)
+      kinds = discover_kinds(context)
+      kinds.each_pair do |key, val|
+        klass = get_static_class(kind: key)
+        next unless klass
+        klass.const_set(:GROUP, val[:group]) unless klass.constants.include?(:GROUP)
+        klass.const_set(:VERSION, val[:version]) unless klass.constants.include?(:VERSION)
+      end
+    end
+
+    def self.discover_kinds(context)
+      kinds = {}
+
+      # At the top level there is the core group (everything below /api/v1),
+      rest_client = v1_kubeclient(context).create_rest_client
+      raw_json = rest_client['v1'].get(rest_client.headers)
+      resource_list = JSON.parse(raw_json)
+      v1_group_version = { group: 'core', version: 'v1' }
+      resource_list['resources'].map do |res|
+        kind = res['kind']
+        kinds[kind] = v1_group_version
+      end
+
+      # ...and the named groups (at path /apis/$NAME/$VERSION)
+      rest_client = apis_kubeclient(context).create_rest_client
+      raw_json = rest_client.get(rest_client.headers)
+      group_list = JSON.parse(raw_json)
+      group_list = group_list['groups']
+
+      # Map out all detected kinds to their (preferred) group version
+      group_list.each do |group|
+        preferred_version = group['preferredVersion']['groupVersion']
+        all_versions = group['versions'].map { |version| version['groupVersion'] }
+        # Make sure the preferred version gets checked first.
+        all_versions.delete(preferred_version)
+        all_versions.unshift(preferred_version)
+
+        # Grab kinds from all versions
+        all_versions.each do |group_version|
+          raw_response = rest_client[group_version].get(rest_client.headers)
+          json_response = JSON.parse(raw_response)
+          resources = json_response['resources']
+          resources.each do |res|
+            kind = res['kind']
+            next if kinds.key?(kind) # Respect the preferred version
+            group, _, version = group_version.rpartition('/')
+            kinds[kind] = { group: group, version: version }
+          end
+        end
+      end
+
+      kinds
+    end
+
+    def self.build(namespace:, context:, definition:, logger:, statsd_tags:)
+      opts = { namespace: namespace, context: context,
+               definition: definition, logger: logger,
+               statsd_tags: statsd_tags }
+      kind = definition["kind"]
+      group, _, version = definition['apiVersion'].rpartition('/')
+
+      raise InvalidTemplateError.new("Template missing 'Kind'",
+        content: definition.to_yaml) if definition["kind"].blank?
+
+      klass = get_static_class(kind: kind)
+      klass = get_dynamic_class(group: group, version: version, kind: kind) unless klass
+      klass.new(**opts)
+    end
+
+    def self.get_static_class(kind:)
+      KubernetesDeploy.const_get(kind) if KubernetesDeploy.const_defined?(kind)
+    end
+
+    def self.get_dynamic_class(group:, version:, kind:)
+      unless DiscoverableResource.const_defined?(kind)
+        generate_resource(group: group, version: version, kind: kind, annotations: {})
+      end
+      DiscoverableResource.const_get(kind)
+    end
+
+    def self.with_retries(retries = 3, backoff = 10)
+      yield
+    rescue KubeException => err
+      if (retries -= 1) > 0
+        logger.warn("Retrying to discover CustomResourceDefinitions: #{err}")
+        sleep(backoff)
+        retry
+      else
+        logger.warn("Unable to discover CustomResourceDefinitions: #{err}")
+      end
+    end
+
+    def self.discover_crd(logger:, client:)
+      @child_classes = Set.new
+      resources = client.get_custom_resource_definitions
+      resources.each do |res|
+        kind = res.spec.names.kind
+        # Remove and redefine the class if it already exists so we can be up to date.
+        if DiscoverableResource.const_defined?(kind)
+          klass = DiscoverableResource.const_get(kind)
+          DiscoverableResource.send(:remove_const, kind)
+          child_classes.delete(klass)
+        end
+        generate_resource(group: res.spec.group,
+                           version: res.spec.version,
+                           kind: kind,
+                           annotations: res.metadata.annotations)
+        logger.info("  - #{res.spec.group}/#{res.spec.version}/#{kind}")
+      end
+    end
+
+    def self.generate_resource(group:, version:, kind:, annotations:)
+      deploy_metadata = annotations[DEPLOY_METADATA_ANNOTATION] || '{}'
+      metadata = JSON.parse(deploy_metadata)
+      unless metadata.is_a?(Hash)
+        raise FatalDeploymentError, "Invalid metadata for #{kind} #{metadata.inspect}"
+      end
+
+      predeploy_dependencies = [] << metadata['predeploy-dependencies']
+      predeploy_dependencies.flatten!
+      predeploy_dependencies.compact!
+
+      prunable = parse_bool(metadata['prunable'])
+      predeploy = parse_bool(metadata['predeploy'])
+      status_query = metadata['status-query']
+      success_status = metadata['status-success']
+
+      resource_template = ERB.new <<-CLASS
+        class #{kind.capitalize} < DiscoverableResource
+          GROUP = '#{group}'
+          VERSION = '#{version}'
+          PREDEPLOY = #{predeploy}
+          PRUNABLE = #{prunable}
+
+          <% unless predeploy_dependencies.empty? %>
+          PREDEPLOY_DEPENDENCIES = #{predeploy_dependencies}
+          <% end %>
+
+          <% if status_query && success_status %>
+          def deploy_succeeded?
+            getter = "get_#{kind.downcase}"
+            @client ||= DiscoverableResource.kubeclient(context: @context, resource_class: self.class)
+            raw_json = @client.send(getter, @name, @namespace, as: :raw)
+            result = query_json(query: STATUS_QUERY, json: raw_json)
+            result == SUCCESS_STATUS
+          end
+          <% end %>
+
+          self
+        end
+      CLASS
+
+      rendered_template = resource_template.result(binding)
+      klass = class_eval(rendered_template)
+      klass.const_set(:STATUS_QUERY, status_query)
+      klass.const_set(:SUCCESS_STATUS, success_status)
+      klass
+    rescue JSON::ParserError => e
+      raise FatalDeploymentError, "Invalid metadata for #{group}/#{version}/#{kind}: #{e}"
+    end
+
+    def self.parse_bool(value)
+      return true if TRUE_VALUES.include?(value)
+      false
+    end
+
+    def query_json(query:, json:)
+      safe_query = Shellwords.escape(query)
+      out, err, status = Open3.capture3("jq -r #{safe_query}", stdin_data: json)
+      unless status.success?
+        str = "Invalid status query '#{query}' for #{self.class.qualified_kind}: #{err}"
+        raise FatalDeploymentError, str
+      end
+      out.strip # Trailing newlines
+    end
+
+    def self.kubeclient(context:, resource_class:)
+      _build_kubeclient(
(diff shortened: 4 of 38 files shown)


kubernetes-deploy.gemspec
```diff
@@ -24,11 +24,12 @@ Gem::Specification.new do |spec|
 
   spec.required_ruby_version = '>= 2.3.0'
   spec.add_dependency "activesupport", ">= 4.2"
-  spec.add_dependency "kubeclient", "~> 2.4"
+  spec.add_dependency "kubeclient", "~> 2.5.1"
```
stefanmb: Required for raw `json` support.

kubernetes-deploy.gemspec
```diff
@@ -24,11 +24,12 @@ Gem::Specification.new do |spec|
...
   spec.add_dependency "googleauth", ">= 0.5"
   spec.add_dependency "ejson", "1.0.1"
   spec.add_dependency "colorize", "~> 0.8"
   spec.add_dependency "statsd-instrument", "~> 2.1"
+  spec.add_dependency "jsonpath", "0.8.8"
```
stefanmb: Required to parse out the custom `status` field.

lib/kubernetes-deploy/discoverable_resource.rb
```diff
@@ -0,0 +1,175 @@
...
+    end
+
+    def self.discover(context:, logger:)
+      logger.info("Discovering custom resources:")
+      @resources = nil
+      @child_classes = nil
+      discover_tpr(v1beta1_kubeclient(context))
+      begin
+        discover_crd(v1beta1_crd_kubeclient(context))
+      rescue KubeException => err
```
stefanmb: This is thrown for Kubernetes < 1.7 - I am also open to either:

1) Eliminating this check and dropping support for < 1.7.
2) Making the check against the version number of the client/server.

lib/kubernetes-deploy/discoverable_resource.rb
```diff
@@ -0,0 +1,175 @@
...
+      resources.each do |res|
+        discovered(group: res.spec.group,
+                   type: res.spec.names.kind,
+                   version: res.spec.version,
+                   annotations: res.metadata.annotations)
+      end
+    end
+
+    def self.discovered(group:, type:, version:, annotations:)
+      resource_class = Class.new(self) do
```
stefanmb: I previously spoke about using meta-programming with @mkobetic. Since that time I've convinced myself this is the right approach.

We are modelling a class/instance relationship (resource definition vs resource instances). Each definition has specific properties shared by all instances (e.g. whether it's prunable, or its configured timeout). This information should be stored somewhere, since Ruby supports OOP it makes sense to generate classes that can contain the specific settings shared by all resources (instances).

lib/kubernetes-deploy/discoverable_resource.rb
```diff
@@ -0,0 +1,175 @@
...
+        @timeout = DiscoverableResource.parse_timeout(type, annotations[TIMEOUT_ANNOTATION])
+
+        status_field = annotations[STATUS_FIELD_ANNOTATION]
+        success_status = annotations[STATUS_SUCCESS_ANNOTATION]
+
+        define_method 'deploy_succeeded?' do
+          getter = "get_#{type.downcase}"
+          @client ||= DiscoverableResource.kubeclient(context: @context, resource_class: self.class)
+          raw_json = @client.send(getter, @name, @namespace, as: :raw)
+          query_path = JsonPath.new(status_field)
```
stefanmb: We allow the CRD implementors to specify how the resources should be queried via arbitrary `jsonpath` to determine its status.

lib/kubernetes-deploy/discoverable_resource.rb
```diff
@@ -0,0 +1,175 @@
...
+          current_status = query_path.first(raw_json)
+          current_status == success_status
+        end if status_field && success_status
+      end
+
+      add_resource(resource_class)
+    end
+
+    def self.add_resource(resource_class)
```
stefanmb: While we generate resource classes dynamically, we still need a way to map resource names to their classes.

lib/kubernetes-deploy/kubernetes_resource/genericresource.rb
```diff
@@ -0,0 +1,26 @@
+# frozen_string_literal: true
+module KubernetesDeploy
+  class GenericResource < KubernetesResource
```
stefanmb: I've extracted this base class that represents a generic resource. This code is shared by several existing resources.

lib/kubernetes-deploy/kubernetes_resource/customresourcedefinition.rb
```diff
@@ -0,0 +1,6 @@
+# frozen_string_literal: true
+module KubernetesDeploy
+  class CustomResourceDefinition < GenericResource
```
stefanmb: This class allows `k8s-deploy` to monitor deployment of custom resource *definitions*.

lib/kubernetes-deploy/kubernetes_resource/thirdpartyresource.rb
```diff
@@ -0,0 +1,12 @@
+# frozen_string_literal: true
+module KubernetesDeploy
+  class ThirdPartyResource < GenericResource
```
stefanmb: This class allows `k8s-deploy` to monitor deployment of third party resource *definitions*.

lib/kubernetes-deploy/discoverable_resource.rb
```diff
@@ -0,0 +1,175 @@
...
+
+    def self.parse_timeout(type, timeout)
+      ActiveSupport::Duration.parse(timeout) if timeout
+    rescue ActiveSupport::Duration::ISO8601Parser::ParsingError
+      raise FatalDeploymentError,
+        "Resource #{type} specified invalid timeout value '#{timeout}', must use ISO8601 duration."
+    end
+
+    def self.kubeclient(context:, resource_class:)
+      _build_kubeclient(
```
stefanmb: This value cannot be memoized because we must refresh what CRD/TPRs are available on each invocation (even if it's during a single lifespan of a`k8s-deploy` process, for example if using this code as a library instead of a standalone executable during test).

test/integration/resource_discovery_test.rb
```diff
@@ -0,0 +1,91 @@
...
+
+  def cleanup(*resources)
+    resources.each do |res|
+      pass unless has_resource?(res)
+      _, err, st = kubectl.run("delete", res, "--all")
+      flunk(err) unless st.success?
+    end
+  end
+
+  def has_resource?(res)
```
stefanmb: Necessary when running without TPR support (k8s > 1.7) or without CRD support (k8s < 1.7).

lib/kubernetes-deploy/discoverable_resource.rb
```diff
@@ -0,0 +1,177 @@
...
+      discover_tpr(v1beta1_kubeclient(context))
+      begin
+        discover_crd(v1beta1_crd_kubeclient(context))
+      rescue KubeException => err
+        logger.warn("Unable to discover CustomResourceDefinitions: #{err}")
+      end
+    end
+
+    def self.build(namespace:, context:, definition:, logger:)
+      return super if KubernetesDeploy.const_defined?(definition["kind"])
```
stefanmb: This allows us to temporarily fall back to the custom classes in `k8s-deploy` if they are present for a given resource. This behaviour is useful for progressively moving the monitoring to the controllers and outside of this gem.

stefanmb:
We may want to also override `failure_message` to read a custom message (and possibly fail fast), for example by introducing:

* (Optional) `kubernetes-deploy.shopify.io/status-failed`: The value that indicates unrecoverable failure.
* (Optional) `kubernetes-deploy.shopify.io/status-failure-message-field`: jsonpath query for detailed human-readable status message. Should be populated when `status-failed` is returned.

Any thoughts? @KnVerey

lib/kubernetes-deploy/discoverable_resource.rb
```diff
@@ -0,0 +1,177 @@
...
+    def self.inherited(child_class)
+      child_classes.add child_class
+    end
+
+    def self.child_classes
+      @child_classes ||= Set.new
+    end
+
+    def self.prunable?
+      return @prunable if defined?(@prunable)
```
ibawt: instance vars that don't exist should still be falsey, just being explicit?

lib/kubernetes-deploy/discoverable_resource.rb
```diff
@@ -0,0 +1,177 @@
...
+      "#{group}/#{version}/#{type}"
+    end
+
+    def self.all
+      child_classes.dup
+    end
+
+    def self.discover(context:, logger:)
+      logger.info("Discovering custom resources:")
+      @resources = nil
```
ibawt: this is also a nop afaik

ibawt: unless you are resetting them

stefanmb:
cc: @kirs I'd like to hear your opinion on this PR as well please.

lib/kubernetes-deploy/discoverable_resource.rb
stefanmb: They should be reset each time we (re)discover.

lib/kubernetes-deploy/discoverable_resource.rb
stefanmb: Just my Java/C#/C background. ;) Simplified, thanks!

stefanmb:
Rebased on master and addressed some cosmetic comments.

lib/kubernetes-deploy/discoverable_resource.rb
```diff
@@ -0,0 +1,171 @@
...
+      group = resource_class.group
+      type = resource_class.type
+      version = resource_class.version
+      @resources ||= {}
+      @resources[group] ||= {}
+      @resources[group][type] ||= {}
+      @resources[group][type][version] = resource_class
+    end
+
+    def self.parse_bool(value)
```
devonboyer: Kubernetes also uses `yes` for the value of a boolean annotation. These annotations are added to persistent volume claims once they are bound.

```
pv.kubernetes.io/bind-completed=yes
pv.kubernetes.io/bound-by-controller=yes
```

This should also look for `yes`.

ibawt: lets do ON, yes, true and 1 wdyt?

devonboyer: The value needs to be a string so "1" might get confusing.

KnVerey: FWIW this is what Rails (well, ActiveModel) does: https://github.com/rails/rails/blob/47eadb68bfcae1641b019e07e051aa39420685fb/activemodel/lib/active_model/type/boolean.rb#L17

stefanmb: Switched to ActiveModel, thanks!

lib/kubernetes-deploy/discoverable_resource.rb
KnVerey: **DISCLAIMER:** I've only worked with this kind of thing a couple times in the past, and I'm going to want to pull in another experienced Rubyist before we merge this PR.

Agreed, but I think we can take it a step further and make the generated classes "normal" rather than using class variables, by using `class_eval`. Here's a working example:

```ruby 
module ScratchPadEphemeral
  class DynamicParent
    require 'active_support/core_ext/class/subclasses'

    def self.define_child(child_name, child_type, some_bool)
      new_class = self.class_eval <<-STRING
        class #{child_name.capitalize} < DynamicParent
          def type
            '#{child_type}'
          end

          def some_bool
            #{some_bool}
          end
        end
      STRING
    end
  end
end
```

```
[35] pry(ScratchPadEphemeral):1> DynamicParent.define_child("foo", "test", true)
=> :some_bool
[36] pry(ScratchPadEphemeral):1> DynamicParent.define_child("bar", "test", false)
=> :some_bool
[37] pry(ScratchPadEphemeral):1> DynamicParent.subclasses
=> [ScratchPadEphemeral::DynamicParent::Bar, ScratchPadEphemeral::DynamicParent::Foo]
[38] pry(ScratchPadEphemeral):1> foo = DynamicParent::Foo.new
=> #<ScratchPadEphemeral::DynamicParent::Foo:0x007fd8e4106fe8>
[39] pry(ScratchPadEphemeral):1> foo.type
=> "test"
[40] pry(ScratchPadEphemeral):1> foo.some_bool
=> true
```

This should incidentally be more performant than `define_method?`, from what I've been told in the past.

lib/kubernetes-deploy/discoverable_resource.rb
KnVerey: It's very possible that I'm missing something, but I believe that if you use the `class_eval` technique I mentioned above, we can subclass from `KubernetesResource` as usual, and the existing lookup technique (`KubernetesDeploy.const_defined?(definition["kind"])`) should work.

lib/kubernetes-deploy/discoverable_resource.rb
```diff
@@ -0,0 +1,171 @@
+# frozen_string_literal: true
+require 'kubernetes-deploy/kubernetes_resource'
+require 'kubernetes-deploy/kubeclient_builder'
+require "active_record"
```
KnVerey: Let's require only the piece we need, or if that isn't possible, copy the one method. I don't want to pull in all of AR.

lib/kubernetes-deploy/discoverable_resource.rb
KnVerey: Is that because we might have deployed CRDs? If we want to make that work, we should predeploy them before doing the discovery (or maybe just accept that this does not work for v1, since that'd be really complicated). Doing discovery every time we need to use the client sounds expensive. Alternatively, could we be using instances of this class and discarding them on every `run`?

lib/kubernetes-deploy/discoverable_resource.rb
```diff
@@ -0,0 +1,171 @@
...
+
+    def self.inherited(child_class)
+      child_classes.add child_class
+    end
+
+    def self.child_classes
+      @child_classes ||= Set.new
+    end
+
+    def self.prunable?
```
KnVerey: Could we add this and `identity` to the other resources classes and build the _entire_ prune list dynamically? It would be awesome if we could be dynamically generating 99% "normal" subclasses of `KuberenetesResource` rather than subtly different classes.

lib/kubernetes-deploy/discoverable_resource.rb
KnVerey: (irrelevant if we can make the instances direct subclasses of KubernetesResource, but...) Doesn't this end up covering the regular resources, i.e. is not temporary?

lib/kubernetes-deploy/discoverable_resource.rb
```diff
@@ -0,0 +1,171 @@
...
+      rescue KubeException => err
+        logger.warn("Unable to discover CustomResourceDefinitions: #{err}")
+      end
+    end
+
+    def self.build(namespace:, context:, definition:, logger:)
+      return super if KubernetesDeploy.const_defined?(definition["kind"])
+
+      # We only discover once per kubernetes-deploy invocation
+      discover(context: context, logger: logger) unless @resources
```
KnVerey: This should assume discovery is has already been done. It would be very heavy to have it actually happen here.

lib/kubernetes-deploy/discoverable_resource.rb
```diff
@@ -0,0 +1,171 @@
...
+    class << self
+      attr_accessor :version, :group, :type
+      attr_reader :timeout
+    end
+
+    STATUS_FIELD_ANNOTATION = 'kubernetes-deploy.shopify.io/status-field'
+    STATUS_SUCCESS_ANNOTATION = 'kubernetes-deploy.shopify.io/status-success'
+    TIMEOUT_ANNOTATION = 'kubernetes-deploy.shopify.io/timeout'
+    PREDEPLOY_ANNOTATION = 'kubernetes-deploy.shopify.io/predeploy'
+    PRUNABLE_ANNOTATION = 'kubernetes-deploy.shopify.io/prunable'
```
KnVerey: There are so many annotations here that I'm wondering if we should instead have a single one that contains JSON.

lib/kubernetes-deploy/kubernetes_resource/genericresource.rb
KnVerey: Hm, `KubernetesResource` (the superclass) is already kinda "generic resource", no? Should we simply change the defaults in that class to what you have here? Regardless, please extract this change to a separate PR. 😄

test/integration/resource_discovery_test.rb
```diff
@@ -0,0 +1,98 @@
...
+          @namespace,
+          KubeclientHelper::MINIKUBE_CONTEXT,
+          $stdout
+        ),
+      log_failure_by_default: false)
+  end
+
+  def cleanup(*resources)
+    resources.each do |res|
+      _, err, st = kubectl.run("delete", res, "--all")
```
KnVerey: Is it impossible/hard to do this with kubeclient?

lib/kubernetes-deploy/discoverable_resource.rb
stefanmb: This is more explicit in the re-written version. If a matching (== kind) symbol exists in the k8s-deploy module it is used, otherwise a class is generated dynamically.

lib/kubernetes-deploy/discoverable_resource.rb
stefanmb: Adopted this approach in the rewritten version, thanks.

lib/kubernetes-deploy/discoverable_resource.rb
stefanmb: This part has been re-written.

stefanmb:
@KnVerey I've updated this branch with the latest approach, it's still a work in progress (+ need some more tests) so feel free to ignore it for now. I'll ping you for review once I feel it's ready.

bin/test
```diff
@@ -27,9 +27,14 @@ if [[ ${PARALLELISM:=0} -lt 1 ]]; then
...
 
+if [[ ${CI:="0"} == "1" ]]; then
+  echo "--- :linux: Install jq"
+  wget -qP /usr/local/bin http://stedolan.github.io/jq/download/linux64/jq && chmod a+x /usr/local/bin/jq
+fi
+
 if [[ ${CI:="0"} == "1" ]]; then
   echo "--- :ruby: Bundle Install"
-  bundle install --jobs 4
+  bundle install --jobs 4 || exit 1
```
stefanmb: The tests should exit if `bundler` fails.

lib/kubernetes-deploy/deploy_task.rb
```diff
@@ -67,22 +61,48 @@ class DeployTask
...
+      predeploy_order = graph.topsort_iterator.to_a
+      predeploy_order.delete(:ROOT_NODE)
+      predeploy_order
+    end
+
+    def all_resources
+      resources = DiscoverableResource.all + KubernetesResource.all
+      # Omit unqualified kinds (they are unsupported on this cluster, or discovery hasn't been performed yet)
+      resources.select do |res|
+        res.constants.include?(:GROUP) && res.constants.include?(:VERSION)
```
stefanmb: This is check is somewhat inelegant, but the `GROUP` and `VERSION` are now detected via discovery for most resources. If these fields are not present we can't query the API server, so we skip over them. This is not an issue if the user run discovery first.

lib/kubernetes-deploy/discoverable_resource.rb
```diff
@@ -0,0 +1,232 @@
...
+
+module KubernetesDeploy
+  class DiscoverableResource < KubernetesResource
+    extend KubernetesDeploy::KubeclientBuilder
+
+    TRUE_VALUES = [true, 1, '1', 't', 'T', 'true', 'TRUE'].to_set
+    DEPLOY_METADATA_ANNOTATION = 'kubernetes-deploy.shopify.io/metadata'
+
+    def self.inherited(child_class)
+      DiscoverableResource.child_classes.add(child_class)
```
stefanmb: Overriding allows us to distinguish dynamically created resources (discovery) versus statically defined resources (direct subclass of `KubernetesResource`).

lib/kubernetes-deploy/discoverable_resource.rb
```diff
@@ -0,0 +1,232 @@
...
+        next unless klass
+        klass.const_set(:GROUP, val[:group]) unless klass.constants.include?(:GROUP)
+        klass.const_set(:VERSION, val[:version]) unless klass.constants.include?(:VERSION)
+      end
+    end
+
+    def self.discover_kinds(context)
+      kinds = {}
+
+      # At the top level there is the core group (everything below /api/v1),
```
stefanmb: Unfortunately for historic reasons not all resources are available under `/apis/$NAME/$VERSION)`, so we have to look in two places.

lib/kubernetes-deploy/discoverable_resource.rb
```diff
@@ -0,0 +1,232 @@
...
+        all_versions.unshift(preferred_version)
+
+        # Grab kinds from all versions
+        all_versions.each do |group_version|
+          raw_response = rest_client[group_version].get(rest_client.headers)
+          json_response = JSON.parse(raw_response)
+          resources = json_response['resources']
+          resources.each do |res|
+            kind = res['kind']
+            next if kinds.key?(kind) # Respect the preferred version
```
stefanmb: For each group we'll attempt to detect resources under the preferred group version, but some resources are only available under newer (beta/alpha) versions, so we have to check those too.

lib/kubernetes-deploy/discoverable_resource.rb
```diff
@@ -0,0 +1,232 @@
...
+          end
+          <% end %>
+
+          self
+        end
+      CLASS
+
+      rendered_template = resource_template.result(binding)
+      klass = class_eval(rendered_template)
+      klass.const_set(:STATUS_QUERY, status_query)
```
stefanmb: I've adopted the `class_eval` approach suggested by @KnVerey, but I did not want to include the `status_query` and `status_success` fields in the `class_eval` input because they can contain arbitrary user-provided data, possibly executable Ruby code. While I do not think there is particular concern for these fields to be used as attack vectors, it doesn't seem prudent to allow user data to be interpreted as arbitrary Ruby code.

lib/kubernetes-deploy/discoverable_resource.rb
```diff
@@ -0,0 +1,232 @@
...
+    end
+
+    def self.parse_bool(value)
+      return true if TRUE_VALUES.include?(value)
+      false
+    end
+
+    def query_json(query:, json:)
+      safe_query = Shellwords.escape(query)
+      out, err, status = Open3.capture3("jq -r #{safe_query}", stdin_data: json)
```
stefanmb: This approach is gross and has poor performance, the rationale:

- I initially used the `jsonpath` gem, but that query language is fairly obscure. `jq` is much more popular.
- I tried using the `ruby-jq` gem but it has two issues:
  - It requires a special shared library version of `jq` to be installed (this was doable, but annoying)
  - The `jq` parser prints errors to `stderr`, so even if you use the gem your program will print out error output from deep within the `jq` library itself. This was a show stopper. It might be possible to disable the output, but I did not see how with configure options and I do not want to have to maintain a fork of `jq`

lib/kubernetes-deploy/kubeclient_builder.rb
```diff
@@ -67,7 +75,7 @@ def _build_kubeclient(api_version:, context:, endpoint_path: nil)
         ssl_options: kube_context.ssl_options,
         auth_options: kube_context.auth_options
       )
-      client.discover
+      client.discover if discover
```
stefanmb: Discovery will crash against certain API endpoints, so we make it explicit so we can still use the client for those paths.

lib/kubernetes-deploy/kubernetes_resource.rb
```diff
@@ -7,7 +7,7 @@
 module KubernetesDeploy
   class KubernetesResource
     attr_reader :name, :namespace, :context
-    attr_writer :type, :deploy_started_at
```
stefanmb: I've removed `type` and replaced it with `kind` so that we match the Kubernetes nomenclature.

lib/kubernetes-deploy/kubernetes_resource.rb
```diff
@@ -95,10 +131,6 @@ def validation_failed?
       @validation_errors.present?
     end
 
-    def id
-      "#{type}/#{name}"
```
stefanmb: The `group/version/kind` (i.e. `qualified_kind`) serves this purpose.

lib/kubernetes-deploy/kubernetes_resource/bugsnag.rb
```diff
@@ -2,10 +2,13 @@
 module KubernetesDeploy
   class Bugsnag < KubernetesResource
     TIMEOUT = 1.minute
+    PREDEPLOY = true
+    GROUP = 'stable.shopify.io'
+    VERSION = 'v1'
```
stefanmb: We provide these values in code for statically defined classes. They can also be provided dynamically by annotations to CRDs.

lib/kubernetes-deploy/kubernetes_resource/horizontalpodautoscaler.rb
```diff
@@ -0,0 +1,29 @@
+# frozen_string_literal: true
+module KubernetesDeploy
+  class HorizontalPodAutoscaler < KubernetesResource
```
stefanmb: The HPA was previously defined only as a string in the `prune` list, we now need an actual class backing it.

lib/kubernetes-deploy/kubernetes_resource/pod.rb
```diff
@@ -2,7 +2,9 @@
 module KubernetesDeploy
   class Pod < KubernetesResource
     TIMEOUT = 10.minutes
-
+    PREDEPLOY = true
+    PREDEPLOY_DEPENDENCIES = %w(ResourceQuota ServiceAccount ConfigMap PersistentVolumeClaim)
```
stefanmb: Extracted from the previous static predeploy dependency list. Note that this is a partial order (the previous list by definition was a total order).

test/integration-serial/resource_discovery_test.rb
```diff
@@ -0,0 +1,103 @@
+# frozen_string_literal: true
+require 'test_helper'
+
+class ResourceDiscoveryTest < KubernetesDeploy::IntegrationTest
```
stefanmb: These tests affect global resources (CRD) so they are not parallel-friendly.

test/unit/kubernetes-deploy/deploy_task_test.rb
```diff
@@ -26,22 +26,83 @@ def test_initializer
...
+    # All resource should be in the list
+    list.each { |r| assert_includes order, r.kind }
+
+    # All preconditions should be respected
+    order.each_with_index do |r, idx|
+      klass = list.detect { |e| e.kind == r }
+      klass.predeploy_dependencies.each do |dep|
+        pos = order.map.with_index { |e, i| e == dep ? i : nil }.compact.first
+        # The current resource is deployed *after* its deps
+        assert_operator idx, :>, pos, "#{r} requires #{dep} but got #{order}"
```
stefanmb: I did not want to hardcode a particular total order here because that's not important. What's important is that the result is in any valid toplogical sort order, which may not be unique.

stefanmb:
@KnVerey Could I ask you to have another look? A lot has changed since the initial version. All tests are currently passing.

lib/kubernetes-deploy/kubernetes_resource.rb
```diff
@@ -28,16 +28,52 @@ class KubernetesResource
...
+    def self.child_classes
+      @child_classes ||= Set.new
+    end
+
+    def self.all
+      child_classes.dup
+    end
+
+    def self.prunable?
+      self::PRUNABLE if defined? self::PRUNABLE
```
KnVerey: nit: these should be `false` rather than `nil` when the constants aren't available

lib/kubernetes-deploy/kubernetes_resource.rb
```diff
@@ -28,16 +28,52 @@ class KubernetesResource
...
+
+    def self.version
+      self::VERSION
+    end
+
+    def self.kind
+      name.demodulize
+    end
+
+    def self.qualified_kind
```
KnVerey: I've seen this creatively referred to as "groupVersionKind" in Go libs

lib/kubernetes-deploy/kubernetes_resource/bugsnag.rb
KnVerey: We should delete this class, actually... we 🔥 this CR a while ago.

lib/kubernetes-deploy/kubernetes_resource/customresourcedefinition.rb
```diff
@@ -0,0 +1,29 @@
...
+    TIMEOUT = 10.seconds
+    PREDEPLOY = true
+
+    def sync
+      _, _err, st = kubectl.run("get", kind, @name)
+      @status = st.success? ? "Available" : "Unknown"
+      @found = st.success?
+    end
+
+    def deploy_succeeded?
```
KnVerey: There's actually a more sophisticated success condition available for this; check out `waitCRDReady` in cloudbuddies.

lib/kubernetes-deploy/kubernetes_resource/customresourcedefinition.rb
```diff
@@ -0,0 +1,29 @@
...
+      exists?
+    end
+
+    def deploy_failed?
```
KnVerey: Might make sense to look at the `NamesAccepted` condition here (see `waitCRDReady` in cloudbuddies)

lib/kubernetes-deploy/kubernetes_resource/job.rb
```diff
@@ -0,0 +1,29 @@
+# frozen_string_literal: true
+module KubernetesDeploy
+  class Job < KubernetesResource
```
KnVerey: I wouldn't insist on this, but it could be helpful if you extracted these new class additions into a separate PR. Making sure each one has appropriate checks and test coverage is not really related to this PR.

lib/kubernetes-deploy/kubernetes_resource/pod.rb
KnVerey: What about the fact that this can depend on custom resources, such as cloudsql/redis?

lib/kubernetes-deploy/discoverable_resource.rb
```diff
@@ -0,0 +1,232 @@
...
+    TRUE_VALUES = [true, 1, '1', 't', 'T', 'true', 'TRUE'].to_set
+    DEPLOY_METADATA_ANNOTATION = 'kubernetes-deploy.shopify.io/metadata'
+
+    def self.inherited(child_class)
+      DiscoverableResource.child_classes.add(child_class)
+    end
+
+    def self.discover(context:, logger:, server_version:)
+      logger.info("Discovering custom resources:")
+      with_retries { discover_groups(context) }
```
KnVerey: Am I reading correctly that this is solely for the purpose of determining which GV to use in the prune whitelist for the statically defined classes?

lib/kubernetes-deploy/discoverable_resource.rb
```diff
@@ -0,0 +1,232 @@
...
+      kinds
+    end
+
+    def self.build(namespace:, context:, definition:, logger:)
+      opts = { namespace: namespace, context: context, definition: definition, logger: logger }
+      kind = definition["kind"]
+      group, _, version = definition['apiVersion'].rpartition('/')
+
+      klass = get_static_class(kind: kind)
+      klass = get_dynamic_class(group: group, version: version, kind: kind) unless klass
```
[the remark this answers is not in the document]
KnVerey: Do I understand correctly that we'll now end up generating dynamic classes for core resource types we don't explicitly support, rather than using `KubernetesResource` directly for those? (Note to self: review the class generation code with that in mind if this is true)

lib/kubernetes-deploy/deploy_task.rb
```diff
@@ -67,22 +61,48 @@ class DeployTask
...
-        extensions/v1beta1/Ingress
-        apps/v1beta1/StatefulSet
-        autoscaling/v1/HorizontalPodAutoscaler
-      )
-      if server_version >= Gem::Version.new('1.8.0')
-        wl << "batch/v1beta1/CronJob"
+      @prune_whitelist ||= _build_prune_whitelist
+    end
+
+    def _build_prune_whitelist
```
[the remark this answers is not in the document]
stefanmb: Sure, was just following the previous examples in the file.

lib/kubernetes-deploy/kubernetes_resource/job.rb
stefanmb: As a prerequisite PR? Without the new classes this PR won't work.

lib/kubernetes-deploy/kubernetes_resource/pod.rb
stefanmb: Then we'll need some way to specify such custom dependencies in the pod spec.

Alternatively we could optionally specify something like `PREDEPLOY_DEPENDENTS` on resources, so for e.g. `CloudSQL` could specify that it is a dependency of `Pod` which is its dependent.

lib/kubernetes-deploy/discoverable_resource.rb
```diff
@@ -0,0 +1,232 @@
...
+      success_status = metadata['status-success']
+
+      resource_template = ERB.new <<-CLASS
+        class #{kind.capitalize} < DiscoverableResource
+          GROUP = '#{group}'
+          VERSION = '#{version}'
+          PREDEPLOY = #{predeploy}
+          PRUNABLE = #{prunable}
+
+          <% unless predeploy_dependencies.empty? %>
```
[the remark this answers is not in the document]
stefanmb: Edit: Deleted misplaced comment here, I think the issue is you can't define constant dynamically. I'll come back to this.

lib/kubernetes-deploy/discoverable_resource.rb
```diff
@@ -0,0 +1,232 @@
...
+    def self.parse_bool(value)
+      return true if TRUE_VALUES.include?(value)
+      false
+    end
+
+    def query_json(query:, json:)
+      safe_query = Shellwords.escape(query)
+      out, err, status = Open3.capture3("jq -r #{safe_query}", stdin_data: json)
+      unless status.success?
+        str = "Invalid status query '#{query}' for #{self.class.qualified_kind}: #{err}"
```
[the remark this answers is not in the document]
stefanmb: Based on IRL discussion this is a race condition, will fix.

airhorns:
@stefanmb this would be so awesome to get in for all of us without all the *Buddies nowadays!!! <3<3

PR #189, opened 2017-10-17
WIP: Fail if server version is less than minimum required

permaninder:


KnVerey:
https://github.com/Shopify/kubernetes-deploy/pull/237 Fixed this same issue

Issue #164, opened 2017-09-11
Better error when the kube config has the wrong master IP


stefanmb:
Currently seeing:

```
[INFO][2017-09-11 20:25:03 +0000]	
[INFO][2017-09-11 20:25:03 +0000]	------------------------------------Phase 1: Initializing deploy------------------------------------
[INFO][2017-09-11 20:25:03 +0000]	All required parameters and files are present
[INFO][2017-09-11 20:25:03 +0000]	Context pipa-test found
[INFO][2017-09-11 20:25:34 +0000]	
[INFO][2017-09-11 20:25:34 +0000]	------------------------------------------Result: FAILURE-------------------------------------------
[FATAL][2017-09-11 20:25:34 +0000]	Namespace pipa-test not found
```

This is a bit confusing to users since it says "Context pipa-test found" which seems to refer to the kube config file, but the actual error occurs at the namespace lookup.

KnVerey:
I think this happens because the context check doesn't actually try to hit the server (just looks for a name in the kubeconfig), making the namespace check the first thing that does. We could fix https://github.com/Shopify/kubernetes-deploy/issues/152 at the same time by changing the context check to a validation that requests the target server version.

KnVerey:
We've also seen namespace not found errors when the namespace actually exists but there was an error in requesting it. We should be more careful about the error returned by kubectl in this validation and retry it if appropriate.

KnVerey:
Fixed by https://github.com/Shopify/kubernetes-deploy/pull/190. The namespace check is no longer the first API call, and both the server validation and the namespace check are now retried.

Issue #152, opened 2017-07-28
Enforce minimum kubernetes version


KnVerey:
We currently require Kubernetes v1.6 and will soon move to v1.7. All three executables should check the version in the target cluster and abort early if it doesn't meet requirements.

KnVerey:
Actually I think it should just warn, and point to a compatibility list (which we don't currently have)

PR #190, opened 2017-10-17
Retry namespace discovery and raise real error

karanthukral:
## What?
- Add rudimentary retry logic to `confirm_namespace_exists`
  - Open to suggestions in case we want to add exponential backoffs
- More importantly this will raise the real error in case of failures. That way it would be easier to track the failures

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,0 +1,9 @@
+### 0.12.9
+
+*Bug Fixes*
+- Retry discovering namespace and kubernetes context
+- Expose real error during namespace discovery
+
+### 0.12.8
+*Bug Fixes*
+- Force deployment to use its own hard timeout instead of relying on the replica set
--- a/lib/kubernetes-deploy/runner.rb
+++ b/lib/kubernetes-deploy/runner.rb
@@ -72,12 +72,15 @@ module KubernetesDeploy
       autoscaling/v1/HorizontalPodAutoscaler
     ).freeze
 
-    def initialize(namespace:, context:, current_sha:, template_dir:, logger:, bindings: {})
+    NOT_FOUND_ERROR = 'NotFound'
+
+    def initialize(namespace:, context:, current_sha:, template_dir:, logger:, kubectl_instance: nil, bindings: {})
       @namespace = namespace
       @context = context
       @current_sha = current_sha
       @template_dir = File.expand_path(template_dir)
       @logger = logger
+      @kubectl = kubectl_instance
       @bindings = bindings
       # Max length of podname is only 63chars so try to save some room by truncating sha to 8 chars
       @id = current_sha[0...8] + "-#{SecureRandom.hex(4)}" if current_sha
@@ -401,12 +404,27 @@ module KubernetesDeploy
       elsif !available_contexts.include?(@context)
         raise FatalDeploymentError, "Context #{@context} is not available. Valid contexts: #{available_contexts}"
       end
+      confirm_cluster_reachable
       @logger.info("Context #{@context} found")
     end
 
+    def confirm_cluster_reachable
+      success = false
+      with_retries(2) do
+        _, _, st = kubectl.run("version", use_namespace: false, log_failure: true)
+        success = st.success?
+      end
+
+      raise FatalDeploymentError, "Failed to reach server for #{@context}" unless success
+    end
+
     def confirm_namespace_exists
-      _, _, st = kubectl.run("get", "namespace", @namespace, use_namespace: false, log_failure: false)
-      raise FatalDeploymentError, "Namespace #{@namespace} not found" unless st.success?
+      st, err = nil
+      with_retries(2) do
+        _, err, st = kubectl.run("get", "namespace", @namespace, use_namespace: false, log_failure: true)
+        st.success? || err.include?(NOT_FOUND_ERROR)
+      end
+      raise FatalDeploymentError, "Failed to find namespace. #{err}" unless st.success?
       @logger.info("Namespace #{@namespace} found")
     end
 
@@ -416,6 +434,15 @@ module KubernetesDeploy
 
     def statsd_tags
       %W(namespace:#{@namespace} sha:#{@current_sha} context:#{@context})
+    end
+
+    def with_retries(limit)
+      retried = 0
+      while retried <= limit
+        success = yield
+        break if success
+        retried += 1
+      end
     end
   end
 end
--- a/test/fixtures/invalid_config.yml
+++ b/test/fixtures/invalid_config.yml
@@ -1,0 +1,20 @@
+apiVersion: v1
+clusters:
+- cluster:
+    insecure-skip-tls-verify: true
+    server: https://192.168.0.1
+  name: minikube
+contexts:
+- context:
+    cluster: minikube
+    namespace: default
+    user: minikube
+  name: minikube
+current-context: ""
+kind: Config
+preferences: {}
+users:
+- name: minikube
+  user:
+    username: username
+    password: password
--- a/test/integration/kubernetes_deploy_test.rb
+++ b/test/integration/kubernetes_deploy_test.rb
@@ -506,7 +506,7 @@ class KubernetesDeployTest < KubernetesDeploy::IntegrationTest
     original_ns = @namespace
     @namespace = 'this-certainly-should-not-exist'
     assert_deploy_failure(deploy_fixtures("hello-cloud", subset: ['configmap-data.yml']))
-    assert_logs_match(/Result: FAILURE.*Namespace this-certainly-should-not-exist not found/m)
+    assert_logs_match(/Result: FAILURE.*namespaces "this-certainly-should-not-exist" not found/m)
   ensure
     @namespace = original_ns
   end
@@ -685,5 +685,24 @@ class KubernetesDeployTest < KubernetesDeploy::IntegrationTest
     ejson_cloud.assert_secret_present('monitoring-token', managed: true)
     ejson_cloud.assert_secret_present('catphotoscom', type: 'kubernetes.io/tls', managed: true)
     ejson_cloud.assert_secret_present('ejson-keys', managed: false)
+  end
+
+  def test_invalid_context
+    old_config = ENV['KUBECONFIG']
+    begin
+      ENV['KUBECONFIG'] = File.join(__dir__, '../fixtures/invalid_config.yml')
+      result = deploy_fixtures('hello-cloud')
+      assert_deploy_failure(result)
+      assert_logs_match_all([
+        'The following command failed: kubectl version',
+        'Unable to connect to the server',
+        'Unable to connect to the server',
+        'Unable to connect to the server',
+        'Result: FAILURE',
+        'Failed to reach server for minikube',
+      ], in_order: true)
+    ensure
+      ENV['KUBECONFIG'] = old_config
+    end
   end
 end
--- a/test/unit/kubernetes-deploy/kubectl_test.rb
+++ b/test/unit/kubernetes-deploy/kubectl_test.rb
@@ -20,7 +20,8 @@ class KubectlTest < KubernetesDeploy::TestCase
   end
 
   def test_run_constructs_the_expected_command_and_returns_the_correct_values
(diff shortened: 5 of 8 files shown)


karanthukral:
@KnVerey whenever you get a chance

test/integration/kubernetes_deploy_test.rb
```diff
@@ -506,7 +506,7 @@ def test_failed_deploy_to_nonexistent_namespace
     original_ns = @namespace
     @namespace = 'this-certainly-should-not-exist'
     assert_deploy_failure(deploy_fixtures("hello-cloud", subset: ['configmap-data.yml']))
-    assert_logs_match(/Result: FAILURE.*Namespace this-certainly-should-not-exist not found/m)
+    assert_logs_match(/namespaces "this-certainly-should-not-exist" not found/m)
```
stefanmb: This changed in `kubectl`?

lib/kubernetes-deploy/runner.rb
```diff
@@ -405,8 +405,12 @@ def confirm_context_exists
     end
 
     def confirm_namespace_exists
-      _, _, st = kubectl.run("get", "namespace", @namespace, use_namespace: false, log_failure: false)
-      raise FatalDeploymentError, "Namespace #{@namespace} not found" unless st.success?
+      st, err = nil
+      3.times do
```
stefanmb: Should we have a small backoff period? Doesn't have to be exponential but I would not retry in a tight loop.

karanthukral: I can add a small backoff 👍

test/integration/kubernetes_deploy_test.rb
karanthukral: This changed in how I am surfacing the error. Since we surface the actual `err` instead of just saying namespace not found, it changes how the logs look

KnVerey:
I agree with retrying the namespace discovery, but fundamentally the reason why we commonly see errors other than "namespace does not exist" here is that it happens to be the first attempt at contacting the server. My suggestion (open for discussion for sure) would be:
1 - Retry the namespace discovery a couple times like you're doing here (but with a slight backoff as Stefan suggested), but only if the error is something other than NotFound. If the namespace actually does not exist, raise the error immediately as before.
2 - As part of `confirm_context_exists`, which happens before `confirm_namespace_exists`, actually try to contact the server to prove the context is valid (fixing https://github.com/Shopify/kubernetes-deploy/issues/164). Use retries here too. I'd suggest using `kubectl version --output=json` for this, as a step towards https://github.com/Shopify/kubernetes-deploy/issues/152. And add a test that uses a kubeconfig with an invalid master IP.

The idea behind this suggestion is make the preliminary checks more resilient to network issues without removing the ability to fail fast on common configuration issues. WDYT?

karanthukral:
I'll work on this right after I clean up my statefulset PR :)

karanthukral:
> I'd suggest using kubectl version --output=json for this, as a step towards #152.

@KnVerey just to confirm you mean test that as a separate test right?

lib/kubernetes-deploy/runner.rb
```diff
@@ -393,22 +395,41 @@ def log_pruning(kubectl_output)
...
       if !st.success?
         raise FatalDeploymentError, err
       elsif !available_contexts.include?(@context)
         raise FatalDeploymentError, "Context #{@context} is not available. Valid contexts: #{available_contexts}"
       end
+      confirm_context_valid_on_server
       @logger.info("Context #{@context} found")
     end
 
+    def confirm_context_valid_on_server
```
karanthukral: @KnVerey Still working on cleaning this up but wanted to see if you had any reservations against this method?

KnVerey: In general this is what I was thinking, yes. I think we should retry the call a few times if we get an error though. It also occurred to me that it could be less complex and just as helpful to use `log_failure: true` to get _all_ the failure messages in the output, and then just raise an error about what we were trying to do on the final failure. 

Other nit: "Context not found on server" isn't exactly accurate, because context is a client-side pairing of a user and a cluster. What we're detecting here is whether the cluster defined by the context in question is reachable. It doesn't terribly matter what command we use as long as we have a test that fails if the command we've chosen stops trying to hit the server.

Suggestion:

```ruby
def confirm_cluster_reachable
  success = false
  with_retries(3) do 
    _, _, st = kubectl.run("version", use_namespace: false, log_failure: true)
    success = st.success?
  end
  
  raise FatalDeploymentError, "Failed to reach server for #{@context}" unless success
end
```

karanthukral: Updated as per your recommendation

test/integration/kubernetes_deploy_test.rb
```diff
@@ -686,4 +686,16 @@ def test_ejson_secrets_respects_no_prune_flag
     ejson_cloud.assert_secret_present('catphotoscom', type: 'kubernetes.io/tls', managed: true)
     ejson_cloud.assert_secret_present('ejson-keys', managed: false)
   end
+
+  def test_invalid_context
+    old_config = ENV['KUBECONFIG']
+    begin
+      ENV['KUBECONFIG'] = File.join(__dir__, '../fixtures/invalid_config.yml')
```
KnVerey: Wouldn't the path this generates have literal `..` in it? Does that actually work?

test/integration/kubernetes_deploy_test.rb
```diff
@@ -686,4 +686,16 @@ def test_ejson_secrets_respects_no_prune_flag
...
+      result = deploy_fixtures('hello-cloud')
+      assert_deploy_failure(result)
+      assert_logs_match_all(['Failed to reach server for minikube'])
```
KnVerey: I'm not getting the errors I'd expect when I try this test. It seems like there's something wrong with the kubeconfig file itself rather than the server. This test needs to prove that the deploy fails specifically when the server is unavailable.

I see this right now:
image


I want to see something like this:
image

And assert on the bad host/port message in addition to the assertion you have

KnVerey:
Have you rebased this since Stefan merged his CI fix?

test/integration/kubernetes_deploy_test.rb
karanthukral: It does work. I can see if it continues to work without the `..`

test/integration/kubernetes_deploy_test.rb
karanthukral: So I looked into this and you were right. There is an issue with the config file. On fixing that I realized that the i/o timeout for the kubectl command is 30s which makes the test fairly time consuming to run. The error is:
`Unable to connect to the server: dial tcp 5.6.7.8:443: i/o timeout`

Should this only be retried once? or just once for the test to make sure the suite doesn't grow in time?

KnVerey: Ugh, that's a really good point. That timeout is supposedly configurable with a flag, though the flag doc seems to be lying about the default, so you'd need to confirm:
> `--request-timeout string`           The length of time to wait before giving up on a single server request. Non-zero values should contain a corresponding time unit (e.g. 1s, 2m, 3h). A value of zero means don't timeout requests. (default "0")

It seems to me we wouldn't want _any_ test waiting 30s, so it'd be nice if we find a solution that lets us tune the timeout for all of them. One option would be to make `KuberentesDeploy::Kubectl#initialize` take a new `default_timeout:` parameter, and then make `kubectl:` an optional argument to `KubernetesDeploy::Runner#initialize` so that we can create our own instance with a custom timeout in the test setup and pass it in. WDYT?

karanthukral: I'll work on that

karanthukral: I did what you recommended. So it seems like they weren't lying about the no timeout since with the timeout in place, the error changes.

```
[INFO][2017-10-26 08:53:34 -0400]
[INFO][2017-10-26 08:53:34 -0400]	------------------------------------Phase 1: Initializing deploy------------------------------------
[INFO][2017-10-26 08:53:34 -0400]	All required parameters and files are present
[WARN][2017-10-26 08:53:44 -0400]	The following command failed: kubectl version --context\=minikube --request-timeout\=10s
[WARN][2017-10-26 08:53:44 -0400]	Unable to connect to the server: net/http: request canceled while waiting for connection (Client.Timeout exceeded while awaiting headers)

[WARN][2017-10-26 08:53:54 -0400]	The following command failed: kubectl version --context\=minikube --request-timeout\=10s
[WARN][2017-10-26 08:53:54 -0400]	Unable to connect to the server: net/http: request canceled while waiting for connection (Client.Timeout exceeded while awaiting headers)

[INFO][2017-10-26 08:53:54 -0400]
[INFO][2017-10-26 08:53:54 -0400]	------------------------------------------Result: FAILURE-------------------------------------------
[FATAL][2017-10-26 08:53:54 -0400]	Failed to reach server for minikube
```

test/fixtures/invalid_config.yml
```diff
@@ -0,0 +1,20 @@
+apiVersion: v1
+clusters:
+- cluster:
+    insecure-skip-tls-verify: true
+    server: https://5.6.7.8
```
KnVerey: Let's use a private IP just in case

test/integration/kubernetes_deploy_test.rb
KnVerey: Please add back the `Result: FAILURE.*` bit... it proves we collected this info into the result section specifically.

test/integration/kubernetes_deploy_test.rb
```diff
@@ -686,4 +686,19 @@ def test_ejson_secrets_respects_no_prune_flag
...
     ejson_cloud.assert_secret_present('ejson-keys', managed: false)
   end
+
+  def test_invalid_context
+    old_config = ENV['KUBECONFIG']
+    begin
+      ENV['KUBECONFIG'] = File.join(__dir__, '../fixtures/invalid_config.yml')
+      result = deploy_fixtures('hello-cloud')
+      assert_deploy_failure(result)
+      assert_logs_match_all([
```
KnVerey: These assertions are a good start, but I'd like to make them more strict re: what we're printing in what order. Eg:
```
[
  "The following command failed: kubectl version",
  "Unable to connect to the server",
  "Result: FAILURE",
  "Failed to reach server for minikube"
], in_order: true
```

And maybe an extra assertion specifically on "Unable to connect to the server" being logged X times to demonstrate that we not only retried but also showed the message each time to help with debugging.

test/unit/kubernetes-deploy/kubectl_test.rb
```diff
@@ -20,7 +20,8 @@ def test_raises_if_initialized_with_null_namespace
   end
 
   def test_run_constructs_the_expected_command_and_returns_the_correct_values
```
KnVerey: Let's add a quick unit test to this file re: your new timeout arg

lib/kubernetes-deploy/runner.rb
```diff
@@ -417,5 +435,14 @@ def kubectl
     def statsd_tags
       %W(namespace:#{@namespace} sha:#{@current_sha} context:#{@context})
     end
+
+    def with_retries(limit)
```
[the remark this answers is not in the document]
karanthukral: Fair enough. I'll get that fixed :)

karanthukral:
Fixed the issues pointed out :D

CHANGELOG.md
```diff
@@ -0,0 +1,11 @@
+### 0.12.9
+*Features*
+- Added a helper function to allow retrying blocks of code
+
+*Bug Fixes*
+- Retry discovering namespace and kubernetes context
+- Expose real error during namespace discovery
+
+### 0.12.8
+*Features*
```
KnVerey: This was actually a bugfix--Make Deployment use its own hard timeout instead of ReplicaSet's by simplifying timeout logic

CHANGELOG.md
```diff
@@ -0,0 +1,11 @@
+### 0.12.9
+*Features*
+- Added a helper function to allow retrying blocks of code
```
KnVerey: Maybe I'm wrong about this, but I'd say the changelog should be for users trying to decide which version to pick (so internal code details are not worth mentioning)

PR #191, opened 2017-10-19
Use correct kubectl version in CI

stefanmb:
**What**
Match the `minikube` `k8s` version in CI by mounting the correct `kubectl` version.

cc: @KnVerey @karanthukral @devonboyer


stefanmb:
Merged to unblock CI - old feature branches need to be rebased.

PR #193, opened 2017-10-26
 Simplify Deployment and ReplicaSet timeout condition 

KnVerey:
There were two extra sources of Deployment timeout with the previous conditions:
- the ReplicaSet hard timeout, which should have matched but didn't
- the Pod hard timeout, which is higher, and which it doesn't even
really make sense to fail Deployments on

This was unnecessarily complex.

### DIFF
--- a/lib/kubernetes-deploy/kubernetes_resource/replica_set.rb
+++ b/lib/kubernetes-deploy/kubernetes_resource/replica_set.rb
@@ -50,10 +50,6 @@ module KubernetesDeploy
       @pods.map(&:timeout_message).compact.uniq.join("\n")
     end
 
-    def deploy_timed_out?
-      super || @pods.present? && @pods.all?(&:deploy_timed_out?)
-    end
-
     def exists?
       @found
     end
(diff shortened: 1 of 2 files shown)


lib/kubernetes-deploy/kubernetes_resource/replica_set.rb
```diff
@@ -50,10 +50,6 @@ def timeout_message
       @pods.map(&:timeout_message).compact.uniq.join("\n")
     end
 
-    def deploy_timed_out?
```
kirs: Can you comment why we should fall back to 

https://github.com/Shopify/kubernetes-deploy/blob/09f6121b2f02f1fbb51fb582e63f077328d74d81/lib/kubernetes-deploy/kubernetes_resource.rb#L113-L116

not to `@pods.all?(&:deploy_timed_out?)`?

KnVerey: The most pragmatic reason is that `@pods.all?(&:deploy_timed_out?)` will never be true before `super` anyway, since the Pod timeout is double the RS timeout; this is just making the behaviour more difficult to understand with no real benefit. In fact, the Pod hard timeout was chosen based on the needs of bare pods, which are assumed to be running a single high-value pseudo-inline task, so referencing it from RS doesn't really make sense. Finally, the supported, k8s-native way to time out a deployment whose pods aren't rolling in time is using `progressDeadlineSeconds`.

Issue #192, opened 2017-10-26
ejson secret base64 encoding issue


mleklund:
Secret generation code is calling `Base64.encode(value)` which which is inserting a newline which breaks the yaml that is being fed to kubectl.

```
{
  "_public_key": "XXXXX",
  "kubernetes_secrets": {
    "mydata": {
      "_type": "Opaque",
      "data": {
        "mypassword": "mypass",
        "mypassword2": "mypass"}
    }
  }
} 
```

results in the following yaml:

```
---
kind: Secret
apiVersion: v1
type: Opaque
metadata:
  name: mydata
  labels:
    name: mydata
  namespace: production
  annotations:
    kubernetes-deploy.shopify.io/ejson-secret: 'true'
data:
  mypassword: 'bXlwYXNz

'
  mypassword2: 'bXlwYXNz

'
```

resulting in the following error:

```
Creation of kubernetes secrets from ejson failed: error from server (badrequest): error when creating "/var/folders/6g/t31p0nvn02l3236fm3v07xsr0000gn/t/mydata20171026-34510-1mbtgtg": secret in version "v1" cannot be handled as a secret: v1.secret: data: decode base64: illegal base64 data at input byte 8, parsing 56 ...xlwyxnz\n"... at {"apiversion":"v1","data":{"mypassword":"bxlwyxnz\n","mypassword2":"bxlwyxnz\n"},"kind":"secret","metadata":{"annotations":{"kubectl.kubernetes.io/last-applied-configuration":"{\"apiversion\":\"v1\",\"data\":{\"mypassword\":\"bxlwyxnz\\n\",\"mypassword2\":\"bxlwyxnz\\n\"},\"kind\":\"secret\",\"metadata\":{\"annotations\":{\"kubernetes-deploy.shopify.io/ejson-secret\":\"true\"},\"labels\":{\"name\":\"mydata\"},\"name\":\"mydata\",\"namespace\":\"production\"},\"type\":\"opaque\"}\n","kubernetes-deploy.shopify.io/ejson-secret":"true"},"labels":{"name":"mydata"},"name":"mydata","namespace":"production"},"type":"opaque"}
```

jshapiro26:
I've just run into this issue as well. Running GKE: 1.8.1-gke.0

alexcurtin:
[...]

PR #196, opened 2017-10-31
Ejson secret base64 encoding issue

alexcurtin:
This PR is to address issue https://github.com/Shopify/kubernetes-deploy/issues/192 .

As Kubernetes >= 1.8 requires the yaml to no longer have a `\n` in it, we'll need to utilize Ruby's `Base64.strict_encode64()` to encode with no new lines. @KnVerey 

```
# irb
:007 > str = "aooajsdjadad"
 => "aooajsdjadad"
 :008 > Base64.encode64(str)
 => "YW9vYWpzZGphZGFk\n"
 :009 > Base64.strict_encode64(str)
 => "YW9vYWpzZGphZGFk"
```

Docs: http://ruby-doc.org/stdlib-2.4.2/libdoc/base64/rdoc/Base64.html#method-i-strict_encode64


jshapiro26:
I've made these changes locally as well and can confirm it has resolved this [issue](https://github.com/Shopify/kubernetes-deploy/issues/192) for me.

PR #197, opened 2017-10-31
Avoid creating base64 encodings with embedded newlines

skaes:
Ruby's Base64.encode method adds newlines to the encoded string after
every 60 characters and also adds one at the end. Kubernetes documentation
clearly states that base64 encoded secrets must not contain newlines (see 
https://kubernetes.io/docs/concepts/configuration/secret/#creating-a-secret-manually).
Also, Kubernetes 1.8 will now reject creating any secrets with embedded
newline characters.

The problem is solved by using Base64.strict_encode64, which does not embed
newline characters in the encoded string.

### DIFF
--- a/lib/kubernetes-deploy/ejson_secret_provisioner.rb
+++ b/lib/kubernetes-deploy/ejson_secret_provisioner.rb
@@ -128,7 +128,7 @@ module KubernetesDeploy
         # Leading underscores in ejson keys are used to skip encryption of the associated value
         # To support this ejson feature, we need to exclude these leading underscores from the secret's keys
         secret_key = key.sub(/\A_/, '')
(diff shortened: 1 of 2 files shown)


KnVerey:
Thanks for the PR! I've merged https://github.com/Shopify/kubernetes-deploy/pull/196 instead since it is identical and was opened first. I'll take a look at the other PR you submitted shortly. 😄

skaes:
Thx for addressing the issue so quickly and also for releasing a new gem version!

BTW, I only discovered the other pull request after I had created mine. Really doesn't matter which one got merged. It was a fairly straightforward.

Oh, and a big thank you for open sourcing this code in the first place. It has really helped a lot with our project.

PR #200, opened 2017-11-01
Update CI Setup for k8s 1.7.x, 1.8.x

stefanmb:
Run CI on:
- k8s 1.6.4 (this is the latest version supported by this CI setup)
- k8s 1.7.x latest
- k8s 1.8.x latest
- k8s 1.9.x latest (nightly builds only)

Latest includes beta/alpha releases.

Requires https://github.com/Shopify/kubernetes-deploy/pull/198 for passing 1.8.x+ builds.

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -1,8 +1,18 @@
-  - name: 'Run Test Suite (:kubernetes: 1.7.5)'
+  - name: 'Run Test Suite (:kubernetes: 1.8-latest)'
     command: bin/ci
     agents:
-      queue: pipa-production-minikube-v1.7.5
+      queue: minikube-v1.8-latest
+    env:
+      LOGGING_LEVEL: 4
+  - name: 'Run Test Suite (:kubernetes: 1.7-latest)'
+    command: bin/ci
+    agents:
+      queue: minikube-v1.7-latest
+    env:
+      LOGGING_LEVEL: 4
   - name: 'Run Test Suite (:kubernetes: 1.6.4)'
     command: bin/ci
     agents:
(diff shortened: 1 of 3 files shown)


.buildkite/pipeline.yml
```diff
@@ -1,8 +1,22 @@
-  - name: 'Run Test Suite (:kubernetes: 1.7.5)'
+  - name: 'Run Test Suite (:kubernetes: 1.8.2)'
     command: bin/ci
     agents:
-      queue: pipa-production-minikube-v1.7.5
-  - name: 'Run Test Suite (:kubernetes: 1.6.4)'
+      queue: minikube-v1.8.2
+    env:
+      LOGGING_LEVEL: 4
```
KnVerey: Is this for buildkite or ultimately for k8s? If the latter, should we set it in `bin/ci` like we do a few other env vars?

stefanmb: `LOGGING_LEVEL` refers to the Kubernetes components. The default (if omitted) is 4. I included it here explicitly as documentation to make it obvious how to get a more verbose log.

Moving it to `bin/ci` is too late as the env var has to be present during the initial k8s bootstrap before actually executing the command.

stefanmb:
@KnVerey I've updated this PR, the latest builds have passed. The 1.9.x-latest builds will happen nightly, see: https://buildkite.com/shopify/kubernetes-deploy-gem-nightly/

This new pipeline has no Github hooks, it is launched via the nightly schedule or manually.

It's possible to set up individual email notifications for that branch when the fail/pass status changes.

stefanmb:
> Could we make those post to slack like we do for Splunk alerts?

Investigating some reasonable way to do that.

PR #201, opened 2017-11-02
Change all CRs to do :replace and add dummy classes for other CRs

KnVerey:
@Shopify/cloudplatform This is my best idea for a workaround for the `apply` failures we're seeing right now. CI has extremely poor coverage of CR types, so I'm working on a way to test this manually right now.


mkobetic:
👍

KnVerey:
Works on tierstaging. Picking a real app to try it with before shipping.

image

PR #206, opened 2017-11-11
support cronjob resources

skaes:
This pull request adds support for cronjob resources. I have run the tests against 1.7 and 1.8 clients and 1.7 and 1.8 clusters. Please let me know if anything needs to be changed.

### DIFF
--- a/lib/kubernetes-deploy/deploy_task.rb
+++ b/lib/kubernetes-deploy/deploy_task.rb
@@ -38,7 +38,7 @@ require 'kubernetes-deploy/kubeclient_builder'
 require 'kubernetes-deploy/ejson_secret_provisioner'
 
 module KubernetesDeploy
-  class Runner
+  class DeployTask
     include KubeclientBuilder
 
     PREDEPLOY_SEQUENCE = %w(
@@ -327,15 +327,15 @@ module KubernetesDeploy
         @logger.info("Deploying resources:")
       else
         resource = resources.first
-        @logger.info("Deploying #{resource.id} (timeout: #{resource.timeout}s)")
+        @logger.info("Deploying #{resource.id} (#{resource.pretty_timeout_type})")
       end
 
       # Apply can be done in one large batch, the rest have to be done individually
       applyables, individuals = resources.partition { |r| r.deploy_method == :apply }
 
       individuals.each do |r|
-        @logger.info("- #{r.id} (timeout: #{r.timeout}s)") if resources.length > 1
-        r.deploy_started = Time.now.utc
+        @logger.info("- #{r.id} (#{r.pretty_timeout_type})") if resources.length > 1
+        r.deploy_started_at = Time.now.utc
         case r.deploy_method
         when :replace
           _, _, replace_st = kubectl.run("replace", "-f", r.file_path, log_failure: false)
@@ -370,9 +370,9 @@ module KubernetesDeploy
 
       command = ["apply"]
       resources.each do |r|
-        @logger.info("- #{r.id} (timeout: #{r.timeout}s)") if resources.length > 1
+        @logger.info("- #{r.id} (#{r.pretty_timeout_type})") if resources.length > 1
         command.push("-f", r.file_path)
-        r.deploy_started = Time.now.utc
+        r.deploy_started_at = Time.now.utc
       end
 
       if prune
--- a/lib/kubernetes-deploy/deploy_task.rb
+++ b/lib/kubernetes-deploy/deploy_task.rb
@@ -79,12 +79,7 @@ module KubernetesDeploy
         autoscaling/v1/HorizontalPodAutoscaler
       )
       if server_version >= Gem::Version.new('1.8.0')
-        wl.concat(%w(
-          apps/v1beta2/DaemonSet
-          apps/v1beta2/Deployment
-          apps/v1beta2/StatefulSet
-          batch/v1beta1/CronJob
-        ))
+        wl << "batch/v1beta1/CronJob"
       end
       wl
     end
--- a/lib/kubernetes-deploy/deploy_task.rb
+++ b/lib/kubernetes-deploy/deploy_task.rb
@@ -70,6 +70,7 @@ module KubernetesDeploy
       core/v1/Pod
       core/v1/Service
       batch/v1/Job
+      batch/v1beta1/CronJob
       extensions/v1beta1/DaemonSet
       extensions/v1beta1/Deployment
       apps/v1beta1/Deployment
--- a/lib/kubernetes-deploy/deploy_task.rb
+++ b/lib/kubernetes-deploy/deploy_task.rb
@@ -90,7 +90,7 @@ module KubernetesDeploy
     end
 
     def server_version
-      @sever_version ||= kubectl.server_version
+      kubectl.server_version
     end
 
     NOT_FOUND_ERROR = 'NotFound'
--- a/lib/kubernetes-deploy/deploy_task.rb
+++ b/lib/kubernetes-deploy/deploy_task.rb
@@ -65,19 +65,33 @@ module KubernetesDeploy
     # core/v1/ReplicationController -- superseded by deployments/replicasets
     # extensions/v1beta1/ReplicaSet -- managed by deployments
     # core/v1/Secret -- should not committed / managed by shipit
-    PRUNE_WHITELIST = %w(
-      core/v1/ConfigMap
-      core/v1/Pod
-      core/v1/Service
-      batch/v1/Job
-      batch/v1beta1/CronJob
-      extensions/v1beta1/DaemonSet
-      extensions/v1beta1/Deployment
-      apps/v1beta1/Deployment
-      extensions/v1beta1/Ingress
-      apps/v1beta1/StatefulSet
-      autoscaling/v1/HorizontalPodAutoscaler
-    ).freeze
+    def prune_whitelist
+      wl = %w(
+        core/v1/ConfigMap
+        core/v1/Pod
+        core/v1/Service
+        batch/v1/Job
+        extensions/v1beta1/DaemonSet
+        extensions/v1beta1/Deployment
+        apps/v1beta1/Deployment
+        extensions/v1beta1/Ingress
+        apps/v1beta1/StatefulSet
+        autoscaling/v1/HorizontalPodAutoscaler
+      )
+      if server_version >= Gem::Version.new('1.8.0')
+        wl.concat(%w(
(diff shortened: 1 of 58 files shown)


KnVerey:
Thanks for the PR! I'm sorry it took me so long to get to it; this is our busiest time of the year at Shopify. 😄 

The resource class logic seems sound; there's not much we can check, since the job doesn't run immediately. Some general comments:
* We don't really want to support alpha versions, so that should let you simplify your code a bunch.
* Since we're still supporting 1.6.x, the tests need to be compatible with that version too. The easiest thing to do is probably to put the cron job file in its own fixture set and have a test just for it that only runs on 1.8+. 
* We probably want to add this type to the `PRUNE_WHITELIST` in `runner.rb`.
* Please pick up the latest changes on our master. I won't be able to use our new CI setup for this PR unless it includes [this commit](https://github.com/Shopify/kubernetes-deploy/commit/3387ee09d45365b667f8a5a1ddec8a3e4edf673c).

skaes:
Let me address your comments and concern, in reverse order:

I merged current master into this PR and got test failures which were unrelated to my code changes. Fixed them as a separate pull request, just in case this one doesn't make it. Merged it into this PR to make progress.

I haven't decided yet what to do about PRUNE_WHITELIST. API versions of Kubernetes resources change over time, so I assume PRUNE_WHITELIST will need to be changed according to client version, just like it probably changes in the kubectl client implementation. For example, all the workload resources [have moved](https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG-1.8.md#workloads-api-appsv1beta2) from extensions/v1beta1 to apps/v1 in 1.8. At some point, they can no longer be accessed by kubectl clients under the olde API version and I suspect will no longer be supported as --prune-whitelist arguments.

The code already supports 1.6 clients and clusters, as CronJob resources have been available since at least version 1.6 of Kubernetes.

I can understand the desire to keep the code simple and minimize support effort. As a maintainer of e.g. ruby-prof, I can certainly relate that. However, in this case, I think am exception to your stated rule can be justified. There is no real logic involved in the CronJob resource class itself, it's all in the test setup and selecting the right API version to talk to the server. I think that API version selection is an inevitable feature which will creep into your code base anyway, very likely with the Kubernetes 1.9/1.10 release.

We have upgraded all our clusters to 1.8, so none of my internal customers will be hurt if we drop Cronjob support for earlier Kubernetes releases. In addition, I won't need to describe to everyone that they can safely ignore the warnings kubernetes-deploy emits . But for benefit of everyone else out there I kindly ask you to reconsider your position regarding this particular feature.

KnVerey:
> At some point, they can no longer be accessed by kubectl clients under the olde API version and I suspect will no longer be supported as --prune-whitelist arguments.

That's correct. For the purposes of the whitelist, it doesn't matter which GV you reference as long as it is recognized by your k8s server version. It so happens that 1.6/1.7/1.8 all share a GV for all the resources we supported so far, but we did have a per-version whitelist in the past, and it seems you'll need to reintroduce it in this PR. [Here is the old implementation](https://github.com/Shopify/kubernetes-deploy/blob/3857718d569a872a0363f02c4051cb66fcbac772/lib/kubernetes-deploy/runner.rb#L171-L177) (note that it predates our nice version helpers in the `Kubectl` class).

> There is no real logic involved in the CronJob resource class itself, it's all in the test setup and selecting the right API version to talk to the server.

If I understand your comment correctly, the alpha GV will actually work in practice (other than pruning?), but we need a bunch of code to get test coverage for it. To clarify, I didn't mean we should explicitly disable the CronJob class on earlier k8s versions. The thing is that alpha tests are only useful if alpha application logic is required, and our current position is that we don't want to support that. I'm open to reconsidering this in the future if alpha support is highly demanded by the community, but for now I would rather not go down that road. Furthermore, deciding to add explicit alpha support is a bit of a separate issue from supporting CronJobs, and would require changes to our CI system. I'd like to get CronJob support in ASAP, so if you want to continue the alpha discussion, let's do it in an issue.

skaes:
I seem to be missing something: you want to make the ConJob resource class available on 1.6, 1.7 , 1.8 and later, but don't want to test it on 1.6 and 1.7? That's a very unusual testing concept.

I can only think of one scenario where this makes sense: you have explicitly disabled alpha features in your kubernetes clusters, including the CI/CD pipeline. Is that the case?

skaes:
Oh, and I forgot to mention: yes, the code works on all versions: 1.6, 1.7 and 1.8

KnVerey:
> I seem to be missing something: you want to make the ConJob resource class available on 1.6, 1.7 , 1.8 and later, but don't want to test it on 1.6 and 1.7? That's a very unusual testing concept.

We don't _want_ to make the alpha resource version (1.6/1.7) work, but if it does, we won't write extra code to block it either. As far as this gem is concerned, the CronJob resource would not be used in 1.6 or 1.7, because the gem does not support alpha resources.

> I can only think of one scenario where this makes sense: you have explicitly disabled alpha features in your kubernetes clusters, including the CI/CD pipeline. Is that the case?

Yes, in CI and in production. That's what I meant by this needing to be a separate discussion from CronJob because supporting alphas "would require changes to our CI system". Even if you change my mind on this 😄  those CI changes are not going to get prioritized before the holidays. The rest of this PR is almost good to go (just the prune whitelist change we were discussing).

test/helpers/fixture_deploy_helper.rb
```diff
@@ -47,14 +47,18 @@ def deploy_raw_fixtures(set, wait: true, bindings: {})
 
   def deploy_dir_without_profiling(dir, wait: true, allow_protected_ns: false, prune: true, bindings: {}, sha: nil)
     current_sha = sha || SecureRandom.hex(6)
+    version_info = {
+      kube_client_version: KubernetesDeploy::IntegrationTest::KUBE_CLIENT_VERSION,
+      kube_server_version: KubernetesDeploy::IntegrationTest::KUBE_SERVER_VERSION
+    }
```
KnVerey: This isn't used anymore; please remove the changes in this file..

test/helpers/fixture_set.rb
```diff
@@ -97,6 +97,14 @@ def assert_configmap_present(cm_name, expected_data)
       assert_equal expected_data, configmaps.first["data"].to_h
     end
 
+    def assert_cronjob_exists(job_name, app_name)
```
KnVerey: `app_name` is a property of these instances with an accessor; you don't need to pass it in.

test/helpers/fixture_deploy_helper.rb
skaes: Will do.

test/helpers/fixture_set.rb
skaes: Thx for the tip.

skaes:
I was a bit handicapped today having only intermittent internet access on a train ride. After looking at the kubernetes source code it seems that kubectl can deal with non-existent kind/version/resource triples. I'm running tests right now. If they're successful, I would simply add the one new resource to the prune whitelist.

skaes:
I think this is good go now.

KnVerey:
I just ran CI for you, and interestingly it passes on 1.6, where cron jobs don't exist at all, but fails on 1.7 with:

```
[FATAL][2017-12-21 19:46:29 +0000]	Invalid templates: See error message
[FATAL][2017-12-21 19:46:29 +0000]	> Error from kubectl:
[FATAL][2017-12-21 19:46:29 +0000]	    error: no matches for batch/, Kind=CronJob
```

Unfortunately we are going to need to version that whitelist. Btw you also have Policial failures.

skaes:
@KnVerey I'm not able to reproduce that failure. The level of support for CronJob resources should be the same in 1.6 and 1.7 (in both versions, CronJobs have api version batch/v2alpha1, which was introduce in 1.5). Maybe your 1.6 CI cluster is configured differently than your 1.7 cluster?

I will try to configure minikube in way that I can reproduce the failure.

BTW, I do sometimes get failures on even on master and when I rerun the test separately it passes. I suspect not all tests leave the cluster in a pristine state.

KnVerey:
I suspect the reason you aren't able to reproduce is that the `batch/v2alpha1` group is enabled in your minikube cluster, leading there to be a `CronJob` kind in the `batch` group. It's possible that our 1.6 CI also has this enabled accidentally (cc @stefanmb ) -- it is in fact using a legacy setup we weren't able to continue with for higher versions. Regardless, this definitely isn't a transient failure: 39 separate tests failed with the exact same message. Re: tests failing on master, yes, there are some flakey tests we need to address (we have a tracking issue for this). OTOH each test operates within a unique namespace, and we haven't seen cluster state being the root cause of any problems so far.

skaes:
Made prune whitelist dependent on server version. The list will need to be revisited when adding support for Kubernetes version 1.9.

lib/kubernetes-deploy/deploy_task.rb
```diff
@@ -64,18 +65,33 @@ class DeployTask
...
+        apps/v1beta1/Deployment
+        extensions/v1beta1/Ingress
+        apps/v1beta1/StatefulSet
+        autoscaling/v1/HorizontalPodAutoscaler
+      )
+      if server_version >= Gem::Version.new('1.8.0')
+        wl.concat(%w(
+          apps/v1beta2/DaemonSet
+          apps/v1beta2/Deployment
+          apps/v1beta2/StatefulSet
```
KnVerey: These three shouldn't be necessary, since the group versions included in the original list still exist in 1.8.

lib/kubernetes-deploy/deploy_task.rb
```diff
@@ -64,18 +65,33 @@ class DeployTask
...
+          batch/v1beta1/CronJob
+        ))
+      end
+      wl
+    end
+
+    def server_version
+      @sever_version ||= kubectl.server_version
```
KnVerey: This separate method / caching is not necessary, as the result of the API call is already cached in `kubectl.version_info`.

stefanmb:
> It's possible that our 1.6 CI also has this enabled accidentally ...

We don't enable alpha features on the 1.6 CI explicitly, so I'm not quite sure why it would be on. I can investigate, but I don't think we should hold up the PR for this in its current form.

stefanmb:
As @KnVerey just pointed out to me the `localkube` boostrapper in Minikube[ enables all APIs](https://github.com/kubernetes/minikube/pull/378/files#diff-460240d91079db5428dca08343274e24R46), hence the behaviour you saw.

Unfortunately this is not configurable. All versions > 1.7.0 are run using `kubeadm` instead.

PR #208, opened 2017-11-20
Return success before fully deployed

dturn:
Support a new annotation on deployments: `kubernetes-deploy.shopify.io/required-rollout`. It changes the behavior of deploy_succesful? 

- `full`: maintain current behavior. 
- `none`: deploy is sucessful as soon as the new rs is created for the deploy.
- `maxUnavailable`: Only for RollingUpdate deploys. The deploy is successful when there are less than `strategy.RollingUpdate.maxUnavailable` (supports both the % and fixed number) updated & ready replicas.

This feature is useful when deployments may take an extended period of time to fully roll out. This could happen when cluster capacity is near usage.

### DIFF
--- a/README.md
+++ b/README.md
@@ -122,10 +122,19 @@ You can add additional variables using the `--bindings=BINDINGS` option. For exa
 
 
 ### Customizing behaviour with annotations
-
 - `kubernetes-deploy.shopify.io/timeout-override`: Override the tool's hard timeout for one specific resource. Both full ISO8601 durations and the time portion of ISO8601 durations are valid. Value must be between 1 second and 24 hours.
   - _Example values_: 45s / 3m / 1h / PT0.25H
   - _Compatibility_: all resource types (Note: `Deployment` timeouts are based on `spec.progressDeadlineSeconds` if present, and that field has a default value as of the `apps/v1beta1` group version. Using this annotation will have no effect on `Deployment`s that time out with "Timeout reason: ProgressDeadlineExceeded".)
+- `kubernetes-deploy.shopify.io/required-rollout`: Modifies how much of the rollout needs to finish
+before the deployment is considered successful.
+  - _Compatibility_: Deployment
+  - `full`: The deploy is successful when all pods in the new ReplicaSet are ready.
+  - `none`: The deploy is successful as soon as the new `replicaSet` is created for the deploy
+  - `maxUnavailable`: The deploy is successful when minimum availability is reached in the new ReplicaSet.
+  In other words, the number of new pods that must be ready is equal to `spec.replicas` - `strategy.RollingUpdate.maxUnavailable`
+  (converted from percentages by rounding up, if applicable). This option is only valid for deployments
+  that use the `RollingUpdate` strategy.
+
 
 ### Running tasks at the beginning of a deploy
 
--- a/README.md
+++ b/README.md
@@ -128,9 +128,9 @@ You can add additional variables using the `--bindings=BINDINGS` option. For exa
 - `kubernetes-deploy.shopify.io/required-rollout`: Modifies how much of the rollout needs to finish
 before the deployment is considered successful.
   - _Compatibility_: Deployment
-  - `full`: The deploy is successful when all pods in the new ReplicaSet are ready.
-  - `none`: The deploy is successful as soon as the new `replicaSet` is created for the deploy
-  - `maxUnavailable`: The deploy is successful when minimum availability is reached in the new ReplicaSet.
+  - `full`: The deployment is successful when all pods in the new `replicaSet` are ready.
+  - `none`: The deployment is successful as soon as the new `replicaSet` is created for the deployment.
+  - `maxUnavailable`: The deploy is successful when minimum availability is reached in the new `replicaSet`.
   In other words, the number of new pods that must be ready is equal to `spec.replicas` - `strategy.RollingUpdate.maxUnavailable`
   (converted from percentages by rounding up, if applicable). This option is only valid for deployments
   that use the `RollingUpdate` strategy.
--- a/lib/kubernetes-deploy/kubernetes_resource/deployment.rb
+++ b/lib/kubernetes-deploy/kubernetes_resource/deployment.rb
@@ -2,6 +2,9 @@
 module KubernetesDeploy
   class Deployment < KubernetesResource
     TIMEOUT = 7.minutes
+    REQUIRED_ROLLOUT_ANNOTATION = 'kubernetes-deploy.shopify.io/required-rollout'
+    REQUIRED_ROLLOUT_TYPES = %w(maxUnavailable full none).freeze
+    DEFAULT_REQUIRED_ROLLOUT = 'full'
 
     def sync
       raw_json, _err, st = kubectl.run("get", type, @name, "--output=json")
@@ -19,6 +22,7 @@ module KubernetesDeploy
         conditions = deployment_data.fetch("status", {}).fetch("conditions", [])
         @progress_condition = conditions.find { |condition| condition['type'] == 'Progressing' }
         @progress_deadline = deployment_data['spec']['progressDeadlineSeconds']
+        @max_unavailable = deployment_data.dig('spec', 'strategy', 'rollingUpdate', 'maxUnavailable')
       else # reset
         @latest_rs = nil
         @rollout_data = { "replicas" => 0 }
@@ -26,6 +30,7 @@ module KubernetesDeploy
         @progress_condition = nil
         @progress_deadline = @definition['spec']['progressDeadlineSeconds']
         @desired_replicas = -1
+        @max_unavailable = @definition.dig('spec', 'strategy', 'rollingUpdate', 'maxUnavailable')
       end
     end
 
@@ -43,10 +48,24 @@ module KubernetesDeploy
     def deploy_succeeded?
       return false unless @latest_rs.present?
 
-      @latest_rs.deploy_succeeded? &&
-      @latest_rs.desired_replicas == @desired_replicas && # latest RS fully scaled up
-      @rollout_data["updatedReplicas"].to_i == @desired_replicas &&
-      @rollout_data["updatedReplicas"].to_i == @rollout_data["availableReplicas"].to_i
+      case required_rollout
+      when 'full'
+        @latest_rs.deploy_succeeded? &&
+        @latest_rs.desired_replicas == @desired_replicas && # latest RS fully scaled up
+        @rollout_data["updatedReplicas"].to_i == @desired_replicas &&
+        @rollout_data["updatedReplicas"].to_i == @rollout_data["availableReplicas"].to_i
+      when 'none'
+        true
+      when 'maxUnavailable'
+        minimum_needed = min_available_replicas
+
+        @latest_rs.desired_replicas >= minimum_needed &&
+        @latest_rs.ready_replicas >= minimum_needed &&
+        @latest_rs.available_replicas >= minimum_needed
+      else
+        raise "#{REQUIRED_ROLLOUT_ANNOTATION}:#{required_rollout} is invalid "\
+        " Acceptable options: #{REQUIRED_ROLLOUT_TYPES.join(',')}"
+      end
     end
 
     def deploy_failed?
@@ -81,6 +100,23 @@ module KubernetesDeploy
       @found
     end
 
+    def validate_definition
+      super
+
+      unless REQUIRED_ROLLOUT_TYPES.include?(required_rollout)
+        @validation_errors << "#{REQUIRED_ROLLOUT_ANNOTATION}:#{required_rollout} is invalid "\
+        "Acceptable options: #{REQUIRED_ROLLOUT_TYPES.join(',')}"
+      end
+
+      if required_rollout.downcase == 'maxunavailable' && @definition.dig('spec', 'strategy').respond_to?(:downcase) &&
+        @definition.dig('spec', 'strategy').downcase == 'recreate'
+        @validation_errors << "#{REQUIRED_ROLLOUT_ANNOTATION}:#{required_rollout} is invalid "\
+        "with strategy 'rollingUpdate'"
+      end
+
+      @validation_errors.empty?
+    end
+
     private
 
     def deploy_failing_to_progress?
@@ -98,18 +134,22 @@ module KubernetesDeploy
       Time.parse(@progress_condition["lastUpdateTime"]).to_i >= (@deploy_started_at - 5.seconds).to_i
     end
 
-    def find_latest_rs(deployment_data)
-      label_string = deployment_data["spec"]["selector"]["matchLabels"].map { |k, v| "#{k}=#{v}" }.join(",")
+    def all_rs_data(match_labels)
+      label_string = match_labels.map { |k, v| "#{k}=#{v}" }.join(",")
       raw_json, _err, st = kubectl.run("get", "replicasets", "--output=json", "--selector=#{label_string}")
-      return unless st.success?
+      return {} unless st.success?
 
-      all_rs_data = JSON.parse(raw_json)["items"]
+      JSON.parse(raw_json)["items"]
+    end
+
+    def find_latest_rs(deployment_data)
       current_revision = deployment_data["metadata"]["annotations"]["deployment.kubernetes.io/revision"]
 
-      latest_rs_data = all_rs_data.find do |rs|
+      latest_rs_data = all_rs_data(deployment_data["spec"]["selector"]["matchLabels"]).find do |rs|
         rs["metadata"]["ownerReferences"].any? { |ref| ref["uid"] == deployment_data["metadata"]["uid"] } &&
         rs["metadata"]["annotations"]["deployment.kubernetes.io/revision"] == current_revision
       end
+
       return unless latest_rs_data.present?
 
       rs = ReplicaSet.new(
@@ -122,6 +162,19 @@ module KubernetesDeploy
       )
       rs.sync(latest_rs_data)
       rs
+    end
+
+    def min_available_replicas
+      if @max_unavailable =~ /%/
+        (@desired_replicas * (100 - @max_unavailable.to_i) / 100.0).ceil
+      else
+        @desired_replicas - @max_unavailable.to_i
+      end
+    end
+
+    def required_rollout
+      @definition.dig('metadata', 'annotations', REQUIRED_ROLLOUT_ANNOTATION).presence ||
+      DEFAULT_REQUIRED_ROLLOUT
     end
   end
 end
--- a/lib/kubernetes-deploy/kubernetes_resource/deployment.rb
+++ b/lib/kubernetes-deploy/kubernetes_resource/deployment.rb
@@ -63,8 +63,7 @@ module KubernetesDeploy
         @latest_rs.ready_replicas >= minimum_needed &&
         @latest_rs.available_replicas >= minimum_needed
       else
-        raise "#{REQUIRED_ROLLOUT_ANNOTATION}:#{required_rollout} is invalid "\
-        " Acceptable options: #{REQUIRED_ROLLOUT_TYPES.join(',')}"
+        raise FatalDeploymentError, rollout_annotation_err_msg
       end
     end
 
@@ -104,20 +103,24 @@ module KubernetesDeploy
       super
 
       unless REQUIRED_ROLLOUT_TYPES.include?(required_rollout)
-        @validation_errors << "#{REQUIRED_ROLLOUT_ANNOTATION}:#{required_rollout} is invalid "\
-        "Acceptable options: #{REQUIRED_ROLLOUT_TYPES.join(',')}"
+        @validation_errors << rollout_annotation_err_msg
       end
 
-      if required_rollout.downcase == 'maxunavailable' && @definition.dig('spec', 'strategy').respond_to?(:downcase) &&
-        @definition.dig('spec', 'strategy').downcase == 'recreate'
-        @validation_errors << "#{REQUIRED_ROLLOUT_ANNOTATION}:#{required_rollout} is invalid "\
-        "with strategy 'rollingUpdate'"
+      strategy = @definition.dig('spec', 'strategy', 'type').to_s
+      if required_rollout.downcase == 'maxunavailable' && strategy.downcase != 'rollingupdate'
+        @validation_errors << "'#{REQUIRED_ROLLOUT_ANNOTATION}: #{required_rollout}' is incompatible "\
+        "with strategy '#{strategy}'"
       end
 
       @validation_errors.empty?
     end
 
     private
+
+    def rollout_annotation_err_msg
+      "'#{REQUIRED_ROLLOUT_ANNOTATION}: #{required_rollout}' is invalid. "\
+        "Acceptable values: #{REQUIRED_ROLLOUT_TYPES.join(', ')}"
+    end
 
     def deploy_failing_to_progress?
       return false unless @progress_condition.present?
--- a/lib/kubernetes-deploy/kubernetes_resource/replica_set.rb
+++ b/lib/kubernetes-deploy/kubernetes_resource/replica_set.rb
@@ -3,13 +3,15 @@ require 'kubernetes-deploy/kubernetes_resource/pod_set_base'
 module KubernetesDeploy
   class ReplicaSet < PodSetBase
     TIMEOUT = 5.minutes
-    attr_reader :desired_replicas, :pods
+    attr_reader :desired_replicas, :ready_replicas, :available_replicas, :pods
 
     def initialize(namespace:, context:, definition:, logger:, parent: nil, deploy_started_at: nil)
       @parent = parent
       @deploy_started_at = deploy_started_at
       @rollout_data = { "replicas" => 0 }
       @desired_replicas = -1
+      @ready_replicas = -1
+      @available_replicas = -1
       @pods = []
       super(namespace: namespace, context: context, definition: definition, logger: logger)
     end
@@ -26,6 +28,8 @@ module KubernetesDeploy
         @rollout_data = { "replicas" => 0 }.merge(
           rs_data["status"].slice("replicas", "availableReplicas", "readyReplicas")
         )
+        @ready_replicas = @rollout_data['readyReplicas'].to_i
+        @available_replicas = @rollout_data["availableReplicas"].to_i
         @status = @rollout_data.map { |state_replicas, num| "#{num} #{state_replicas.chop.pluralize(num)}" }.join(", ")
         @pods = find_pods(rs_data)
       else # reset
@@ -34,6 +38,8 @@ module KubernetesDeploy
         @status = nil
         @pods = []
         @desired_replicas = -1
+        @ready_replicas = -1
+        @available_replicas = -1
       end
     end
 
--- a/test/fixtures/slow-cloud/web.yml.erb
+++ b/test/fixtures/slow-cloud/web.yml.erb
@@ -1,0 +1,37 @@
+apiVersion: apps/v1beta1
+kind: Deployment
+metadata:
+  name: web
+  annotations:
+    shipit.shopify.io/restart: "true"
+    kubernetes-deploy.shopify.io/required-rollout: maxUnavailable
+spec:
+  replicas: 3
+  strategy:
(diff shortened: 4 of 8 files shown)


dturn:
cc: @dwradcliffe @ibawt since we were just chatting about this. its still a wip, I'm looking for feedback before finishing it up

dwradcliffe:
I think it probably needs to be an annotation because each deployment might have different requirements

KnVerey:
I agree that an annotation makes more sense. I'd go with a generic name like `kubernetes-deploy.shopify.io/required-rollout` so we could theoretically expand it to multiple values in the future, e.g. `maxUnavailable` (this PR), `none` (could be a more flexible replacement for the `--skip-wait` feature)

lib/kubernetes-deploy/kubernetes_resource/deployment.rb
```diff
@@ -74,7 +81,7 @@ def pretty_timeout_type
 
     def deploy_timed_out?
       # Do not use the hard timeout if progress deadline is set
-      @progress_condition.present? ? deploy_failing_to_progress? : super
+      super #@progress_condition.present? ? deploy_failing_to_progress? : super
```
dturn: Are we assuming that deployments always use progressDeadlineSeconds and never the timeout in this file?

lib/kubernetes-deploy/kubernetes_resource/deployment.rb
```diff
@@ -74,7 +73,11 @@ def pretty_timeout_type
 
     def deploy_timed_out?
       # Do not use the hard timeout if progress deadline is set
-      @progress_condition.present? ? deploy_failing_to_progress? : super
+      if @definition['spec']['progressDeadlineSeconds'].present? && @progress_condition.present?
```
dturn: Looks like `k8s` merges in a default, this will check to see if you manually set one. This might be a behavior change

KnVerey: Yes, it would be a behaviour change. Why do you think that would be better? IMO getting the smarter timeout logic shouldn't depend on whether you liked the default value the server gives you. (interestingly, we originally intended to look at `@definition` but made a mistake... and I ended up thinking it was a happy accident 😄 )

dturn: I don't have strong feelings either way. Though the behavior being different on first vs subsequent deploys seems less than ideal and makes me wonder if there are cases the test aren't capturing. 

My bigger issue is without changing this behavior, I have no idea how to write a reliable test here.

KnVerey: > the behavior being different on first vs subsequent deploys

What do you mean by this? Why would that be the case?

> My bigger issue is without changing this behavior, I have no idea how to write a reliable test here.

Maybe I'm missing the obvious, but why does changing this behaviour help? I agree this is tricky to test reliably... you'd almost want a way to prevent the whole RS from coming up... I'll think about it.

lib/kubernetes-deploy/kubernetes_resource/deployment.rb
```diff
@@ -98,18 +101,53 @@ def deploy_failing_to_progress?
...
+        desired = min_unavailable_rollout.to_i
+      end
+
+      desired.to_i
+    end
+
+    def partial_deployment_success
+      minimum_needed = minimum_unavailable_replicas_to_succeeded
+
+      running_rs.size <= 2 &&
```
KnVerey: I can understand the motivation for this check, but couldn't it make the new feature ineffective for deployments with long-running jobs (like core has) that take a long time to terminate? We actually have a test asserting that terminating pods from older revisions do not impede deploy success. I'm not certain what that state looks like from a RS standpoint, but I'd expect it to be `desired = 0` and `replicas > 0`.

dturn:
ready for 👀 again.

lib/kubernetes-deploy/kubernetes_resource/deployment.rb
```diff
@@ -123,5 +126,35 @@ def find_latest_rs(deployment_data)
       rs.sync(latest_rs_data)
       rs
     end
+
+    def min_unavailable_rollout
```
KnVerey: I'd call this something like `required_rollout_type`, have it always return a value, i.e.:

```ruby
def required_rollout_type
  strategy = @definition.dig('metadata', 'annotations', 'kubernetes-deploy.shopify.io/required-rollout')
  strategy.presence || REQUIRED_ROLLOUT_TYPES.fetch(:full_availability)
end
```

The validation of the annotation value also needs to happen earlier, before the deploy might have run. I'd suggest overwriting `validate_definition` and calling `super` first, then validating required_rollout_type against a list in a constant if that passed.

lib/kubernetes-deploy/kubernetes_resource/deployment.rb
```diff
@@ -42,6 +40,7 @@ def fetch_logs
 
     def deploy_succeeded?
       return false unless @latest_rs.present?
+      return partial_deployment_success unless min_unavailable_rollout.blank?
```
KnVerey: I think the new concept you're introducing should have first-class visibility here. I.e. I'd move the case statement on `required_rollout_type` (or whatever you name the method) right in here, and have `minimum_unavailable_replicas` assume it's being asked in the context of the relevant rollout strategy.

lib/kubernetes-deploy/kubernetes_resource/deployment.rb
```diff
@@ -123,5 +126,35 @@ def find_latest_rs(deployment_data)
...
+    def min_unavailable_rollout
+      @definition.dig('metadata', 'annotations', 'kubernetes-deploy.shopify.io/required-rollout')
+    end
+
+    def minimum_unavailable_replicas_to_succeeded
+      desired = @desired_replicas
+
+      case min_unavailable_rollout
+      when 'maxUnavailable'
+        max_unavailable = @definition.dig('spec', 'strategy', 'rollingUpdate', 'maxUnavailable')
```
KnVerey: We need to use the value from `sync` rather than `@definition`, possibly with a fallback to `@definition` (like we do for `@progress_deadline`), since this value can be server-side defaulted.

lib/kubernetes-deploy/kubernetes_resource/deployment.rb
```diff
@@ -123,5 +126,35 @@ def find_latest_rs(deployment_data)
...
+        if max_unavailable =~ /%/
+          (desired * (100 - max_unavailable.to_i) / 100.0).to_i
```
KnVerey: This won't match k8s behaviour, because k8s rounds up (well, it rounds the maxUnavailable num down) and `Float.to_i` simply truncates (i.e. rounds down).

e.g. desired = 5, maxUnavailable = 25%, we should return 4 here and we'll return 3

lib/kubernetes-deploy/kubernetes_resource/deployment.rb
```diff
@@ -123,5 +126,35 @@ def find_latest_rs(deployment_data)
...
+          desired - max_unavailable.to_i
+        end
+      when 'none'
+        0
+      else
+        raise "#{min_unavailable_rollout} is not a valid value for required-rollout"
+      end
+    end
+
+    def partial_deployment_success
```
KnVerey: nit: returns a boolean so should end in a `?`

test/fixtures/slow-cloud/web.yml.erb
```diff
@@ -0,0 +1,75 @@
...
+        app: slow-cloud
+    spec:
+      containers:
+      - name: app
+        readinessProbe:
+          exec:
+            command:
+              - sleep
+              - '4'
+        image: nginx:alpine
```
KnVerey: we're trying to use busybox for everything now (with `imagePullPolicy: IfNotPresent` and `command: ["tail", "-f", "/dev/null"]`) so that the suite only pulls one tiny image.

test/fixtures/slow-cloud/web.yml.erb
```diff
@@ -0,0 +1,75 @@
...
+      maxUnavailable: 1
+  template:
+    metadata:
+      labels:
+        name: web
+        app: slow-cloud
+    spec:
+      containers:
+      - name: app
+        readinessProbe:
```
KnVerey: Instead of having the probe sleep, would it work to set `initialDelaySeconds` to `4`? I'd be surprised if this probe would ever pass as is, because `timeoutSeconds` is `1` by default. As a rule I've also been setting `periodSeconds` to `1` in fixtures, so that we don't get delayed 10s by the probe if something weird happens.

test/integration/kubernetes_deploy_test.rb
```diff
@@ -632,6 +632,18 @@ def test_can_deploy_deployment_with_zero_replicas
     ])
   end
 
+  def test_can_deploy_deployment_with_partial_rollout_success
+    result = deploy_fixtures("slow-cloud", subset: ["configmap-data.yml", "web.yml.erb"])
```
KnVerey: Since this test is the purpose of this fixture set, you shouldn't need to deploy a subset.

test/fixtures/slow-cloud/web.yml.erb
```diff
@@ -0,0 +1,75 @@
+apiVersion: extensions/v1beta1
+kind: Ingress
```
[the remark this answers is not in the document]
KnVerey: Doesn't look like you really use the configmap either.

test/fixtures/slow-cloud/web.yml.erb
```diff
@@ -0,0 +1,75 @@
...
+          name: http
+        env:
+        - name: GITHUB_REV
+          value: <%= current_sha %>
+        - name: CONFIG
+          valueFrom:
+            configMapKeyRef:
+              name: hello-cloud-configmap-data
+              key: datapoint1
+      - name: sidecar
```
KnVerey: This isn't doing anything useful, so I'd remove it.

lib/kubernetes-deploy/kubernetes_resource/deployment.rb
KnVerey: I know there's not much in the way of infrastructure for this, but we should probably get some unit tests on this logic.

test/integration/kubernetes_deploy_test.rb
```diff
@@ -632,6 +632,18 @@ def test_can_deploy_deployment_with_zero_replicas
     ])
   end
 
+  def test_can_deploy_deployment_with_partial_rollout_success
```
KnVerey: nit: the point is more that a deployment with the maxUnavailable required-rollout annotation does not wait for full availability (and the copy in `restart_task_test` should talk about restarts)

test/integration/kubernetes_deploy_test.rb
```diff
@@ -632,6 +632,18 @@ def test_can_deploy_deployment_with_zero_replicas
...
+    result = deploy_fixtures("slow-cloud", subset: ["configmap-data.yml", "web.yml.erb"])
+    assert_deploy_success(result)
+
+    result = deploy_fixtures("slow-cloud", subset: ["configmap-data.yml", "web.yml.erb"])
+    assert_deploy_success(result)
+
+    assert_logs_match_all(
+      [%r{Deployment\/web\s+[34] replicas, 3 updatedReplicas, 2 availableReplicas, [12] unavailableReplica}]
```
KnVerey: could `maxSurge: 0` remove the variability here?

lib/kubernetes-deploy/kubernetes_resource/deployment.rb
stefanmb: I like Katrina's suggestion above. I think this function as is requires too much knowledge from users (i.e. calling `.blank?` on its result).

lib/kubernetes-deploy/kubernetes_resource/deployment.rb
```diff
@@ -8,17 +8,15 @@ def sync
       @found = st.success?
 
       if @found
-        deployment_data = JSON.parse(raw_json)
-        @desired_replicas = deployment_data["spec"]["replicas"].to_i
-        @latest_rs = find_latest_rs(deployment_data)
-
-        @rollout_data = { "replicas" => 0 }.merge(deployment_data["status"]
+        @deployment_data = JSON.parse(raw_json)
```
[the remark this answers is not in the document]
dturn: Its going to have to be an instance var if we want to use it in `minimum_unavailable_replicas_to_succeeded` (or whatever it gets called).

test/fixtures/slow-cloud/web.yml.erb
dturn: >I'd be surprised if this probe would ever pass as is ...
I don't understand this the tests pass so are you suggesting I'm not testing what I think I am?

lib/kubernetes-deploy/kubernetes_resource/deployment.rb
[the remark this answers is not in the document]
KnVerey: > Well that or an instance var that just captures the value

Yeah, if it's only one more value that you need, I'd be explicit about capturing that (and potentially resetting it on `sync` cycles that err) rather than the whole blob.

test/fixtures/slow-cloud/web.yml.erb
KnVerey: Yeah, I'm confused by how this works. It could be the test, or it could just be me. If the probe should time out after 1 second by default (maybe that's not the default you're getting, somehow?), and your probe sleeps for 4, how can that work?

test/integration/kubernetes_deploy_test.rb
dturn: I thought it might, but after trying it doesn't, just changes the combinations.

test/fixtures/slow-cloud/web.yml.erb
dturn: Looks like a k8s bug https://github.com/kubernetes/kubernetes/issues/26895

test/integration/kubernetes_deploy_test.rb
```diff
@@ -632,6 +632,18 @@ def test_can_deploy_deployment_with_zero_replicas
...
 
+  def test_deploy_successful_with_partial_availability
+    result = deploy_fixtures("slow-cloud")
+    assert_deploy_success(result)
+
+    result = deploy_fixtures("slow-cloud")
+    assert_deploy_success(result)
+
+    assert_logs_match_all(
+      [%r{Deployment\/web\s+[34] replicas, 3 updatedReplicas, 2 availableReplicas, [12] unavailableReplica}]
```
dturn: This test still isn't that reliable. But I'm skeptical that increasing the sleep or using `initialDelaySeconds` is going to make it better. The rest of the feedback is addressed.

dturn:
@KnVerey We may need to revisit the annotation not taking a number or `%` . It seems reasonable to me that you might not want to loose availability during a deployment so have 0 for `maxUnavailable` but still be ok when 75% of the replicas are for the new SHA

KnVerey:
Why would you want to declare it successful when you don't have acceptable capacity in the target revision? I'm worried it's because you just want to be able to twist the tool's arm when your deploys aren't rolling as fast as you'd like, not because that result is actually ok. That doesn't sound like the right thing to do, and I don't want to make the wrong thing easy.

test/integration/kubernetes_deploy_test.rb
KnVerey: The reason I think tuning the sleep/delay could help is that the gem only polls for statuses every 3 seconds. So e.g. if we make a scaling operation take, say, 5-7 seconds, we should have a much better chance of succeeding at a consistent point in the scaling, because we'll see every step.

If I'm wrong about that, or it makes the test hella slow, let's use a fancier regex (e.g. `Deployment\/web\s+(?<replicas>\d+) replicas?, (?<updatedReplicas>\d+) updatedReplicas?, (?<availableReplicas>\d+) availableReplicas?`) to parse the numbers we want out of the logs and make looser assertions (e.g. `availableReplicas < replicas`, `updatedReplicas < replicas`). It's shitty, but having multiple flaky tests is worse.

lib/kubernetes-deploy/kubernetes_resource/deployment.rb
```diff
@@ -123,5 +157,20 @@ def find_latest_rs(deployment_data)
...
+    def min_available_replicas
+      max_unavailable = @definition.dig('spec', 'strategy', 'rollingUpdate', 'maxUnavailable')
+      if max_unavailable =~ /%/
+        (@desired_replicas * (100 - max_unavailable.to_i) / 100.0).ceil
+      else
+        @desired_replicas - max_unavailable.to_i
+      end
+    end
+
+    def required_rollout
```
dturn: I put this back because `validate_definition` gets called before `sync` and I wasn't happy extracting the definition value there.

KnVerey: Wait a second, why don't we just use the value we see in the definition all the time? This annotation isn't something that'd be server-side defaulted.

lib/kubernetes-deploy/kubernetes_resource/deployment.rb
```diff
@@ -123,5 +156,19 @@ def find_latest_rs(deployment_data)
       rs.sync(latest_rs_data)
       rs
     end
+
+    def min_available_replicas
+      max_unavailable = @definition.dig('spec', 'strategy', 'rollingUpdate', 'maxUnavailable')
```
KnVerey: This one _could_ very well be server-side defaulted. IIRC you originally cached the entire resource blob from the `sync` to get this, and I asked you to save the specific value instead. Why are we not using it at all now? Am I forgetting something from earlier review passes?

lib/kubernetes-deploy/kubernetes_resource/deployment.rb
```diff
@@ -123,5 +156,19 @@ def find_latest_rs(deployment_data)
...
+      if max_unavailable =~ /%/
+        (@desired_replicas * (100 - max_unavailable.to_i) / 100.0).ceil
+      else
```
KnVerey: So if `max_unavailable` is nil, we'll effectively return `@desired_replicas`. If that's on purpose, I think it might be worth a comment. Because of the possibility of defaulting (see my other comment), we can't really validate this at template parsing time. However, it could be worth having a separate case that prints a warning (just once, not on every polling cycle). 

I think the main time this would happen is for deployments with `strategy: Recreate`, which I hadn't been considering previously. The only time I'm aware of us using them is single-replica deployments with PVCs, but there's nothing preventing others from using them with larger deployments. That would be a compelling use case for having the `%/#` option on the annotation as you suggested. That said, I still think `maxUnavailable` is the correct option for `rollingUpdate` deployments, and `%/#` options could be added in future PRs instead of now.

dturn: I've added a validation that `'kubernetes-deploy.shopify.io/required-rollout = maxUnavailable` isn't in use with `strategy: Recreate`. Since that's a bit nonSensical.

KnVerey: Works for me. The strategy isn't necessarily present in the templates, but the default has been RollingUpdate for many versions, so a validation using the definition should catch all cases in practice.

lib/kubernetes-deploy/kubernetes_resource/deployment.rb
```diff
@@ -43,10 +46,24 @@ def fetch_logs
     def deploy_succeeded?
```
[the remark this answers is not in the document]
dturn: Added unit tests

lib/kubernetes-deploy/kubernetes_resource/deployment.rb
```diff
@@ -43,10 +46,24 @@ def fetch_logs
...
-      @rollout_data["updatedReplicas"].to_i == @desired_replicas &&
-      @rollout_data["updatedReplicas"].to_i == @rollout_data["availableReplicas"].to_i
+      case required_rollout
+      when 'full'
+        @latest_rs.deploy_succeeded? &&
+        @latest_rs.desired_replicas == @desired_replicas && # latest RS fully scaled up
+        @rollout_data["updatedReplicas"].to_i == @desired_replicas &&
+        @rollout_data["updatedReplicas"].to_i == @rollout_data["availableReplicas"].to_i
+      when 'none'
+        true
```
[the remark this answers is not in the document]
dturn: I added a unit test for this

lib/kubernetes-deploy/kubernetes_resource/replica_set.rb
```diff
@@ -3,13 +3,15 @@
...
-    attr_reader :desired_replicas, :pods
+    attr_reader :desired_replicas, :ready_replicas, :available_replicas, :pods
 
     def initialize(namespace:, context:, definition:, logger:, parent: nil, deploy_started_at: nil)
       @parent = parent
       @deploy_started_at = deploy_started_at
       @rollout_data = { "replicas" => 0 }
       @desired_replicas = -1
+      @ready_replicas = -1
+      @available_replicas = -1
```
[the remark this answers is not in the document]
dturn: done

test/integration/kubernetes_deploy_test.rb
dturn: The last two commits have passed all tests on the first pass, so I'd comfortable calling them reliable.

KnVerey: I just ran this one 100 times locally, and it failed 3 times:
* `Deployment/web                                    3 replicas, 3 updatedReplicas, 3 availableReplicas`
* `Deployment/web                                    3 replicas, 3 updatedReplicas, 3 availableReplicas`
* `Deployment/web                                    4 replicas, 2 updatedReplicas, 3 availableReplicas, 1 unavailableReplica`

That's not _horrible_, but considering that there are three of them, they could end up noticeably increasing the fail rate overall. I'm thinking we should make sure we have good unit coverage of the `%` option and remove that copy of the test. Maybe the restart copy too. @stefanmb wdyt? Do you have any ideas for improving the reliability here?

lib/kubernetes-deploy/kubernetes_resource/deployment.rb
```diff
@@ -81,6 +100,24 @@ def exists?
...
+    def validate_definition
+      valid = super
+
+      unless REQUIRED_ROLLOUT_TYPES.include?(required_rollout)
+        @validation_errors << "#{REQUIRED_ROLLOUT_ANNOTATION}:#{required_rollout} is invalid "\
+        "Acceptable options: #{REQUIRED_ROLLOUT_TYPES.join(',')}"
+        valid = false
+      end
+
+      if required_rollout == 'maxUnavailable' && @definition.dig('spec', 'strategy') == 'recreate'
```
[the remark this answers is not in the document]
dturn: Good question, I have no idea. I'll just downcase everything.

test/integration/kubernetes_deploy_test.rb
dturn: We could change the restart one to be `%` based and take out the `%` one here. That would save us a test w/o any loss of coverage.

dturn:
Feedback is in. Let me know if there is anything else you'd like tweaked.

KnVerey:
The new tests already flaked on one of the runs on my branch. Test flakes are already a problem on this repo, so I think it isn't actually ok to merge two new ones that are known to be unreliable. I put some more thought into this, and here is a faster (runs in ~19s instead of ~36s) test that passed 100 times in a row for me locally:

<details>
<summary>Test</summary>

```ruby
  def test_deploy_successful_with_partial_availability
      result = deploy_fixtures("slow-cloud", sha: "deploy1")
      assert_deploy_success(result)

      result = deploy_fixtures("slow-cloud", sha: "deploy2") do |fixtures|
        dep = fixtures["web.yml.erb"]["Deployment"].first
        container = dep["spec"]["template"]["spec"]["containers"].first
        container["readinessProbe"] = {
          "exec" => { "command" => %w(sleep 5) },
          "timeoutSeconds" => 6
        }
      end
      assert_deploy_success(result)

      new_pods = kubeclient.get_pods(namespace: @namespace, label_selector: 'name=web,app=slow-cloud,sha=deploy2')
      assert new_pods.length >= 1, "Expected at least one new pod, saw #{new_pods.length}"

      new_ready_pods = new_pods.select do |pod|
        pod.status.phase == "Running" &&
        pod.status.conditions.any? { |condition| condition["type"] == "Ready" && condition["status"] == "True" }
      end
      assert_equal 1, new_ready_pods.length, "Expected exactly one new pod to be ready, saw #{new_ready_pods.length}"
  end
```
</details>

<details>
<summary>Fixture</summary>

```yaml
apiVersion: apps/v1beta1
kind: Deployment
metadata:
  name: web
  annotations:
    shipit.shopify.io/restart: "true"
    kubernetes-deploy.shopify.io/required-rollout: maxUnavailable
spec:
  replicas: 2
  selector:
    matchLabels:
      name: web
      app: slow-cloud
  strategy:
    type: RollingUpdate
    rollingUpdate:
      maxSurge: 0
      maxUnavailable: 1
  template:
    metadata:
      labels:
        name: web
        app: slow-cloud
        sha: "<%= current_sha %>"
    spec:
      terminationGracePeriodSeconds: 0
      containers:
      - name: app
        image: busybox
        imagePullPolicy: IfNotPresent
        command: ["tail", "-f", "/dev/null"]
        ports:
        - containerPort: 80
          name: http
```
</details>

```
Finished in 1859.656208s, 0.0538 runs/s, 0.4302 assertions/s.

100 runs, 800 assertions, 0 failures, 0 errors, 0 skips
```

Unless you see a reason why the test isn't valid, let's use this instead (some adjustments will need to be made for the restart version).

test/integration/restart_task_test.rb
```diff
@@ -206,6 +206,37 @@ def test_restart_failure
...
+      }
+    end
+    assert_deploy_success(result)
+
+    restart = build_restart_task
+    assert_restart_success(restart.perform(["web"]))
+
+    pods = kubeclient.get_pods(namespace: @namespace, label_selector: 'name=web,app=slow-cloud')
+    new_pods = pods.select do |pod|
+      pod.spec.containers.select { |c| c["name"] == "app" && c.env&.find { |n| n.name == "RESTARTED_AT" } }
```
KnVerey: This will always be truthy, because a `select` with no results is `[]`. So I'd expect that `new_pods` would always include all pods. I'd use `any?` instead.

test/integration/restart_task_test.rb
```diff
@@ -206,6 +206,37 @@ def test_restart_failure
...
+    end
+    assert new_pods.length >= 1, "Expected at least one new pod, saw #{new_pods.length}"
+
+    new_ready_pods = new_pods.select do |pod|
+      pod.status.phase == "Running" &&
+      pod.status.conditions.any? { |condition| condition["type"] == "Ready" && condition["status"] == "True" } &&
+      pod.spec.containers.detect { |c| c["name"] == "app" && c.env&.find { |n| n.name == "RESTARTED_AT" } }
```
KnVerey: You shouldn't need this if L225 filters properly

README.md
```diff
@@ -122,10 +122,19 @@ You can add additional variables using the `--bindings=BINDINGS` option. For exa
 
 
 ### Customizing behaviour with annotations
-
 - `kubernetes-deploy.shopify.io/timeout-override`: Override the tool's hard timeout for one specific resource. Both full ISO8601 durations and the time portion of ISO8601 durations are valid. Value must be between 1 second and 24 hours.
   - _Example values_: 45s / 3m / 1h / PT0.25H
   - _Compatibility_: all resource types (Note: `Deployment` timeouts are based on `spec.progressDeadlineSeconds` if present, and that field has a default value as of the `apps/v1beta1` group version. Using this annotation will have no effect on `Deployment`s that time out with "Timeout reason: ProgressDeadlineExceeded".)
+- `kubernetes-deploy.shopify.io/required-rollout`: Modifies how much of the rollout needs to finish
+before the deployment is considered successful.
+  - _Compatibility_: Deployment
```
KnVerey: We should be consistent about how we reference objects and whether they're in backticks. The new uses both "ReplicaSet" and "`replicaSet`", and has "deployment" and "Deployment" and "deploy". The existing text above uses "`Deployment`".

dturn: Good eye. I'm going to standardize on deployment and `replicaSet`

PR #211, opened 2017-11-24
Fix crashes with Service of type ExternalName

maxlaverse:
`Service` of type `ClusterIP` have a `selector` field but those of type `ExternalName` don't.
Trying to deploy such s service with `kubernetes-deploy` leads to a crash:

```
[FATAL][2017-11-24 16:53:19 +0100]	No actions taken
.../lib/kubernetes-deploy/kubernetes_resource/service.rb:62:in `selector': undefined method `map' for nil:NilClass (NoMethodError)
	from .../lib/kubernetes-deploy/kubernetes_resource/service.rb:72:in `fetch_related_replica_count'
	from .../lib/kubernetes-deploy/kubernetes_resource/service.rb:15:in `sync'
	from .../lib/kubernetes-deploy/concurrency.rb:13:in `each'
```

This PR modifies the behavior of the `kubernetes-deploy` service resources to succeed automatically by not searching to look at inexistent corresponding pods.

### DIFF
--- a/lib/kubernetes-deploy/kubernetes_resource/service.rb
+++ b/lib/kubernetes-deploy/kubernetes_resource/service.rb
@@ -48,7 +48,12 @@ module KubernetesDeploy
     end
 
     def requires_endpoints?
-      return false if external_name_svc? || @related_deployment_replicas.blank?
+      # Service of type External don't have endpoints
+      return false if external_name_svc?
+
+      # problem counting replicas - by default, assume endpoints are required
+      return true if @related_deployment_replicas.blank?
+
       @related_deployment_replicas > 0
     end
 
--- a/lib/kubernetes-deploy/kubernetes_resource/service.rb
+++ b/lib/kubernetes-deploy/kubernetes_resource/service.rb
@@ -18,7 +18,7 @@ module KubernetesDeploy
       elsif selects_some_pods?
         "Selects at least 1 pod"
       else
-        "Unknown"
+        "Selects 0 pods"
       end
     end
 
@@ -48,7 +48,7 @@ module KubernetesDeploy
     end
 
     def requires_endpoints?
-      # Service of type External don't have endpoints
+      # service of type External don't have endpoints
       return false if external_name_svc?
 
       # problem counting replicas - by default, assume endpoints are required
--- a/lib/kubernetes-deploy/kubernetes_resource/service.rb
+++ b/lib/kubernetes-deploy/kubernetes_resource/service.rb
@@ -11,16 +11,19 @@ module KubernetesDeploy
     end
 
     def status
-      if @num_pods_selected.blank?
+      if !requires_endpoints?
+        "Doesn't require any endpoint"
+      elsif @num_pods_selected.blank?
         "Failed to count related pods"
       elsif selects_some_pods?
         "Selects at least 1 pod"
       else
-        "Selects 0 pods"
+        "Unknown"
       end
     end
 
     def deploy_succeeded?
+      return exists? unless requires_endpoints?
       # We can't use endpoints if we want the service to be able to fail fast when the pods are down
       exposes_zero_replica_deployment? || selects_some_pods?
     end
@@ -44,28 +47,39 @@ module KubernetesDeploy
       @related_deployment_replicas == 0
     end
 
+    def requires_endpoints?
+      return false if external_name_svc? || @related_deployment_replicas.blank?
+      @related_deployment_replicas > 0
+    end
+
     def selects_some_pods?
       return false unless @num_pods_selected
       @num_pods_selected > 0
     end
 
     def selector
-      @selector ||= @definition["spec"]["selector"].map { |k, v| "#{k}=#{v}" }.join(",")
+      @selector ||= @definition["spec"].fetch("selector", []).map { |k, v| "#{k}=#{v}" }.join(",")
     end
 
     def fetch_related_pod_count
+      return 0 unless selector.present?
       raw_json, _err, st = kubectl.run("get", "pods", "--selector=#{selector}", "--output=json")
       return unless st.success?
       JSON.parse(raw_json)["items"].length
     end
 
     def fetch_related_replica_count
+      return 0 unless selector.present?
       raw_json, _err, st = kubectl.run("get", "deployments", "--selector=#{selector}", "--output=json")
       return unless st.success?
 
       deployments = JSON.parse(raw_json)["items"]
       return unless deployments.length == 1
       deployments.first["spec"]["replicas"].to_i
+    end
+
+    def external_name_svc?
+      @definition["spec"]["type"] == "ExternalName"
     end
   end
 end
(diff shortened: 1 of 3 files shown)


peiranliushop:
We can address the "undefined method" exception for selector generically.
In stead of targeting a special case of services without selectors, the ExternalName service, we can check if selector exists before mapping. If so, set it to "". In effect, kubectl will ignore option selector when its value is an empty string.

The change in service.rb can be focused on a single method as such. I am new to Ruby, the code may have space for improvement.

```
class Service < KubernetesResource
...
    def selector
      # If service is without selectors, set selector to empty string which is the default value for
      # rubectl option "--selector". 
      if @definition["spec"].key?("selector")
        @selector ||= @definition["spec"]["selector"].map { |k, v| "#{k}=#{v}" }.join(",")
      else
        @selector ||= ""
      end
      @selector
    end
```

KnVerey:
@peiranliushop's solution would still require a change in `deploy_succeeded?` as well so that it doesn't incorrectly return false, but I agree that fixing `selector` is more elegant than early returns in the `_count` methods. The method body could be: `@definition["spec"].fetch("selector", []).map { |k, v| "#{k}=#{v}" }.join(",")`.

maxlaverse:
Changed

maxlaverse:
Hi @KnVerey 
I hadn't thought about a custom status and that's a good idea. It looks better with it.
I rebased and tested this PR by deploying normal services and headless services. Everything went fine.

I'm must confess I have read this class a few time and I forget everytime how it works :/

Why would we declare that endpoints **are required** if `@related_deployment_replicas` is blank ?
The three reasons for a blank `@related_deployment_replicas` are:
* `sync` not called (isn't it always called?)
* the command to get the deployment failed
* no deployments were found (or more than 1)

KnVerey:
> Why would we declare that endpoints are required if @related_deployment_replicas is blank ?

You're exactly right about the reasons it might be blank. All three cases involved unexpected results, so what we're talking about here is really what the default should be. My reasoning is the most common use cases for services do require endpoints, making that the saner default. Also, if we are missing data for some reason (e.g. the deployment command failed), the logic overall should make us wait to try again on the next `sync` rather than potentially succeeding/failing prematurely.

test/integration/kubernetes_deploy_test.rb
```diff
@@ -11,7 +11,7 @@ def test_full_hello_cloud_set_deploy_succeeds
       "Deploying ConfigMap/hello-cloud-configmap-data (timeout: 30s)",
       "Hello from the command runner!", # unmanaged pod logs
       "Result: SUCCESS",
-      "Successfully deployed 14 resources"
+      "Successfully deployed 15 resources"
     ], in_order: true)
 
     assert_logs_match_all([
```
KnVerey: Can you please add an assertion for the logs about your new fixture here? (ie. `%r{Service/redis-external\s+Doesn't require any endpoint}`)

lib/kubernetes-deploy/kubernetes_resource/service.rb
```diff
@@ -44,28 +47,39 @@ def exposes_zero_replica_deployment?
       @related_deployment_replicas == 0
     end
 
+    def requires_endpoints?
+      return false if external_name_svc? || @related_deployment_replicas.blank?
```
KnVerey: Do you disagree with my reasoning that this should be
```ruby
return false if external_name_svc?
return true if @related_deployment_replicas.blank? # problem counting replicas - by default, assume endpoints are required
```

KnVerey:
If @peiranliushop has no objections, I'll merge and release this tomorrow morning.

lib/kubernetes-deploy/kubernetes_resource/service.rb
```diff
@@ -11,16 +11,19 @@ def sync
     end
 
     def status
-      if @num_pods_selected.blank?
+      if !requires_endpoints?
```
peiranliushop: Is it true that an endpoint is not required when the service is in type ExternalName or deployment replica is 0?
The if statement here is not quite straightforward, some comments will help to clarify.

lib/kubernetes-deploy/kubernetes_resource/service.rb
```diff
@@ -11,16 +11,19 @@ def sync
...
+        "Doesn't require any endpoint"
+      elsif @num_pods_selected.blank?
         "Failed to count related pods"
       elsif selects_some_pods?
         "Selects at least 1 pod"
       else
-        "Selects 0 pods"
+        "Unknown"
```
peiranliushop: The elsif condition selects_some_pods? applies to when pods are found. 
The else condition should cover the case when no pods are found for the service selector while replica is > 0. Besides that, are there are cases that will fall into the else clause? Message "unknown" is not clear to me.

lib/kubernetes-deploy/kubernetes_resource/service.rb
```diff
@@ -11,16 +11,19 @@ def sync
...
       end
     end
 
     def deploy_succeeded?
+      return exists? unless requires_endpoints?
```
peiranliushop: The service must be found no matter if the service is external or not. Line 26 says that if the service is external then it is deployed successfully. It may not be true.

lib/kubernetes-deploy/kubernetes_resource/service.rb
KnVerey: > Is it true that an endpoint is not required when the service is in type ExternalName or deployment replica is 0?

Yes, that's exactly it, and that already seems pretty clear to me by reading this method's name and definition. Here, making the status say "Doesn't require any endpoints" when `!requires_endpoints?` also seems straightforward. I wouldn't add any more comments.

lib/kubernetes-deploy/kubernetes_resource/service.rb
KnVerey: Hm, it seems like we'll only arrive here if we expected to select some pods but saw <= 0. If that's true, `Selects 0 pods` might still be a better message, but otoh I don't hate `Unknown` since this is a catch-all fallback case. I think you have a point, but I wouldn't block merging the PR on this if the author feels differently.

lib/kubernetes-deploy/kubernetes_resource/service.rb
KnVerey: > The service must be found no matter if the service is external or not.

Do you mean L28 should be `exists? && (exposes_zero_replica_deployment? || selects_some_pods?)`? It's true that the pod selection doesn't inherently prove the server-side service resource was created, though that problem would predate this PR. Come to think of it, this could now just be `exists? && selects_some_pods?` since `exposes_zero_replica_deployment?` is already covered by the `requires_endpoints?` check.

>  Line 26 says that if the service is external then it is deployed successfully. It may not be true.

What do you mean? If the service is external, afaik there's no good way to check whether it is in fact available, so server-side existence is the best indicator we have.

lib/kubernetes-deploy/kubernetes_resource/service.rb
maxlaverse: I replaced the message with `Unknown` to be consistent with the other services, but I did realized we could have this message for good reasons.

lib/kubernetes-deploy/kubernetes_resource/service.rb
peiranliushop: Please ignore my comment about line 26. I was confused. The code is right.

maxlaverse: 👍

KnVerey:
@maxlaverse FYI this fix is included in v0.14.1, which we just released.

PR #213, opened 2017-11-28
Fix insta-timeout bug

KnVerey:
This PR has a bunch of changes related to fixing the "status: Recreate bug" (which turned out to a progress deadline problem) and making stuff like that easier to debug in the future (or now, if the workaround doesn't work 😄 ). I believe this may be the same bug we've seen when doing a rollback on core. I've added test cases for `type: Recreate` and rollbacks, though unfortunately I wasn't able to figure out how to make them true regression tests (the buggy k8s behaviour seems to be inconsistent, and these tests do not fail on master).

Highlights:
- Work around [this k8s bug](https://github.com/kubernetes/kubernetes/issues/49637) in versions < 1.7.7 by refusing to declare a failure to progress within 10 seconds of the resource being deployed.
- Make sure Deployment failures and timeouts always log _something_ about why they happened, and include the name of the replicaset we looked at (if applicable)
- Make the resets used when a particular poll cycle fails more robust.
- Make `deploy_timed_out?` more explicit (it was sometimes a fallback `else` after the other two checks)
- Make it possible to use debug-level logging in production by disabling kubectl output logs for the EjsonSecretProvisioner (previously, the debug-level logging would have logged everything we got back from the server, including base64-encoded secret data).

It also does two renames to fix frequent sources of confusion: `Runner` (often confused with `RunnerTask`) -> `DeployTask`, and `@deploy_started` -> `@deploy_started_at`/`deploy_started?` (it's a timestamp). These admittedly aren't necessary for this PR, so I can extract them (they're separate commits) if they're making it hard to review. Same goes for the `apps/v1beta` change in the fixtures.

@Shopify/cloudplatform

### DIFF
--- a/lib/kubernetes-deploy/ejson_secret_provisioner.rb
+++ b/lib/kubernetes-deploy/ejson_secret_provisioner.rb
@@ -23,7 +23,13 @@ module KubernetesDeploy
       @ejson_file = "#{template_dir}/#{EJSON_SECRETS_FILE}"
       @logger = logger
       @prune = prune
-      @kubectl = Kubectl.new(namespace: @namespace, context: @context, logger: @logger, log_failure_by_default: false)
+      @kubectl = Kubectl.new(
+        namespace: @namespace,
+        context: @context,
+        logger: @logger,
+        log_failure_by_default: false,
+        output_is_sensitive: true # output may contain ejson secrets
+      )
     end
 
     def secret_changes_required?
--- a/lib/kubernetes-deploy/kubernetes_resource.rb
+++ b/lib/kubernetes-deploy/kubernetes_resource.rb
@@ -6,7 +6,7 @@ require 'kubernetes-deploy/kubectl'
 
 module KubernetesDeploy
   class KubernetesResource
-    attr_reader :name, :namespace, :file, :context, :validation_error_msg
+    attr_reader :name, :namespace, :context, :validation_error_msg
     attr_writer :type, :deploy_started_at
 
     TIMEOUT = 5.minutes
--- a/lib/kubernetes-deploy/kubernetes_resource.rb
+++ b/lib/kubernetes-deploy/kubernetes_resource.rb
@@ -107,11 +107,7 @@ module KubernetesDeploy
     end
 
     def type
-      @type || self.class.name.split('::').last
-    end
-
-    def deploy_finished?
-      deploy_failed? || deploy_succeeded? || deploy_timed_out?
+      @type || self.class.name.demodulize
     end
 
     def deploy_timed_out?
@@ -129,9 +125,14 @@ module KubernetesDeploy
       if deploy_failed?
         helpful_info << ColorizedString.new("#{id}: FAILED").red
         helpful_info << failure_message if failure_message.present?
-      else
+      elsif deploy_timed_out?
         helpful_info << ColorizedString.new("#{id}: TIMED OUT").yellow + " (limit: #{timeout}s)"
         helpful_info << timeout_message if timeout_message.present?
+      else
+        # Arriving in debug_message when we neither failed nor timed out is very unexpected. Dump all available info.
+        helpful_info << ColorizedString.new("#{id}: MONITORING ERROR").red
+        helpful_info << failure_message if failure_message.present?
+        helpful_info << timeout_message if timeout_message.present? && timeout_message != STANDARD_TIMEOUT_MESSAGE
       end
       helpful_info << "  - Final status: #{status}"
 
--- a/lib/kubernetes-deploy/kubernetes_resource.rb
+++ b/lib/kubernetes-deploy/kubernetes_resource.rb
@@ -7,7 +7,7 @@ require 'kubernetes-deploy/kubectl'
 module KubernetesDeploy
   class KubernetesResource
     attr_reader :name, :namespace, :file, :context, :validation_error_msg
-    attr_writer :type, :deploy_started
+    attr_writer :type, :deploy_started_at
 
     TIMEOUT = 5.minutes
     LOG_LINE_COUNT = 250
@@ -86,8 +86,12 @@ module KubernetesDeploy
       false
     end
 
+    def deploy_started?
+      @deploy_started_at.present?
+    end
+
     def deploy_succeeded?
-      if @deploy_started && !@success_assumption_warning_shown
+      if deploy_started? && !@success_assumption_warning_shown
         @logger.warn("Don't know how to monitor resources of type #{type}. Assuming #{id} deployed successfully.")
         @success_assumption_warning_shown = true
       end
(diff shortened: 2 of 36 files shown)


ibawt:
LGTM I'd probably split them out but I'm not very concerned.

lib/kubernetes-deploy/kubernetes_resource/pod.rb
```diff
@@ -62,19 +63,23 @@ def timeout_message
...
+        else
+          "The following containers are in a state that is unlikely to be recoverable:\n"
+        end
+        doomed_containers.each do |c|
+          red_name = ColorizedString.new(c.name).red
+          container_problems += "> #{red_name}: #{c.doom_reason}\n"
+        end
       end
-      intro + "\n" + container_messages.join("\n") + "\n"
+      "#{phase_problem}#{container_problems}".presence
```
KnVerey: Here's an example where both are present:

image

lib/kubernetes-deploy/kubernetes_resource/pod_set_base.rb
```diff
@@ -2,10 +2,12 @@
 module KubernetesDeploy
   class PodSetBase < KubernetesResource
     def failure_message
+      "Pods in #{name} #{self.class.name.demodulize} are failing. " +
```
KnVerey: Example of Daemonset with this message:

image

test/integration/kubernetes_deploy_test.rb
```diff
@@ -252,23 +252,42 @@ def test_invalid_k8s_spec_that_is_valid_yaml_but_has_no_template_path_in_error_p
     ], in_order: true)
   end
 
-  def test_bad_container_image_on_run_once_halts_and_fails_deploy
+  def test_bad_container_image_on_unmanaged_pod_halts_and_fails_deploy
```
KnVerey: This test didn't test what it claimed to. I fixed it and added the test below to replace the thing this one was actually testing.

KnVerey:
Here's an example of the timeout logging after the commit I just pushed:

image

image

image

lib/kubernetes-deploy/ejson_secret_provisioner.rb
```diff
@@ -23,7 +23,13 @@ def initialize(namespace:, context:, template_dir:, logger:, prune: true)
       @ejson_file = "#{template_dir}/#{EJSON_SECRETS_FILE}"
       @logger = logger
       @prune = prune
-      @kubectl = Kubectl.new(namespace: @namespace, context: @context, logger: @logger, log_failure_by_default: false)
+      @kubectl = Kubectl.new(
+        namespace: @namespace,
+        context: @context,
+        logger: @logger,
+        log_failure_by_default: false,
+        output_is_sensitive: true # output may contain ejson secrets
```
stefanmb: Good catch/addition!

lib/kubernetes-deploy/kubernetes_resource.rb
```diff
@@ -103,16 +111,12 @@ def status
     end
 
     def type
-      @type || self.class.name.split('::').last
-    end
-
-    def deploy_finished?
-      deploy_failed? || deploy_succeeded? || deploy_timed_out?
+      @type || self.class.name.demodulize
```
stefanmb: TIL `demodulize` - Ruby does actually have a function for everything.

lib/kubernetes-deploy/kubernetes_resource.rb
```diff
@@ -125,9 +129,14 @@ def debug_message
       if deploy_failed?
         helpful_info << ColorizedString.new("#{id}: FAILED").red
         helpful_info << failure_message if failure_message.present?
-      else
-        helpful_info << ColorizedString.new("#{id}: TIMED OUT").yellow + " (limit: #{timeout}s)"
+      elsif deploy_timed_out?
```
stefanmb: I can't think of a way this would break, it seems like a safe fix.

lib/kubernetes-deploy/kubernetes_resource/deployment.rb
```diff
@@ -11,15 +11,21 @@ def sync
...
         conditions = deployment_data.fetch("status", {}).fetch("conditions", [])
-        @progress = conditions.find { |condition| condition['type'] == 'Progressing' }
+        @progress_condition = conditions.find { |condition| condition['type'] == 'Progressing' }
+        @progress_deadline = deployment_data['spec']['progressDeadlineSeconds']
       else # reset
         @latest_rs = nil
         @rollout_data = { "replicas" => 0 }
         @status = nil
+        @progress_condition = nil
+        @progress_deadline = @definition['spec']['progressDeadlineSeconds']
```
stefanmb: Will this always be present in the spec (it's filled by server side defaults)?

lib/kubernetes-deploy/kubernetes_resource/deployment.rb
```diff
@@ -44,26 +50,31 @@ def deploy_succeeded?
...
       else
-        @latest_rs&.timeout_message
+        "Timeout reason: hard deadline for #{type}"
       end
+      return reason_msg unless @latest_rs.present?
+      "#{reason_msg}\nLatest ReplicaSet: #{@latest_rs.name}\n\n#{@latest_rs.timeout_message}"
+    end
+
+    def pretty_timeout_type
+      @progress_deadline.present? ? "progress deadline: #{@progress_deadline}s" : super
```
stefanmb: When will `progress_deadline?` not be present?

We require its presence in https://github.com/Shopify/kubernetes-deploy/pull/213/files#diff-4709b5257970978173e422904663525fR92

lib/kubernetes-deploy/kubernetes_resource/pod.rb
stefanmb: TIL `presence` - clever.

lib/kubernetes-deploy/kubernetes_resource.rb
KnVerey: This is actually coming from ActiveSupport, but you're still not wrong 😉

lib/kubernetes-deploy/kubernetes_resource/deployment.rb
KnVerey: If it is filled by server-side defaults, it will be in the live copy (i.e. `deployment_data` above) but not this one (`@definition`, which is just the template the user gave us). I figured that if we couldn't access the server version for some reason, getting what the template had is slightly better than defaulting to `nil`.

lib/kubernetes-deploy/kubernetes_resource/deployment.rb
KnVerey: There is no server-side default in `extensions/v1beta1`, so it will not be present in an `extensions/v1beta1` deployment that doesn't explicitly include the field. I have a test that uses such a deployment, actually: 

image

The method you linked actually has a `return false unless @progress_condition.present?` guard, though there's a giant comment making that hard to notice. 😄

KnVerey:
🎩  using the executable was successful. Will merge on green.

PR #215, opened 2017-11-29
Ignore nightly build in Shipit

KnVerey:
@stefanmb


KnVerey:
I'm going to merge without CI since this change is just to the shipit file

Issue #214, opened 2017-11-29
Slow test: test_pruning_works


KnVerey:
`test_pruning_works` often runs for 2+ minutes, and it turns out to be blocking on the `kubectl apply` itself. I noticed that it finally succeeded once the statefulset disappeared, and sure enough, if I keep the statefulset in the second deploy (i.e. don't let apply prune it), the test is fast again.

@Shopify/cloudplatform have you ever seen this happen to a production statefulset deploy?

PR #216, opened 2017-11-30
Make statefulset shut down faster to speed up test

KnVerey:
Fixes https://github.com/Shopify/kubernetes-deploy/issues/214. `test_pruning` sometimes takes over a minute because `apply` blocks on the statefulset scaling down for some reason. A few recent results:

`KubernetesDeployTest#test_pruning_works = 89.72 s`
`KubernetesDeployTest#test_pruning_works = 93.05 s`
`KubernetesDeployTest#test_pruning_works = 106.93 s`

After this change, `apply` waits for ~6s, and the test runs in ~25s.

### DIFF
--- a/test/fixtures/hello-cloud/stateful_set.yml
+++ b/test/fixtures/hello-cloud/stateful_set.yml
@@ -11,9 +11,9 @@ spec:
         app: hello-cloud
         name: nginx-ss
     spec:
+      terminationGracePeriodSeconds: 0
       containers:
       - name: nginx
-        image: gcr.io/google_containers/nginx-slim:0.8
-        command: ["sleep", "40"]
+        image: nginx:alpine
         ports:
         - containerPort: 80


test/fixtures/hello-cloud/stateful_set.yml
```diff
@@ -11,9 +11,9 @@ spec:
         app: hello-cloud
         name: nginx-ss
     spec:
+      terminationGracePeriodSeconds: 0
       containers:
       - name: nginx
-        image: gcr.io/google_containers/nginx-slim:0.8
-        command: ["sleep", "40"]
```
KnVerey: This is actually the biggest speed improvement, getting the `apply` speed down around 12s. I believe sleep doesn't handle SIGTERM, so we end up waiting for the pod to die or hit the default termination grace period.

test/fixtures/hello-cloud/stateful_set.yml
```diff
@@ -11,9 +11,9 @@ spec:
         app: hello-cloud
         name: nginx-ss
     spec:
+      terminationGracePeriodSeconds: 0
```
KnVerey: This cuts the time from ~12s to ~6s by SIGKILL the pods immediately instead of letting nginx shut down nicely. Note that k8s v1.6 does not have the field required to shut down all SS pods at once, so they're still being shut down in an ordered fashion. Interestingly, decreasing the replicas to 1 doesn't seem to help.

test/fixtures/hello-cloud/stateful_set.yml
```diff
@@ -11,9 +11,9 @@ spec:
...
       containers:
       - name: nginx
-        image: gcr.io/google_containers/nginx-slim:0.8
-        command: ["sleep", "40"]
+        image: nginx:alpine
```
KnVerey: Many other tests use this image, so why pull a different one from GCR just for this test?

stefanmb: Do we actually care that it be Nginx? If not, we could use an even smaller image (e.g. `busybox`) but this is probably good enough.

KnVerey: No, we don't really care. OTOH it should only be pulled once per run, so the pull time's impact should be minimal. Is there a particular busybox command you think would work nicely? I'm going to merge this now since it is a big improvement, but I can follow up to switch images on everything.

stefanmb: If you just want it to stay up forever (like Nginx would) then `tail -f /dev/null` is a good trick. IIRC the Busybox image is 2MB.

PR #219, opened 2017-12-04
More powerful bindings parser

fw42:
* Fix a bug in the `kubernetes-deploy --bindings` parser. Previously, it was not possible to pass bindings for which the value contains commas (because the comma was interpreted as a separator). This is now possible via something like `--bindings='"foo=bar,42",bla=42'`.
* Add possibility to pass a JSON encoded hash, e.g. `--bindings='{ "foo": 42, "bar": 17 }'`.
* Add some tests for the parser.

### DIFF
--- a/lib/kubernetes-deploy/bindings_parser.rb
+++ b/lib/kubernetes-deploy/bindings_parser.rb
@@ -1,0 +1,54 @@
+# frozen_string_literal: true
+require 'json'
+require 'csv'
+
+module KubernetesDeploy
+  module BindingsParser
+    extend self
+
+    def parse(string)
+      bindings = parse_json(string) || parse_csv(string)
+
+      unless bindings
+        raise ArgumentError, "Failed to parse bindings."
+      end
+
+      bindings
+    end
+
+    private
+
+    def parse_json(string)
+      bindings = JSON.parse(string)
+
+      unless bindings.is_a?(Hash)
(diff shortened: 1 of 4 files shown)


lib/kubernetes-deploy/bindings_parser.rb
```diff
@@ -0,0 +1,54 @@
...
+        raise ArgumentError, "Expected JSON data to be a hash."
+      end
+
+      bindings
+    rescue JSON::ParserError
+      nil
+    end
+
+    def parse_csv(string)
+      lines = CSV.parse(string)
```
pushrax: This might be baseless but I'm not 100% certain that this will parse things that worked before [(which just used `split(',')`)](https://github.com/ruby/ruby/blob/5a8a2706d33849cf42cd56714f519108a4ededfc/lib/optparse.rb#L1924-L1929). The most obvious example would be a newline in a value.

However, the additional quote parsing flexibility would be nice, so maybe the answer is "ship and see"

ibawt:
can you add some json to the integration tests for the bindings?

fw42:
Hm I don't see an integration test that exercises the `exe/kubernetes-deploy` script. Are there any? The integration tests I see are passing the bindings in as a hash (i.e. already parsed).

ibawt:
ok should be fine then.

Issue #179, opened 2017-09-26
ServiceAccount resources should be predeployed like ConfigMaps


albertrdixon:
The use case here is attaching a service account to the pre-deploy unmanaged Pods. Currently these pods fail if they have a service account attached since ServiceAccount objects are deployed after unmanaged Pods.

KnVerey:
Seems reasonable to me. @karanthukral, you implemented the SA class; do you agree?

karanthukral:
Seems reasonable. Should be easy to implement!

PR #221, opened 2017-12-05
ServiceAccount resources should be predeployed like ConfigMaps

peiranliushop:
https://github.com/Shopify/kubernetes-deploy/issues/179

Force deploy service accounts before unmanaged pods in redeployment.
Tested a service account is deployed before a pod that refers to it.
Tested that an image pull secret that a service account refers is deployed before all redeployed resources.

### DIFF
--- a/lib/kubernetes-deploy/ejson_secret_provisioner.rb
+++ b/lib/kubernetes-deploy/ejson_secret_provisioner.rb
@@ -116,6 +116,7 @@ module KubernetesDeploy
 
       secret_yaml = generate_secret_yaml(secret_name, secret_type, data)
       file = Tempfile.new(secret_name)
+
       file.write(secret_yaml)
       file.close
 
--- a/lib/kubernetes-deploy/ejson_secret_provisioner.rb
+++ b/lib/kubernetes-deploy/ejson_secret_provisioner.rb
@@ -116,7 +116,6 @@ module KubernetesDeploy
 
       secret_yaml = generate_secret_yaml(secret_name, secret_type, data)
       file = Tempfile.new(secret_name)
-
       file.write(secret_yaml)
       file.close
 
--- a/test/fixtures/ejson-cloud/service-account.yml
+++ b/test/fixtures/ejson-cloud/service-account.yml
@@ -1,7 +1,0 @@
-apiVersion: v1
-kind: ServiceAccount
-metadata:
-  name: build-robot
-imagePullSecrets:
-- name: image-pull-secret
-automountServiceAccountToken: false
--- a/test/fixtures/ejson-cloud/service-account.yml
+++ b/test/fixtures/ejson-cloud/service-account.yml
@@ -1,0 +1,7 @@
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: build-robot
+imagePullSecrets:
+- name: image-pull-secret
+automountServiceAccountToken: false
--- a/test/integration/kubernetes_deploy_test.rb
+++ b/test/integration/kubernetes_deploy_test.rb
@@ -27,6 +27,51 @@ class KubernetesDeployTest < KubernetesDeploy::IntegrationTest
     assert_logs_match(%r{ConfigMap/hello-cloud-configmap-data\s+Available}, 1)
   end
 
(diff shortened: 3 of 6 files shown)


lib/kubernetes-deploy/ejson_secret_provisioner.rb
```diff
@@ -116,6 +116,7 @@ def create_or_update_secret(secret_name, secret_type, data)
 
       secret_yaml = generate_secret_yaml(secret_name, secret_type, data)
       file = Tempfile.new(secret_name)
+
```
stefanmb: Spurious newline - let's aim for a minimal change set.

test/integration/kubernetes_deploy_test.rb
```diff
@@ -27,6 +27,51 @@ def test_full_hello_cloud_set_deploy_succeeds
...
+    service_account_name = "unknown-service-account"
+    result = deploy_fixtures("hello-cloud",
+      subset: ["configmap-data.yml", "unmanaged-pod.yml.erb"]) do |fixtures|
+      pod = fixtures["unmanaged-pod.yml.erb"]["Pod"].first
+      pod["spec"]["serviceAccountName"] = service_account_name
+      pod["spec"]["automountServiceAccountToken"] = false
+    end
+    # Expect deploying the unmanaged pod fails due to missing service account
+    assert_deploy_failure(result)
+    # if KUBE_CLIENT_VERSION < Gem::Version.new("1.8.0")
```
stefanmb: Is this a leftover from testing?

test/integration/kubernetes_deploy_test.rb
```diff
@@ -27,6 +27,51 @@ def test_full_hello_cloud_set_deploy_succeeds
     assert_logs_match(%r{ConfigMap/hello-cloud-configmap-data\s+Available}, 1)
   end
 
+  def test_service_account_predeployed_before_unmanaged_pod
+    # Add an undefined service account in unmanaged pod
+    service_account_name = "unknown-service-account"
```
stefanmb: Is the first part of this test adding much value?

The failure here comes from Kubernetes - we know the deployment will fail if it mentions a missing service account, so we're not testing k8s-deploy specific functionality. Unless I'm missing something, I'd suggest deleting this test case and keeping only the success case.

KnVerey: I agree with @stefanmb on this.

test/fixtures/ejson-cloud/service-account.yml
```diff
@@ -0,0 +1,7 @@
+apiVersion: v1
```
KnVerey: I don't think this new service account, and by extension your image pull secret, is ever being used, since your test uses the `hello-cloud` fixture set, not the `ejson-cloud` set. If you switch your test to use ejson-cloud, one option would be to make ejson-cloud's web.yaml _always_ mount this new service account and have your test assert that it does so.

peiranliushop: Added another test "test_pod_get_image_pull_secret_from_service_account" in ejson-cloud to verify that if a pod does not contain any ImagePullSecrets, ImagePullSecrets of the pod's ServiceAccount should be added to the pod.
Test "test_service_account_predeployed_before_unmanaged_pod" in hello_cloud is to verify that service accounts are deployed before unmanaged pods.

test/integration/kubernetes_deploy_test.rb
peiranliushop: Agree. Removed the first part of the test which does not test feature provided kubernetes-deploy

test/integration/kubernetes_deploy_test.rb
peiranliushop: Removed.

test/fixtures/ejson-cloud/service-account.yml
KnVerey: Isn't that basically just testing a kubernetes feature again? That's not something the gem is doing itself.

peiranliushop: Following our discuss, I have removed the test for image pull secret.

Issue #210, opened 2017-11-22
Support KUBECONFIG referencing multiple files


KnVerey:
If your `KUBECONFIG` env var points to multiple files, you'll currently get an error that the full value of the var is not a file.

PR #222, opened 2017-12-06
Support KUBECONFIG referencing multiple files

peiranliushop:
https://github.com/Shopify/kubernetes-deploy/issues/210

Support getting context from multiple config files in env "KUBECONFIG" for building kubeclient.


lib/kubernetes-deploy/deploy_task.rb
```diff
@@ -277,8 +277,14 @@ def validate_configuration(allow_protected_ns:, prune:)
       errors = []
       if ENV["KUBECONFIG"].blank?
         errors << "$KUBECONFIG not set"
-      elsif !File.file?(ENV["KUBECONFIG"])
-        errors << "Kube config not found at #{ENV['KUBECONFIG']}"
+      elsif config_files.empty?
+        errors << "Kube config file names not set in $KUBECONFIG"
```
stefanmb: Nitpick: "Kube config file name(s) ... "

lib/kubernetes-deploy/kubeclient_builder.rb
```diff
@@ -44,11 +45,18 @@ def build_apps_v1beta1_kubeclient(context)
     end
 
     def _build_kubeclient(api_version:, context:, endpoint_path: nil)
-      config = GoogleFriendlyConfig.read(ENV.fetch("KUBECONFIG"))
-      unless config.contexts.include?(context)
+      # Find a context defined in kube conf files that matches the input context by name
+      kube_context = nil
+      config_files.each do |f|
```
stefanmb: This works, but in Ruby a more idiomatic approach may be:

```ruby
friendly_configs = config_files.map { |f| GoogleFriendlyConfig.read(f) }
config = friendly_configs.find { |c| c.contexts.include?(context) }

raise ContextMissingError, context unless config

kube_context = config.context(context)
```

(You could get fancier with lazy enumerators and the like.)

lib/kubernetes-deploy/kubeclient_builder.rb
```diff
@@ -59,5 +67,10 @@ def _build_kubeclient(api_version:, context:, endpoint_path: nil)
       client.discover
       client
     end
+
+    def config_files
+      # Split the list by colon for Linux and Mac, and semicolon for Windows.
+      ENV.fetch("KUBECONFIG").tr(":;", " ").split
```
stefanmb: Nitpick: 
```ruby
ENV.fetch("KUBECONFIG").split(/[:;]/)
```

peiranliushop: Revised as suggested and added code to trim spaces and remove empty items from the array of file names.

lib/kubernetes-deploy/kubeclient_builder.rb
peiranliushop: Good point. Added lazy loading.

lib/kubernetes-deploy/deploy_task.rb
peiranliushop: Updated.

lib/kubernetes-deploy/kubeclient_builder.rb
```diff
@@ -44,10 +45,12 @@ def build_apps_v1beta1_kubeclient(context)
     end
 
     def _build_kubeclient(api_version:, context:, endpoint_path: nil)
-      config = GoogleFriendlyConfig.read(ENV.fetch("KUBECONFIG"))
-      unless config.contexts.include?(context)
-        raise ContextMissingError, context
-      end
+      # Find a context defined in kube conf files that matches the input context by name
+      friendly_configs = config_files.map { |f| GoogleFriendlyConfig.read(f) }
+      config = friendly_configs.each.lazy.find { |c| c.contexts.include?(context) }
```
KnVerey: What does `friendly_configs.each.lazy.find` buy us vs `friendly_configs.find` directly? I must admit I'm not super familiar with `.lazy`, but it seems like `.each` will return `self` in this case, and `.find` is invoked immediately regardless, and already stops traversing the array once it finds a match.

test/integration/kubernetes_deploy_test.rb
```diff
@@ -806,6 +806,46 @@ def test_invalid_context
...
+      assert_deploy_failure(result)
+      assert_logs_match_all([
+        'Result: FAILURE',
+        'Configuration invalid',
+        "$KUBECONFIG not set"
+      ], in_order: true)
+
+      valid_config = File.join(__dir__, '../fixtures/kube-config/valid_config.yml')
+      ENV['KUBECONFIG'] = "#{old_config}:#{valid_config}"
+      result = deploy_fixtures('hello-cloud')
```
KnVerey: We should deploy only a tiny subset (suggestion: the configmap) to make the test faster, since the contents are actually irrelevant.

test/integration/kubernetes_deploy_test.rb
```diff
@@ -806,6 +806,46 @@ def test_invalid_context
...
+      assert_logs_match_all([
+        'Result: FAILURE',
+        'Configuration invalid',
+        "Kube config not found at #{config_file}"
+      ], in_order: true)
+
+      ENV['KUBECONFIG'] = " : "
+      result = deploy_fixtures('hello-cloud')
+      assert_deploy_failure(result)
+      assert_logs_match_all([
```
KnVerey: We should probably reset the logger between these subtests. Otherwise, technically the first two assertions here could pass based on the output of the previous subtest. Basically you need to do `@logger_stream = StringIO.new`. Could be nice to make a `reset_logger` helper for this in `test_helper.rb`.

test/integration/kubernetes_deploy_test.rb
```diff
@@ -806,6 +806,46 @@ def test_invalid_context
...
+        'Result: FAILURE',
+        'Configuration invalid',
+        "Kube config file name(s) not set in $KUBECONFIG"
+      ], in_order: true)
+
+      ENV['KUBECONFIG'] = nil
```
KnVerey: FYI this case is also covered by `test_error_message_when_kubeconfig_not_set` in `test/unit/kubernetes-deploy/deploy_task_test.rb`. Technically most of your new test could be split into separate unit tests in that file, since only the successful case needs to be able to hit the api if I'm not mistaken. I'm fine either way though.

test/unit/kubernetes-deploy/kubeclient_builder_test.rb
```diff
@@ -0,0 +1,35 @@
+# frozen_string_literal: true
+require 'test_helper'
+require 'kubernetes-deploy/kubeclient_builder'
+
+class KubeClientBuilder < KubernetesDeploy::TestCase
```
KnVerey: This class name should end in `Test` by convention

exe/kubernetes-deploy
```diff
@@ -12,7 +12,8 @@ bindings = {}
 verbose_log_prefix = false
 
 ARGV.options do |opts|
-  opts.on("--bindings=BINDINGS", "Expose additional variables to ERB templates (format: k1=v1,k2=v2 or JSON)") do |binds|
+  opts.on("--bindings=BINDINGS", "Expose additional variables to ERB templates " \
```
KnVerey: Same comment as on your other PR (fine to fix this, but we should 🎩  every time we change this file since it is untested)

test/integration/kubernetes_deploy_test.rb
peiranliushop: This new test function is focused on ENV KUBECONFIG. It does not verify other mis config cases which were covered in test_error_message_when_kubeconfig_not_set. Besides, it tested multi-config in integrated test case, so I would prefer to leave it in kubernetes_deploy_test.

test/unit/kubernetes-deploy/kubeclient_builder_test.rb
peiranliushop: Updated class name.

test/integration/kubernetes_deploy_test.rb
peiranliushop: Couldn't agree more. The test time is reduced from 14sec to 1.5sec after the suggested change.

test/integration/kubernetes_deploy_test.rb
peiranliushop: Helper function reset_logger is added to clean StringIO buffer.

lib/kubernetes-deploy/kubeclient_builder.rb
peiranliushop: You are right, find/detect will return the first element found instead of loading all contexts.
lazy is to wrapper of the Enumerator returned by each to generate anther enumerator Enumerator::Lazy to load element on an as-needed basis. Here, lazy is redundant.

exe/kubernetes-deploy
peiranliushop: 🎩 tested.

lib/kubernetes-deploy/kubeclient_builder.rb
```diff
@@ -44,10 +45,12 @@ def build_apps_v1beta1_kubeclient(context)
     end
 
     def _build_kubeclient(api_version:, context:, endpoint_path: nil)
-      config = GoogleFriendlyConfig.read(ENV.fetch("KUBECONFIG"))
-      unless config.contexts.include?(context)
-        raise ContextMissingError, context
-      end
+      # Find a context defined in kube conf files that matches the input context by name
+      friendly_configs = config_files.map { |f| GoogleFriendlyConfig.read(f) }
+      config = friendly_configs.each.find { |c| c.contexts.include?(context) }
```
KnVerey: You don't even need the `.each` here. `Array` is an Enumerable, so you can call `.find` on it directly

test/integration/kubernetes_deploy_test.rb
```diff
@@ -806,6 +806,49 @@ def test_invalid_context
     end
   end
 
+  def test_multiple_configuration_files
+    old_config = ENV['KUBECONFIG']
+    begin
```
KnVerey: you actually don't need to `begin...end` here if you outdent the `ensure` to match `def`; the body of a ruby method definition is an implicit `begin...end` block.

lib/kubernetes-deploy/kubeclient_builder.rb
peiranliushop: Updated.

test/integration/kubernetes_deploy_test.rb
peiranliushop: Another trick to learn :-)

test/unit/kubernetes-deploy/kubeclient_builder_test.rb
```diff
@@ -0,0 +1,35 @@
...
+  include KubernetesDeploy::KubeclientBuilder
+
+  def setup
+    Kubeclient::Client.any_instance.stubs(:discover)
+    super
+  end
+
+  def test_build_client_from_multiple_config_files
+    old_config = ENV['KUBECONFIG']
+    begin
```
KnVerey: Same comment about the begin/end here (sorry, should have commented here at the same time

peiranliushop: Updated.

Issue #155, opened 2017-08-01
Bad output when scaling Deployment to zero


KnVerey:
![image](https://user-images.githubusercontent.com/4789493/28834373-0fad6780-76b0-11e7-9204-e931de7f76b6.png)

The Deployment was in fact successfully scaled to zero. The zero-replica integration test deploys zero replicas to begin with; we need one that deploys 1 and then scales to 0 in a second deploy.

peiranliushop:
The originally reported error log could not be reproduced although scale down was successful. A test is added in PR https://github.com/Shopify/kubernetes-deploy/pull/224
If the error logs appear again in future, the test should be able to catch it, then we can reopen the issue.
I will close the issue.

PR #224, opened 2017-12-13
Bad output when scaling Deployment to zero

peiranliushop:
Issue: https://github.com/Shopify/kubernetes-deploy/issues/155
Fix:
Add a single integration test for deployment scale down to 0 replica

### DIFF
--- a/exe/kubernetes-deploy
+++ b/exe/kubernetes-deploy
@@ -12,7 +12,8 @@ bindings = {}
 verbose_log_prefix = false
 
 ARGV.options do |opts|
-  opts.on("--bindings=BINDINGS", "Expose additional variables to ERB templates (format: k1=v1,k2=v2 or JSON)") do |binds|
+  opts.on("--bindings=BINDINGS", "Expose additional variables to ERB templates " \
+    "(format: k1=v1,k2=v2 or JSON)") do |binds|
     bindings = KubernetesDeploy::BindingsParser.parse(binds)
   end
 
--- a/test/integration/kubernetes_deploy_test.rb
+++ b/test/integration/kubernetes_deploy_test.rb
@@ -645,7 +645,7 @@ invalid type for io.k8s.apimachinery.pkg.apis.meta.v1.ObjectMeta.labels:",
     assert_logs_match_all([
       %r{Service/web\s+Selects at least 1 pod},
       %r{Deployment/web\s+1 replica, 1 updatedReplica, 1 availableReplica},
-      %r{Service/web\s+Selects 0 pods},
+      %r{Service/web\s+Doesn't require any endpoint},
       %r{Deployment/web\s+0 replicas}
     ], in_order: true)
   end
--- a/test/integration/kubernetes_deploy_test.rb
+++ b/test/integration/kubernetes_deploy_test.rb
@@ -617,6 +617,39 @@ invalid type for io.k8s.apimachinery.pkg.apis.meta.v1.ObjectMeta.labels:",
     ], in_order: true)
   end
 
+  def test_scale_existing_deployment_down_to_zero
+    pod_name = nil
+    pod_status = "Running"
+    # Create a deployement with 1 pod
+    source = 1
+    result = deploy_fixtures("hello-cloud", subset: ["configmap-data.yml", "web.yml.erb"]) do |fixtures|
+      web = fixtures["web.yml.erb"]["Deployment"].first
+      web["spec"]["replicas"] = source
+      # Disable grace period in deleting pods to speed up test
+      web["spec"]["template"]["spec"]["terminationGracePeriodSeconds"] = 0
+      pod_name = web["spec"]["template"]["metadata"]["labels"]["name"]
+    end
+    assert_deploy_success(result)
+    hello_cloud = FixtureSetAssertions::HelloCloud.new(@namespace)
+    hello_cloud.assert_pod_status(pod_name, pod_status, source)
+    # Scale down to 0 pod
+    target = 0
+    result = deploy_fixtures("hello-cloud", subset: ["configmap-data.yml", "web.yml.erb"]) do |fixtures|
+      web = fixtures["web.yml.erb"]["Deployment"].first
+      web["spec"]["replicas"] = target
+      web["spec"]["template"]["spec"]["terminationGracePeriodSeconds"] = 0
+    end
+    assert_deploy_success(result)
+    hello_cloud.assert_pod_status(pod_name, pod_status, target)
+
+    assert_logs_match_all([
+      %r{Service/web\s+Selects at least 1 pod},
+      %r{Deployment/web\s+1 replica, 1 updatedReplica, 1 availableReplica},
+      %r{Service/web\s+Selects 0 pods},
+      %r{Deployment/web\s+0 replicas}
+    ], in_order: true)
+  end
+
   def test_can_deploy_deployment_with_zero_replicas
     result = deploy_fixtures("hello-cloud", subset: ["configmap-data.yml", "web.yml.erb"]) do |fixtures|
       web = fixtures["web.yml.erb"]["Deployment"].first


test/integration/kubernetes_deploy_test.rb
```diff
@@ -616,6 +616,42 @@ def test_output_when_switching_labels_incorrectly
...
+      pod_name = web["spec"]["template"]["metadata"]["labels"]["name"]
+    end
+    assert_deploy_success(result)
+    hello_cloud = FixtureSetAssertions::HelloCloud.new(@namespace)
+    hello_cloud.assert_pod_status(pod_name, pod_status, 3)
+    assert_logs_match_all([
+      %r{Service/web\s+Selects at least 1 pod},
+      %r{Deployment/web\s+3 replicas, 3 updatedReplicas, 3 availableReplicas}
+    ])
+    # Scale down to 1 pod
```
KnVerey: Why not do a second deploy here with `deploy_fixtures`, setting `web["spec"]["replicas"] = 0`? The problem was that a deploy that scaled replicas to zero would give the wrong output (claimed the deploy timed out), which we're not really able to test if the scaling is done externally. Looking back at the issue, it looks possible that it was actually another incarnation of the instant "timeout" bug we fixed two weeks ago, so if you're now unable to replicate the output pasted in the issue, that could be why.

peiranliushop: Just updated the test code to use deploy_fixtures to apply the change of replicas instead of calling kubectl scale directly.

I couldn't reproduce the logs originally posted in the issue.
I ran test test_can_deploy_deployment_with_zero_replicas in current master and in a commit back to Sept 28, but still couldn't reproduce it. 
Do you still remember about the timeout issue that was fixed 2 weeks ago? I can pull out the specific commit and try it out again. Was the problem logs generated consistently in CI only or locally as well?

KnVerey: > I ran test test_can_deploy_deployment_with_zero_replicas in current master and in a commit back to Sept 28, but still couldn't reproduce it.

That test doesn't reproduce the issue; it's something we saw in production (the screenshot is of Shipit logs). I don't think anyone had a chance to try to reproduce locally or with a test. We opened the issue to remind ourselves to try that, but you're the first to pick it up. 😄 

This is the PR that fixed the timeout issue I mentioned: https://github.com/Shopify/kubernetes-deploy/pull/213. One approach would be to check out  a commit from the beginning of August, when the issue you're investigating was filed, and make sure your new test reproduces the bug (i.e. fails) against that code. If it doesn't, please ping me and we can decide what to do about this. I'm not sure the bug from #213 was the cause, but if it was, it would make the issue nearly impossible to reproduce reliably.

test/integration/kubernetes_deploy_test.rb
```diff
@@ -616,6 +616,52 @@ def test_output_when_switching_labels_incorrectly
...
+      web["spec"]["replicas"] = source
+      # Disable grace period in deleting pods to speed up test
+      web["spec"]["template"]["spec"]["terminationGracePeriodSeconds"] = 0
+      dep_name = web["metadata"]["name"]
+      pod_name = web["spec"]["template"]["metadata"]["labels"]["name"]
+    end
+    assert_deploy_success(result)
+    hello_cloud = FixtureSetAssertions::HelloCloud.new(@namespace)
+    hello_cloud.assert_pod_status(pod_name, pod_status, 3)
+    # Scale down to 1 pod
```
KnVerey: I don't think this step is really necessary. The problem seemed to be going from X replicas to zero. I don't have any reason to believe that 1 would be a special case.

test/integration/kubernetes_deploy_test.rb
```diff
@@ -616,6 +616,52 @@ def test_output_when_switching_labels_incorrectly
...
+    hello_cloud.assert_watch_scale_down(pod_name, pod_status, target: target)
+    # Scale down to 0 pod
+    target = 0
+    result = deploy_fixtures("hello-cloud", subset: ["configmap-data.yml", "web.yml.erb"]) do |fixtures|
+      web = fixtures["web.yml.erb"]["Deployment"].first
+      web["spec"]["replicas"] = target
+      web["spec"]["template"]["spec"]["terminationGracePeriodSeconds"] = 0
+    end
+    assert_deploy_success(result)
+    hello_cloud.assert_watch_scale_down(pod_name, pod_status, target: target)
```
KnVerey: This shouldn't be necessary either. The deploy shouldn't succeed until the desired scale is reached, so you should be able to assert that there are no replicas present immediately.

test/integration/kubernetes_deploy_test.rb
```diff
@@ -616,6 +616,52 @@ def test_output_when_switching_labels_incorrectly
...
+
+    assert_logs_match_all([
+      %r{Service/web\s+Selects at least 1 pod},
+      %r{Deployment/web\s+3 replicas, 3 updatedReplicas, 3 availableReplicas},
+      %r{Service/web\s+Selects at least 1 pod},
+      %r{Deployment/web\s+1 replica, 1 updatedReplica, 1 availableReplica},
+      %r{Service/web\s+Selects 0 pods},
+      %r{Deployment/web\s+0 replicas}
+    ])
```
KnVerey: let's use the `in_order: true` option here

exe/kubernetes-deploy
```diff
@@ -12,7 +12,8 @@ bindings = {}
 verbose_log_prefix = false
 
 ARGV.options do |opts|
-  opts.on("--bindings=BINDINGS", "Expose additional variables to ERB templates (format: k1=v1,k2=v2 or JSON)") do |binds|
+  opts.on("--bindings=BINDINGS", "Expose additional variables to ERB templates " \
+    "(format: k1=v1,k2=v2 or JSON)") do |binds|
```
KnVerey: I don't mind having this whitespace fix in, but you'll need to 🎩  the executable just in case if we keep this change. We have no test coverage for the executables files.

test/integration/kubernetes_deploy_test.rb
```diff
@@ -616,6 +616,39 @@ def test_output_when_switching_labels_incorrectly
     ], in_order: true)
   end
 
+  def test_scale_down_deployment
```
KnVerey: nit: `test_scale_existing_deployment_down_to_zero` (even though this isn't a regression test in the end, I'd still assume the "to zero" part is what made this scenario sensitive to the premature timeouts bug)

Btw how long does this test take to run?

test/integration/kubernetes_deploy_test.rb
```diff
@@ -616,6 +616,39 @@ def test_output_when_switching_labels_incorrectly
...
+      web["spec"]["replicas"] = target
+      web["spec"]["template"]["spec"]["terminationGracePeriodSeconds"] = 0
+    end
+    assert_deploy_success(result)
+    hello_cloud.assert_pod_status(pod_name, pod_status, target)
+
+    assert_logs_match_all([
+      %r{Service/web\s+Selects at least 1 pod},
+      %r{Deployment/web\s+1 replica, 1 updatedReplica, 1 availableReplica},
+      %r{Service/web\s+Selects 0 pods},
```
KnVerey: Beware that https://github.com/Shopify/kubernetes-deploy/pull/211 is going to change the wording of this log entry and is similarly close to merging.

exe/kubernetes-deploy
peiranliushop: Manually tested kubernetes-deploy executable with option --bindings=BINDINGS and applied two variables in ERA template with success.

test/integration/kubernetes_deploy_test.rb
peiranliushop: Updated the test function name.

Total time for the test is 7 sec in which the first deployment of 1 pod took 4.9sec and the scaledown to 0 pod took the rest 2.1 sec.

test/integration/kubernetes_deploy_test.rb
peiranliushop: Thanks for reminding me. I will not merge until #211 is done, and I will rebase and test it again.

peiranliushop:
This pull request is to create a new test for scaling down deployment to 0 replica. The originally reported error log could not be reproduced although scale down was successful. If the error logs appear again in future, this test should be able to catch it, then we can reopen the issue.

Issue #44, opened 2017-02-27
Support Helm-style templating


KnVerey:
- {env}.yaml with most values app maintainers will change parameterized
- partials
- probably not ERB (cloudplatform team did prototypes of this in both ERB and Golang)

KnVerey:
A few notes: 
- We will probably stick with ERB after all
- We need to re-survey the state of the community tooling and opinion on this before implementing anything. Relevant:
  - [Whitebox COTS application management](https://docs.google.com/document/d/1S3l2F40LCwFKg6WG0srR6056IiZJBwDmDvzHWRffTWk/edit#heading=h.gh6cf96u8mlr) (need to join sig-apps google group for access)
  - [Templates design proposal](https://github.com/kubernetes/community/blob/master/contributors/design-proposals/templates.md) (notably argues against partials)
  - [Rust prototype based on that proposal](https://github.com/InQuicker/ktmpl)
  - PodTemplate may eventually become independently usable: [issue1](https://github.com/kubernetes/kubernetes/issues/42789), [issue2](https://github.com/kubernetes/kubernetes/issues/170)

cc @xldenis

errordeveloper:
Have you seen my [`kubegen`](http://github.com/errordeveloper/kubegen) project? Lately, I've been playing around with some ideas around a macro syntax that would be native to YAML/JSON (somewhat alike CloudFormation), happy to chat more. I feel like plain text templating engines make the whole problem much more complex at times, and don't really guarantee valid output.

KnVerey:
Hi @errordeveloper! I did check out your project at some point and it looks pretty cool. 😄  I've also got my eye on the [ksonnet](https://github.com/ksonnet/ksonnet-lib) project's solution to this problem. Shopify is still using regular yamls for the time being, but we're actively exploring other options. As far as `kubernetes-deploy` is concerned, I'm currently thinking we'll either expect users to generate regular k8s defs with the tool of their choice before invoking it or at most add some sort of preprocessing plug-ins hooks. Making it super opinionated about a proprietary template format seems like the wrong way to go.

skaes:
The main pain point for us are cronjobs. One of our teams has 20 of them. Just writing them down is an ordeal. Keeping them consistent after changes, for example in env variables, is a nightmare. I would love to have partials to solve this problem.

skaes:
[...]

PR #228, opened 2017-12-19
Allow template rendering from the command line

maxlaverse:
Hey!
Would it be acceptable to have a feature to only render templates and not applying them ? Similar to what can be done with Helm ?

This allows parsing the generated template to extract some information for other tools, like all the images involved.

### DIFF
--- a/README.md
+++ b/README.md
@@ -50,6 +50,9 @@ This repo also includes related tools for [running tasks](#kubernetes-run) and [
 * [Prerequisites](#prerequisites-1)
 * [Usage](#usage-2)
 
+**KUBERNETES-RENDER**
+* [Usage](#usage-3)
+
 **DEVELOPMENT**
 * [Setup](#setup)
 * [Running the test suite locally](#running-the-test-suite-locally)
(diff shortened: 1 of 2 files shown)


KnVerey:
Hi! I think this is a great idea, and it fits in nicely with some thoughts I was having about the future of rendering in this tool (https://github.com/Shopify/kubernetes-deploy/issues/44#issuecomment-352906890). Can you make it a completely separate command though? Maybe `kubernetes-render`?  That would also require extracting a renderer class that could be called from both the new executable and `DeployTask`, which I think makes a lot of sense, especially as the renderer gets more complex. IMO the class extraction is also required for https://github.com/Shopify/kubernetes-deploy/pull/207, so please coordinate with @skaes (who isn't at Shopify either--it's great to have all these contributions! ❤️ ) about who will have time to do that piece first.

maxlaverse:
Sure @KnVerey , I'll talk to him

maxlaverse:
Hi @KnVerey.
I rebased the PR and introduced a new command as suggest.
Could you point me in the direction you want this feature to go ?

maxlaverse:
I won't really have time to work on this pull-request anymore.
I prefer to close it and re-open it in the future if want to invest in that direction again.

Issue #225, opened 2017-12-15
Timeout on stateful sets


Teots:
Hi,

I tried to deploy a stateful set with a high number of replicas and each of them takes a few minutes to start. Eventually, I ran into this error: `StatefulSet/<my_service_name>: TIMED OUT (limit: 600s)`.

Is there any way to override this timeout? I've seen [this PR](https://github.com/Shopify/kubernetes-deploy/pull/130), but the `progressDeadlineSeconds` is only available for the `Deployment` type and not for `StatefulSet` as far as I know.

KnVerey:
Hi @Teots ! This is something we've run into internally at Shopify as well, and we've been discussing possible solutions. It's true that progressDeadlineSeconds / the "progressing" status condition is not currently available on StatefulSet. 

What would be the ideal behaviour, in your experience?
A - The gem should keep watching the SS as long as it is progressing, even if the rollout might take hours. This could be done by supporting a progressDeadlineSeconds-like annotation and watching progress in the gem. We'd deprecate that annotation if/when SS get a proper progress condition.
B  - The gem should ignore SS that take a long time to roll out, as indicated by a special annotation. This supposes that these slow SS are better monitored by other means, and that it isn't useful for the deploy to hang for minutes/hours watching them.
C - The gem should declare success after a configurable (via annotation) portion of the rollout has completed successfully. This view also supposes some external monitoring of the rollout, since partial success is less of a sure signal of eventual completion for differentiated pods than undifferentiated ones (like deployment pods).
D - The gem should let you change the hard timeout for a specific SS via an annotation. Like A, this could result in extremely long watches.

*A* would be hairiest to implement/maintain. *B*  or *D* could be implemented generically for any resource type. *C* is similar to something we're already working on for deployments, except that SS don't have an existing concept of minimum availability for us to leverage (whereas for deployments, that is already part of the rollingUpdate configuration).

cc @psycotica0-shopify @dturn

Teots:
Hi @KnVerey,

thanks for your response. In our current setup option D would suffice completely. Deployment is executed by our CD pipeline and thus a long watch isn't an issue as long as we still could specify a hard timeout. Just to avoid getting stuck forever.

KnVerey:
Timeout customization is now available for all resources as of version 0.15.0. Documentation is at https://github.com/Shopify/kubernetes-deploy#customizing-behaviour-with-annotations.

PR #231, opened 2017-12-21
increase timeout to 15mins

katdrobnjakovic:
Fixes https://github.com/Shopify/redis-memcached/issues/195 

cc @shopify/redis-memcached @kwestein

### DIFF
--- a/lib/kubernetes-deploy/kubernetes_resource/redis.rb
+++ b/lib/kubernetes-deploy/kubernetes_resource/redis.rb
@@ -1,7 +1,7 @@
 # frozen_string_literal: true
 module KubernetesDeploy
   class Redis < KubernetesResource
-    TIMEOUT = 5.minutes
+    TIMEOUT = 15.minutes
     UUID_ANNOTATION = "redis.stable.shopify.io/owner_uid"
 
     def sync


KnVerey:
Looking at the linked issue, is this high timeout something you want generically for all Redis resources, or is it a small minority that are expected to take this long? I ask because in https://github.com/Shopify/kubernetes-deploy/issues/225 we were discussing the possibility of adding a generic feature that would enable you to override the timeout on any individual resource via an annotation.

katdrobnjakovic:
> Looking at the linked issue, is this high timeout something you want generically for all Redis resources, or is it a small minority that are expected to take this long?

Thanks for reading the linked PR 👍 We have only seen this timeout happen with the iq_* app redis resources; however, we are okay with this being global for all redis resources for now to solve the issues they are seeing. 

> I ask because in #225 we were discussing the possibility of adding a generic feature that would enable you to override the timeout on any individual resource via an annotation.

What is the timeframe for #225 to be shipped? We would definitely change to only overriding an individual resource using the annotation when that PR has shipped 👍

KnVerey:
Nobody had started it, but I expect it to be quite easy to implement, so I'll whip up a PR now.

KnVerey:
Opened that PR here: https://github.com/Shopify/kubernetes-deploy/pull/232. PTAL. As mentioned there, a third option is to make the `Redis` class use `progressDeadlineSeconds` if available, and set that to a really high value for the problematic resource.

katdrobnjakovic:
Okay, Ill take a look. At first glance https://github.com/Shopify/kubernetes-deploy/pull/232 is addressing our problem in a better way (individual resource increase of the timeout rather than global) so I will close this PR and focus on reviewing https://github.com/Shopify/kubernetes-deploy/pull/232.

Issue #226, opened 2017-12-15
Flakey Tests


stefanmb:
Please add new flakes to [this spreadsheet](https://docs.google.com/spreadsheets/d/16NeL3Q76n1IrRMmHwSJN6UjIKRDM4eABFN2keTjgYTw/edit?usp=sharing)

----
**Original issue**

Based on manual analysis of failed nightly builds:

- test_bad_container_on_stateful_sets_fails_with_rolling_update
    - This test seems 100% broken on k8s 1.9
    - It has also failed on [1.7](https://buildkite.com/shopify/kubernetes-deploy-gem-nightly/builds/64#5ee3391e-c2ea-4d9a-b518-ff9267a5da00)
- test_crashing_container_on_deployment_fails_quickly
   - 1.8 fails sporadically
   - [1.6 fails sporadically](https://buildkite.com/shopify/kubernetes-deploy-gem-nightly/builds/43#ddf287e7-a7a4-43b5-b13f-d4a8f2aa0fc7)
- test_deploy_result_logging_for_mixed_result_deploy
   - 1.8 fails sporadically
   - 1.6 fails sporadically
- test_deployment_with_progress_times_out_for_short_duration
   - 1.8 fails sporadically
- test_output_when_switching_labels_incorrectly
   - 1.6 fails sporadically

KnVerey:
* test_bad_init_container_on_deployment_fails_quickly
  * 1.8 fails sporadically
* test_output_of_failed_unmanaged_pod
  * 1.8 failed once (missing "The following containers encountered errors:")
* test_bad_container_on_daemon_sets_fails
  * all versions fail sporadically [1.6](https://buildkite.com/shopify/kubernetes-deploy-gem/builds/256#29f5826e-643d-4703-8a28-b6bfcd6aabc3), [1.7](https://buildkite.com/shopify/kubernetes-deploy-gem/builds/254#8017159a-87d9-4960-9783-796e54640fe1), [1.8](https://buildkite.com/shopify/kubernetes-deploy-gem-nightly/builds/19#4a398717-8ec7-4192-9a60-8a69fe8585f5) 
* test_restart_failure
[...]

PR #233, opened 2017-12-22
Improve test speed and reliability

KnVerey:
@Shopify/cloudplatform 

## What

- Tries to fix as many flakey tests as possible (https://github.com/Shopify/kubernetes-deploy/issues/226)
- Cuts test run time from ~13 minutes to ~6 minutes

**EDIT:** This also removes 1.6 from our CI and adds 1.9 in its place. We were intending to roughly follow GKE's version support, and we noticed yesterday that 1.6 is no longer available for new clusters on GKE. I'm doing it in this PR because it means it is 1.9 not 1.6 that needs to show stability now.

## How

- Runs unit tests (which can't easily be run in parallel because of rampant `any_instance` stubbing) on their own. Even if they could be run in parallel, they would need to run separately from integration tests because they should have net connect disabled, which is a global setting.
- Splits integration tests according to whether or not they can be run in parallel, and run the ones that can with N=7 (we needed to upgrade the CI agent size to do this).
  - There are only four that can't, and two of those will be able to after a planned (but low-priority) refactor
- Introduces `minitest-reporters`, which provides some helpful options like the ability to print failure logs immediately and get a list of the slowest tests in the run
- Turns most ProgressDeadlineSeconds up to 60. The shorter value was intended to protect the test cluster, but I believe it was actually causing many of the test flakes by not giving the resource enough time to get into the desired state, particularly when the cluster is overloaded. The easiest way to reproduce several of the failures was actually to set the parallelism way too high.
  - A couple tests/fixtures need to keep lower PDS values because they are actually testing timeouts and will always wait all the way until the deadline.
- Adds a few new unit tests on pods (and gives them their own fixture). I wrote these while investigating a theory that didn't pan out, but I figure we might as well improve coverage.
- Fixes up some tests that were slow for dumb reasons.

### DIFF
--- a/README.md
+++ b/README.md
@@ -74,6 +74,7 @@ This repo also includes related tools for [running tasks](#kubernetes-run) and [
 * Each app managed by `kubernetes-deploy` must have its own exclusive Kubernetes namespace.<sup>2</sup>
 
 <sup>1</sup> We run integration tests against these Kubernetes versions. Kubernetes v1.6 was officially supported in gem versions < 0.16. Kubernetes v1.5 was officially supported in gem versions < 0.12.
+
 <sup>2</sup> This requirement can be bypassed with the `--no-prune` option, but it is not recommended.
 
 
--- a/README.md
+++ b/README.md
@@ -68,17 +68,18 @@ This repo also includes related tools for [running tasks](#kubernetes-run) and [
 ## Prerequisites
 
 * Ruby 2.3+
-* Your cluster must be running Kubernetes v1.6.0 or higher
+* Your cluster must be running Kubernetes v1.7.0 or higher<sup>1</sup>
 * Each app must have a deploy directory containing its Kubernetes templates (see [Templates](#templates))
 * You must remove the` kubectl.kubernetes.io/last-applied-configuration` annotation from any resources in the namespace that are not included in your deploy directory. This annotation is added automatically when you create resources with `kubectl apply`. `kubernetes-deploy` will prune any resources that have this annotation and are not in the deploy directory.**
-* Each app managed by `kubernetes-deploy` must have its own exclusive Kubernetes namespace.**
+* Each app managed by `kubernetes-deploy` must have its own exclusive Kubernetes namespace.<sup>2</sup>
 
+<sup>1</sup> We run integration tests against these Kubernetes versions. Kubernetes v1.6 was officially supported in gem versions < 0.16. Kubernetes v1.5 was officially supported in gem versions < 0.12.
+<sup>2</sup> This requirement can be bypassed with the `--no-prune` option, but it is not recommended.
 
-**This requirement can be bypassed with the `--no-prune` option, but it is not recommended.
 
 ## Installation
 
-1. [Install kubectl](https://kubernetes.io/docs/tasks/tools/install-kubectl/#install-kubectl-binary-via-curl) (requires v1.6.0 or higher) and make sure it is available in your $PATH
+1. [Install kubectl](https://kubernetes.io/docs/tasks/tools/install-kubectl/#install-kubectl-binary-via-curl) (requires v1.7.0 or higher) and make sure it is available in your $PATH
 2. Set up your [kubeconfig file](https://kubernetes.io/docs/tasks/access-application-cluster/authenticate-across-clusters-kubeconfig/) for access to your cluster(s).
 3. `gem install kubernetes-deploy`
 
@@ -249,7 +250,7 @@ With this done, you can use the following command to restart all of them:
 
 ## Prerequisites
 
-* You've already deployed a [`PodTemplate`](https://kubernetes.io/docs/api-reference/v1.6/#podtemplate-v1-core) object with field `template` containing a `Pod` specification that does not include the `apiVersion` or `kind` parameters. An example is provided in this repo in `test/fixtures/hello-cloud/template-runner.yml`.
+* You've already deployed a [`PodTemplate`](https://kubernetes.io/docs/api-reference/v1.9/#podtemplate-v1-core) object with field `template` containing a `Pod` specification that does not include the `apiVersion` or `kind` parameters. An example is provided in this repo in `test/fixtures/hello-cloud/template-runner.yml`.
 * The `Pod` specification in that template has a container named `task-runner`.
 
 Based on this specification `kubernetes-run` will create a new pod with the entrypoint of the `task-runner ` container overridden with the supplied arguments.
@@ -275,7 +276,7 @@ Based on this specification `kubernetes-run` will create a new pod with the entr
 
 If you work for Shopify, just run `dev up`, but otherwise:
 
-1. [Install kubectl version 1.6.0 or higher](https://kubernetes.io/docs/user-guide/prereqs/) and make sure it is in your path
+1. [Install kubectl version 1.7.0 or higher](https://kubernetes.io/docs/user-guide/prereqs/) and make sure it is in your path
 2. [Install minikube](https://kubernetes.io/docs/getting-started-guides/minikube/#installation) (required to run the test suite)
 3. Check out the repo
 4. Run `bin/setup` to install dependencies
--- a/bin/ci
+++ b/bin/ci
@@ -14,6 +14,7 @@ docker run --rm \
     -e CI=1 \
     -e CODECOV_TOKEN=$CODECOV_TOKEN \
     -e COVERAGE=1 \
+    -e VERBOSE=1 \
     -w /usr/src/app \
     ruby:2.3 \
     bin/test
--- a/bin/ci
+++ b/bin/ci
@@ -15,6 +15,7 @@ docker run --rm \
     -e CODECOV_TOKEN=$CODECOV_TOKEN \
     -e COVERAGE=1 \
     -e VERBOSE=1 \
+    -e PARALLELISM=$PARALLELISM \
     -w /usr/src/app \
     ruby:2.3 \
     bin/test
--- a/bin/test
+++ b/bin/test
@@ -2,6 +2,13 @@
 set -euo pipefail
 
 echo "--- :ruby: Bundle Install"
-bundle install --jobs 4 
-echo "+++ :kubernetes: Run Tests"
-bundle exec rake test TESTOPTS='-v'
+bundle install --jobs 4
+
+echo "+++ :kubernetes: Run Unit Tests"
+bundle exec rake unit_test
+
+echo "+++ :kubernetes: Run Non-Parallel Integration Tests"
+bundle exec rake serial_integration_test
+
+echo "+++ :kubernetes: Run Parallel Integration Tests"
+PARALLELIZE_ME=1 N=7 bundle exec rake integration_test
--- a/bin/test
+++ b/bin/test
@@ -1,14 +1,44 @@
 #!/bin/bash
-set -euo pipefail
+set -uo pipefail
 
-echo "--- :ruby: Bundle Install"
-bundle install --jobs 4
+err=0
+trap 'err=1' ERR
+trap 'exit 1' SIGINT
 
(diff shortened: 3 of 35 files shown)


test/helpers/parallelizable_reporter.rb
```diff
@@ -0,0 +1,29 @@
...
+            record_failure(test)
+          end
+          print(result)
+        end
+
+        # Print fast_fail information
+        return unless @fast_fail && test.failure # test.failure includes skips
+        return if test.skipped? && !@detailed_skip
+        puts
+        puts
```
dturn: Two puts in a row is odd. Is this code copied from somewhere?

test/helpers/fixture_deploy_helper.rb
```diff
@@ -26,7 +26,7 @@ module FixtureDeployHelper
   #     pod = fixtures["unmanaged-pod.yml.erb"]["Pod"].first
   #     pod["spec"]["containers"].first["image"] = "hello-world:thisImageIsBad"
   #   end
-  def deploy_fixtures(set, subset: nil, wait: true, allow_protected_ns: false, prune: true, bindings: {}, sha: nil)
+  def deploy_fixtures(set, subset: nil, **args) # extra args are passed through to deploy_dir_without_profiling
```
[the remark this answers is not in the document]
dturn: I've usually seen it done where the last arg is a hash that gets passed down. But probably not worth refactoring everything that calls this method

test/helpers/parallelizable_reporter.rb
KnVerey: [Kinda](https://github.com/kern/minitest-reporters/blob/master/lib/minitest/reporters/default_reporter.rb), yes (I started by copying things and then changed most of it). I could `puts "\n"` instead.

test/test_helper.rb
```diff
@@ -128,7 +158,8 @@ def assert_logs_match_all(entry_list, in_order: false)
 
     def refute_logs_match(regexp)
       logging_assertion do |logs|
-        refute_match regexp, logs
```
dturn: would [assert_contains](http://www.rubydoc.info/gems/minitest-rails-shoulda/0.4.1/MiniTest/Assertions:assert_contains) be better?

test/helpers/parallelizable_reporter.rb
dturn: I think that would be better. When I see something like this, I start wondering if its a work around or a bug or something.

test/test_helper.rb
KnVerey: That's in the `minitest-rails-shoulda` gem. The problem with `refute_match` is that it escapes the newline characters, so the logs in the error message are rather unreadable. You can add an additional error message, but then the logs are printed twice (which is a lot of text, half of which is useless).

test/integration/kubernetes_deploy_test.rb
```diff
@@ -794,7 +729,7 @@ def test_bad_container_on_stateful_sets_fails_with_rolling_update
       "StatefulSet/stateful-busybox: FAILED",
       "app: Crashing repeatedly (exit 1). See logs for more information.",
       "Events (common success events excluded):",
-      "[Pod/stateful-busybox-0]	FailedSync: Error syncing pod",
+      %r{\[Pod/stateful-busybox-\d\]	BackOff: Back-off restarting failed container},
```
stefanmb: Always a single digit?

README.md
```diff
@@ -273,7 +275,7 @@ Based on this specification `kubernetes-run` will create a new pod with the entr
 
 If you work for Shopify, just run `dev up`, but otherwise:
 
-1. [Install kubectl version 1.6.0 or higher](https://kubernetes.io/docs/user-guide/prereqs/) and make sure it is in your path
+1. [Install kubectl version 1.7.0 or higher](https://kubernetes.io/docs/user-guide/prereqs/) and make sure it is in your path
```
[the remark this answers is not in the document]
KnVerey: Why would we say "1.8 or higher" if we're still supporting 1.7+?

README.md
```diff
@@ -68,17 +68,19 @@ This repo also includes related tools for [running tasks](#kubernetes-run) and [
 ## Prerequisites
 
 * Ruby 2.3+
-* Your cluster must be running Kubernetes v1.6.0 or higher
+* Your cluster must be running Kubernetes v1.7.0 or higher<sup>1</sup>
```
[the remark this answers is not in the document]
KnVerey: I had been doing it on master right before releasing, but it would indeed make particular sense to add it up-front in the PR in this case, since this PR adds documentation that mentions version 0.16.

bin/ci
```diff
@@ -14,6 +14,7 @@ docker run --rm \
     -e CI=1 \
     -e CODECOV_TOKEN=$CODECOV_TOKEN \
     -e COVERAGE=1 \
+    -e VERBOSE=1 \
```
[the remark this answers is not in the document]
KnVerey: This basically has the same effect as `TESTOPTS='-v'` did before I added Minitest::Reporters (moved here from `bin/test` to reduce the duplication required)

test/fixtures/invalid/bad_probe.yml
```diff
@@ -45,7 +47,7 @@ spec:
         readinessProbe:
           exec:
             command: ["test", "0", "-eq", "0"]
-          initialDelaySeconds: 1
-          periodSeconds: 1
+          initialDelaySeconds: 2
```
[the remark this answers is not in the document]
KnVerey: Good question. An occasional problem with this test was that this container, which should be ready basically instantly, was listed as not ready alongside the other two (which are not ready on purpose). I had a vague idea that the probe starting too early could contribute to this, but [one of the failures on 1.6 on this branch](https://buildkite.com/shopify/kubernetes-deploy-gem/builds/456#6d430396-ff90-4a3c-87c5-7722c86e5da0) was this problem. It's hard to say whether it did help the other k8s versions pass 20 times in a row though, so I'm not sure what to do with it. 😕

test/fixtures/kube-config/invalid_config.yml
```diff
@@ -2,7 +2,7 @@ apiVersion: v1
 clusters:
 - cluster:
     insecure-skip-tls-verify: true
-    server: https://192.168.0.1
+    server: https://192.168.5.1
```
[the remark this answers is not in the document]
KnVerey: It doesn't matter. Good idea.

test/integration/kubernetes_deploy_test.rb
KnVerey: SS pods are assigned a stable index based on the total number of replicas, so the test statefulset would need to have 11+ replicas for there to potentially be an additional digit

README.md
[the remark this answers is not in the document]
stefanmb: Misread the paragraph, it appeared to be aimed at Shopify employees (and we deprecated 1.7). Please ignore.

bin/test
```diff
@@ -2,6 +2,13 @@
...
+bundle install --jobs 4
+
+echo "+++ :kubernetes: Run Unit Tests"
+bundle exec rake unit_test
+
+echo "+++ :kubernetes: Run Non-Parallel Integration Tests"
+bundle exec rake serial_integration_test
+
+echo "+++ :kubernetes: Run Parallel Integration Tests"
+PARALLELIZE_ME=1 N=7 bundle exec rake integration_test
```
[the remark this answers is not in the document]
KnVerey: This file is currently only used for CI; dev does `bundle exec rake test`, but we could change that. It'd need some conditionals to not bundle install locally and look nice... I'll push something. 

As for the number, isn't `7` equal to `cpus - 1` with our current machine size? (btw I had tried `N=8` and saw timeout-related failures on random tests) Also, embarrassing question: how do I detect which to use? `if uname == Darwin; else`?

test/fixtures/kube-config/invalid_config.yml
```diff
@@ -2,7 +2,7 @@ apiVersion: v1
 clusters:
 - cluster:
     insecure-skip-tls-verify: true
-    server: https://192.168.0.1
+    server: https://127.0.0.2
```
KnVerey: Actually, this results in `The connection to the server 127.0.0.2 was refused - did you specify the right host or port?` on CI (connection refused instead of timeout). I could change the test assertion to look for that error, but it needs to happen consistently on dev/ci. What do you recommend @stefanmb ?

bin/test
```diff
@@ -1,7 +1,24 @@
 #!/bin/bash
-set -euo pipefail
```
KnVerey: I removed the `e` so that all the suites will always run

bin/test
```diff
@@ -1,7 +1,24 @@
...
+set -uo pipefail
 
-echo "--- :ruby: Bundle Install"
-bundle install --jobs 4 
-echo "+++ :kubernetes: Run Tests"
-bundle exec rake test TESTOPTS='-v'
\ No newline at end of file
+function print_header() {
```
KnVerey: Here's what it looks like locally (CI remains the same):

image

bin/test
```diff
@@ -1,7 +1,24 @@
...
+fi
+
+print_header "Run Unit Tests"
+bundle exec rake unit_test
+
+print_header "Run Non-Parallel Integration Tests"
+bundle exec rake serial_integration_test
+
+print_header "Run Parallel Integration Tests\n"
+PARALLELIZE_ME=1 N=7 bundle exec rake integration_test
```
KnVerey: My questions disappeared even though I didn't change this line.

> As for the number, isn't 7 equal to cpus - 1 with our current machine size? (btw I had tried N=8 and saw timeout-related failures on random tests) Also, embarrassing question: how do I detect which to use? if uname == Darwin; else?

bin/test
KnVerey: And now of course only the last suite's exit status is used by Buildkite. I'll fix it on Monday. Tips appreciated... I'm not great at bash.

test/fixtures/invalid/bad_probe.yml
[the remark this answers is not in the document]
stefanmb: 2 isn't that much more (or less) arbitrary than 1 so I'm okay with the new delay.

bin/test
stefanmb: It is, my thought was we could do `[num_cpus / 2, 1].max` (to account for vcpus). For our purposes the `Darwin` check would work.

This is a nice-to-have suggestion but feel free to merge without it, especially given how fickle the tests have proven with respect to concurrency. The idea behind the suggestion was to avoid having to ever adjust the parallelism values if the machines change.

bin/test
stefanmb: I would use the approach here: https://stackoverflow.com/a/42219754

Example:

```
#!/bin/bash

err=0
trap 'err=1' ERR

echo "Good Command"
badcommand
echo "Good Command"

test $err = 0
```

This way if any one command errors out the script will return an error code on exit, but only after it reaches the last line.

test/fixtures/kube-config/invalid_config.yml
stefanmb: Interesting, I tried locally with the 127.* addresses and I don't get a different error. I wonder if this is related to the Minikube setup on CI. We can leave as-is unless anyone reports breakage with their (test) setup.

bin/test
KnVerey: As discussed on Slack, what I just pushed checks if `ENV[PARALLELISM]` is set, and uses that if so. So we can set it and easily tweak it from Buildkite. If it isn't set, it essentially uses num_cpu/2.

KnVerey:
@dturn @stefanmb can you please take a look at https://github.com/Shopify/kubernetes-deploy/pull/233/commits/ac0a3b4877c2fd00404f70a8bd4966aaf8c64cd5?

dturn:
:shipit:

PR #236, opened 2018-01-10
Deployment unit tests

KnVerey:
@dturn @stefanmb 
This PR is based on https://github.com/Shopify/kubernetes-deploy/pull/208. It proposes an alternative way to unit test the new changes to `Deployment`, and adds a few unrelated tests as well to try to demonstrate the usefulness of the new helper. The main idea is to avoid calling/stubbing private methods, or setting private instance variables. It does this using a helper that calls `sync` and stubs kubectl responses. Getting `sync` involved in every test is kinda 💩 , but all the methods worth testing depend heavily on its data. LMK what you think.

### DIFF
--- a/test/fixtures/for_unit_tests/deployment_test.yml
+++ b/test/fixtures/for_unit_tests/deployment_test.yml
@@ -1,0 +1,62 @@
+---
+apiVersion: apps/v1beta1
+kind: Deployment
+metadata:
+  name: web
+  uid: foobar
+  annotations:
+    "deployment.kubernetes.io/revision": "1"
+spec:
+  replicas: 3
+  progressDeadlineSeconds: 10
+  strategy:
+    type: RollingUpdate
+    rollingUpdate:
+      maxSurge: 1
+      maxUnavailable: 1
+  selector:
+    matchLabels:
+      name: web
+      app: hello-cloud
+  template:
+    metadata:
+      labels:
+        name: web
+        app: hello-cloud
+    spec:
+      containers:
+      - name: app
+        image: busybox
+status:
+  replicas: 3
+  conditions:
+  - type: Progressing
+    status: True
+    lastUpdateTime: "2018-01-09 22:56:45 UTC"
+
+---
+apiVersion: apps/v1beta1
+kind: ReplicaSet
+metadata:
+  name: web-1
+  annotations:
+    "deployment.kubernetes.io/revision": "1"
+  ownerReferences:
+    - uid: foobar
+spec:
+  replicas: 3
+  selector:
+    matchLabels:
+      name: web
+      app: hello-cloud
+  template:
+    metadata:
+      labels:
+        name: web
+        app: hello-cloud
+    spec:
+      containers:
+      - name: app
+        image: busybox
+status:
+  replicas: 3
--- a/test/unit/kubernetes-deploy/kubernetes_resource/deployment_test.rb
+++ b/test/unit/kubernetes-deploy/kubernetes_resource/deployment_test.rb
@@ -2,118 +2,332 @@
 require 'test_helper'
 
 class DeploymentTest < KubernetesDeploy::TestCase
+  def setup
+    KubernetesDeploy::Kubectl.any_instance.expects(:run).never
+    super
+  end
+
   def test_deploy_succeeded_with_none_annotation
-    rollout = {
-      'metadata' => {
-        'name' => 'fake',
-        'annotations' => { KubernetesDeploy::Deployment::REQUIRED_ROLLOUT_ANNOTATION => 'none' }
-      }
+    deployment_status = {
+      "replicas" => 3,
+      "updatedReplicas" => 1,
+      "unavailableReplicas" => 1,
+      "availableReplicas" => 0
     }
 
-    deploy = KubernetesDeploy::Deployment.new(namespace: "", context: "", logger: logger, definition: rollout)
-    deploy.instance_variable_set(:@latest_rs, true)
-
+    rs_status = {
+      "replicas" => 3,
+      "availableReplicas" => 0,
+      "readyReplicas" => 0
+    }
+    dep_template = build_deployment_template(status: deployment_status, rollout: 'none',
+      strategy: 'RollingUpdate', max_unavailable: 1)
+    deploy = build_synced_deployment(template: dep_template, replica_sets: [build_rs_template(status: rs_status)])
     assert deploy.deploy_succeeded?
+  end
+
+  def test_deploy_succeeded_is_false_with_none_annotation_before_new_rs_created
+    deployment_status = {
+      "replicas" => 3,
+      "updatedReplicas" => 3,
+      "unavailableReplicas" => 0,
+      "availableReplicas" => 3
+    }
+    deploy = build_synced_deployment(
+      template: build_deployment_template(status: deployment_status, rollout: 'none'),
+      replica_sets: []
+    )
+    refute deploy.deploy_succeeded?
   end
 
   def test_deploy_succeeded_with_max_unavailable
-    rollout = {
-      'metadata' => {
-        'name' => 'fake',
-        'annotations' => { KubernetesDeploy::Deployment::REQUIRED_ROLLOUT_ANNOTATION => 'maxUnavailable' }
-      }
+    deployment_status = {
+      "replicas" => 3, # one terminating in old rs, one starting in new rs, one up in new rs
+      "updatedReplicas" => 2,
+      "unavailableReplicas" => 2,
+      "availableReplicas" => 1
     }
 
-    deploy = KubernetesDeploy::Deployment.new(namespace: "", context: "", logger: logger, definition: rollout)
-    mock_rs = Minitest::Mock.new
-    needed = 2
-    mock_rs.expect :present?, true
-    mock_rs.expect :desired_replicas, needed
-    mock_rs.expect :ready_replicas, needed
-    mock_rs.expect :available_replicas, needed
-    deploy.instance_variable_set(:@max_unavailable, 0)
-    deploy.instance_variable_set(:@latest_rs, mock_rs)
-    deploy.instance_variable_set(:@desired_replicas, needed)
+    rs_status = {
+      "replicas" => 2,
+      "availableReplicas" => 1,
+      "readyReplicas" => 1
+    }
+    replica_sets = [build_rs_template(status: rs_status)]
 
+    deploy = build_synced_deployment(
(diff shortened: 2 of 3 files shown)


test/fixtures/for_unit_tests/deployment_test.yml
```diff
@@ -0,0 +1,62 @@
...
+      - name: app
+        image: busybox
+status:
+  replicas: 3
+  conditions:
+  - type: Progressing
+    status: True
+    lastUpdateTime: "2018-01-09 22:56:45 UTC"
+
+---
```
dturn: Seems like this should be two files.

test/unit/kubernetes-deploy/kubernetes_resource/deployment_test.rb
```diff
@@ -2,118 +2,277 @@
...
-    deploy = KubernetesDeploy::Deployment.new(namespace: "", context: "", logger: logger, definition: rollout)
-    deploy.instance_variable_set(:@kubectl, kubectl_mock)
+      refute deploy.deploy_timed_out?
+    end
+  end
 
-    refute deploy.validate_definition
+  private
+
+  def build_synced_deployment(status:, new_rs_status:, rollout: nil, max_unavail:)
```
dturn: I found `status` and `new_rs_status` confusing. How would you feel about `status` => `deployment_status` and `new_rs_status` => `rs_status`.

Does it make sense to default `max_unavail` to `false` and the other two params to `{}`?

test/unit/kubernetes-deploy/kubernetes_resource/deployment_test.rb
```diff
@@ -2,118 +2,277 @@
...
+    else
+      spec["spec"]["strategy"] = { "type" => "Recreate" }
+    end
+
+    deploy = KubernetesDeploy::Deployment.new(namespace: "test", context: "nope", logger: logger, definition: spec)
+    deploy.kubectl.expects(:run).with("get", "Deployment", "web", "--output=json").returns(
+      [spec.to_json, "", SystemExit.new(0)]
+    )
+
+    replicasets = { "items" => [] }
```
dturn: Should the building of the `rs` and `deployment` be their own functions?

test/unit/kubernetes-deploy/kubernetes_resource/deployment_test.rb
```diff
@@ -2,118 +2,277 @@
...
+  def test_validation_fails_with_invalid_rollout_annotation
+    deploy = build_synced_deployment(status: {}, new_rs_status: {}, rollout: 'bad', max_unavail: false)
+    deploy.kubectl.expects(:run).with('create', '-f', anything, '--dry-run', '--output=name', anything).returns(
+      ["", "super failed", SystemExit.new(1)]
+    )
+    refute deploy.validate_definition
+
+    expected = <<~STRING.strip
+      super failed
+      'kubernetes-deploy.shopify.io/required-rollout: bad' is invalid. Acceptable values: maxUnavailable, full, none
```
dturn: Its a bit brittle to test against a string. You could expect `rollout_annotation_err_msg` to be called, though I have mixed feelings about testing for a private function being called too

test/fixtures/for_unit_tests/deployment_test.yml
KnVerey: My thinking is to have one fixture file per unit test file so it's clear what they're intended for, but admittedly I have no idea if that will meet our real needs when we have more of these. Another reason to keep them together is that they collectively represent a single deployed resource (the Deployment), including the cluster-side results (the RS) of shipping that resource. So they actually contain cross references (ownerRef, uid).

test/unit/kubernetes-deploy/kubernetes_resource/deployment_test.rb
KnVerey: True, but IMO the value of asserting something about it is seeing that the full string makes sense. I actually caught a few errors when I wrote these assertions (spacing/punctuation problems, but more importantly that "is invalid with strategy 'rollingUpdate'" was saying the opposite of what it meant, which I didn't realize until I read the whole thing in the test). I can at least reference the constant instead of hard-coding the annotation though.

test/unit/kubernetes-deploy/kubernetes_resource/deployment_test.rb
KnVerey: Good points. I'll try to rework this a bit.

KnVerey:
Updated

test/unit/kubernetes-deploy/kubernetes_resource/deployment_test.rb
```diff
@@ -2,118 +2,332 @@
...
-        'annotations' => { KubernetesDeploy::Deployment::REQUIRED_ROLLOUT_ANNOTATION => 'maxUnavailable' }
-      }
-    }
+  def test_deploy_succeeded_raises_with_invalid_rollout_annotation
+    deploy = build_synced_deployment(
+      template: build_deployment_template(rollout: 'bad'),
+      replica_sets: [build_rs_template]
+    )
+    msg = "'#{KubernetesDeploy::Deployment::REQUIRED_ROLLOUT_ANNOTATION}: bad' is "\
+      "invalid. Acceptable values: maxUnavailable, full, none"
```
dturn: Can you use `REQUIRED_ROLLOUT_TYPES` here instead of `maxUnavailable, full, none`.

test/unit/kubernetes-deploy/kubernetes_resource/deployment_test.rb
```diff
@@ -2,118 +2,332 @@
...
-    deploy.instance_variable_set(:@latest_rs, true)
-
+    rs_status = {
+      "replicas" => 3,
+      "availableReplicas" => 0,
+      "readyReplicas" => 0
+    }
+    dep_template = build_deployment_template(status: deployment_status, rollout: 'none',
+      strategy: 'RollingUpdate', max_unavailable: 1)
+    deploy = build_synced_deployment(template: dep_template, replica_sets: [build_rs_template(status: rs_status)])
```
dturn: Should have made this more clear. I was thinking that `build_synced_deployment` would call `build_rs_template` & `build_deployment_template`. But this works if you like it better.

KnVerey: I think this will be more flexible, in that a single test with esoteric needs could modify a field in the template before passing it into `build_synced_deployment`. This als makes it easy to write tests that need more than one replica set.

test/unit/kubernetes-deploy/kubernetes_resource/deployment_test.rb
KnVerey: I can if you feel strongly, but my opinion is that the further these get from showing the human-readable message, the less valuable the assertion is. The fact that they were written out made me notice that our `join` was missing a space.

dturn:
I'm good with the PR as is.

test/unit/kubernetes-deploy/kubernetes_resource/deployment_test.rb
```diff
@@ -2,118 +2,332 @@
...
+    base_rs_manifest = fixtures.find { |fixture| fixture["kind"] == "ReplicaSet" }
+    result = base_rs_manifest.deep_merge("status" => status)
+
+    if spec_override = status["replicas"].presence # ignores possibility of surge; need a spec_replicas arg for that
+      result["spec"]["replicas"] = spec_override
+    end
+    result
+  end
+
+  def build_synced_deployment(template:, replica_sets:)
```
stefanmb: I would rename this, perhaps:

> build_deployment_from_test_fixture(...)

As a lay person it took me several read throughs to understand that this function creates deployment objects from static test assets.

I think the name indicates that it is syncing something with the cluster.

test/unit/kubernetes-deploy/kubernetes_resource/deployment_test.rb
```diff
@@ -2,118 +2,332 @@
...
+  def test_validation_fails_with_invalid_rollout_annotation
+    deploy = build_synced_deployment(template: build_deployment_template(rollout: 'bad'), replica_sets: [])
+    deploy.kubectl.expects(:run).with('create', '-f', anything, '--dry-run', '--output=name', anything).returns(
+      ["", "super failed", SystemExit.new(1)]
+    )
+    refute deploy.validate_definition
+
+    expected = <<~STRING.strip
+      super failed
+      '#{KubernetesDeploy::Deployment::REQUIRED_ROLLOUT_ANNOTATION}: bad' is invalid. Acceptable values: maxUnavailable, full, none
```
stefanmb: Perhaps:

```
 Acceptable values: #{REQUIRED_ROLLOUT_TYPES.join(', ')}
```

... alternatively don't test for "Acceptable values" at all?

KnVerey: Danny had the same suggestion. My opinion is that the assertion isn't as valuable if it is pretty much the same as the originating code. By writing it out the way the user would read it, you can notice things like spacing/punctuation errors much more easily. But I'm clearly in the  minority on this one so I'll change it. 😄

test/unit/kubernetes-deploy/kubernetes_resource/deployment_test.rb
KnVerey: Hm, it doesn't actually build it from a fixture though, now that `build_deployment_template` is a separate helper... It is creating a resource instance, and making that instance look as though `sync` returned the `template` arg. So it is in fact supposed to be simulating a `sync` with the cluster.

test/unit/kubernetes-deploy/kubernetes_resource/deployment_test.rb
stefanmb: > My opinion is that the assertion isn't as valuable if it is pretty much the same as the originating code. 

True, which is why I wonder if it's worthwhile checking for the "Acceptable values".
Effectively that tests that the list hasn't changed (and also punctuation).

I'm ok with it either way.

test/unit/kubernetes-deploy/kubernetes_resource/deployment_test.rb
stefanmb: > So it is in fact supposed to be simulating a sync with the cluster.

Right, it just wasn't immediately obvious which way stuff was syncing (in this case fixture -> object, as opposed to cluster -> object). Feel free to ignore this comment.

KnVerey:
I changed the `REQUIRED_ROLLOUT_TYPES` thing and squashed my commits. @dturn I'll let you merge this since it is going into your PR. 😄

PR #238, opened 2018-01-12
Fetch debug info in parallel

KnVerey:
Problem: We noticed that when a bunch (~10) of core's large deployments fail, there can be a 5+ minute delay between the polling loop completing and the statsd metric for completion being emitted. The only interesting thing that happens in that period is the fetching of the debug information (i.e. logs and events). Unlike all the other API call points (validation, syncing), we're still doing this one in serial. We may also want to introduce an option to skip this altogether (or automatically disable it in some circumstance?), but I think not parallelizing here was an oversight to begin with.

@dturn @stefanmb 
cc @Shopify/cloudplatform

### DIFF
--- a/lib/kubernetes-deploy/kubernetes_resource.rb
+++ b/lib/kubernetes-deploy/kubernetes_resource.rb
@@ -142,7 +142,7 @@ module KubernetesDeploy
 
     def sync_debug_info
       @events = fetch_events
-      @logs = fetch_logs if type_supports_logs?
+      @logs = fetch_logs if supports_logs?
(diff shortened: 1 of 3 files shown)


lib/kubernetes-deploy/kubernetes_resource.rb
```diff
@@ -322,6 +328,10 @@ def create_definition_tempfile
       file&.close
     end
 
+    def type_supports_logs?
```
dturn: Its a nit but just `supports_logs?`

PR #239, opened 2018-01-12
Skip collecting logs and events on failure via env var

dturn:
When a bunch (~10) of deployments fail, there can be a 5+ minute delay between the polling loop completing and the statsd metric for completion being emitted.  https://github.com/Shopify/kubernetes-deploy/pull/238 sped this up by collecting logs and events in parallel. This PR goes one more step and provides a way to skip this altogether.

Open question: Should this be an env var?

### DIFF
--- a/lib/kubernetes-deploy/kubernetes_resource.rb
+++ b/lib/kubernetes-deploy/kubernetes_resource.rb
@@ -168,35 +168,37 @@ module KubernetesDeploy
 
       if ENV['NO_DEBUG_INFO']
         helpful_info << NO_DEBUG_INFO_MESSAGE
-      else
-        if @events.present?
-          helpful_info << "  - Events (common success events excluded):"
-          @events.each do |identifier, event_hashes|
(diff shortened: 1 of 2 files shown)


dturn:
This is ready for a second look.

lib/kubernetes-deploy/kubernetes_resource.rb
```diff
@@ -169,12 +173,16 @@ def debug_message
         @events.each do |identifier, event_hashes|
           event_hashes.each { |event| helpful_info << "      [#{identifier}]\t#{event}" }
         end
+      elsif ENV[DISABLE_FETCHING_EVENT_INFO]
+        helpful_info << DISABLED_EVENT_INFO_MESSAGE
```
KnVerey: We should probably indent this (and the log message) with "  - " like we do with the actual info. Looks a bit out of place:

image

image

(I'm using `test_deploy_result_logging_for_mixed_result_deploy` to see that)

PR #240, opened 2018-01-18
required-rollout takes a percent

dturn:
#208 added the `kubernetes-deploy.shopify.io/required-rollout` annotation. There is a corner case when  `maxUnavailable` is `0%` and the annotation becomes a no-op. This PR allows a user to specify a number or a percent (e.g. `10` or `15%` ) to the annotation instead of using the value from `maxUnavailable`. 

Previously there was a concern that allowing users to provide their own value would allow the user to  🔫 👣  . Would it be better if this feature required `maxUnavailable` to be `0%` / only took a `%`? That would limit the blast radius. 

cc: @dwradcliffe

### DIFF
--- a/lib/kubernetes-deploy/kubernetes_resource/deployment.rb
+++ b/lib/kubernetes-deploy/kubernetes_resource/deployment.rb
@@ -55,7 +55,7 @@ module KubernetesDeploy
         @rollout_data["updatedReplicas"].to_i == @rollout_data["availableReplicas"].to_i
       elsif required_rollout == 'none'
         true
-      elsif required_rollout == 'maxUnavailable' || number_or_percent(required_rollout)
+      elsif required_rollout == 'maxUnavailable' || percent?(required_rollout)
         minimum_needed = min_available_replicas
 
         @latest_rs.desired_replicas >= minimum_needed &&
@@ -101,7 +101,7 @@ module KubernetesDeploy
     def validate_definition
       super
 
-      unless REQUIRED_ROLLOUT_TYPES.include?(required_rollout) || number_or_percent(required_rollout)
+      unless REQUIRED_ROLLOUT_TYPES.include?(required_rollout) || percent?(required_rollout)
         @validation_errors << rollout_annotation_err_msg
       end
 
@@ -167,6 +167,8 @@ module KubernetesDeploy
     end
 
     def min_available_replicas
+      max_unavailable = percent?(required_rollout) ? "#{100 - required_rollout.to_i}%" : @max_unavailable
+
       if max_unavailable =~ /%/
         (@desired_replicas * (100 - max_unavailable.to_i) / 100.0).ceil
       else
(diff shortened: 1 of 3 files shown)


lib/kubernetes-deploy/kubernetes_resource/deployment.rb
```diff
@@ -168,16 +167,27 @@ def find_latest_rs(deployment_data)
     end
 
     def min_available_replicas
```
dturn: I need to rework this function I dropped in the value from the annotation, but that’s not the right.

dturn: Fixed the logic. Ready for 👀

lib/kubernetes-deploy/kubernetes_resource/deployment.rb
```diff
@@ -168,16 +167,21 @@ def find_latest_rs(deployment_data)
...
     end
 
     def required_rollout
-      @definition.dig('metadata', 'annotations', REQUIRED_ROLLOUT_ANNOTATION).presence ||
-      DEFAULT_REQUIRED_ROLLOUT
+      @definition.dig('metadata', 'annotations', REQUIRED_ROLLOUT_ANNOTATION).presence || DEFAULT_REQUIRED_ROLLOUT
+    end
+
+    def percent?(value)
+      value =~ /%/
```
KnVerey: how about `/d+%`?

test/unit/kubernetes-deploy/kubernetes_resource/deployment_test.rb
```diff
@@ -80,6 +80,34 @@ def test_deploy_succeeded_with_max_unavailable
...
+    replica_sets = [build_rs_template(status: rs_status)]
+
+    deploy = build_synced_deployment(
+      template: build_deployment_template(status: deployment_status, rollout: '33%', max_unavailable: 2),
+      replica_sets: replica_sets
+    )
+    assert deploy.deploy_succeeded?
+
+    deploy = build_synced_deployment(
+      template: build_deployment_template(status: deployment_status, rollout: '34%', max_unavailable: 2),
```
KnVerey: let's add `0%` and `100%` cases for good measure. We could also leave out `max_unavailable` in this test, since it shouldn't matter. Or we could set it to `0%` to prove it isn't being referenced.

test/unit/kubernetes-deploy/kubernetes_resource/deployment_test.rb
```diff
@@ -147,6 +175,17 @@ def test_validation_fails_with_invalid_rollout_annotation
     assert_equal expected, deploy.validation_error_msg
   end
 
+  def test_validation_with_percent_rollout_annotation
+    deploy = build_synced_deployment(template: build_deployment_template(rollout: '10%'), replica_sets: [])
+    deploy.kubectl.expects(:run).with('create', '-f', anything, '--dry-run', '--output=name', anything).returns(
+      ["", "super failed", SystemExit.new(1)]
```
KnVerey: If we're trying to prove that this value is valid, why not make this command return 0 and `assert deploy_validate_definition`/`assert_empty deploy.validation_error_msg`?

test/unit/kubernetes-deploy/kubernetes_resource/deployment_test.rb
```diff
@@ -80,6 +80,34 @@ def test_deploy_succeeded_with_max_unavailable
     refute deploy.deploy_succeeded?
   end
 
+  def test_deploy_succeeded_with_annotation_as_number
```
KnVerey: `_as_percent`?

Issue #229, opened 2017-12-19
Version 1.0 requirements tracker


KnVerey:
@Shopify/cloudplatform @kirs I'm creating this issue to track what a v1.0 of this gem might look like, particularly with regard to breaking changes we might like to make.

## Breaking changes

- [x] New name that doesn't infringe on the "Kubernetes" trademark (https://github.com/Shopify/kubernetes-deploy/issues/30)
- [x] Single CLI command with subcommands for deploy/run/restart/render (https://github.com/Shopify/kubernetes-deploy/issues/256)
- [x] [tentative] Deploy command renders ERB only when given a flag, or maybe never. (must use the render command we've split out; related to https://github.com/Shopify/kubernetes-deploy/issues/44)
- [x] Different exit status for failures and timeouts
- [ ] [nice to have] Support a config file that can be used in place of flags: https://github.com/Shopify/kubernetes-deploy/issues/301

## Features/fixes

- [x] accepts $stdin as template dir (https://github.com/Shopify/kubernetes-deploy/issues/410)
- [x] has separate renderer command (related externally contributed proposal: https://github.com/Shopify/kubernetes-deploy/pull/228)
- [x] supports all major resource types (https://github.com/Shopify/kubernetes-deploy/issues/46)
- [x] stable CI for all GA k8s versions >= 1.6 (https://github.com/Shopify/kubernetes-deploy/issues/226) 
- [x] generic CR support (https://github.com/Shopify/kubernetes-deploy/issues/128)
- [x] generic timeout override annotation (https://github.com/Shopify/kubernetes-deploy/pull/232)
- [ ] (nice to have) generic monitoring opt-out annotation (related: https://github.com/Shopify/kubernetes-deploy/pull/208)
- [x] warning for unsupported k8s versions (https://github.com/Shopify/kubernetes-deploy/issues/152)
- [x] Use $TASK_ID as deploy ID if available (i.e when running in Shipit)  (https://github.com/Shopify/kubernetes-deploy/issues/136)
- [x] Support selector-restricted pruning / sub-namespace deploys (https://github.com/Shopify/kubernetes-deploy/issues/411 / https://github.com/Shopify/kubernetes-deploy/pull/235)
- [x] Support AWS EKS (https://github.com/Shopify/kubernetes-deploy/issues/396 / https://github.com/Shopify/kubernetes-deploy/issues/406)
[...]

Issue #137, opened 2017-07-18
Summary section inaccurate when deploy aborted


KnVerey:
If you kill the deploy somewhere between the initiation and completion of an action, the deploy summary printed will be inaccurate. For example, if you abort the deploy between the predeploy and the main deploy, the summary will say "No actions taken". Ideas for solutions:

1) Introduce a mechanism for tracking the action currently being attempted and reporting it on failure
2) Don't print the summary section when the process has been killed
3) Change the message printed to be more ¯\\\_(ツ)_/¯

benlangfeld:
@KnVerey Should this have been closed by https://github.com/Shopify/kubernetes-deploy/pull/244/files#r177486770 , or is there something else to be done here?

dturn:
#244 definitely implemented (2) . The UX would be better if we'd done (1) but it hasn't come up so I think we can close this out and re-open if it comes up again.

PR #244, opened 2018-02-22
Use a new exit code for failures due to timeouts

dturn:
Pulled from https://github.com/Shopify/kubernetes-deploy/issues/229 `Different exit status for failures and timeouts`. This is technically a breaking change, but probably not one that will impact many people.

### DIFF
--- a/exe/kubernetes-deploy
+++ b/exe/kubernetes-deploy
@@ -63,9 +63,14 @@ runner = KubernetesDeploy::DeployTask.new(
   logger: logger
 )
 
-success = runner.run(
+success, err = runner.run(
   verify_result: !skip_wait,
   allow_protected_ns: allow_protected_ns,
   prune: prune
 )
-exit 1 unless success
+
+if err == :timeout
+  exit 2
+elsif !success
+  exit 1
+end
--- a/exe/kubernetes-deploy
+++ b/exe/kubernetes-deploy
@@ -63,11 +63,14 @@ runner = KubernetesDeploy::DeployTask.new(
   logger: logger
 )
 
-success, error = runner.run(
-  verify_result: !skip_wait,
-  allow_protected_ns: allow_protected_ns,
-  prune: prune
-)
-
-exit 2 if error.is_a?(KubernetesDeploy::DeploymentTimeoutError)
-exit 1 unless success
+begin
+  runner.run!(
+    verify_result: !skip_wait,
+    allow_protected_ns: allow_protected_ns,
+    prune: prune
+  )
+rescue KubernetesDeploy::DeploymentTimeoutError
+  exit 124
+rescue KubernetesDeploy::FatalDeploymentError
+  exit 1
+end
--- a/exe/kubernetes-deploy
+++ b/exe/kubernetes-deploy
@@ -69,8 +69,5 @@ success, error = runner.run(
   prune: prune
 )
 
-if error == :timeout
-  exit 2
-elsif !success
-  exit 1
-end
+exit 2 if error.is_a?(KubernetesDeploy::DeploymentTimeoutError)
+exit 1 unless success
--- a/exe/kubernetes-deploy
+++ b/exe/kubernetes-deploy
@@ -70,7 +70,7 @@ begin
     prune: prune
   )
 rescue KubernetesDeploy::DeploymentTimeoutError
-  exit 124
+  exit 70
 rescue KubernetesDeploy::FatalDeploymentError
   exit 1
 end
--- a/exe/kubernetes-deploy
+++ b/exe/kubernetes-deploy
@@ -63,13 +63,13 @@ runner = KubernetesDeploy::DeployTask.new(
   logger: logger
 )
 
-success, err = runner.run(
+success, error = runner.run(
   verify_result: !skip_wait,
   allow_protected_ns: allow_protected_ns,
   prune: prune
 )
 
-if err == :timeout
+if error == :timeout
   exit 2
 elsif !success
   exit 1
--- a/exe/kubernetes-restart
+++ b/exe/kubernetes-restart
@@ -17,7 +17,10 @@ context = ARGV[1]
 logger = KubernetesDeploy::FormattedLogger.build(namespace, context)
 
 restart = KubernetesDeploy::RestartTask.new(namespace: namespace, context: context, logger: logger)
-success, error = restart.perform(raw_deployments)
-
-exit 2 if error.is_a?(KubernetesDeploy::DeploymentTimeoutError)
-exit 1 unless success
+begin
+  restart.perform!(raw_deployments)
+rescue KubernetesDeploy::DeploymentTimeoutError
+  exit 124
+rescue KubernetesDeploy::FatalDeploymentError
+  exit 1
+end
--- a/exe/kubernetes-restart
+++ b/exe/kubernetes-restart
@@ -19,8 +19,5 @@ logger = KubernetesDeploy::FormattedLogger.build(namespace, context)
 restart = KubernetesDeploy::RestartTask.new(namespace: namespace, context: context, logger: logger)
 success, error = restart.perform(raw_deployments)
 
-if error == :timeout
-  exit 2
-elsif !success
-  exit 1
-end
+exit 2 if error.is_a?(KubernetesDeploy::DeploymentTimeoutError)
+exit 1 unless success
--- a/exe/kubernetes-restart
+++ b/exe/kubernetes-restart
@@ -20,7 +20,7 @@ restart = KubernetesDeploy::RestartTask.new(namespace: namespace, context: conte
 begin
   restart.perform!(raw_deployments)
 rescue KubernetesDeploy::DeploymentTimeoutError
-  exit 124
+  exit 70
 rescue KubernetesDeploy::FatalDeploymentError
   exit 1
 end
--- a/exe/kubernetes-restart
+++ b/exe/kubernetes-restart
@@ -17,5 +17,10 @@ context = ARGV[1]
 logger = KubernetesDeploy::FormattedLogger.build(namespace, context)
 
 restart = KubernetesDeploy::RestartTask.new(namespace: namespace, context: context, logger: logger)
-success = restart.perform(raw_deployments)
-exit 1 unless success
+success, error = restart.perform(raw_deployments)
+
+if error == :timeout
+  exit 2
+elsif !success
+  exit 1
+end
--- a/lib/kubernetes-deploy/deferred_summary_logging.rb
+++ b/lib/kubernetes-deploy/deferred_summary_logging.rb
@@ -41,7 +41,7 @@ module KubernetesDeploy
         level = :info
       elsif status == :timed_out
         heading("Result: ", status_string, :yellow)
-        level = :warn
+        level = :fatal
       else
         heading("Result: ", status_string, :red)
         level = :fatal
--- a/lib/kubernetes-deploy/deferred_summary_logging.rb
+++ b/lib/kubernetes-deploy/deferred_summary_logging.rb
@@ -39,6 +39,9 @@ module KubernetesDeploy
       if status == :success
         heading("Result: ", status_string, :green)
         level = :info
+      elsif status == :timed_out
+        heading("Result: ", status_string, :yellow)
+        level = :warn
       else
         heading("Result: ", status_string, :red)
         level = :fatal
--- a/lib/kubernetes-deploy/deferred_summary_logging.rb
+++ b/lib/kubernetes-deploy/deferred_summary_logging.rb
@@ -34,12 +34,13 @@ module KubernetesDeploy
     end
 
     # Outputs the deferred summary information saved via @logger.summary.add_action and @logger.summary.add_paragraph
-    def print_summary(success)
-      if success
-        heading("Result: ", "SUCCESS", :green)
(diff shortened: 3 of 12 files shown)


dturn:
As far as I can tell every piece of software defines their own exit status numbers:

- The timeout command uses exit code [124](https://www.gnu.org/software/coreutils/manual/html_node/timeout-invocation.html) for a timeout 
- Curl uses [28](https://ec.haxx.se/usingcurl-returns.html)
- wget doesn't even bother having a specific timeout [exit status](https://www.gnu.org/software/wget/manual/html_node/Exit-Status.html)

Given the lack of standardization, I'd prefer not to cargo cult something.

Technically this is a breaking change, but not one that makes me thing we need to bump to 1.0 immediately.  (a) we're pre 1.0 so anything goes and (b) 2 is still a failing exit code and I can't image people have `exit_status == 1` instead of `> 0`

Tests coming soon.

lib/kubernetes-deploy/deploy_task.rb
```diff
@@ -159,14 +161,15 @@ def run(verify_result: true, allow_protected_ns: false, prune: true)
...
+      [success, error]
     rescue FatalDeploymentError => error
       @logger.summary.add_action(error.message)
       success = false
+      [success, error]
     ensure
       @logger.print_summary(success)
       status = success ? "success" : "failed"
       ::StatsD.measure('all_resources.duration', StatsD.duration(start), tags: statsd_tags << "status:#{status}")
-      success
```
dturn: ensure blocks only have return values if you explicitly use the `return` keyword

test/helpers/fixture_deploy_helper.rb
```diff
@@ -26,18 +26,24 @@ module FixtureDeployHelper
   #     pod = fixtures["unmanaged-pod.yml.erb"]["Pod"].first
   #     pod["spec"]["containers"].first["image"] = "hello-world:thisImageIsBad"
   #   end
-  def deploy_fixtures(set, subset: nil, **args) # extra args are passed through to deploy_dir_without_profiling
+  def deploy_fixtures(set, subset: nil, **args, &block) # extra args are passed through to deploy_dir_without_profiling
```
dturn: I didn't want to convert every test that uses deploy_fixtures to expect to return args....

dturn:
@KnVerey This is ready for 👀 again. I like to test the executables, but I think they need to be re factored to make that easy and that should go into a different PR

lib/kubernetes-deploy/deploy_task.rb
```diff
@@ -107,6 +107,7 @@ def initialize(namespace:, context:, current_sha:, template_dir:, logger:, kubec
 
     def run(verify_result: true, allow_protected_ns: false, prune: true)
       start = Time.now.utc
+      error = nil
```
klautcomputing: Should error be a symbol or do we want to create error classes?

lib/kubernetes-deploy/deploy_task.rb
```diff
@@ -149,6 +150,7 @@ def run(verify_result: true, allow_protected_ns: false, prune: true)
         deploy_resources(resources, prune: prune, verify: true)
         ::StatsD.measure('normal_resources.duration', StatsD.duration(start_normal_resource), tags: statsd_tags)
         success = resources.all?(&:deploy_succeeded?)
+        error = :timeout if !success && resources.any?(&:deploy_timed_out?)
```
klautcomputing: And then here add which resource(s) timed out?

test/integration/kubernetes_deploy_test.rb
```diff
@@ -376,20 +383,30 @@ def test_wait_false_still_waits_for_priority_resources
...
     ])
   end
 
+  def test_failed_deployment_error_nil_when_not_timeout
+    result, error = deploy_fixtures_with_error("invalid", subset: ["cannot_run.yml"]) do |fixtures|
+      container = fixtures["cannot_run.yml"]["Deployment"].first["spec"]["template"]["spec"]["containers"].first
+      container["image"] = "some-invalid-image:badtag"
+    end
+    assert_deploy_failure(result)
+    refute_equal error, :timeout
```
klautcomputing: The name of the test suggests that the `error` is `nil` when it didn't time out, but then it doesn't test for it.

In case we want to make this also work in the future (when we maybe support other errors than `:timeout`) do we want to make the check `assert_equal error, nil`? or change the name of the test?

test/integration/restart_task_test.rb
```diff
@@ -100,6 +100,14 @@ def test_restart_not_existing_deployment
       in_order: true)
   end
 
+  def test_restart_error_nil_when_not_timeout
```
klautcomputing: same here as above

exe/kubernetes-restart
```diff
@@ -17,5 +17,10 @@ context = ARGV[1]
 logger = KubernetesDeploy::FormattedLogger.build(namespace, context)
 
 restart = KubernetesDeploy::RestartTask.new(namespace: namespace, context: context, logger: logger)
-success = restart.perform(raw_deployments)
-exit 1 unless success
+success, error = restart.perform(raw_deployments)
```
klautcomputing: same here as above

lib/kubernetes-deploy/restart_task.rb
```diff
@@ -43,14 +44,18 @@ def perform(deployments_names = nil)
...
     rescue FatalDeploymentError => error
       @logger.summary.add_action(error.message)
       success = false
+      [success, error]
     ensure
       @logger.print_summary(success)
       status = success ? "success" : "failed"
       tags = %W(namespace:#{@namespace} context:#{@context} status:#{status} deployments:#{deployments.to_a.length}})
       ::StatsD.measure('restart.duration', StatsD.duration(start), tags: tags)
+      [success, error]
```
klautcomputing: As you mentioned in your comment above `ensure`s don't return, so this can be deleted, right?

test/helpers/fixture_deploy_helper.rb
```diff
@@ -26,18 +26,24 @@ module FixtureDeployHelper
...
+    success, _ = deploy_fixtures_with_error(set, subset: subset, **args, &block)
```
klautcomputing: shouldn't we make sure that `_` is `nil`, so:

```
success, error = deploy_fixtures_with_error...
assert_equal error, nil
success
```

test/integration/kubernetes_deploy_test.rb
dturn: I think you're right,  `assert_nil error` makes sense here. I also think the test name matches the current behavior so should stay put. If / when we add more conditions we'll need to change what we assert here and the test name.

test/helpers/fixture_deploy_helper.rb
dturn: Turns out several (20) tests break with this change.

dturn:
@klautcomputing Thanks for the review. Adding an error class was a good idea. I've made the changes you suggested with the exception of the one that broke 20 tests. I'd be open to refactor that in a separate PR after this, but think its too much churn for 1 pr.

test/helpers/fixture_deploy_helper.rb
klautcomputing: I think we should change that in the future, as you mentioned. And I am fine with not doing this it in this PR.

KnVerey: This helper doesn't make any assumptions about the desired result; most (all?) tests that use `deploy_fixtures` should be consuming the return value with either `assert_deploy_failure` or `assert_deploy_success` (those helpers give you nice failure output when the result is incorrect). The 20 tests that fail are probably the ones where the deploy is  _expected_ to fail.

dturn:
Changes are in, part of me is worried this has made things worse and not better, but maybe I've just spent too much time staring at this code today.

exe/kubernetes-deploy
```diff
@@ -63,9 +63,14 @@ runner = KubernetesDeploy::DeployTask.new(
...
-)
-exit 1 unless success
+begin
+  runner.run!(
+    verify_result: !skip_wait,
+    allow_protected_ns: allow_protected_ns,
+    prune: prune
+  )
+rescue KubernetesDeploy::DeploymentTimeoutError
+  exit 124
```
KnVerey: Should we use the code Jean already configured in our Shipit? (70)

lib/kubernetes-deploy/deploy_task.rb
```diff
@@ -159,14 +171,20 @@ def run(verify_result: true, allow_protected_ns: false, prune: true)
...
       @logger.summary.add_action(error.message)
+      ::StatsD.measure('all_resources.duration', StatsD.duration(start), tags: statsd_tags << "status:timeout")
+      raise
+    rescue FatalDeploymentError => error
+      @logger.summary.add_action(error.message) if error.message.present?
+      ::StatsD.measure('all_resources.duration', StatsD.duration(start), tags: statsd_tags << "status:failed")
       success = false
+      raise
     ensure
       @logger.print_summary(success)
```
KnVerey: Did you miss this comment? https://github.com/Shopify/kubernetes-deploy/pull/244#discussion_r174644083
I'm not married to that specific implementation, but the problem described is still an issue in this version. I realized the other day that getting rid of the `ensure` would have the nice side-effect of not printing the result banner in the case of unexpected exceptions that are dumping backtraces.

test/test_helper.rb
```diff
@@ -102,13 +102,14 @@ def reset_logger
...
 
-    def assert_deploy_failure(result)
+    def assert_deploy_failure(result, cause = nil)
       if ENV["PRINT_LOGS"]
         assert_equal false, result, "Deploy succeeded when it was expected to fail"
         return
       end
 
       logging_assertion do |logs|
+        assert_match(/failed due to timeouts./, logs) if cause == :timeout
```
KnVerey: This should be asserting on `Result: TIMED OUT` (which isn't happening yet in this PR), like L114

lib/kubernetes-deploy/deploy_task.rb
dturn: I misunderstood what you were suggesting since `print_summary` takes a bool right now. I'll fix this up.

lib/kubernetes-deploy/deferred_summary_logging.rb
```diff
@@ -34,12 +34,13 @@ def heading(text, secondary_msg = '', secondary_msg_color = :cyan)
...
-      if success
-        heading("Result: ", "SUCCESS", :green)
+    def print_summary(status)
+      status_string = status.to_s.humanize.upcase
+      if status == :success
+        heading("Result: ", status_string, :green)
         level = :info
       else
-        heading("Result: ", "FAILURE", :red)
+        heading("Result: ", status_string, :red)
```
KnVerey: Would it be more consistent to make this piece of heading yellow? (Just a thought; I don't have a strong opinion)

image

lib/kubernetes-deploy/errors.rb
```diff
@@ -8,6 +8,15 @@ def initialize(name, context)
       super("Namespace `#{name}` not found in context `#{context}`")
     end
   end
+
+  class DeploymentTimeoutError < FatalDeploymentError
+    attr_reader :resources
+    def initialize(resources)
+      @resources = resources
+      super("Resources #{@resources.map(&:name).join(',')} failed due to timeouts.")
```
KnVerey: For consistency with the other messaging (e.g. right now we see `Successfully deployed 1 resource, failed to deploy 1 resource, and resources web failed due to timeouts.`), and because this could be a long list, I think this should be a count, not a list. 

Since we're trying to clarify that timeouts aren't exactly failures, might it be more helpful to not use "failed" in this message? e.g. "gave up watching X resources", "timed out waiting for X resources", "reached the watch timeout for X resources", I dunno.

lib/kubernetes-deploy/deploy_task.rb
```diff
@@ -157,16 +169,19 @@ def run(verify_result: true, allow_protected_ns: false, prune: true)
...
         @logger.summary.add_paragraph(ColorizedString.new(warning).yellow)
-        success = true
       end
-    rescue FatalDeploymentError => error
+      ::StatsD.measure('all_resources.duration', StatsD.duration(start), tags: statsd_tags << "status:success")
+      @logger.print_summary(:success)
+    rescue DeploymentTimeoutError => error
       @logger.summary.add_action(error.message)
-      success = false
-    ensure
```
KnVerey: By getting rid of the `ensure`, this PR is implementing solution 2 for https://github.com/Shopify/kubernetes-deploy/issues/137.

lib/kubernetes-deploy/deferred_summary_logging.rb
```diff
@@ -34,12 +34,16 @@ def heading(text, secondary_msg = '', secondary_msg_color = :cyan)
...
-      if success
-        heading("Result: ", "SUCCESS", :green)
+    def print_summary(status)
+      status_string = status.to_s.humanize.upcase
+      if status == :success
+        heading("Result: ", status_string, :green)
         level = :info
+      elsif status == :timed_out
+        heading("Result: ", status_string, :yellow)
+        level = :warn
```
KnVerey: I think we still want `:fatal` for this, since it did in fact kill the deploy... but also because `error` and `warn` with this particular logger make the whole lined coloured:

image

PR #245, opened 2018-02-26
Release 0.16.0

klautcomputing:
Let's release a new version!

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,11 @@
 ### 0.16.0
-
 **Important:** This release changes the officially supported Kubernetes versions to v1.7 through v1.9. Other versions may continue to work, but we are no longer running our test suite against them.
+
+*Features*
+- Support partials to reduce duplication in yaml files ([#207](https://github.com/Shopify/kubernetes-deploy/pull/207))
+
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -1,7 +1,12 @@
 ### 0.16.0
-
 **Important:** This release changes the officially supported Kubernetes versions to v1.7 through v1.9. Other versions may continue to work, but we are no longer running our test suite against them.
 
+*Features*
```
dturn: Its a nit but this has been `Enhancements` in all previous versions, is the change intentional?

KnVerey: I originally intended for there to be three things: Features (new thing not previously possible), Enhancements (existing thing made better) and Bug Fixes. Looks like we only used "Features" in 0.13.0 so far though. Do you think it's a useful distinction?

klautcomputing: Intentional because of e.g:

```
### 0.13.0
*Features*
- Added support for StatefulSets for kubernetes 1.7+ using RollingUpdate
```

dturn: I like the distinction.  tI just wanted to make sure we were being consistent.

PR #246, opened 2018-02-28
Support deploying identical images during a single deployment.

skaes:
Docker image tags, such as "latest", or a version numbering scheme like
"v1", "v2", etc., are not guaranteed to be stable. This means that when
someone pushes to the docker registry during an ongoing deployment, one
might end up with different containers on different hosts, which makes
it really hard to reason about system properties.

Fortunately, this problem can be circumvented by using docker's sha256
notation for image references. However, sha256 references are unwieldy
for humans.

This commit introduces a helper function which can used in ERB snippets
inside YAML templates to convert image references into sha256 form
automatically.

Example:

<%= image_with_digest("busybox:latest") %>

gets transformed into something like this:

busybox@sha256:187c294487772d71eb5e6675f31e8863fcd655e44f634ffbc92ed30f407a1b9d

The feature is implemented using the docker registry API to retrieve the
image shas exactly once during a deployment and works for docker hub and
private registries.

### DIFF
--- a/lib/kubernetes-deploy.rb
+++ b/lib/kubernetes-deploy.rb
@@ -19,6 +19,7 @@ require 'kubernetes-deploy/statsd'
 require 'kubernetes-deploy/concurrency'
 require 'kubernetes-deploy/bindings_parser'
 require 'kubernetes-deploy/duration_parser'
+require 'kubernetes-deploy/docker_registry'
 
 module KubernetesDeploy
   MIN_KUBE_VERSION = '1.7.0'
--- a/lib/kubernetes-deploy/docker_registry.rb
+++ b/lib/kubernetes-deploy/docker_registry.rb
@@ -1,0 +1,95 @@
+# frozen_string_literal: true
+require 'net/https'
+require 'json'
+
+module KubernetesDeploy
+  module DockerRegistry
+    extend self
+
+    def image_digest(image_url)
+      response = authenticated_request(manifest_url(image_url))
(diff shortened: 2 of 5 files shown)


dturn:
Can you please provide more context on why you're using tags this way (reusing existing tags in production), how this problem is impacting you and why you believe that a template function in this gem is the best solution? For example, alternatives might include using a pre-processor in combination with the `--bindings` option, or using an admission controller to enforce this even for manually created resources (related: https://github.com/kubernetes/kubernetes/issues/1697). Our vision for kubernetes-deploy is one where it focusses on its primary responsibilty ("[helping] you ship changes to a Kubernetes namespace and understand the result") and plays well with other interoperable tools in the Kubernetes community. As such, we're hesitant to add specialized templating functions to the renderer in general. For this particular case, we're also concerned about the pattern of making http calls from within templates, and the overhead of supporting something other than DockerHub (e.g. http://cri-o.io/) in a sustainable way.

skaes:
@dturn thx for your detailed comment. The motivation for the pull request comes from our existing development and deployment process. Most of our teams use two branches deployed to two environments: staging and production, multiple times per day. It felt unnatural to introduce new tags for every release and the proposed strategy felt like the easiest way to achieve out goal of providing safe rollback and consistent deployments. However, we since moved to Quay as a registry and this registry implementation does only allow to pull images via digest only if they also have a tag, which now forces us to introduce the tags we wanted to avoid and made this pull request obsolete.

PR #252, opened 2018-03-19
fix broken template link

doodzik:
I'm guessing that the link is supposed to point at `#using-templates-and-variables`.
The `#templates` section doesn't exist anymore.

### DIFF
--- a/README.md
+++ b/README.md
@@ -70,7 +70,7 @@ This repo also includes related tools for [running tasks](#kubernetes-run) and [
 
 * Ruby 2.3+
 * Your cluster must be running Kubernetes v1.7.0 or higher<sup>1</sup>
-* Each app must have a deploy directory containing its Kubernetes templates (see [Templates](#templates))
+* Each app must have a deploy directory containing its Kubernetes templates (see [Templates](#using-templates-and-variables))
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -70,7 +70,7 @@ This repo also includes related tools for [running tasks](#kubernetes-run) and [
 
 * Ruby 2.3+
 * Your cluster must be running Kubernetes v1.7.0 or higher<sup>1</sup>
-* Each app must have a deploy directory containing its Kubernetes templates (see [Templates](#templates))
+* Each app must have a deploy directory containing its Kubernetes templates (see [Templates](#using-templates-and-variables))
 * You must remove the` kubectl.kubernetes.io/last-applied-configuration` annotation from any resources in the namespace that are not included in your deploy directory. This annotation is added automatically when you create resources with `kubectl apply`. `kubernetes-deploy` will prune any resources that have this annotation and are not in the deploy directory.**
```
doodzik: @KnVerey correct me if I'm wrong, but it seems that `<sub>2</sub>` is supposed to be at the end of this line and not in line 75

KnVerey: You're correct!

PR #253, opened 2018-03-19
Unify render errors and handle an edge case

KnVerey:
Started off fixing an edge case where a `.yml` file that actually _starts_ with some ERB would raise an exception instead of failing gracefully with a nice message. Ended up shaving a small yak upon noticing that we displayed errors about invalid templates several different ways.

## Before
`test_deploy_fails_if_required_binding_not_present`
image

`test_missing_nested_partial_prints_helpful_error`
image

`test_invalid_yaml_in_partial_prints_helpful_error`
image

`test_invalid_yaml_fails_fast`
image

`test_multiple_invalid_k8s_specs_fails_on_apply_and_prints_template`
image

## After

`test_deploy_fails_if_required_binding_not_present`
image

`test_missing_nested_partial_correctly_identifies_invalid_template_and_its_parents` [improved test]
image

`test_invalid_yaml_in_partial_prints_helpful_error`
image

`test_invalid_yaml_fails_fast`
image

`test_multiple_invalid_k8s_specs_fails_on_apply_and_prints_template`
image

### DIFF
--- a/lib/kubernetes-deploy/deploy_task.rb
+++ b/lib/kubernetes-deploy/deploy_task.rb
@@ -187,15 +187,6 @@ module KubernetesDeploy
 
     private
 
-    # Inspect the file referenced in the kubectl stderr
-    # to make it easier for developer to understand what's going on
-    def find_bad_files_from_kubectl_output(stderr)
-      # stderr often contains one or more lines like the following, from which we can extract the file path(s):
-      # Error from server (TypeOfError): error when creating "/path/to/service-gqq5oh.yml": Service "web" is invalid:
-      matches = stderr.scan(%r{"(/\S+\.ya?ml\S*)"})
-      matches&.flatten
-    end
-
     def deploy_has_priority_resources?(resources)
       resources.any? { |r| PREDEPLOY_SEQUENCE.include?(r.type) }
     end
@@ -225,7 +216,8 @@ module KubernetesDeploy
       return unless failed_resources.present?
 
       failed_resources.each do |r|
-        record_invalid_template(r.validation_error_msg, file_paths: [r.file_path])
+        content = File.read(r.file_path) if File.file?(r.file_path)
+        record_invalid_template(err: r.validation_error_msg, filename: File.basename(r.file_path), content: content)
       end
       raise FatalDeploymentError, "Template validation failed"
     end
@@ -244,10 +236,6 @@ module KubernetesDeploy
         end
       end
       resources
-    rescue InvalidTemplateError => e
-      debug_msg = paragraph_for_invalid_templates(e.message, filenames: [e.filename], content: e.content)
-      @logger.summary.add_paragraph(debug_msg)
-      raise FatalDeploymentError, "Failed to render and parse template"
     end
 
     def split_templates(filename)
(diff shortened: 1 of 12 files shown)


test/integration/kubernetes_deploy_test.rb
```diff
@@ -865,4 +879,8 @@ def test_cronjobs_can_be_deployed
     cronjobs = FixtureSetAssertions::CronJobs.new(@namespace)
     cronjobs.assert_cronjob_present("my-cronjob")
   end
+
+  def test_friendly_error_on_misidentified_erb_file
```
KnVerey: This is the new one:

image

.rubocop.yml
```diff
@@ -3,3 +3,9 @@ inherit_from:
 
 AllCops:
   TargetRubyVersion: 2.3
+
+Style/TrailingCommaInArrayLiteral:
```
KnVerey: I'm not opposed to adding these (seems to be new in the Shopify style guide), but doing so now would cause rebase hell for the several people with huge PRs in flight.

lib/kubernetes-deploy/deploy_task.rb
```diff
@@ -225,24 +225,25 @@ def discover_resources
         end
       end
       resources
+    rescue InvalidTemplateError => e
+      debug_msg = paragraph_for_invalid_templates(e.message, filenames: [e.filename], content: e.content)
+      @logger.summary.add_paragraph(debug_msg)
```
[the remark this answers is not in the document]
dturn: Why log here instead of `split_templates`

lib/kubernetes-deploy/renderer.rb
```diff
@@ -60,12 +60,14 @@ def render_partial(partial, locals)
       # Note that JSON is a subset of YAML.
       JSON.generate(docs.children.first.to_ruby)
     rescue PartialNotFound => err
-      raise InvalidPartialError, err.message
+      # get the filename from the first parent, not the missing partial itself
+      raise err if err.filename == partial
+      raise InvalidPartialError.new(err.message, filename: partial, content: expanded_template || template)
```
klautcomputing: Shouldn't the underlying code raise the correct error type? Instead of rescuing one type and raising either one or the other?

lib/kubernetes-deploy/deploy_task.rb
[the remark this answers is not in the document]
KnVerey: > Why log here instead of split_templates

I think I did that because one of the three places that can raise this is the `rescue` of `split_templates`, so the alternative is:

```ruby
    def split_templates(filename)
      ...
    rescue Psych::SyntaxError => e
      log_invalid_templates(e.message, filenames: [filename], content: rendered_content)
      raise FatalDeploymentError, "Failed to render and parse template"
    rescue InvalidTemplateError => e
      log_invalid_templates(e.message, filenames: [e.filename], content: e.content)
      raise FatalDeploymentError, "Failed to render and parse template"
    end
```

It's a bit of duplication, but I don't feel strongly about it. Do you think that would be easier to follow?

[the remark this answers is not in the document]
dturn: I think moving it to split_templates is easer to read / understand.

lib/kubernetes-deploy/renderer.rb
KnVerey: This is transforming the data, not just the error type. `rescue PartialNotFound` needs to be distinct because it is the _first parent_ (potentially of a long chain) that has the relevant filename and template content. So we hit this case in the missing child (in which case we skip it, L64) and the _first_ parent (in which case we grab the filename and content, L65), and then after that we just add the additional parent name to the list (L67), like we would for any other type of error.

lib/kubernetes-deploy/deploy_task.rb
```diff
@@ -405,6 +379,41 @@ def log_pruning(kubectl_output)
       @logger.summary.add_action("pruned #{pruned.length} #{'resource'.pluralize(pruned.length)}")
     end
 
+    def record_apply_failure(err)
+      warn_msg = "WARNING: Any resources not mentioned in the error(s) below were likely created/updated. " \
+        "You may wish to roll back this deploy."
+      @logger.summary.add_paragraph(ColorizedString.new(warn_msg).yellow)
+
+      unidentified_errors = []
```
KnVerey: This might be a bad idea. If there were a multi-line error message where one of the middle lines has a filename in it, we'd be splitting it up strangely. I haven't seen that, but I certainly haven't seen everything. An alternative is to always dump the complete error before the individual cases we were able to extract and enhance.

lib/kubernetes-deploy/deploy_task.rb
```diff
@@ -405,6 +379,41 @@ def log_pruning(kubectl_output)
...
+    # Inspect the file referenced in the kubectl stderr
+    # to make it easier for developer to understand what's going on
+    def find_bad_files_from_kubectl_output(line)
+      # stderr often contains one or more lines like the following, from which we can extract the file path(s):
+      # Error from server (TypeOfError): error when creating "/path/to/service-gqq5oh.yml": Service "web" is invalid:
+
+      line.scan(%r{"(/\S+\.ya?ml\S*)"}).each_with_object([]) do |matches, bad_files|
+        matches.each do |path|
+          content = File.read(path) if File.file?(path)
+          bad_files << { filename: File.basename(path), err: line, content: content }
```
KnVerey: This lets us split up the various errors we've identified appropriately.

Example before this commit:

image


Example after:

image

dturn:
LGTM, I'm not worried about the edge case with multiline errors and filenames.

PR #257, opened 2018-04-02
Directory based kubectl apply when there are more than 1k resources

dturn:
Deploys with thousands of resources can fail with `Argument list too long - kubectl (Errno::E2BIG). This happens because we render each resource to its own file and then pass all of the files to kubectl apply via the `-f` flag. This pr detects when there are greater than 1k resources copies them to a temp directory and passes only the directory to kubectl apply.

cc: @dwradcliffe

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,10 @@
 deploy fails due to one or more resources failing to deploy in time.
 ([#244](https://github.com/Shopify/kubernetes-deploy/pull/244))
 
+*Bug Fixes*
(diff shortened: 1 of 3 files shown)


dturn:
Changes are in. Would like feedback on the ensure block vs wrapping almost the entire `apply_all` method in a block.

KnVerey:
> Would like feedback on the ensure block vs wrapping almost the entire apply_all method in a block.

I don't have a strong preference since the nesting doesn't get too crazy.

PR #258, opened 2018-04-02
Fix action summaries with failures/timeouts

KnVerey:
Fixes a bug introduced in #244 (which is on master but has not yet been released) that I noticed in while 🎩 a different PR. @dturn @klautcomputing we need to either get this in or revert #244 so we can unblock a release for #257 

## Problem

image

image

## Solution

Add a bunch more test coverage for the action line in existing tests, in addition to fixing the bug.

image

image

### DIFF
--- a/lib/kubernetes-deploy/deploy_task.rb
+++ b/lib/kubernetes-deploy/deploy_task.rb
@@ -158,7 +158,7 @@ module KubernetesDeploy
         failed_resources = resources.reject(&:deploy_succeeded?)
         success = failed_resources.empty?
         if !success && failed_resources.all?(&:deploy_timed_out?)
-          raise DeploymentTimeoutError, failed_resources
+          raise DeploymentTimeoutError
         end
         raise FatalDeploymentError unless success
       else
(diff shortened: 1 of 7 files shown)


lib/kubernetes-deploy/deploy_task.rb
```diff
@@ -172,13 +172,12 @@ def run!(verify_result: true, allow_protected_ns: false, prune: true)
...
       @logger.print_summary(:success)
-    rescue DeploymentTimeoutError => error
-      @logger.summary.add_action(error.message)
+    rescue DeploymentTimeoutError
       @logger.print_summary(:timed_out)
       ::StatsD.measure('all_resources.duration', StatsD.duration(start), tags: statsd_tags << "status:timeout")
       raise
     rescue FatalDeploymentError => error
-      @logger.summary.add_action(error.message) if error.message.present?
+      @logger.summary.add_action(error.message) if error.message != error.class.to_s
```
KnVerey: `error.message.present?` is always true because it defaults to the class name

This condition is a bit of a smell... we should probably make all error sources responsible for adding action snippets if appropriate. However, I tried briefly and it's not a trivial change, and I don't think it should block us releasing the timeout stuff.

PR #262, opened 2018-04-04
Fix bug where redeploy hangs forever because PDS condition is ignored

KnVerey:
There's a logic error in `deploy_failing_to_progress?`: it assumes that the deploy currently doing the watching also initiated the latest changes to the deployment. Because of this, no-op redeploys of bad code will hang forever, as seen in this shipit stack:

This PR proposes a new assumption: that if the status generation matches the resource generation, the conditions in that status must refer to the latest version. I'm not entirely sure this is a safe assumption to make. It could reintroduce something like https://github.com/Shopify/kubernetes-deploy/pull/213 (though that case was caused by a Kubernetes bug). Let me know if you have any better ideas. Another thing we could do is always return false if `Time.now - deploy_started_at` is less than PDS, without even checking the condition (like we do for <=1.7.7).

### DIFF
--- a/lib/kubernetes-deploy/kubernetes_resource/deployment.rb
+++ b/lib/kubernetes-deploy/kubernetes_resource/deployment.rb
@@ -96,6 +96,16 @@ module KubernetesDeploy
 
     private
 
+    def current_generation
+      return -2 unless exists? # different default than observed
+      @instance_data.dig('metadata', 'generation')
+    end
+
+    def observed_generation
+      return -1 unless exists? # different default than current
+      @instance_data.dig('status', 'observedGeneration')
+    end
+
     def desired_replicas
       return -1 unless exists?
       @instance_data["spec"]["replicas"].to_i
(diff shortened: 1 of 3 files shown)


KnVerey:
From [kubernetes community docs](https://github.com/kubernetes/community/blob/be9eeca6ee3becfa5b4c96bedf62b5b3ff5b1f8d/contributors/devel/controllers.md) about writing controllers:

> If the primary resource your controller is reconciling supports ObservedGeneration in its status, make sure you correctly set it to metadata.Generation whenever the values between the two fields mismatches.
>
> This lets clients know that the controller has processed a resource. Make sure that your controller is the main controller that is responsible for that resource, otherwise if you need to communicate observation via your own controller, you will need to create a different kind of ObservedGeneration in the Status of the resource.

And from the [community api conventions doc](https://github.com/kubernetes/community/blob/e5aeecb27cd2475007b4e81960604f26cd35cd48/contributors/devel/api-conventions.md):

> Some resources report the observedGeneration, which is the generation most recently observed by the component responsible for acting upon changes to the desired state of the resource. This can be used, for instance, to ensure that the reported status reflects the most recent desired status.

KnVerey:
Do you mean in the original condition, i.e. `Time.parse(@progress_condition["lastUpdateTime"]).to_i >= (@latest_rs.created_at - 5.seconds).to_i`? The danger there would be that we'd never be able to time out when then the deployment controller is failing to create a new replicaset in the first place (which is something we've actually seen in production due to a k8s bugs with `type: Recreate` deployments).

KnVerey:
@mkobetic can you weigh in on the logic here? i.e. what do you think about having the deploy make this assumption:

> This PR proposes a new assumption: that if the status generation matches the resource generation, the conditions in that status must refer to the latest version.

KnVerey:
Chatted with Martin IRL and he agrees that the generation-based assumption is good. I've documented it in an inline comment. Based on his feedback, I've also removed the check I had added on the reason text. The current list of possible reasons is [here](https://github.com/kubernetes/kubernetes/blob/827b3d77cde4a5d28bea5b2adece154d198ef85d/pkg/controller/deployment/util/deployment_util.go#L66-L89), and we'd want to time out on all of the negative ones.

PR #269, opened 2018-04-05
Adds file parsing to --bindings option

pallan:
On occasion the config for a system could be more complex than just a few variables. In those cases it may be feasible to use a separate JSON formatted configuration file that you can supply to the `--bindings` argument.

This uses a similar convention as cURL where you can specify you want to load a file by prefixing the argument value with an `@`. For example, 
```
kubernetes-deploy my-app cluster1 --bindings='@config/production.json'
```
would load and parse the production.json file in the same manner as a JSON string. An `ArgumentError` would be raised if the `@` prefix is present but the file does not exist or cannot be read.

Documentation, README and CLI help have been updated accordingly to reflect the change.


exe/kubernetes-deploy
```diff
@@ -13,7 +13,7 @@ verbose_log_prefix = false
 
 ARGV.options do |opts|
   opts.on("--bindings=BINDINGS", "Expose additional variables to ERB templates " \
-    "(format: k1=v1,k2=v2 or JSON)") do |binds|
+    "(format: k1=v1,k2=v2, JSON string or JSON file path)") do |binds|
```
fw42: maybe worth mentioning the `@` prefix here?

pallan: Added

lib/kubernetes-deploy/bindings_parser.rb
```diff
@@ -19,6 +19,14 @@ def parse(string)
     private
 
     def parse_json(string)
+      if string =~ /\A@/
```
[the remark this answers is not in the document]
pallan: No reason not to support YAML. I will update this PR to validate the extension and create a separate PR for YAML integration.  I will raise the same argument error if the extension is invalid.

pallan:
Updated to support YAML with some file extension verification.

pallan:
Rebased to master and updated the changelog

Issue #272, opened 2018-04-06
Timing error in watcher's sleep


KnVerey:
Seen in version 0.18.1, but the lines of code in question hadn't changed for 7 months as of that version, so it's likely this is a rare edge case that has been possible for  a long time.

```
lib/kubernetes-deploy/resource_watcher.rb:23:in `sleep': time interval must be positive (ArgumentError)
  from lib/kubernetes-deploy/resource_watcher.rb:23:in `run'
  from lib/kubernetes-deploy/deploy_task.rb:380:in `deploy_resources'
  from lib/kubernetes-deploy/deploy_task.rb:155:in `run!'
  from exe/kubernetes-deploy:67:in `<top (required)>'

```

dturn:
Code in question is:
`        if Time.now.utc < delay_sync_until
          sleep(delay_sync_until - Time.now.utc)
        end
`

Easy fix ideas are saving the first call to Time.now.utc or `[delay_sync_until, 0.1].max` in the sleep call

KnVerey:
I'm fine with either of those fixes. Interested in making the PR?

PR #273, opened 2018-04-06
[bug fix] Make sure we sleep for positive duration

dturn:
Due to calling `Time.now.utc` twice, it was possible for sleep to receive a negative value. Fix that by calling `Time.now.utc` once and ensure the sleep duration is `> 0`.

Fixes: https://github.com/Shopify/kubernetes-deploy/issues/272#issuecomment-379361407

Attempted to improve: https://github.com/Shopify/kubernetes-deploy/issues/274

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,8 @@
+### Master
+
+*Bug Fixes*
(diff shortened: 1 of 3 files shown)


klautcomputing:
I got the tests to pass but one of the test was very flakey see #274

Issue #274, opened 2018-04-06
New resource watcher test is flakey


klautcomputing:
### Bug report
We observed that the test `test_resource_watcher_reports_failed_after_timeout` introduced in #73 is very flakey.

dturn:
Bumped `progressDeadlineSeconds` for the deployment we want to hit the global timeout, it was hitting progressDeadlineSeconds deadline about 50% of the time:  https://github.com/Shopify/kubernetes-deploy/pull/273/commits/1817314ccca384ebc3ce7456d31c116713a7f355

Leaving this open for a bit to make sure that was the only flakey part of the test.

PR #275, opened 2018-04-11
Bump minikube to k8s v1.9.4, and fix tests + deprecation

kvs:
This ensures tests are run against a current version of Kubernetes, and fixes a test that was failing due to updated resource-names after an API-versioning change


kvs:
The hyperkit change could potentially be viewed as contentious, but minikube loudly complains that xhyve is deprecated, hence switching to that.

Tests still flake out from time to time, and I can't get it to hit green on all three versions, but this doesn't seem to have been introduced here :-/

dturn:
Flakey test is known and I'm working on it: https://github.com/Shopify/kubernetes-deploy/issues/274

kvs:
@dturn honestly, I have no idea. Both 1.9.4 and 1.9.6 start to bootstrap on my machine, but then fail for network-related reasons that I haven't solved yet. CI is using `1.9-latest`, but I don't know how it resolves that to 1.9.6 (which the logs indicate it does)

test/test_helper.rb
```diff
@@ -24,7 +24,7 @@
 require 'minitest/autorun'
 require 'minitest/stub/const'
 require 'webmock/minitest'
-require 'mocha/mini_test'
+require 'mocha/minitest'
```
dturn: Please bump https://github.com/Shopify/kubernetes-deploy/blob/master/kubernetes-deploy.gemspec#L39 to 1.5.0. This blows up pre that version https://github.com/freerange/mocha/commit/55d66ed4f5e6cad95e1605341ec3c608be2c6833

dev.yml
```diff
@@ -8,7 +8,7 @@ up:
   - custom:
       name: Minikube Cluster
       met?: test $(minikube status | grep Running | wc -l) -eq 2 && $(minikube status | grep -q 'Correctly Configured')
-      meet: minikube start --vm-driver=xhyve --kubernetes-version=v1.7.5
+      meet: minikube start --kubernetes-version=v1.9.6 --vm-driver=hyperkit
```
dturn: How would you feel about 1.9.4? Its the most recent version supported by minikube v0.25.2 see: `minikube get-k8s-versions`.  When using minikube v0.26.0 the most recent version I hit https://github.com/kubernetes/kubeadm/issues/733

PR #276, opened 2018-04-11
Bump global timeout to fix flakey test

dturn:
test_resource_watcher_reports_failed_after_timeout fails about a third of the time because one of the resources timesouts via the global timeout instead of just failing. Bumping the global timeout to try and make this test more reliable. 

closes: https://github.com/Shopify/kubernetes-deploy/issues/274


dturn:
Tests all passed the first time. I hit the re-build button and they all passed the second time as well. Feeling good about this fix.

PR #278, opened 2018-04-13
Add namespace labels to statsd tags

Fattouche:
We need this so that we can can pass custom statsd tags through the namespace. The test isn't great because half of the statsd metrics that are being measured don't have the tag from the namespace. A potential solution would be to add these same tags to the `kubernetes_resource.rb` however I am not sure if we need them in that location. What are your thoughts on this @stefanmb.

Edit: After looking further at what we have on datadog, we need the tags on `kubernetes_resource.rb` as well, I will update the test to check for the tag on all metrics.

Another thing I have noticed is that in some cases the metric from `kubernetes_resource.rb` will contain `sha:` with no value. I am not sure how this happens because it uses `ENV[Revision]` just like the `deploy_task.rb` does. Does anyone have any insight into this?

### DIFF
--- a/test/integration/kubernetes_deploy_test.rb
+++ b/test/integration/kubernetes_deploy_test.rb
@@ -986,6 +986,4 @@ unknown field \"myKey\" in io.k8s.apimachinery.pkg.apis.meta.v1.ObjectMeta",
       assert_empty desired_tags - metric.tags
     end
   end
-
-
 end


test/integration/kubernetes_deploy_test.rb
```diff
@@ -927,4 +927,23 @@ def test_resource_watcher_raises_after_timeout_seconds
...
+    desired_tags = %W(context:#{KubeclientHelper::MINIKUBE_CONTEXT} namespace:#{@namespace} statsd_tag:test)
+    hello_cloud = FixtureSetAssertions::HelloCloud.new(@namespace)
+    kubeclient.patch_namespace(hello_cloud.namespace, metadata: { labels: { statsd_tag: 'test' } })
+    metrics = capture_statsd_calls do
+      assert_deploy_success(deploy_fixtures("hello-cloud"))
+    end
+    metrics.each do |metric|
+      applicable = false
+      metric.tags.each do |metric_tag|
+        applicable = true if metric_tag.include?("test_adds_namespace_labels_to_statsd_tags")
```
Fattouche: Since the way statsd-instrument captures its metrics is just by mocking a backend, we can't ensure that all the metrics we grab are from this specific test because they are running in parallel. The hacky solution I have is to make sure that the tag includes our test name.

Source: https://github.com/Shopify/statsd-instrument/blob/46df2c67ec9a9384b9b667df13497a884d560278/lib/statsd/instrument/helpers.rb#L3

lib/kubernetes-deploy/deploy_task.rb
```diff
@@ -455,7 +456,8 @@ def confirm_cluster_reachable
     def confirm_namespace_exists
       st, err = nil
       with_retries(2) do
-        _, err, st = kubectl.run("get", "namespace", @namespace, use_namespace: false, log_failure: true)
+        @namespace_info, err, st = kubectl.run("get", "namespace", @namespace, "-o", "json", use_namespace: false,
```
dturn: Let's parse the JSON here

lib/kubernetes-deploy/kubernetes_resource.rb
```diff
@@ -356,7 +359,12 @@ def statsd_tags
       else
         "unknown"
       end
-      %W(context:#{context} namespace:#{namespace} resource:#{id} type:#{type} sha:#{ENV['REVISION']} status:#{status})
+      tags = %W(context:#{context} namespace:#{namespace} resource:#{id}
+                type:#{type} sha:#{ENV['REVISION']} status:#{status})
+      @extra_statsd_tags&.each do |tag|
```
dturn: I think you can get the same behavior via `tags | (@extra_statsd_tags || [])`

lib/kubernetes-deploy/deploy_task.rb
```diff
@@ -239,7 +240,7 @@ def discover_resources
 
         split_templates(filename) do |r_def|
           r = KubernetesResource.build(namespace: @namespace, context: @context, logger: @logger,
-                                       definition: r_def, extra_statsd_tags: statsd_tags)
+                                       definition: r_def, statsd_tags: statsd_tags)
```
Fattouche: We might as well only pass in the `@namespace_tags` if we have them, this avoids any potential overwrite in the `KubernetesResource` class.

lib/kubernetes-deploy/deploy_task.rb
```diff
@@ -456,26 +457,35 @@ def confirm_cluster_reachable
...
       end
       raise FatalDeploymentError, "Failed to find namespace. #{err}" unless st.success?
       @logger.info("Namespace #{@namespace} found")
     end
 
+    def tags_from_namespace_labels
+      namespace_info = nil
+      with_retries(2) do
+        namespace_info, err, st = kubectl.run("get", "namespace", @namespace, "-o", "json", use_namespace: false,
+          log_failure: true)
```
Fattouche: What is the point of duplicating the kubectl stuff?

stefanmb: `confirm_namespace_exists` should not have unexpected side effects like getting the tags, let's separate each distinct bit of functionality into its own method.

lib/kubernetes-deploy/deploy_task.rb
stefanmb: Not sure I follow, how would you change it?

stefanmb: Addressed IRL, thanks!

stefanmb:
@dturn Can we get another look? Thank you!

test/integration/kubernetes_deploy_test.rb
```diff
@@ -927,4 +927,22 @@ def test_resource_watcher_raises_after_timeout_seconds
...
+      assert_deploy_success(deploy_fixtures("hello-cloud"))
+    end
+
+    # We can't ensure that all the metrics we grab are from this specific test because they are running in parallel
+    metrics = metrics.select { |m| m.tags.include? "namespace:#{@namespace}" }
+    assert !metrics.empty?
+
+    metrics.each do |metric|
+      difference = desired_tags - metric.tags
+      assert difference.empty?
```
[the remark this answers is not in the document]
stefanmb: `metric.tags` is a superset, but the `assert_empty` works.

test/integration/kubernetes_deploy_test.rb
```diff
@@ -927,4 +927,22 @@ def test_resource_watcher_raises_after_timeout_seconds
...
+    desired_tags = %W(context:#{KubeclientHelper::MINIKUBE_CONTEXT} namespace:#{@namespace} foo:bar)
+    hello_cloud = FixtureSetAssertions::HelloCloud.new(@namespace)
+    kubeclient.patch_namespace(hello_cloud.namespace, metadata: { labels: { foo: 'bar' } })
+    metrics = capture_statsd_calls do
+      assert_deploy_success(deploy_fixtures("hello-cloud"))
+    end
+
+    # We can't ensure that all the metrics we grab are from this specific test because they are running in parallel
+    metrics = metrics.select { |m| m.tags.include? "namespace:#{@namespace}" }
+    assert !metrics.empty?
```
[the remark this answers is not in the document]
stefanmb: Done

test/integration/kubernetes_deploy_test.rb
```diff
@@ -927,4 +927,21 @@ def test_resource_watcher_raises_after_timeout_seconds
       "If you expected it to take longer than 5 seconds for your deploy to roll out, increase --max-watch-seconds."
     ], in_order: true)
   end
+
+  def test_adds_namespace_labels_to_statsd_tags
+    desired_tags = %W(context:#{KubeclientHelper::MINIKUBE_CONTEXT} namespace:#{@namespace} foo:bar)
+    hello_cloud = FixtureSetAssertions::HelloCloud.new(@namespace)
+    kubeclient.patch_namespace(hello_cloud.namespace, metadata: { labels: { foo: 'bar' } })
+    metrics = capture_statsd_calls do
+      assert_deploy_success(deploy_fixtures("hello-cloud"))
```
KnVerey: this fixture set is HUGE so don't deploy the whole thing unless you need to. You can select specific files from it using the `subset:` arg. If you don't need anything specific, I'd suggest `"configmap-data.yml"`, which will deploy instantly.

lib/kubernetes-deploy/kubernetes_resource.rb
```diff
@@ -356,7 +358,9 @@ def statsd_tags
       else
         "unknown"
       end
-      %W(context:#{context} namespace:#{namespace} resource:#{id} type:#{type} sha:#{ENV['REVISION']} status:#{status})
+      tags = %W(context:#{context} namespace:#{namespace} resource:#{id}
+                type:#{type} sha:#{ENV['REVISION']} status:#{status})
+      tags | @optional_statsd_tags
```
KnVerey: TIL about  `Array#|` ✨

lib/kubernetes-deploy/deploy_task.rb
```diff
@@ -462,12 +465,28 @@ def confirm_namespace_exists
...
+    def tags_from_namespace_labels
+      namespace_info = nil
+      with_retries(2) do
+        namespace_info, _, st = kubectl.run("get", "namespace", @namespace, "-o", "json", use_namespace: false,
+          log_failure: true)
+        st.success?
+      end
+      tags = []
+      return tags if namespace_info.blank?
+      namespace_labels = JSON.parse(namespace_info, symbolize_names: true).dig(:metadata, :labels)
```
KnVerey: TIL about the `symbolize_names` option. Two new tricks in one small PR, thanks! ✨

lib/kubernetes-deploy/deploy_task.rb
```diff
@@ -462,12 +465,28 @@ def confirm_namespace_exists
...
     end
 
+    def tags_from_namespace_labels
+      namespace_info = nil
+      with_retries(2) do
+        namespace_info, _, st = kubectl.run("get", "namespace", @namespace, "-o", "json", use_namespace: false,
+          log_failure: true)
+        st.success?
+      end
+      tags = []
```
KnVerey: I think that this line and following could be more concisely expressed as:
```ruby
  return [] if namespace_info.blank?
  namespace_labels = JSON.parse(namespace_info, symbolize_names: true).fetch(:metadata, {}).fetch(:labels, {})
  namespace_labels.map { |key, value| "#{key}:#{value}" }
```
Just a suggestion. I don't feel strongly about it.

test/integration/kubernetes_deploy_test.rb
```diff
@@ -927,4 +927,21 @@ def test_resource_watcher_raises_after_timeout_seconds
...
+    end
+
+    # We can't ensure that all the metrics we grab are from this specific test because they are running in parallel
+    metrics = metrics.select { |m| m.tags.include? "namespace:#{@namespace}" }
+    refute_empty metrics
```
KnVerey: Could we assert the precise number of matches we expect?

test/integration/kubernetes_deploy_test.rb
Fattouche: Yea we just want to check that it is grabbing the labels from the namespace, we don't care about what it is deploying. Thanks :)

PR #282, opened 2018-04-17
Add an annotation that skips rollout verification

dturn:
Allow a resource to skip rollout verification via a new annotation `'kubernetes-deploy.shopify.io/no-rollout-verification'`

This prevents the deploy from being a failure if the resource fails to deploy, by causing `ResourceWatcher` to ignore the resource though it does print a  message saying its ignoring the resource.

Feature wanted from: https://github.com/Shopify/kubernetes-deploy/issues/229

screen shot 2018-04-18 at 10 17 50 am

### DIFF
--- a/test/integration/restart_task_test.rb
+++ b/test/integration/restart_task_test.rb
@@ -154,7 +154,7 @@ class RestartTaskTest < KubernetesDeploy::IntegrationTest
   end
 
   def test_restart_failure
-    success = deploy_fixtures("hello-cloud", subset: ["configmap-data.yml", "web.yml.erb"]) do |fixtures|
+    result = deploy_fixtures("hello-cloud", subset: ["configmap-data.yml", "web.yml.erb"]) do |fixtures|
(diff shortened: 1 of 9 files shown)


test/integration/restart_task_test.rb
```diff
@@ -190,6 +190,32 @@ def test_restart_failure
       in_order: true)
   end
 
+  def test_restart_failure_can_be_ignored
+    success = deploy_fixtures("hello-cloud", subset: ["configmap-data.yml", "web.yml.erb"]) do |fixtures|
```
klautcomputing: This is my ultimate nitpick: `success` is what we like, but so far we only know it's a `result`

dturn:
@KnVerey No rush on this, but want to make sure its on your radar

dturn:
We're closing this pr unmerged because it doesn't seem important enough to justify making the state tracking code more cluttered. We have an issue to improve this code, but its not high priority. If this feature becomes high priority we can either bring this back as is or fix up the state tracking and then build this the right way.

PR #283, opened 2018-04-17
Only runs the right tests when TEST is set

dturn:
Setting the `TEST=foo` on the command line overrides `test_files`.  Currently `TEST=test/integration/kubernetes_deploy_test.rb be rake test` causes the file to be run three times. This PR fixes things up so that its only run once.

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -2,28 +2,42 @@
 require "bundler/gem_tasks"
 require "rake/testtask"
 
+files = {
+  integration_test: FileList['test/integration/**/*_test.rb'],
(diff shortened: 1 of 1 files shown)


dturn:
Closing since it looks like the real issue is I don't know how to minitest very well.

kirs:
Raw Minitest is not very user friendly 😭
One thing that we can do here is add the [m](https://github.com/qrush/m) gem, generate a binstub (`bundle binstub m`), and you'll be able to use `bin/m test/whatever_test.rb` as well as `bin/m test/whatever_test.rb:100` run the exact test line.

PR #284, opened 2018-04-19
Bump kubeclient to 3.0

dturn:
Bumping kubeclient to 3.0. 

Two major changes  (https://github.com/abonas/kubeclient/blob/master/CHANGELOG.md):
- Entity classes (`Kubeclient::Namespace`) have been dropped  everything is now a `Kubeclient::Resource` with a kind. This only impacted tests.
- patch_xxx now returns a `Kubeclient::Resource`. Failure to update the resource will raise an error.

Minor Change:
- Was able to remove the dependecy on `"rest-client"` since the min version was fixed in kubeclient

### DIFF
--- a/kubernetes-deploy.gemspec
+++ b/kubernetes-deploy.gemspec
@@ -24,7 +24,7 @@ Gem::Specification.new do |spec|
 
   spec.required_ruby_version = '>= 2.3.0'
   spec.add_dependency "activesupport", ">= 5.0"
-  spec.add_dependency "kubeclient", "~> 2.4"
+  spec.add_dependency "kubeclient", "~> 3.0"
   spec.add_dependency "rest-client", ">= 1.7" # Minimum required by kubeclient. Remove when kubeclient releases v3.0.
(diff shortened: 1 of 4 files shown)


kubernetes-deploy.gemspec
```diff
@@ -24,7 +24,7 @@ Gem::Specification.new do |spec|
 
   spec.required_ruby_version = '>= 2.3.0'
   spec.add_dependency "activesupport", ">= 5.0"
-  spec.add_dependency "kubeclient", "~> 2.4"
+  spec.add_dependency "kubeclient", "~> 3.0"
   spec.add_dependency "rest-client", ">= 1.7" # Minimum required by kubeclient. Remove when kubeclient releases v3.0.
```
timothysmith0609: Per the comment, is this needed anymore?

dturn: Nope fixed upstream https://github.com/abonas/kubeclient/blob/v3.0.0/kubeclient.gemspec#L30

Issue #285, opened 2018-04-19
Mutually exclusive / consistent reporting


dturn:
`_succeeded?`, `_failed?`, `_timed_out?` and (effectively) [`_ignored?` ](https://github.com/Shopify/kubernetes-deploy/pull/282) should all be mutually exclusive, but there's nothing forcing them to be, and if they aren't in practice, what we'll show depends on what the caller decided to check first. 

We should improve this situation since its likely to lead to reporting bugs / our reporting code is already getting complicated.

1 idea is to carefully define states and have the `sync` method move us through them..

cc: @KnVerey

PR #286, opened 2018-04-19
[bug fix] A failed deployment hasn't timed out

dturn:
If both `deploy_timed_out?` and `deploy_failed?` return true the result is `Result: TIMED OUT` not `Result: FAILURE` . This can only happen for deployments since other resources have deploy_timed_out? dependent on !deploy_failed? . 

This is a short term patch. The long term solution is https://github.com/Shopify/kubernetes-deploy/issues/285

Creating a test to show this behavior was easy:
```
    KubernetesDeploy::Deployment.any_instance.stubs(:deploy_timed_out?).returns(true)
    KubernetesDeploy::Deployment.any_instance.stubs(:deploy_failed?).returns(true)
    result = deploy_fixtures("invalid", subset: ["cannot_run.yml"])
   ...
```

However, the fix is in `deploy_timed_out?` so the test doesn't go green with the fix in place.

### DIFF
--- a/lib/kubernetes-deploy/kubernetes_resource/deployment.rb
+++ b/lib/kubernetes-deploy/kubernetes_resource/deployment.rb
@@ -74,6 +74,7 @@ module KubernetesDeploy
     end
 
     def deploy_timed_out?
+      return false if deploy_failed?
       # Do not use the hard timeout if progress deadline is set
       progress_condition.present? ? deploy_failing_to_progress? : super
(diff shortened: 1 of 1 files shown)


KnVerey:
We could probably make a unit test on deployment that replicates the situation (deployment fixture has pds exceeded condition, selects a pod fixture with an error), but I wouldn't insist on it in this case.

PR #287, opened 2018-05-04
Allow for multiple `--bindings` arguments.

kvs:
We developed a need to pass in multiple files with bindings in it. This allows for that usecase.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
 ([#280](https://github.com/Shopify/kubernetes-deploy/pull/280))
 
 *Enhancements*
+- Merge multiple `--bindings` arguments, to allow a composite bindings map (multiple arguments or files)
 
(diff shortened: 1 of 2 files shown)


dturn:
Simple enough, can you please add a test

kvs:
Hmm, I don't think there are any tests that currently run `exe/kubernetes-deploy`. I'll see if I can find time to figure out how to best test that part of the code, or if we'll find a workaround :)

dturn:
I think for now adding something to `test/unit/kubernetes-deploy/bindings_parser_test.rb ` would be good enough. Testing the exe's is on the roadmap, but requires a refactor.

kvs:
There's nothing to test in that file, since BindingsParser never gets to see the multiple bindings - it all happens in the exe.

kvs:
Cool, thanks @KnVerey

Issue #230, opened 2017-12-20
Improve pod eviction handling


KnVerey:
The following happened after 8.7s:

![image](https://user-images.githubusercontent.com/4789493/34219285-acaa9552-e57e-11e7-9237-6d42241a2dba.png)

That is a single-replica deployment with `spec.strategy.type=Recreate`. The deploy selected the correct replicaset, and it appears that its pods are being evicted immediately on creation because they were being repeatedly scheduled on a node that is under disk pressure. Evictions should be recoverable, so this result should probably have been a timeout that showed the evictions.

KnVerey:
I just saw this on a more typical deployment:

```
Deployment/web: FAILED
Latest ReplicaSet: web-6c4c87894

Pod status: Failed (Reason: Evicted). 
- Final status: 20 replicas, 20 availableReplicas, 5 unavailableReplicas
```

The deployment was also hitting resource quotas, which could explain why it could fail on a single eviction despite having scaled the new RS to 5 (i.e. maybe only one of the five was successfully created).

KnVerey:
This can affect DaemonSets too, and even there the replacement pod usually comes up fine seconds later. Fundamentally, evictions are a temporary issue and we shouldn't consider them failures at the pod level.

PR #293, opened 2018-05-21
Deploys are failed if pod is evicted

dturn:
Evictions should be recoverable so don't let them cause a failure.

Fixes: https://github.com/Shopify/kubernetes-deploy/issues/230

### DIFF
--- a/test/unit/kubernetes-deploy/kubernetes_resource/pod_test.rb
+++ b/test/unit/kubernetes-deploy/kubernetes_resource/pod_test.rb
@@ -155,6 +155,21 @@ class PodTest < KubernetesDeploy::TestCase
     assert_equal expected_msg, pod.failure_message
   end
 
+  def test_deploy_failed_is_false_for_evicted_
+    container_state = {
+      "state" => {
+        "waiting" => {
+          "message" => "Filler",
+          "reason" => "Evicted"
+        }
+      }
+    }
+    pod = build_synced_pod(build_pod_template(container_state: container_state))
+
+    refute pod.deploy_failed?
+    assert_nil pod.failure_message
+  end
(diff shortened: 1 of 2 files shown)


test/unit/kubernetes-deploy/kubernetes_resource/pod_test.rb
```diff
@@ -155,6 +155,21 @@ def test_deploy_failed_is_true_for_container_cannot_run_error
     assert_equal expected_msg, pod.failure_message
   end
 
+  def test_deploy_failed_is_false_for_evicted_
```
KnVerey: nit: you've got a trailing underscore (maybe you meant to add "pods"?)

test/unit/kubernetes-deploy/kubernetes_resource/pod_test.rb
```diff
@@ -155,6 +155,21 @@ def test_deploy_failed_is_true_for_container_cannot_run_error
...
+    container_state = {
```
KnVerey: Did this test really fail without your change? This is setting container status info, but eviction data (as your business logic correctly assumes) is at the top level of the pod status, like this:

```
status:
  message: 'The node was low on resource: nodefs.'
  phase: Failed
  reason: Evicted
  startTime: 2018-05-04T21:17:08Z
```

dturn:
> Did this test really fail without your change? 

It didn't, probably because I guessed at what the status looks like for an evicted pod. I've updated the test based on `kubectl --context tier4 get pod  web-5fb98b959-qflv6  -n wedge-viewer-production  -o json   ` not sure if its realistic enough, but at least the test fails without my change.

PR #296, opened 2018-06-05
Add Job Resource

dturn:
Add minimal Job Resource and a test to make sure it deploys.


dturn:
Added new success and failure conditions, ready for 👀 again

test/integration/kubernetes_deploy_test.rb
```diff
@@ -917,6 +919,19 @@ def test_cronjobs_can_be_deployed
     cronjobs.assert_cronjob_present("my-cronjob")
   end
 
+  def test_jobs_can_be_successful
+    assert_deploy_success(deploy_fixtures("jobs"))
```
KnVerey: Why not deploy the job subset of the hello-cloud fixture set instead of having a second fixture set?

test/integration/kubernetes_deploy_test.rb
```diff
@@ -21,7 +21,8 @@ def test_full_hello_cloud_set_deploy_succeeds
       %r{Service/web\s+Selects at least 1 pod},
       %r{DaemonSet/ds-app\s+1 updatedNumberScheduled, 1 desiredNumberScheduled, 1 numberReady},
       %r{StatefulSet/stateful-busybox},
-      %r{Service/redis-external\s+Doesn't require any endpoint}
+      %r{Service/redis-external\s+Doesn't require any endpoint},
+      %r{Job/hello-job\s+Exists}
```
KnVerey: Can we be more helpful about this? Our success condition is more complex than this implies. For example, we might say "Succeeded" or "Started" as applicable.

test/integration/kubernetes_deploy_test.rb
```diff
@@ -917,6 +919,19 @@ def test_cronjobs_can_be_deployed
...
+    assert_deploy_success(deploy_fixtures("jobs"))
+  end
+
+  def test_jobs_can_fail
+    fixtures = deploy_fixtures("jobs") do |f|
+      spec = f["job.yml"]["Job"].first["spec"]["template"]["spec"]
+      spec["backoffLimit"] = 1
+      spec["containers"].first["args"] = %w(FAKE)
+    end
+    assert_deploy_failure(fixtures)
```
KnVerey: What does this look like? What status are we printing? Is the failure information we're printing as helpful as it could be?

test/integration/kubernetes_deploy_test.rb
```diff
@@ -917,6 +919,19 @@ def test_cronjobs_can_be_deployed
     cronjobs.assert_cronjob_present("my-cronjob")
   end
 
+  def test_jobs_can_be_successful
+    assert_deploy_success(deploy_fixtures("jobs"))
+  end
+
+  def test_jobs_can_fail
+    fixtures = deploy_fixtures("jobs") do |f|
```
KnVerey: - Same comment about using the fixture from the hello-cloud set instead. 
- Can you run this a whole bunch of times in a row to confirm it isn't flakey? I'm worried that the deploy will race that first pod crash. If that's the case, we could deploy it with something that is guaranteed to take longer (e.g. a deployment that hits an annotation-based timeout).

lib/kubernetes-deploy/kubernetes_resource/job.rb
```diff
@@ -0,0 +1,25 @@
...
+    end
+
+    def deploy_failed?
+      return false unless deploy_started?
+      return false unless @instance_data.dig("spec", "backoffLimit").present?
+      (@instance_data.dig("status", "failed") || 0) >= @instance_data.dig("spec", "backoffLimit")
+    end
+
+    def timeout_message
+      UNUSUAL_FAILURE_MESSAGE
```
KnVerey: This isn't an unusual failure. Please also adjust `status` as appropriate ("exists" isn't sufficiently descriptive here).

lib/kubernetes-deploy/kubernetes_resource/job.rb
```diff
@@ -0,0 +1,25 @@
+# frozen_string_literal: true
+module KubernetesDeploy
+  class Job < KubernetesResource
+    TIMEOUT = 30.seconds
```
KnVerey: Is this generous enough? What if someone's pod takes several minutes to start? I'd recommend setting this closer to the timeouts on other pod managers.

lib/kubernetes-deploy/kubernetes_resource/job.rb
```diff
@@ -0,0 +1,47 @@
...
+    private
+
+    def done?
+      (@instance_data.dig("status", "succeeded") || 0) == @instance_data.dig("spec", "completions")
+    end
+
+    def running?
+      now = Time.now.utc
+      start_time = @instance_data.dig("status", "startTime")
+      # Wait 5 seconds to ensure job doesn't immediately fail.
```
KnVerey: I'm not sure I understand this; `deploy_failed?` doesn't look at `running?`, so afaict the effect of this is to potentially delay declaring success.

test/integration/kubernetes_deploy_test.rb
```diff
@@ -917,6 +919,33 @@ def test_cronjobs_can_be_deployed
     cronjobs.assert_cronjob_present("my-cronjob")
   end
 
+  def test_jobs_can_be_successful
+    assert_deploy_success(deploy_fixtures("hello-cloud", subset: ["job.yml"]))
+    assert_logs_match_all([
+      "Deploying Job/hello-job (timeout: 600s)",
```
KnVerey: I'd suggest adding this assertion to `full_hello...` instead, since it's the only thing that test doesn't already cover. Generally speaking I'd consider that sufficient for asserting the basic success case.

lib/kubernetes-deploy/kubernetes_resource/job.rb
```diff
@@ -0,0 +1,47 @@
...
+
+    def deploy_failed?
+      return false unless deploy_started?
+      return false unless @instance_data.dig("spec", "backoffLimit").present?
+      (@instance_data.dig("status", "failed") || 0) >= @instance_data.dig("spec", "backoffLimit")
+    end
+
+    def status
+      if !exists?
+        super
```
KnVerey: This is a rather minor point, but it applies to all these resource PRs:  I noticed in 🎩 that `super` results in "Unknown", whereas some of our resources print "Not found" for the same thing, which is inconsistent. Do you have an opinion on which is better? Perhaps the solution is to change the superclass to "Not found" too.

dturn: From a quick grep it looks like only `Service` uses 'Not found'. I don't have strong feeling on which is better, but I suspect fixing service would be easier.

lib/kubernetes-deploy/kubernetes_resource/job.rb
dturn: This actually prevents declaring success before k8s has tried to run the command. `test_jobs_can_fail` which uses the fake command `/not/a/command` will actually report as running for a brief period (& therefore successful) if we don't introduce a slight delay.

KnVerey: Oh, I see. Requiring it to run for at least 5 seconds seems reasonable. I wonder if adding a readiness probe to the container would work too, but this will make our tool work better in more cases regardless.

lib/kubernetes-deploy/kubernetes_resource/job.rb
KnVerey: I have a slight preference for "Not found" because it more directly reflects what we actually did to reach this conclusion. Don't need to fix in this PR though.

dturn: Created an issue https://github.com/Shopify/kubernetes-deploy/issues/321

PR #297, opened 2018-06-05
Fix CI for 1.9

dturn:
Something has changed with CI 
screen shot 2018-06-05 at 12 46 45 pm

TLDR: test_failure_logs_from_unmanaged_pod_appear_in_summary_section no longer has logs available, so the test was failing, but only on 1.9 and only recently....

### DIFF
--- a/lib/kubernetes-deploy/kubeclient_builder/google_friendly_config.rb
+++ b/lib/kubernetes-deploy/kubeclient_builder/google_friendly_config.rb
@@ -13,7 +13,7 @@ module KubernetesDeploy
       end
 
       def self.read(filename)
-        new(YAML.safe_load(File.read(filename)), File.dirname(filename))
+        new(YAML.safe_load(File.read(filename), [Time]), File.dirname(filename))
       end
 
       def new_token
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -5,6 +5,7 @@
     env:
       LOGGING_LEVEL: 4
       KUBERNETES_VERSION: v1.9-latest
+      PARALLELISM: 1
   - name: 'Run Test Suite (:kubernetes: 1.8-latest)'
     command: bin/ci
     agents:
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -5,7 +5,6 @@
     env:
       LOGGING_LEVEL: 4
       KUBERNETES_VERSION: v1.9-latest
-      PARALLELISM: 1
   - name: 'Run Test Suite (:kubernetes: 1.8-latest)'
     command: bin/ci
     agents:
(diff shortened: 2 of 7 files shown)


lib/kubernetes-deploy/kubeclient_builder/google_friendly_config.rb
```diff
@@ -13,7 +13,7 @@ def fetch_user_auth_options(user)
       end
 
       def self.read(filename)
-        new(YAML.safe_load(File.read(filename)), File.dirname(filename))
+        new(YAML.safe_load(File.read(filename), [Time]), File.dirname(filename))
```
dturn: Not sure why I needed this now to run tests, didn't need to when I 🎩 the PR modifying this line. My kubeconfig has changed since then though.

KnVerey:
Are you sure this isn't a legit bug, i.e. our log-fetching request really isn't working with the latest 1.9 version for some reason? The name of the test suggests that its purpose is to assert the presence of those log lines, so I don't want to remove those assertions.

dturn:
Testing against 1.9.8 I can't replicate the behavior or CI. Added puts and the command to fetch the logs looks correct and doesn't fail:

screen shot 2018-06-05 at 4 57 46 pm

I agree this is 💩 given the purpose of the test, I'll keep digging

dturn:
Upgraded to minikube version 0.27.0 and I'm now able to reproduce. Not to figure it out.

dturn:
@KnVerey If I change the command in `test_failure_logs_from_unmanaged_pod_appear_in_summary_section` to `["sh", "-c", "/some/bad/path"]` we get logs. But that of course changes the `containerStatuses` from 
```
"state": {
                    "terminated": {
                        "containerID": "docker://b157da68133423007ef7247074ba9cb8f168c2372ba84530514da0e03500aa48",
                        "exitCode": 127,
                        "finishedAt": "2018-06-07T18:14:01Z",
                        "message": "OCI runtime create failed: container_linux.go:348: starting container process caused \"exec: \\\"/some/bad/path\\\": stat /some/bad/path: no such file or directory\": unknown",
                        "reason": "ContainerCannotRun",
                        "startedAt": "2018-06-07T18:14:01Z"
                    }
                }
```

to

```
"state": {
                    "terminated": {
                        "containerID": "docker://a4dedf0835c2dd0d106273d4138d2989ba06f391113b3508f6f776ca9237ef3c",
                        "exitCode": 127,
                        "finishedAt": "2018-06-07T18:12:54Z",
                        "reason": "Error",
                        "startedAt": "2018-06-07T18:12:54Z"
                    }
                }

```

Would that be acceptable or are we really trying to test something specific to OCI

KnVerey:
> Would that be acceptable or are we really trying to test something specific to OCI

The test is just trying to make sure log fetching works and appears in the correct place in the output. So changing the command is totally fine. 👍

dturn:
ping

PR #303, opened 2018-06-22
Resource deploy status states

dturn:
Deploy statuses has been rather tricky to maintain. Currently a resource's status is not mutually exclusive, e.g.  it could be failed, started, timed_out, and successful. While initially this might have been helpful (e.g. an resource was eventually successful, but not before the timeout hit), it has actually made writing the logging code difficult. This PR is an attempt to make make statuses mutually exclusive. 

Resources now expose a single method: `deploy_status` which returns the status (currently as a string, could be a symbol) from a list of statuses in order of precedence. This ensures mutually exclusive states.

We also push the responsibility for the state of the deploy into the resource manager so that end users don't need to be concerned with checking each resource to determine the final state. 

Part of the motivation for this PR is making changes like https://github.com/Shopify/kubernetes-deploy/pull/282/files much smaller. 

Tests fail because I haven't updated the unit tests, I wanted 👀 before going all the way.

### DIFF
--- a/lib/kubernetes-deploy/deploy_task.rb
+++ b/lib/kubernetes-deploy/deploy_task.rb
@@ -157,14 +157,14 @@ module KubernetesDeploy
 
       if verify_result
         start_normal_resource = Time.now.utc
-        deploy_resources(resources, prune: prune, verify: true)
+        watcher = deploy_resources(resources, prune: prune, verify: true)
         ::StatsD.measure('normal_resources.duration', StatsD.duration(start_normal_resource), tags: statsd_tags)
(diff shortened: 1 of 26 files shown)


dturn:
@KnVerey / @klautcomputing this is ready for high level thoughts

KnVerey:
Note from https://github.com/Shopify/kubernetes-deploy/pull/325:  There is probably an additional state along the lines of "unobserved", where we've deployed the resource but its status is state, so we cannot yet make a determination.

dturn:
Closing since this is over a year old

PR #305, opened 2018-06-27
Add HPA

dturn:
Add HPA Resource (part of splitting of https://github.com/Shopify/kubernetes-deploy/pull/188/files)

 In v2beta1 we get access to three[condition statuses](https://kubernetes.io/docs/tasks/run-application/horizontal-pod-autoscale-walkthrough/#appendix-horizontal-pod-autoscaler-status-conditions)`AbleToScale, ScalingActive , and ScalingLimited`. I think it makes sense to succeeded when `AbleToScale` is true.

>The first, AbleToScale, indicates whether or not the HPA is able to fetch and update scales, as well as whether or not any backoff-related conditions would prevent scaling.  

The downside of using v2beta is no support in 1.7. I think its worth the tradeoff

### DIFF
--- a/lib/kubernetes-deploy/kubernetes_resource.rb
+++ b/lib/kubernetes-deploy/kubernetes_resource.rb
@@ -115,7 +115,7 @@ module KubernetesDeploy
     end
 
     def sync(mediator)
-      @instance_data = mediator.get_instance(type, name)
+      @instance_data = mediator.get_instance(kubectl_resource_type, name)
     end
 
     def deploy_failed?
@@ -145,6 +145,10 @@ module KubernetesDeploy
 
     def type
       @type || self.class.kind
+    end
+
+    def kubectl_resource_type
+      type
     end
 
     def deploy_timed_out?
--- a/lib/kubernetes-deploy/kubernetes_resource/horizontal_pod_autoscaler.rb
+++ b/lib/kubernetes-deploy/kubernetes_resource/horizontal_pod_autoscaler.rb
@@ -1,0 +1,57 @@
+# frozen_string_literal: true
+module KubernetesDeploy
+  class HorizontalPodAutoscaler < KubernetesResource
+    TIMEOUT = 3.minutes
+    RECOVERABLE_CONDITIONS = %w(ScalingDisabled FailedGet)
+
+    def deploy_succeeded?
+      scaling_active_condition["status"] == "True"
+    end
+
+    def deploy_failed?
+      return false unless exists?
+      recoverable = RECOVERABLE_CONDITIONS.any? { |c| scaling_active_condition.fetch("reason", "").start_with?(c) }
+      scaling_active_condition["status"] == "False" && !recoverable
+    end
+
+    def kubectl_resource_type
+      'hpa.v2beta1.autoscaling'
+    end
+
+    def status
+      if !exists?
+        super
+      elsif deploy_succeeded?
+        "Configured"
+      elsif scaling_active_condition.present? || able_to_scale_condition.present?
+        condition = scaling_active_condition.presence || able_to_scale_condition
+        condition['reason']
+      else
+        "Unknown"
+      end
+    end
+
+    def failure_message
+      condition = scaling_active_condition.presence || able_to_scale_condition.presence || {}
+      condition['message']
+    end
+
+    def timeout_message
+      failure_message.presence || super
+    end
+
+    private
+
+    def conditions
+      @instance_data.dig("status", "conditions") || []
+    end
+
+    def able_to_scale_condition
+      conditions.detect { |c| c["type"] == "AbleToScale" } || {}
+    end
+
+    def scaling_active_condition
+      conditions.detect { |c| c["type"] == "ScalingActive" } || {}
+    end
+  end
+end
--- a/test/fixtures/hpa/deployment.yml
+++ b/test/fixtures/hpa/deployment.yml
@@ -19,4 +19,3 @@ spec:
         resources:
           requests:
             cpu: 200m
-            memory: 1Gi
--- a/test/fixtures/hpa/deployment.yml
+++ b/test/fixtures/hpa/deployment.yml
@@ -1,0 +1,22 @@
+apiVersion: apps/v1beta1
+kind: Deployment
+metadata:
+  name: hpa-deployment
+spec:
+  replicas: 1
+  progressDeadlineSeconds: 60
+  template:
+    metadata:
+      labels:
+        name: hpa
+        app: hpa-deployment
+    spec:
+      containers:
+      - name: app
+        image: busybox
+        imagePullPolicy: IfNotPresent
+        command: ["tail", "-f", "/dev/null"]
+        resources:
+          requests:
+            cpu: 200m
+            memory: 1Gi
--- a/test/helpers/kubeclient_helper.rb
+++ b/test/helpers/kubeclient_helper.rb
@@ -33,4 +33,8 @@ module KubeclientHelper
   def apiextensions_v1beta1_kubeclient
     @apiextensions_v1beta1_kubeclient ||= build_apiextensions_v1beta1_kubeclient(MINIKUBE_CONTEXT)
   end
+
+  def autoscaling_v1_kubeclient
+    @autoscaling_v1_kubeclient ||= build_autoscaling_v1_kubeclient(MINIKUBE_CONTEXT)
+  end
 end
--- a/test/integration/kubernetes_deploy_test.rb
+++ b/test/integration/kubernetes_deploy_test.rb
@@ -1053,4 +1053,21 @@ unknown field \"myKey\" in io.k8s.apimachinery.pkg.apis.meta.v1.ObjectMeta",
       "  datapoint: value1"
     ], in_order: true)
   end
+
+  def test_hpa_can_be_successful
+    skip if KUBE_SERVER_VERSION < Gem::Version.new('1.8.0')
+    assert_deploy_success(deploy_fixtures("hpa"))
+    assert_logs_match_all([
+      "Deploying resources:",
+      "HorizontalPodAutoscaler/hello-hpa (timeout: 180s)",
+      %r{HorizontalPodAutoscaler/hello-hpa\s+Configured}
+    ])
+  end
+
+  def test_hpa_can_be_pruned
+    skip if KUBE_SERVER_VERSION < Gem::Version.new('1.8.0')
+    assert_deploy_success(deploy_fixtures("hpa"))
+    assert_deploy_success(deploy_fixtures("hpa", subset: ["deployment.yml"]))
+    assert_logs_match_all(['The following resources were pruned: horizontalpodautoscaler "hello-hpa"'])
+  end
 end
--- a/test/integration/kubernetes_deploy_test.rb
+++ b/test/integration/kubernetes_deploy_test.rb
@@ -1055,7 +1055,7 @@ unknown field \"myKey\" in io.k8s.apimachinery.pkg.apis.meta.v1.ObjectMeta",
   end
 
   def test_hpa_can_be_successful
-    skip if KUBE_SERVER_VERSION < Gem::Version.new('1.8.0')
+    skip if KUBE_SERVER_VERSION < Gem::Version.new('1.9.0')
     assert_deploy_success(deploy_fixtures("hpa"))
     assert_logs_match_all([
       "Deploying resources:",
@@ -1065,7 +1065,7 @@ unknown field \"myKey\" in io.k8s.apimachinery.pkg.apis.meta.v1.ObjectMeta",
   end
 
   def test_hpa_can_be_pruned
-    skip if KUBE_SERVER_VERSION < Gem::Version.new('1.8.0')
+    skip if KUBE_SERVER_VERSION < Gem::Version.new('1.9.0')
     assert_deploy_success(deploy_fixtures("hpa"))
     assert_deploy_success(deploy_fixtures("hpa", subset: ["deployment.yml"]))
     assert_logs_match_all([/The following resources were pruned: horizontalpodautoscaler(.autoscaling)? "hello-hpa"/])
--- a/test/integration/kubernetes_deploy_test.rb
+++ b/test/integration/kubernetes_deploy_test.rb
@@ -1068,6 +1068,6 @@ unknown field \"myKey\" in io.k8s.apimachinery.pkg.apis.meta.v1.ObjectMeta",
     skip if KUBE_SERVER_VERSION < Gem::Version.new('1.8.0')
     assert_deploy_success(deploy_fixtures("hpa"))
     assert_deploy_success(deploy_fixtures("hpa", subset: ["deployment.yml"]))
-    assert_logs_match_all(['The following resources were pruned: horizontalpodautoscaler "hello-hpa"'])
+    assert_logs_match_all([/The following resources were pruned: horizontalpodautoscaler(.autoscaling)? "hello-hpa"/])
   end
 end
--- a/test/unit/sync_mediator_test.rb
+++ b/test/unit/sync_mediator_test.rb
@@ -143,6 +143,12 @@ class SyncMediatorTest < KubernetesDeploy::TestCase
     mediator.get_instance('FakeConfigMap', @fake_cm.name)
   end
 
+  def test_sync_uses_kubectl_resource_type
+    hpa = FakeHPA.new('fake')
+    stub_kubectl_response('get', hpa.kubectl_resource_type, *@params, resp: { "items" => [] }, times: 1)
+    mediator.sync([hpa] * (KubernetesDeploy::SyncMediator::LARGE_BATCH_THRESHOLD + 1))
+  end
+
   private
 
   def mediator
@@ -162,6 +168,10 @@ class SyncMediatorTest < KubernetesDeploy::TestCase
 
     def type
       self.class.name.demodulize
+    end
+
+    def kubectl_resource_type
+      type
     end
 
     def kubectl_response
@@ -187,6 +197,12 @@ class SyncMediatorTest < KubernetesDeploy::TestCase
   class BadCitizen < MockResource
     def sync(mediator)
       mediator.sync([]) # clears the cache
+    end
+  end
+
+  class FakeHPA < MockResource
+    def kubectl_resource_type
+      'Not-HPA-Type'
     end
   end
 end
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,11 +2,12 @@
 
 *Enhancements*
 - Add Job resource class ([#295](https://github.com/Shopify/kubernetes-deploy/pull/296))
-- Add CustomResourceDefinition resource class ([#305](https://github.com/Shopify/kubernetes-deploy/pull/306))
+- Add CustomResourceDefinition resource class ([#306](https://github.com/Shopify/kubernetes-deploy/pull/306))
 - Officially support Kubernetes 1.10 ([#308](https://github.com/Shopify/kubernetes-deploy/pull/308))
 - SyncMediator will only batch fetch resources when there is a sufficiently large enough set of resources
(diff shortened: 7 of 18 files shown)


KnVerey:
I found the difference between the intent of ScalingActive and AbleToScale a bit confusing too, so I looked up all the possible reasons for them ([this file](https://github.com/kubernetes/kubernetes/blob/a30f4591609301bd6160abeb22731f70e7dae2e0/pkg/controller/podautoscaler/horizontal.go)):

AbleToScale

- `FailedGetScale`: the HPA controller was unable to get the target's current scale
- `BackoffDownscale`: the time since the previous scale is still within the downscale forbidden window
- `BackoffBoth`: the time since the previous scale is still within both the downscale and upscale forbidden windows
- `BackoffUpscale`: the time since the previous scale is still within the upscale forbidden window
- `ReadyForNewScale`: the last scale time was sufficiently old as to warrant a new scale
- `FailedUpdateScale`: the HPA controller was unable to update the target scale
- `SucceededRescale`: the HPA controller was able to update the target scale to %d

ScalingActive

- `InvalidSelector`: the HPA target's scale is missing a selector
- `InvalidSelector`: couldn't convert selector into a corresponding internal selector object
- `FailedGetObjectMetric/FailedGetPodsMetric/FailedGetResourceMetric/FailedGetExternalMetric`: the HPA was unable to compute the replica count
- `InvalidMetricSourceType`: the HPA was unable to compute the replica count
- `ValidMetricFound`: the HPA was able to successfully calculate a replica count from %s
- `ScalingDisabled`: scaling is disabled since the replica count of the target is zero


Based on that, AbleToScale is talking about ability to scale right now in particular, whereas ScalingActive tells us whether the HPA is correctly configured. So the latter is what we want. We should probably ignore false values with that last reason though, or else deploys will fail when people using HPAs scale to zero manually.

lib/kubernetes-deploy/kubernetes_resource/horizontal_pod_autoscaler.rb
```diff
@@ -0,0 +1,25 @@
+# frozen_string_literal: true
+module KubernetesDeploy
+  class HorizontalPodAutoscaler < KubernetesResource
+    PRUNABLE = true
```
KnVerey: I do like the idea of defining this at the class level, but this doesn't do anything right now. Your test passes because, oddly enough, HPA was already on the whitelist.

lib/kubernetes-deploy/kubernetes_resource/horizontal_pod_autoscaler.rb
```diff
@@ -0,0 +1,25 @@
...
+    TIMEOUT = 30.seconds
+
+    def deploy_succeeded?
+      conditions = @instance_data.dig("status", "conditions") || []
+      able_to_scale = conditions.detect { |c| c["type"] == "AbleToScale" } || {}
+      able_to_scale["status"] == "True"
+    end
+
+    def deploy_failed?
+      !exists?
```
KnVerey: I think this should be checking the `ScalingActive` condition too (but should the deploy should succeed when that's false because `ScalingDisabled`).

lib/kubernetes-deploy/kubernetes_resource/horizontal_pod_autoscaler.rb
```diff
@@ -0,0 +1,25 @@
...
+    end
+
+    def timeout_message
+      UNUSUAL_FAILURE_MESSAGE
```
KnVerey: I'm not sure this message (`It is very unusual for this resource type to fail to deploy. Please try the deploy again. If that new deploy also fails, contact your cluster administrator.`) is fitting, since HPAs have legit failure mode (unlike say configmaps).

lib/kubernetes-deploy/kubernetes_resource/horizontal_pod_autoscaler.rb
```diff
@@ -0,0 +1,25 @@
...
+    end
+
+    def type
+      'hpa.v2beta1.autoscaling'
```
KnVerey: Why this? Don't we want it called `HorizontalPodAutoscaler`?

test/fixtures/hpa/deployment.yml
```diff
@@ -0,0 +1,20 @@
...
+  annotations:
+    shipit.shopify.io/restart: "true"
+spec:
+  replicas: 1
+  progressDeadlineSeconds: 60
+  template:
+    metadata:
+      labels:
+        name: web
+        app: hello-cloud
```
KnVerey: nit: these labels should be "hpa" and "hpa-deployment"

test/integration/kubernetes_deploy_test.rb
```diff
@@ -1006,4 +1006,21 @@ def test_raise_on_yaml_missing_kind
       "  datapoint: value1"
     ], in_order: true)
   end
+
+  def test_hpa_can_be_successful
+    skip if KUBE_SERVER_VERSION < Gem::Version.new('1.8.0')
```
KnVerey: Your PR description says `The downside of using v2beta is no support in 1.8.`; did you mean 1.7? If that's true, our unofficial deprecation policy (following GKE) would allow us to drop 1.7 at this point, and then the hpa could be folded into `hello-cloud`, as could the cronjob tests. If that isn't the case (i.e. the conditions aren't present in a version we need to support), don't we need to have alternate success/failure conditions in the code itself rather than here, to avoid breaking deploys for people with hpas on that version?

test/helpers/kubeclient_helper.rb
```diff
@@ -25,4 +25,8 @@ def apps_v1beta1_kubeclient
   def batch_v1beta1_kubeclient
     @batch_v1beta1_kubeclient ||= build_batch_v1beta1_kubeclient(MINIKUBE_CONTEXT)
   end
+
+  def autoscaling_v2beta1_kubeclient
+    @autoscaling_v2beta1_kubeclient ||= build_autoscaling_v2beta1_kubeclient(MINIKUBE_CONTEXT)
```
KnVerey: I'm not seeing these clients being used

test/integration/kubernetes_deploy_test.rb
```diff
@@ -1006,4 +1006,21 @@ def test_raise_on_yaml_missing_kind
...
+    assert_deploy_success(deploy_fixtures("hpa"))
```
KnVerey: Please try to include at least one relevant logging assertion with your tests. The logs are our UI, so we should always check (`PRINT_LOGS=1`) what they look like and assert that we haven't broken them. Typically I'll assert on what gets printed about the resource in the summary

lib/kubernetes-deploy/kubernetes_resource/horizontal_pod_autoscaler.rb
dturn: We do, but this method determines how kubectl fetches the resource. And we don't get the status conditions until v2beta1. Do we want to add a new method?

test/integration/kubernetes_deploy_test.rb
dturn: It was added in 1.8 and I've updated the description. What does dropping support for 1.7 mean past not running 1.7 CI?

dturn:
It looks like I used `AbleToScale` vs `ScalingActive` because `scalingActive` is false in minikube:
```
            "status": {
                "conditions": [
                    {
                        "lastTransitionTime": "2018-07-17T00:15:37Z",
                        "message": "the HPA controller was able to get the target's current scale",
                        "reason": "SucceededGetScale",
                        "status": "True",
                        "type": "AbleToScale"
                    },
                    {
                        "lastTransitionTime": "2018-07-17T00:15:37Z",
                        "message": "the HPA was unable to compute the replica count: unable to get metrics for resource cpu: unable to fetch metrics from resource metrics API: the server could not find the requested resource (get pods.metrics.k8s.io)",
                        "reason": "FailedGetResourceMetric",
                        "status": "False",
                        "type": "ScalingActive"
                    }
                ],
                "currentMetrics": null,
                "currentReplicas": 1,
                "desiredReplicas": 0
            }
        }
```

lib/kubernetes-deploy/kubernetes_resource/horizontal_pod_autoscaler.rb
```diff
@@ -0,0 +1,36 @@
+# frozen_string_literal: true
+module KubernetesDeploy
+  class HorizontalPodAutoscaler < KubernetesResource
+    TIMEOUT = 30.seconds
```
KnVerey: Should we make this a little longer? Seemingly some metrics check is happening, and I have no idea how long that can take.

dturn:
Turns out the issue was https://github.com/kubernetes/kubernetes/issues/57673 e.g. that we need to deploy the metric server from https://github.com/kubernetes-incubator/metrics-server.

This is ready for 👀 again

test/integration/kubernetes_deploy_test.rb
```diff
@@ -1053,4 +1053,44 @@ def test_raise_on_yaml_missing_kind
...
+    skip if KUBE_SERVER_VERSION < Gem::Version.new('1.8.0')
+    deploy_metric_server
+    assert_deploy_success(deploy_fixtures("hpa/hpa"))
+    assert_deploy_success(deploy_fixtures("hpa/hpa", subset: ["deployment.yml"]))
+    assert_logs_match_all(['The following resources were pruned: horizontalpodautoscaler "hello-hpa"'])
+  end
+
+  private
+
+  def deploy_metric_server
```
dturn: Is there a helper file we've been putting methods like this?

lib/kubernetes-deploy/kubernetes_resource/horizontal_pod_autoscaler.rb
```diff
@@ -0,0 +1,46 @@
...
+    RECOVERABLE_CONDITIONS = %w(ScalingDisabled FailedGet)
+
+    def deploy_succeeded?
+      scaling_active_condition["status"] == "True"
+    end
+
+    def deploy_failed?
+      return false unless exists?
+      recoverable = RECOVERABLE_CONDITIONS.any? { |c| scaling_active_condition.fetch("reason", "").start_with?(c) }
+      able_to_scale_condition["status"] == "False" || (scaling_active_condition["status"] == "False" && !recoverable)
```
KnVerey: I don't think the "able to scale" condition is relevant at all (same in `status`). It's talking about whether or not the target in question can be scaled at this particular moment (based on things like whether or not it was _just_ scaled a moment ago), not on the general validity of the autoscaling configuration. I see you have a test that implies that "able to scale - false; FailedGet" means the target resource may not exist, but couldn't failing on that cause a race condition, since the HPA doesn't get deployed after deployments? Couldn't it also mean the request for that resource (transiently) failed, i.e. should be excluded on the same grounds as other FailedGetXs? Is "Scaling active" actually true when the target doesn't exist?

lib/kubernetes-deploy/kubernetes_resource.rb
```diff
@@ -114,7 +114,7 @@ def file_path
     end
 
     def sync(mediator)
-      @instance_data = mediator.get_instance(type, name)
+      @instance_data = mediator.get_instance(fetch_type, name)
```
KnVerey: The sync mediator's caching should also use the new method, and we should make sure we add a test that would catch that.

https://github.com/Shopify/kubernetes-deploy/blob/7b47ed81f040922ad877fc7e3035f6b82a71ea87/lib/kubernetes-deploy/sync_mediator.rb#L39

lib/kubernetes-deploy/kubernetes_resource/horizontal_pod_autoscaler.rb
```diff
@@ -0,0 +1,46 @@
...
+
+    def fetch_type
+      'hpa.v2beta1.autoscaling'
+    end
+
+    def status
+      if !exists?
+        super
+      elsif deploy_succeeded?
+        "Succeeded"
```
KnVerey: "Configured"? "Succeeded" sounds like we're claiming it scaled something.

test/fixtures/hpa/hpa/deployment.yml
```diff
@@ -0,0 +1,24 @@
+apiVersion: apps/v1beta1
```
KnVerey: Why is this fixture set two layers deep? i.e. why not `test/fixtures/hpa/thing.yml`?

test/fixtures/hpa/hpa/deployment.yml
```diff
@@ -0,0 +1,24 @@
...
+kind: Deployment
+metadata:
+  name: hpa-deployment
+  annotations:
+    shipit.shopify.io/restart: "true"
```
KnVerey: no need for this annotation

test/integration/kubernetes_deploy_test.rb
KnVerey: Sort of, yes. This is being deployed globally, so I think it should happen during the initial test suite setup, which is taken care of inside `test_helper` itself. I'd add it to this module:

https://github.com/Shopify/kubernetes-deploy/blob/7b47ed81f040922ad877fc7e3035f6b82a71ea87/test/test_helper.rb#L243-L254

And then call it with the PV setup, i.e. when the file is loaded

https://github.com/Shopify/kubernetes-deploy/blob/7b47ed81f040922ad877fc7e3035f6b82a71ea87/test/test_helper.rb#L289

test/integration/kubernetes_deploy_test.rb
```diff
@@ -1053,4 +1053,44 @@ def test_raise_on_yaml_missing_kind
...
+    # Set-up the metric server that the HPA needs https://github.com/kubernetes-incubator/metrics-server
+    ns = @namespace
+    @namespace = "kube-system"
+    assert_deploy_success(deploy_fixtures("hpa/kube-system", allow_protected_ns: true, prune: false))
```
KnVerey: `hpa/kube-system` isn't really a fixture set; it's more like test infra components. I'd move all those configs somewhere else, like `test/setup/metrics-server` or something. Maybe we should even use KubeClient or `kubectl apply` to create it... seems a little weird to have the test setup itself depend on the correct functioning of the core code.

lib/kubernetes-deploy/kubernetes_resource/horizontal_pod_autoscaler.rb
```diff
@@ -0,0 +1,46 @@
+# frozen_string_literal: true
+module KubernetesDeploy
+  class HorizontalPodAutoscaler < KubernetesResource
+    TIMEOUT = 5.minutes
```
KnVerey: Is this arbitrary or based on testing? 30s seemed short to me, but this seems really long 😄

dturn:
>Did you have to tweak the metrics server configs at all to get them working, or are they a direct copy-paste from https://github.com/kubernetes-incubator/metrics-server?

Direct copy-paste

lib/kubernetes-deploy/kubernetes_resource/horizontal_pod_autoscaler.rb
dturn: I actually think `FailedGetScale` is relevant since when we are failing with that reason there isn't an  'AbleToScale' condition to look at. I'll scope this down to just that condition

test/integration/kubernetes_deploy_test.rb
```diff
@@ -1053,4 +1053,31 @@ def test_raise_on_yaml_missing_kind
...
+      "HorizontalPodAutoscaler/hello-hpa (timeout: 180s)",
+      %r{HorizontalPodAutoscaler/hello-hpa\s+Configured}
+    ])
+  end
+
+  def test_hpa_times_out_when_no_matching_deployment
+    skip if KUBE_SERVER_VERSION < Gem::Version.new('1.8.0')
+    assert_deploy_failure(deploy_fixtures("hpa", subset: ["hpa.yml"]), :timed_out)
+    assert_logs_match_all([
+      "Deploying HorizontalPodAutoscaler/hello-hpa (timeout: 180s)",
```
KnVerey: This test is guaranteed to take at least 180s 😢 
There's some value in asserting that the message displayed on timeout is helpful, but I'm not sure it's worth 3 minutes of CI. Timeouts are fallback behaviour, not something we're generally striving for. I'd be inclined to replace this with a test for a case that actually fails.

lib/kubernetes-deploy/kubernetes_resource/horizontal_pod_autoscaler.rb
```diff
@@ -0,0 +1,48 @@
...
+    end
+
+    def status
+      if !exists?
+        super
+      elsif deploy_succeeded?
+        "Configured"
+      elsif scaling_active_condition.present? || able_to_scale_condition.present?
+        condition = scaling_active_condition.presence || able_to_scale_condition
+        "#{condition['reason']} (#{condition['message']})"
```
KnVerey: Looking at the test output, this makes for a realllly long status string. Usually those are a word or two. I'd suggest using the reason as the status and moving the message to `failure_message`/`timeout_message` when it is present. We're not currently setting those at all, and as a result we're getting the default "Kubernetes will continue to deploy this..." timeout message, which isn't great.

test/unit/sync_mediator_test.rb
```diff
@@ -189,4 +195,10 @@ def sync(mediator)
       mediator.sync([]) # clears the cache
     end
   end
+
+  class FakeHPA < MockResource
+    def kubectl_resource_type
+      'fakeHPA'
```
KnVerey: Isn't this the same as the value of `type`?  If so the test isn't proving anything

dturn: [1] pry(#<SyncMediatorTest>)> hpa.kubectl_resource_type
=> "fakeHPA"
[2] pry(#<SyncMediatorTest>)> hpa.type
=> "FakeHPA"

I'll update to make it more clear.

test/integration/kubernetes_deploy_test.rb
dturn: I'm having a hard time figuring out how to make the HPA pass validation, have AbleToScale be true, and have ScalingActive false with out the cause being in `RECOVERABLE_CONDITIONS`. 

The k8s tests (https://github.com/kubernetes/kubernetes/blob/master/pkg/controller/podautoscaler/horizontal_test.go) weren't very instructional, any suggestions?

KnVerey: I can't find anything useful either. Kinda weird that so many conditions have been dedicated to situations that are not possible to create from the outside. My final observation is that in playing around with this locally, there seems to be a substantial delay before the conditions get populated, so maybe it would be safe to fail on `ScalingActive/FailedGetThingMetric` after all, i.e. the benefit of failing fast for those misconfigurations would outweigh the (small) possibility of a race condition causing spurious failures. Without much direct HPA experience, I'm not really sure what is best. 

If we can't reproduce a failure scenario, maybe we can set a short timeout on the hpa resource? That won't really work if the condition message we'd look for in the logs can also take 3 minutes to appear though.

dturn: How would you feel about using `Timecop.scale` to make this test faster in wallclock time? 

In general I dislike trading correctness for an optimization, failing fast.

dturn:
1.8 is failing but not 1.9 or 1.10. Do you think https://github.com/kubernetes-incubator/metrics-server supports 1.7 and > 1.8 (not >=1.8)?

dturn:
1.8 failing appears to be related to minikube. And as far as I can tell a change in k8s default from 1.8 to 1.9. To make it work I had to run add `--extra-config=controller-manager.horizontal-pod-autoscaler-use-rest-clients=true` to the minikube start command `minikube addons enable metrics-server`  to make it pass. I'm going to disable the tests for 1.8 rather than mess with this more. 

https://github.com/kubernetes/kubernetes/issues/57673
https://stackoverflow.com/questions/48325627/minikube-horizontal-pod-autoscaling-error-unable-to-get-metrics-for-resource-cp

PR #306, opened 2018-06-28
Add crd

dturn:
Add CRD Resource (part of splitting of https://github.com/Shopify/kubernetes-deploy/pull/188/files)

Success of failure based on the `NamesAccepted` status condition. Is not very sophisticated, but better than nothing.

screen shot 2018-07-17 at 10 05 31 am


karanthukral:
Looks good to me. Simple to start 👍

lib/kubernetes-deploy/kubernetes_resource/custom_resource_definition.rb
```diff
@@ -0,0 +1,40 @@
...
+    def deploy_succeeded?
+      names_accepted_status == "True"
+    end
+
+    def deploy_failed?
+      names_accepted_status == "False"
+    end
+
+    def timeout_message
+      UNUSUAL_FAILURE_MESSAGE
```
KnVerey: Let's be more specific about what happened here, e.g. "The names this CRD is attempting to register were neither accepted nor rejected in time"

lib/kubernetes-deploy/kubernetes_resource/custom_resource_definition.rb
```diff
@@ -0,0 +1,40 @@
+# frozen_string_literal: true
+module KubernetesDeploy
+  class CustomResourceDefinition < KubernetesResource
+    TIMEOUT = 30.seconds
```
KnVerey: I'd suggest being a little more generous here, since something actually does need to happen (vs. configmaps etc. where we're really just creating them and seeing that they showed up)

test/helpers/fixture_sets/hello_cloud.rb
```diff
@@ -94,5 +95,9 @@ def assert_daemon_set_up
     def assert_stateful_set_up
       assert_stateful_set_present("stateful-busybox")
     end
+
+    def assert_crd_up
+      assert_crd_present("mails.stable.example.io")
```
KnVerey: The `mails` plural still irks me :P

test/integration/kubernetes_deploy_test.rb
```diff
@@ -1006,4 +1007,32 @@ def test_raise_on_yaml_missing_kind
...
+      crd = f.dig("crd.yml", "CustomResourceDefinition").first
+      names = crd.dig("spec", "names")
+      crd["metadata"]["name"] = 'mis-matched.stable.example.io'
+      names["plural"] = 'mis-matched'
+    end
+    assert_deploy_failure(result)
+    assert_logs_match_all([
+      "Deploying CustomResourceDefinition/mails.stable.example.io (timeout: 30s)",
+      "CustomResourceDefinition/mis-matched.stable.example.io: FAILED",
+      "Final status: ListKindConflict"
```
KnVerey: This is a little cryptic... Does the condition's message have additional information we're not printing?

lib/kubernetes-deploy/deploy_task.rb
```diff
@@ -247,6 +248,10 @@ def discover_resources
           @logger.info "  - #{r.id}"
         end
       end
+      if (global = resources.select(&:global?).presence)
+        @logger.info("Detected non-namespaced #{'resouce'.pluralize(global.count)} which will never be pruned:")
```
KnVerey: let's print this in yellow, similar to these warnings:

https://github.com/Shopify/kubernetes-deploy/blob/a300fc604ba9c6315179b304287d0272bebb91bd/lib/kubernetes-deploy/deploy_task.rb#L319-L322

KnVerey: Typo: `resouce`-> `resource`

test/integration/kubernetes_deploy_test.rb
```diff
@@ -1006,4 +1007,32 @@ def test_raise_on_yaml_missing_kind
...
+      "Detected non-namespaced resouce which will never be pruned:",
+      " - CustomResourceDefinition/mails.stable.example.io",
+      "Phase 2: Checking initial resource statuses",
+      "Deploying CustomResourceDefinition/mails.stable.example.io (timeout: 30s)",
+      %r{CustomResourceDefinition/mails.stable.example.io\s+Succeeded}
+    ])
+  end
+
+  def test_crd_can_fail
+    assert_deploy_success(deploy_fixtures("hello-cloud", subset: ["crd.yml"]))
```
KnVerey: Is this first deploy necessary in this case? Won't mismatched names be rejected even on the first registration?

lib/kubernetes-deploy/kubernetes_resource/custom_resource_definition.rb
```diff
@@ -0,0 +1,40 @@
...
+
+    def timeout_message
+      UNUSUAL_FAILURE_MESSAGE
+    end
+
+    def status
+      if !exists?
+        super
+      elsif deploy_succeeded?
+        "Succeeded"
```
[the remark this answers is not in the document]
dturn: I wasn't sure if this was a place to be specific or match what other resources. Given the timeout_message "Names accepted" seems good to me

test/integration/kubernetes_deploy_test.rb
dturn: We're actually creating a conflict between the `listKind` field in the fixture vs. the slightly modified fixture. A mismatch between the name & plural is caught at validation time and isn't really what we want to test.

dturn:
> We also need to do something to clean up the CRDs after every test, otherwise the success tests in particular won't be valid (because the CRD will actually predate them if any of the hello-cloud test have run against that cluster).

Because the CRD resource is global, I think the CRD test either need to create unique CRDs or not be run in parallel

timothysmith0609:
>Because the CRD resource is global, I think the CRD test either need to create unique CRDs or not be run in parallel. I've gone with the unique route, we're polluting minikube, but I don't think that's a real issue.

Could we not just delete the CRD at the end of each test run?

dturn:
> Could we not just delete the CRD at the end of each test run?

We can't because of the parallelism.

KnVerey:
> We can't because of the parallelism. 

We have a file for tests that should not be run in parallel: `test/integration-serial/run_serial_test.rb`. Let's move these tests there.

KnVerey:
I'm 👍 on the code and output now though.

PR #307, opened 2018-06-28
[WIP] Add a state machine for deploy states

dturn:
Deploy statuses has been rather tricky to maintain. Currently a resource's status is not mutually exclusive, e.g. it could be failed, started, timed_out, and successful. While initially this might have been helpful (e.g. an resource was eventually successful, but not before the timeout hit), it has actually made writing the logging code difficult. To fix this introduce a [state_machine](https://github.com/state-machines/state_machines)

Part of the motivation for this PR is making changes like https://github.com/Shopify/kubernetes-deploy/pull/282/files much smaller.

Tests are failing...


dturn:
Closing as this is over a year old

PR #299, opened 2018-06-06
Enable Kubernetes 1.10 Tests

stefanmb:
**What**
Enable k8s 1.10 tests.

PR #308, opened 2018-06-29
K8s 110 ci with fixes

dturn:
In 1.10 progressDeadlineSeconds is in instance data even if its not in the spec. As a result, our logic for progressDeadlineSeconds is triggered instead of the hard-timeout logic.

This is https://github.com/Shopify/kubernetes-deploy/pull/299 with a commit to fix the issue.

### DIFF
--- a/dev.yml
+++ b/dev.yml
@@ -8,7 +8,7 @@ up:
   - custom:
       name: Minikube Cluster
       met?: test $(minikube status | grep Running | wc -l) -eq 2 && $(minikube status | grep -q 'Correctly Configured')
-      meet: minikube start --kubernetes-version=v1.9.4 --vm-driver=hyperkit
+      meet: minikube start --kubernetes-version=v1.10.4 --vm-driver=hyperkit
       down: minikube stop
 commands:
   reset-minikube: minikube delete && rm -rf ~/.minikube
--- a/dev.yml
+++ b/dev.yml
@@ -8,7 +8,7 @@ up:
   - custom:
       name: Minikube Cluster
       met?: test $(minikube status | grep Running | wc -l) -eq 2 && $(minikube status | grep -q 'Correctly Configured')
-      meet: minikube start --kubernetes-version=v1.10.4 --vm-driver=hyperkit
+      meet: minikube start --kubernetes-version=v1.10.0 --vm-driver=hyperkit
       down: minikube stop
 commands:
   reset-minikube: minikube delete && rm -rf ~/.minikube
--- a/lib/kubernetes-deploy/kubernetes_resource/deployment.rb
+++ b/lib/kubernetes-deploy/kubernetes_resource/deployment.rb
@@ -119,14 +119,14 @@ module KubernetesDeploy
     end
 
     def progress_condition
-      return unless exists?
+      return unless exists? && progress_deadline
       conditions = @instance_data.fetch("status", {}).fetch("conditions", [])
       conditions.find { |condition| condition['type'] == 'Progressing' }
     end
 
     def progress_deadline
       if exists?
-        @instance_data['spec']['progressDeadlineSeconds']
+        @instance_data['spec']['progressDeadlineSeconds'] if @definition['spec']['progressDeadlineSeconds'].present?
       else
         @definition['spec']['progressDeadlineSeconds']
       end
--- a/lib/kubernetes-deploy/kubernetes_resource/deployment.rb
+++ b/lib/kubernetes-deploy/kubernetes_resource/deployment.rb
@@ -119,14 +119,14 @@ module KubernetesDeploy
     end
 
     def progress_condition
-      return unless exists? && progress_deadline
+      return unless exists?
       conditions = @instance_data.fetch("status", {}).fetch("conditions", [])
       conditions.find { |condition| condition['type'] == 'Progressing' }
     end
 
     def progress_deadline
       if exists?
-        @instance_data['spec']['progressDeadlineSeconds'] if @definition['spec']['progressDeadlineSeconds'].present?
+        @instance_data['spec']['progressDeadlineSeconds']
       else
         @definition['spec']['progressDeadlineSeconds']
       end
--- a/test/integration/kubernetes_deploy_test.rb
+++ b/test/integration/kubernetes_deploy_test.rb
@@ -543,17 +543,28 @@ unknown field \"myKey\" in io.k8s.apimachinery.pkg.apis.meta.v1.ObjectMeta",
       %r{ConfigMap/test\s+Available}
     ], in_order: true)
 
+    if KUBE_SERVER_VERSION < Gem::Version.new("1.10.0")
+      start_bad_probe_logs = [
+        %r{Deployment/bad-probe: TIMED OUT \(timeout: \d+s\)},
+        "Timeout reason: hard deadline for Deployment"
+      ]
+      end_bad_probe_logs = [%r{Unhealthy: Readiness probe failed: .* \(\d+ events\)}] # event
+    else
+      start_bad_probe_logs = [
+        %r{Deployment/bad-probe: TIMED OUT \(progress deadline: \d+s\)},
+        "Timeout reason: ProgressDeadlineExceeded"
+      ]
+      end_bad_probe_logs = ["Scaled up replica set bad-probe-"] # event
+    end
+
     # Debug info for bad probe timeout
-    assert_logs_match_all([
-      %r{Deployment/bad-probe: TIMED OUT \(timeout: \d+s\)},
-      "Timeout reason: hard deadline for Deployment",
+    assert_logs_match_all(start_bad_probe_logs + [
       /Latest ReplicaSet: bad-probe-\w+/,
       "The following containers have not passed their readiness probes on at least one pod:",
       "http-probe must respond with a good status code at '/bad/ping/path'",
       "exec-probe must exit 0 from the following command: 'test 0 -eq 1'",
       "Final status: 1 replica, 1 updatedReplica, 1 unavailableReplica",
-      "Scaled up replica set bad-probe-", # event
-    ], in_order: true)
+    ] + end_bad_probe_logs, in_order: true)
     refute_logs_match("sidecar must exit 0") # this container is ready
 
     # Debug info for missing volume timeout
@@ -930,12 +941,18 @@ unknown field \"myKey\" in io.k8s.apimachinery.pkg.apis.meta.v1.ObjectMeta",
     end
     assert_deploy_failure(result)
 
+    bad_probe_timeout = if KUBE_SERVER_VERSION < Gem::Version.new("1.10.0")
+      "Deployment/bad-probe: TIMED OUT (timeout: 5s)"
+    else
+      "Deployment/bad-probe: GLOBAL WATCH TIMEOUT (20 seconds)"
+    end
+
     assert_logs_match_all([
       "Successfully deployed 1 resource, timed out waiting for 2 resources to deploy, and failed to deploy 1 resource",
       "Successful resources",
       "ConfigMap/test",
       "Deployment/cannot-run: FAILED",
-      "Deployment/bad-probe: TIMED OUT (timeout: 5s)",
+      bad_probe_timeout,
       "Deployment/missing-volumes: GLOBAL WATCH TIMEOUT (20 seconds)"
     ])
   end
--- a/test/integration/kubernetes_deploy_test.rb
+++ b/test/integration/kubernetes_deploy_test.rb
@@ -535,7 +535,7 @@ unknown field \"myKey\" in io.k8s.apimachinery.pkg.apis.meta.v1.ObjectMeta",
   def test_deploy_result_logging_for_mixed_result_deploy
     subset = ["bad_probe.yml", "init_crash.yml", "missing_volumes.yml", "config_map.yml"]
     result = deploy_fixtures("invalid", subset: subset) do |f|
-      if KUBE_SERVER_VERSION >= Gem::Version.new("1.10.0")
+      if KUBE_SERVER_VERSION >= Gem::Version.new("1.10.0") # https://github.com/kubernetes/kubernetes/issues/66135
         f["bad_probe.yml"]["Deployment"].first["spec"]["progressDeadlineSeconds"] = 20
       end
     end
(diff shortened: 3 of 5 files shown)


stefanmb:
👀 Thanks for looking at this, I dropped the ball with following up.

lib/kubernetes-deploy/kubernetes_resource/deployment.rb
```diff
@@ -119,14 +119,14 @@ def rollout_data
     end
 
     def progress_condition
-      return unless exists?
+      return unless exists? && progress_deadline
```
stefanmb: I'm a bit rusty on this code - I understand why the `progress_deadline` body had to be changed, but why/how was `progress_condition` working with just the `exists?` check before this PR?

dturn: In 1.10 the `Progressing` condition (added by a k8s controller) is now being added to deployments even when `progressDeadlineSeconds` isn't in the spec deployed.  This isn't happening in 1.9.  (I'm not sure why it's happening. A guess, who knows if its valid since I'm not sure where in the k8s diff between 1.9-1.10 to look, is that the default value of `progressDeadlineSeconds` is getting merged in sooner than before?)

KnVerey: The test in question uses an extensions/v1beta1 Deployment. PDS is not supposed to get defaulted at all in that GV... maybe 1.10 added a bug? Can you confirm whether PDS itself is being defaulted now, or whether it's just the Progressing condition being added (and never expiring, I guess?)? 

If we change our progress deadline to depend on `if @definition['spec']['progressDeadlineSeconds'].present?` as in this PR, we'll have a behaviour change: previously, if your Deployment got a PDS via defaulting (e.g. an apps/v1 Deployment with no PDS specified), we'd use a PDS-based timeout. After this change, cases like that would use the hard timeout.

KnVerey: Note that the [1.10 docs](https://v1-10.docs.kubernetes.io/docs/reference/generated/kubernetes-api/v1.10/#deploymentspec-v1beta1-extensions) still claim there is no default value in extensions/v1beta1. And the controller code removes the Progressing condition if no PDS is set: https://github.com/kubernetes/kubernetes/blob/74bcefc8b2bf88a2f5816336999b524cc48cf6c0/pkg/controller/deployment/progress.go#L38-L41 😕

timothysmith0609: I took a look at an older deployment that didn't have PDS set and also using `extensions/v1beta1`. It does indeed get its PDS set with the default 600 seconds in its spec. Could it be that the apiserver is using `apps/v1` under the hood and the default is being injected into the `extensions/v1beta1` spec?

The spec definition: https://github.com/Shopify/u2/blob/73c0ed1f461b319bd5adb35a5865f9beacf82a9e/config/deploy/staging/web.yml.erb#L62

Inside the cluster (PDS has been set automatically somewhere):
```
spec:
  progressDeadlineSeconds: 600
  replicas: 3
  revisionHistoryLimit: 2
  selector:
     ...
```

dturn: Yep, this change behavior change happened somewhere between 1.9.9 and 1.10.0 (https://gist.github.com/dturn/d0a180e011e0ca3c47987818a245ed4d#file-gistfile1-txt-L25), I'd guess in this pr https://github.com/kubernetes/kubernetes/pull/61419 .

> if your Deployment got a PDS via defaulting (e.g. an apps/v1 Deployment with no PDS specified), we'd use a PDS-based timeout. After this change, cases like that would use the hard timeout.

I'm ok with this change. Its rather trivial for people to explicitly set a pds if they want to maintain the previous behavior. I also suspect a pds of 10 minutes wasn't something anyone really wanted.

KnVerey: Given that the 1.10 docs still say these values have no default in extensions/v1beta1, that maintainers have [repeatedly](https://github.com/kubernetes/kubernetes/pull/46576#pullrequestreview-41367176) [said](https://github.com/kubernetes/kubernetes/pull/61221) defaults cannot be changed in released API versions, and that changing defaults [was the whole reason](https://github.com/kubernetes/kubernetes/pull/39683) they introduced apps/v1beta1 in the first place, this seems like a Kubernetes bug. However, I can't find any related bug reports other than [this one](https://github.com/kubernetes/kubernetes/issues/43589), which suggests turning up verbosity to see which groupversion is being used in the requests. If we do that and confirm that requests to `extensions/v1beta1` endpoints are, with bare kubectl, inheriting the wrong defaults, I think we should open a bug report upstream.

I don't think the behaviour change on our side is the worst thing, but it's a bit of an unintuitive way to provide the option of a hard timeout. This is the behaviour I would prefer:
1 - If the resource has the `kubernetes-deploy.shopify.io/timeout-override` annotation on it, use that. This is currently supported for all resources _except_ deployment.
2 - If the live resource has a PDS, use that (can still fall back on local template when live is unavailable for some reason, but live should be the authority).
3 - If none of the above apply, use the class's hard timeout.

Given the k8s bug, the test for (3) would need to be version-conditional in our CI.

dturn: This looks like a bug to me: https://gist.github.com/dturn/c250ec22837765b3df9cd07e657ea55d

> Given the k8s bug, the test for (3) would need to be version-conditional in our CI.

I'm a bit confused about this, I think you're suggesting reverting the code changes and modifying the tests to accept the bug.

KnVerey: That field is _present_ in both versions, and changing the access version should not change the value. What we need to prove is that the `kubectl --context minikube -n test create -f test/fixtures/invalid/bad_probe.yml` used the extensions/v1beta1 endpoint and got a default value anyway. If it didn't use the extensions/v1beta1 endpoint, the bug is in kubectl not the defaulting.

> I'm a bit confused about this, I think you're suggesting reverting the code changes and modifying the tests to accept the bug.

Yes, accepting that in 1.10 it is not possible to test a PDS-less deployment, because it isn't possible to create one (assuming we prove that). IMO this is the right thing to do because if the resource has a progress deadline, we should use it, and the live resource (when available) should be the source of truth. The correct way to make a hard deadline feature available for deployments is to make the annotation we already offer work for them, not ignore defaulted PDS values.

dturn: After a reset:
```
kubectl -v8 --context minikube -n test create -f test/fixtures/invalid/bad_probe.yml
...
I0711 13:56:39.759594    5689 request.go:991] Request Body: {"apiVersion":"extensions/v1beta1","kind":"Deployment","metadata":{"annotations":{"kubernetes-deploy.shopify.io/timeout-override":"25s"},"name":"bad-probe","namespace":"test"},"spec":{"replicas":1,"template":{"metadata":{"labels":{"app":"crash-app","name":"bad-probe"}},"spec":{"containers":[{"command":["tail","-f","/dev/null"],"image":"busybox","imagePullPolicy":"IfNotPresent","name":"http-probe","ports":[{"containerPort":80,"name":"http"}],"readinessProbe":{"failureThreshold":1,"httpGet":{"path":"/bad/ping/path","port":80},"initialDelaySeconds":0,"periodSeconds":1,"timeoutSeconds":1}},{"command":["tail","-f","/dev/null"],"image":"busybox","imagePullPolicy":"IfNotPresent","name":"exec-probe","readinessProbe":{"exec":{"command":["test","0","-eq","1"]},"failureThreshold":1,"initialDelaySeconds":0,"periodSeconds":1,"timeoutSeconds":1}},{"command":["tail","-f","/dev/null"],"image":"busybox","imagePullPolicy":"IfNotPresent","name":"sidecar","readinessProbe":{"exec":{"command":["test","0","-eq","0"]},"failureThreshold":100,"initialDelaySeconds":2,"periodSeconds":2,"successThreshold":1}}]}}}}
I0711 13:56:39.759679    5689 round_trippers.go:383] POST https://192.168.64.66:8443/apis/extensions/v1beta1/namespaces/test/deployments
I0711 13:56:39.759690    5689 round_trippers.go:390] Request Headers:
I0711 13:56:39.759696    5689 round_trippers.go:393]     Accept: application/json
I0711 13:56:39.759701    5689 round_trippers.go:393]     Content-Type: application/json
I0711 13:56:39.759707    5689 round_trippers.go:393]     User-Agent: kubectl/v1.7.8 (darwin/amd64) kubernetes/bc6162c
I0711 13:56:39.768511    5689 round_trippers.go:408] Response Status: 201 Created in 8 milliseconds
I0711 13:56:39.768531    5689 round_trippers.go:411] Response Headers:
I0711 13:56:39.768541    5689 round_trippers.go:414]     Content-Type: application/json
I0711 13:56:39.768549    5689 round_trippers.go:414]     Content-Length: 2080
I0711 13:56:39.768556    5689 round_trippers.go:414]     Date: Wed, 11 Jul 2018 20:56:39 GMT
I0711 13:56:39.768600    5689 request.go:991] Response Body: {"kind":"Deployment","apiVersion":"extensions/v1beta1","metadata":{"name":"bad-probe","namespace":"test","selfLink":"/apis/extensions/v1beta1/namespaces/test/deployments/bad-probe","uid":"e8693345-854c-11e8-a514-5278130fe47d","resourceVersion":"560","generation":1,"creationTimestamp":"2018-07-11T20:56:39Z","labels":{"app":"crash-app","name":"bad-probe"},"annotations":{"kubernetes-deploy.shopify.io/timeout-override":"25s"}},"spec":{"replicas":1,"selector":{"matchLabels":{"app":"crash-app","name":"bad-probe"}},"template":{"metadata":{"creationTimestamp":null,"labels":{"app":"crash-app","name":"bad-probe"}},"spec":{"containers":[{"name":"http-probe","image":"busybox","command":["tail","-f","/dev/null"],"ports":[{"name":"http","containerPort":80,"protocol":"TCP"}],"resources":{},"readinessProbe":{"httpGet":{"path":"/bad/ping/path","port":80,"scheme":"HTTP"},"timeoutSeconds":1,"periodSeconds":1,"successThreshold":1,"failureThreshold":1},"terminationMessagePath":"/dev/termination-log","terminationMessagePolicy":"File","imagePullPolicy":"IfNotPresent"},{"name":"exec-probe","image":"busybox","command":["tail","-f","/dev/null"],"resources":{},"readinessProbe":{"exec":{"command":["test","0","-eq","1"]},"timeoutSeconds":1,"periodSeconds":1,"successThreshold":1,"failureThreshold":1},"terminationMessagePath":"/dev/termination-log","terminationMessagePolicy":"File","imagePullPolicy":"IfNotPresent"},{"name":"sidecar","image":"busybox","command":["tail","-f","/dev/null"],"resources":{},"readinessProbe":{"exec":{"command":["test","0","-eq","0"]},"initialDelaySeconds":2,"timeoutSeconds":1,"periodSeconds":2,"successThreshold":1,"failureThreshold":100},"terminationMessagePath":"/dev/termination-log","terminationMessagePolicy":"File","imagePullPolicy":"IfNotPresent"}],"restartPolicy":"Always","terminationGracePeriodSeconds":30,"dnsPolicy":"ClusterFirst","securityContext":{},"schedulerName":"default-scheduler"}},"strategy":{"type":"RollingUpdate","rollingUpdate":{"maxUnavailable":1,"maxSurge":1}},"revisionHistoryLimit":10,"progressDeadlineSeconds":600},"status":{}}
...
```

KnVerey: 👍 that looks like good evidence. Can you open the issue upstream?

dturn: Issue created https://github.com/kubernetes/kubernetes/issues/66135

dturn: @KnVerey 

> 1 - If the resource has the kubernetes-deploy.shopify.io/timeout-override annotation on it, use that. This is currently supported for all resources except deployment.

I agree we should add this, but I'd like to keep the scope of this PR to be just whats needed to make CI work with 1.10

KnVerey: SGTM. We can just open an issue.

test/integration/kubernetes_deploy_test.rb
```diff
@@ -543,17 +543,28 @@ def test_deploy_result_logging_for_mixed_result_deploy
       %r{ConfigMap/test\s+Available}
     ], in_order: true)
 
+    if KUBE_SERVER_VERSION < Gem::Version.new("1.10.0")
+      start_bad_probe_logs = [
+        %r{Deployment/bad-probe: TIMED OUT \(timeout: \d+s\)},
+        "Timeout reason: hard deadline for Deployment"
+      ]
+      end_bad_probe_logs = [%r{Unhealthy: Readiness probe failed: .* \(8 events \)}] # event
```
dturn: I think its ok to ignore Policial here since we've been using the `%r{}` syntax for regex and I don't think we want to start using the `//` syntax.

lib/kubernetes-deploy/kubernetes_resource/deployment.rb
dturn: Issue created https://github.com/Shopify/kubernetes-deploy/issues/315

dturn:
CI is passing, can I get 👀 again

dev.yml
```diff
@@ -8,7 +8,7 @@ up:
   - custom:
       name: Minikube Cluster
       met?: test $(minikube status | grep Running | wc -l) -eq 2 && $(minikube status | grep -q 'Correctly Configured')
-      meet: minikube start --kubernetes-version=v1.9.4 --vm-driver=hyperkit
+      meet: minikube start --kubernetes-version=v1.10.4 --vm-driver=hyperkit
```
KnVerey: 1.10.0 is still the latest available in minikube afaik

test/integration/kubernetes_deploy_test.rb
KnVerey: The rule it is enforcing is the "mixed" style if you look at the rubocop cache file. If we don't want that, we need to change our rubocop rules. But in general I'm inclined to stick with the standard style guide.

test/integration/kubernetes_deploy_test.rb
```diff
@@ -930,12 +941,18 @@ def test_resource_watcher_reports_failed_after_timeout
     end
     assert_deploy_failure(result)
 
+    bad_probe_timeout = if KUBE_SERVER_VERSION < Gem::Version.new("1.10.0")
+      "Deployment/bad-probe: TIMED OUT (timeout: 5s)"
+    else
+      "Deployment/bad-probe: GLOBAL WATCH TIMEOUT (20 seconds)"
```
KnVerey: In this version, `bad-probe` has no purpose since `missing-volumes` is already covering the global watch timeout.

test/integration/kubernetes_deploy_test.rb
```diff
@@ -534,7 +534,11 @@ def test_pruning_of_existing_managed_secrets_when_ejson_file_has_been_deleted
 
   def test_deploy_result_logging_for_mixed_result_deploy
     subset = ["bad_probe.yml", "init_crash.yml", "missing_volumes.yml", "config_map.yml"]
-    result = deploy_fixtures("invalid", subset: subset)
+    result = deploy_fixtures("invalid", subset: subset) do |f|
+      if KUBE_SERVER_VERSION >= Gem::Version.new("1.10.0")
```
KnVerey: Can you please add an inline comment by the first of these changes in each test, linking to the issue you opened?

PR #312, opened 2018-07-06
Pruneable CRs

dturn:
Follow-up to https://github.com/Shopify/kubernetes-deploy/pull/306 to support prunable CRDs.

Thoughts / questions:

- Should  `custom_resource_definitions` be a class method on CRD?
- `prunable?` checks the labels as well as the annotations because the annotation wasn't sticking between deploys


lib/kubernetes-deploy/deploy_task.rb
```diff
@@ -198,6 +202,20 @@ def run!(verify_result: true, allow_protected_ns: false, prune: true)
...
     private
 
+    def cluster_custom_resource_definitions
+      @_cluster_custom_resource_definitions ||= begin
+        @sync_mediator.get_all(CustomResourceDefinition.kind).map do |r_def|
+          crd = KubernetesResource.build(namespace: @namespace, context: @context, logger: @logger,
+            definition: r_def, statsd_tags: @namespace_tags)
+
+          unless KubernetesDeploy.const_defined?(crd.kind)
+            KubernetesDeploy.const_set crd.kind, Class.new(KubernetesDeploy::CustomResource)
```
dturn: Can we avoid making classes? This could cause problems when used as a library against two different clusters.

Also this functions should be moved to its own class.

test/fixtures/resource-discovery/widgets.yml
```diff
@@ -0,0 +1,17 @@
+---
+apiVersion: apiextensions.k8s.io/v1beta1
+kind: CustomResourceDefinition
+metadata:
+  name: widgets.api.foobar.com
+  annotations:
+    kubernetes-deploy.shopify.io/metadata:
+      prunable: "true"
```
dturn: Is this malformed? (Is that why I needed the label)

lib/kubernetes-deploy/deploy_task.rb
```diff
@@ -27,6 +27,8 @@
   bucket
   stateful_set
   cron_job
+  custom_resource_definition
+  custom_resource
```
stefanmb: In the initial PR (https://github.com/Shopify/kubernetes-deploy/pull/188/files) this static list was removed in favour of a DAG representation. Do we plan to reintroduce the graph based representation?

Some advantages of the DAG:

- Only expresses ordering between dependencies that actually matter (e.g. multiple solutions to topological sort are valid)
- Can express dependencies between custom resources, a list that considers all CRs of equal deploy priority cannot.

lib/kubernetes-deploy/resource_discovery.rb
```diff
@@ -0,0 +1,20 @@
...
+    def initialize(namespace:, context:, logger:, namespace_tags:)
+      @namespace = namespace
+      @context = context
+      @logger = logger
+      @namespace_tags = namespace_tags
+      @cr_mapping = {}
+    end
+
+    def crds(sync_mediator)
+      sync_mediator.get_all(CustomResourceDefinition.kind).map do |r_def|
```
stefanmb: Since we are not walking the entire API tree we can only express dependencies between resources that are explicitly modelled in KubernetesDeploy (i.e. there is a class defined for them in this project) or CRDs.

dturn:
>If it were up to me I would keep the DAG representation of the the deploy list, I think it's more conceptually correct and a better way to represent the problem than a linearly walking through the list. The list may be sufficient for our immediate purposes, but it seems like a gross simplification to only save a bit of code.

The goal of this PR is to be the minimal set of features to make pruning work. I think the dag is enough code to justify its own PR, if/when we decide to take it on.

lib/kubernetes-deploy/kubernetes_resource/custom_resource_definition.rb
```diff
@@ -0,0 +1,41 @@
...
+      version = @definition.dig("spec", "version")
+      "#{group}/#{version}/#{kind}"
+    end
+
+    def kind
+      @definition.dig("spec", "names", "kind")
+    end
+
+    def prunable?
+      json_annotation = @definition.dig("metadata", "annotations", "kubernetes-deploy.shopify.io/metadata")
```
KnVerey: What are your thoughts (@stefanmb too) on one big annotation vs. many smaller ones? Or an inbetween option that groups related items (e.g. predeploy/dependencies, success/failure/timeout)?

lib/kubernetes-deploy/resource_discovery.rb
```diff
@@ -0,0 +1,20 @@
+# frozen_string_literal: true
+
+module KubernetesDeploy
+  class ResourceDiscovery
+    def initialize(namespace:, context:, logger:, namespace_tags:)
+      @namespace = namespace
+      @context = context
+      @logger = logger
+      @namespace_tags = namespace_tags
+      @cr_mapping = {}
```
KnVerey: this seems unused

lib/kubernetes-deploy/resource_discovery.rb
```diff
@@ -0,0 +1,20 @@
...
+    end
+
+    def crds(sync_mediator)
+      sync_mediator.get_all(CustomResourceDefinition.kind).map do |r_def|
+        KubernetesResource.build(namespace: @namespace, context: @context, logger: @logger,
```
KnVerey: I'd just do `CustomResourceDefinition.new` directly. `build` is basically just a class lookup mechanism.

test/fixtures/hello-cloud/crd.yml
```diff
@@ -0,0 +1,13 @@
+apiVersion: apiextensions.k8s.io/v1beta1
+kind: CustomResourceDefinition
+metadata:
+  name: mails.stable.example.io
+spec:
+  group: stable.example.io
+  names:
+    kind: Mail
+    listKind: MailList
+    plural: mails
```
KnVerey: no need for the example to carry over our incorrect plural, lol 😄

test/fixtures/hello-cloud/crd.yml
```diff
@@ -0,0 +1,13 @@
+apiVersion: apiextensions.k8s.io/v1beta1
```
KnVerey: Adding this to hello-cloud (which strictly speaking is correct--I usually want all supported resources to be in this set) makes me want to build a real demo directory like we've been asked to, so that we don't _encourage_ this gem to be used for globals.

test/integration-serial/resource_discovery_test.rb
```diff
@@ -0,0 +1,22 @@
+# frozen_string_literal: true
+require 'test_helper'
+
+class ResourceDiscoveryTest < KubernetesDeploy::IntegrationTest
+  def test_non_prunable_crd
+    assert_deploy_success(deploy_fixtures("resource-discovery", subset: ["shredders.yml"]))
+    assert_deploy_success(deploy_fixtures("resource-discovery", subset: ["shredders_cr.yml"]))
```
KnVerey: Are these two separate deploys because CRDs aren't in the priority list?

test/integration/kubernetes_deploy_test.rb
```diff
@@ -1006,4 +1007,19 @@ def test_raise_on_yaml_missing_kind
       "  datapoint: value1"
     ], in_order: true)
   end
+
+  def test_crd_can_be_successful
+    assert_deploy_success(deploy_fixtures("hello-cloud", subset: ["crd.yml"]))
```
KnVerey: I don't think this test adds anything over the `full_hello` test you modified above.

lib/kubernetes-deploy/kubernetes_resource/custom_resource_definition.rb
dturn: I'd prefer to use smaller ones (e.x. `kubernetes-deploy.shopify.io/prunable` & kubernetes-deploy.shopify.io/predeploy`) that don't require us to store json strings.

test/integration-serial/resource_discovery_test.rb
dturn: No, the CRD has to be added to the cluster before the CR can pass validation. Since we do validation before we deploy even pre-deploy resources we need to deployments.

test/integration/kubernetes_deploy_test.rb
dturn: It doesn't, but this file is actually from the base branch (https://github.com/Shopify/kubernetes-deploy/pull/306/files#diff-43c27b92ee93912981b34f7d8b622714R1011) where it now is a legit test.  I'll fix all of the up once that PR merges.

test/integration-serial/resource_discovery_test.rb
KnVerey: Ah, that makes sense.

dturn:
This is ready for 👀 again.

test/fixtures/resource-discovery/widgets_cr.yml
```diff
@@ -0,0 +1,6 @@
+---
+apiVersion: "api.foobar.com/v1"
+kind: Widget
+metadata:
+  name: my-first-widget
+status: "ok"
```
KnVerey: Why include status? Once it is a proper subresource (1.11+), this will be ignored.

test/integration-serial/resource_discovery_test.rb
```diff
@@ -0,0 +1,24 @@
+# frozen_string_literal: true
+require 'test_helper'
+
+class ResourceDiscoveryTest < KubernetesDeploy::IntegrationTest
+  def test_non_prunable_crd
```
KnVerey: I'd recommend merging these two tests into a single test that: 
- deploys CRDs of both kinds and asserts that all is well
- deploys the rest of the fixture set (CRs of all kinds + some non-CR) and asserts that results look good
- deploys the non-CRs and asserts that the correct CRs were pruned/left alone

Technically that could easily cover the existing CRD success test as well. Since these tests are slow and can be flakey (inherently, because they hit a live minikube), I try to keep tests that are 100% overlapped by others to a minimum. I think we should move those CRD tests to this file regardless.

test/integration-serial/resource_discovery_test.rb
```diff
@@ -0,0 +1,24 @@
...
+    assert_deploy_success(deploy_fixtures("resource-discovery", subset: ["shredders.yml"]))
+    assert_deploy_success(deploy_fixtures("resource-discovery", subset: ["shredders_cr.yml"]))
+    # Deploy any other non-priority (predeployable) resource to trigger pruning
+    assert_deploy_success(deploy_fixtures("hello-cloud", subset: ["configmap-data.yml"]))
+
+    refute_logs_match("The following resources were pruned: widget \"my-first-shredder\"")
+  end
```
KnVerey: don't forget to clean up the CRDs. I'm thinking we should make test teardown do this generically by listing and deleting CRDs. That could just be in this suite if we move the other CRD tests over.

test/integration-serial/resource_discovery_test.rb
```diff
@@ -0,0 +1,24 @@
+# frozen_string_literal: true
+require 'test_helper'
+
+class ResourceDiscoveryTest < KubernetesDeploy::IntegrationTest
+  def test_non_prunable_crd
+    assert_deploy_success(deploy_fixtures("resource-discovery", subset: ["shredders.yml"]))
+    assert_deploy_success(deploy_fixtures("resource-discovery", subset: ["shredders_cr.yml"]))
+    # Deploy any other non-priority (predeployable) resource to trigger pruning
+    assert_deploy_success(deploy_fixtures("hello-cloud", subset: ["configmap-data.yml"]))
```
KnVerey: let's add a configmap (or whatever) to the fixture set we're using instead of deploying an unrelated one. The fixture sets are intended to be representing apps, so deploying a different one to the same ns is semantically odd.

test/fixtures/resource-discovery/widgets.yml
```diff
@@ -0,0 +1,14 @@
+---
+apiVersion: apiextensions.k8s.io/v1beta1
+kind: CustomResourceDefinition
+metadata:
+  name: widgets.api.foobar.com
+  annotations:
+    kubernetes-deploy.shopify.io/prunable: "true"
```
KnVerey: Should we have a third one that has the annotation, but set to "false"?

lib/kubernetes-deploy/resource_discovery.rb
```diff
@@ -0,0 +1,19 @@
...
+module KubernetesDeploy
+  class ResourceDiscovery
+    def initialize(namespace:, context:, logger:, namespace_tags:)
+      @namespace = namespace
+      @context = context
+      @logger = logger
+      @namespace_tags = namespace_tags
+    end
+
+    def crds(sync_mediator)
```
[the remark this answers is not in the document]
dturn: Are you worried about the sync mediator's cache? I'd assumed we wanted to discourage direct calls to kubeclient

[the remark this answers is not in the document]
KnVerey: Not really worried about the cache, since we clear that at the beginning of every sync cycle. We definitely want to discourage direct calls to kubeclient from resource classes--they should only be making calls during the sync cycle, and therefore should only need the sync mediator (and can all benefit from its cache if they do). But what we're doing here has nothing to do with the sync cycle and the mediator gives us no benefit. EjsonSecretProvisioner is an example of another piece of independent functionality that also takes care of its own calls.

test/fixtures/resource-discovery/widgets.yml
dturn: I think its overkill, we already test the case where its not pruned.

test/integration-serial/run_serial_test.rb
```diff
@@ -158,4 +144,28 @@ def test_crd_can_fail
...
+      " - CustomResourceDefinition/mail.stable.example.io",
+      "Phase 3: Deploying all resources",
+      "CustomResourceDefinition/mail.stable.example.io (timeout: 120s)",
+      %r{CustomResourceDefinition/mail.stable.example.io\s+Names accepted}
+    ])
+    assert_deploy_success(deploy_fixtures("crd", subset: %w(mail_cr.yml widgets_cr.yml configmap-data.yml)))
+    # Deploy any other non-priority (predeployable) resource to trigger pruning
+    assert_deploy_success(deploy_fixtures("crd", subset: %w(configmap-data.yml configmap2.yml)))
+
+    refute_logs_match("The following resources were pruned: widget \"my-first-mail\"")
```
KnVerey: wouldn't it say `mail "my-first-mail"` if it had been? This brings up that we should add an assertion that actually confirms the thing is still there by looking at the namespace.

KnVerey:
In case you missed it on my previous review, we'll need docs for this feature. That can be a follow-up if you prefer.

Issue #314, opened 2018-07-09
Bypass batched fetching for small numbers of resources


KnVerey:
# Bug report

Reported in https://github.com/Shopify/kubernetes-deploy/issues/313. The root cause there is that the namespace being deployed to has over 4000 pods, most of which are old evicted pods. Since we are unconditionally using the batched fetching strategy, our sync cycles fetch and initialize all 4000+ pods even though in the example given only a single one was needed.

**Expected behavior:** Polling is instant for small resource groups.

**Actual behavior:** Polling can take forever for small resource groups, if the cluster contains a large number of resources of that type.

**Version(s) affected:** Latest.

# Proposed solution

Wrap https://github.com/Shopify/kubernetes-deploy/blob/master/lib/kubernetes-deploy/sync_mediator.rb#L32-L37 in something along the lines of `if resources.length < LARGE_BATCH_THRESHOLD` and add/adjust tests.

cc @dwradcliffe @csfrancis (likely affects overall core deploy speed too) @Shopify/cloudx

dturn:
The PR has been merged. It still need to be released (waiting to ship 2 or 3 more PRS) then bumped in core.

PR #316, opened 2018-07-12
Only batch fetch when there is a sufficiently large number requested

dturn:
Polling can take a long time for small resource groups when the cluster contains a large number of resources of that type. Don't batch fetch when number of resources being tracked is small.

fixes: https://github.com/Shopify/kubernetes-deploy/issues/314

### DIFF
--- a/lib/kubernetes-deploy/sync_mediator.rb
+++ b/lib/kubernetes-deploy/sync_mediator.rb
@@ -1,6 +1,8 @@
 # frozen_string_literal: true
 module KubernetesDeploy
   class SyncMediator
+    LARGE_BATCH_THRESHOLD = 5
+
     def initialize(namespace:, context:, logger:)
       @namespace = namespace
       @context = context
@@ -29,11 +31,14 @@ module KubernetesDeploy
 
     def sync(resources)
       clear_cache
-      dependencies = resources.map(&:class).uniq.flat_map do |c|
-        c::SYNC_DEPENDENCIES if c.const_defined?('SYNC_DEPENDENCIES')
+
+      if resources.count < LARGE_BATCH_THRESHOLD
+        dependencies = resources.map(&:class).uniq.flat_map do |c|
+          c::SYNC_DEPENDENCIES if c.const_defined?('SYNC_DEPENDENCIES')
+        end
+        kinds = (resources.map(&:type) + dependencies).compact.uniq
+        kinds.each { |kind| fetch_by_kind(kind) }
       end
-      kinds = (resources.map(&:type) + dependencies).compact.uniq
-      kinds.each { |kind| fetch_by_kind(kind) }
 
       KubernetesDeploy::Concurrency.split_across_threads(resources) do |r|
         r.sync(dup)
--- a/lib/kubernetes-deploy/sync_mediator.rb
+++ b/lib/kubernetes-deploy/sync_mediator.rb
@@ -32,7 +32,7 @@ module KubernetesDeploy
     def sync(resources)
       clear_cache
 
-      if resources.count < LARGE_BATCH_THRESHOLD
+      if resources.count > LARGE_BATCH_THRESHOLD
         dependencies = resources.map(&:class).uniq.flat_map do |c|
           c::SYNC_DEPENDENCIES if c.const_defined?('SYNC_DEPENDENCIES')
         end
--- a/lib/kubernetes-deploy/sync_mediator.rb
+++ b/lib/kubernetes-deploy/sync_mediator.rb
@@ -1,7 +1,7 @@
 # frozen_string_literal: true
 module KubernetesDeploy
   class SyncMediator
-    LARGE_BATCH_THRESHOLD = 5
+    LARGE_BATCH_THRESHOLD = Concurrency::MAX_THREADS * 3
 
     def initialize(namespace:, context:, logger:)
       @namespace = namespace
--- a/test/unit/sync_mediator_test.rb
+++ b/test/unit/sync_mediator_test.rb
@@ -90,7 +90,8 @@ class SyncMediatorTest < KubernetesDeploy::TestCase
     all_cm = [@fake_cm.kubectl_response, @fake_cm2.kubectl_response]
     stub_kubectl_response('get', 'FakeConfigMap', *@params, resp: { "items" => all_cm }, times: 1)
     stub_kubectl_response('get', 'FakePod', *@params, resp: { "items" => [@fake_pod.kubectl_response] }, times: 1)
-    mediator.sync([@fake_cm, @fake_pod])
+    # Cache is only warmed if we batch fetch
+    mediator.sync([@fake_cm, @fake_pod] * KubernetesDeploy::SyncMediator::LARGE_BATCH_THRESHOLD)
 
     # These should use the warm cache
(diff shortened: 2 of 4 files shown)


dturn:
PR that added batch fetching https://github.com/Shopify/kubernetes-deploy/pull/251. Doesn't look like the tests had major changes.

When `LARGE_BATCH_THRESHOLD` is set to 5 all tests pass.  

When set to `-1`
- 5 unit tests fail with errors expected errors e.x.:
```
Minitest::Assertion: unexpected invocation: #<AnyInstance:KubernetesDeploy::Kubectl>.run("get", "BadCitizen", "foo", "-a", "--output=json")
unsatisfied expectations:
- expected exactly once, not yet invoked: #<AnyInstance:KubernetesDeploy::Kubectl>.run("get", "BadCitizen", "-a", "--output=json")
```

lib/kubernetes-deploy/sync_mediator.rb
```diff
@@ -1,6 +1,8 @@
 # frozen_string_literal: true
 module KubernetesDeploy
   class SyncMediator
+    LARGE_BATCH_THRESHOLD = 5
```
KnVerey: What makes you choose 5? I would expect a much higher number. We fetch 8-way parallel, so I'd be inclined to define the threshold in relation to how many batches of concurrent requests it will take... maybe 3? 

On the other hand, what do you think about applying the threshold per class rather than globally? E.g. maybe you have 15 things, but they're each of a different kind, so the batched fetching doesn't actually help. I'm thinking it would be more accurate in terms of using the batch strategy when it actually helps, but it would need to take sync dependencies into account and definitely be more complicated.

test/unit/sync_mediator_test.rb
```diff
@@ -118,18 +121,31 @@ def test_calling_instance_sync_does_not_allow_instances_to_affect_the_global_cac
...
-    all_cm = [@fake_cm.kubectl_response, @fake_cm2.kubectl_response]
-    stub_kubectl_response('get', 'FakeConfigMap', *@params, resp: { "items" => all_cm }, times: 1)
+    test_resources = [@fake_pod, @fake_cm, @fake_deployment]
+    test_resources.each { |r| r.expects(:sync).once }
+    mediator.sync(test_resources)
+  end
+
+  def test_sync_does_not_batch_with_few_resources
+    stub_kubectl_response('get', 'FakeConfigMap', @fake_cm.name, *@params,
+      resp: { "items" => [@fake_cm.kubectl_response] }, times: 0)
```
KnVerey: `times: 0` seems to still expect a request, but not return a valid response from it... why? I'd think we'd want a `.expects.never` here on a list call (the one here is an instance call for `@fake_cm`), and have an actual stub on the instance API call we expect rather than the `expects` below, to show what happened to the cache (it wasn't populated).

test/unit/sync_mediator_test.rb
```diff
@@ -118,18 +121,31 @@ def test_calling_instance_sync_does_not_allow_instances_to_affect_the_global_cac
...
+
+    test_resources = [@fake_cm]
+    test_resources.each { |r| r.expects(:sync).once }
+    mediator.sync(test_resources)
+  end
+
+  def test_sync_does_batch_with_enough_resources
```
KnVerey: What value does this add beyond `test_sync_caches_the_types_of_the_dependencies_of_the_instances_given` and `test_sync_caches_the_types_of_the_instances_given` (which should probably have their names amended)? Because this test is stubbing the instance syncs, it isn't actually proving anything about the cache, just that the list api call was made (which those existing tests also do).

test/unit/sync_mediator_test.rb
```diff
@@ -118,18 +121,31 @@ def test_calling_instance_sync_does_not_allow_instances_to_affect_the_global_cac
...
-      resp: { "items" => [@fake_deployment.kubectl_response] }, times: 1)
-    stub_kubectl_response('get', 'FakePod', *@params, resp: { "items" => [@fake_pod.kubectl_response] }, times: 1)
-    all_cm = [@fake_cm.kubectl_response, @fake_cm2.kubectl_response]
-    stub_kubectl_response('get', 'FakeConfigMap', *@params, resp: { "items" => all_cm }, times: 1)
+    test_resources = [@fake_pod, @fake_cm, @fake_deployment]
+    test_resources.each { |r| r.expects(:sync).once }
+    mediator.sync(test_resources)
+  end
+
+  def test_sync_does_not_batch_with_few_resources
```
KnVerey: nit: I know I talked about this in terms of batching, but the real behaviour difference might be better described in terms of whether or not sync warms the type cache

dturn: That makes a lot more sense.

lib/kubernetes-deploy/sync_mediator.rb
dturn: 5 was a number out of a hat, I didn't want to get bogged down with this number before seeing how the code looked. I think 3 x parallelism makes sense. 

I agree we could get a lot more sophisticated here, but it seems like you're a small app who wont batch or your giant and the extra logic wont make a difference in most cases.

test/unit/sync_mediator_test.rb
```diff
@@ -118,18 +121,21 @@ def test_calling_instance_sync_does_not_allow_instances_to_affect_the_global_cac
...
-    stub_kubectl_response('get', 'FakePod', *@params, resp: { "items" => [@fake_pod.kubectl_response] }, times: 1)
-    all_cm = [@fake_cm.kubectl_response, @fake_cm2.kubectl_response]
-    stub_kubectl_response('get', 'FakeConfigMap', *@params, resp: { "items" => all_cm }, times: 1)
+    test_resources = [@fake_pod, @fake_cm, @fake_deployment]
+    test_resources.each { |r| r.expects(:sync).once }
+    mediator.sync(test_resources)
+  end
+
+  def test_sync_does_not_warm_cache_with_few_resources
+    KubernetesDeploy::Kubectl.any_instance.expects(:run).with('get', 'FakeConfigMap', @fake_cm.name, *@params).never
```
KnVerey: I guess it doesn't really matter because an exception will be raised if either request is attempted, but technically the cache warming would be requesting the list, not the instance.

test/unit/sync_mediator_test.rb
```diff
@@ -118,18 +121,21 @@ def test_calling_instance_sync_does_not_allow_instances_to_affect_the_global_cac
...
 
-    test_resources = [@fake_pod, @fake_cm, @fake_cm2, @fake_deployment]
+    test_resources = [@fake_cm]
```
KnVerey: Let's make this test the relevant edge by multiplying it by `KubernetesDeploy::SyncMediator::LARGE_BATCH_THRESHOLD`

test/unit/sync_mediator_test.rb
```diff
@@ -118,18 +121,21 @@ def test_calling_instance_sync_does_not_allow_instances_to_affect_the_global_cac
...
     test_resources.each { |r| r.expects(:sync).once }
     mediator.sync(test_resources)
```
KnVerey: For additional proof, we could add a request stub after this and do `mediator.get_instance('FakeConfigMap', @fake_cm.name)` to show the cache is not warm (kinda the opposite of what the cache-is-warm tests above do).

PR #317, opened 2018-07-16
Don't consider pod preempting a failure

DazWorrall:
Similar to #293, pod/node preempting is a recoverable state (the desired pod count will reduce once the node has been reaped), so isn't a valid condition for fast failure.

### DIFF
--- a/test/unit/kubernetes-deploy/kubernetes_resource/pod_test.rb
+++ b/test/unit/kubernetes-deploy/kubernetes_resource/pod_test.rb
@@ -170,6 +170,21 @@ class PodTest < KubernetesDeploy::TestCase
     assert_nil pod.failure_message
   end
 
+  def test_deploy_failed_is_false_for_preempting
+    container_state = pod_spec.merge(
+      "status" => {
+        "message" => "Preempted in order to admit critical pod",
+        "phase" => "Failed",
+        "reason" => "Preempting",
+        "startTime" => "2018-04-13T22:43:23Z"
+      }
+    )
+    pod = build_synced_pod(container_state)
+
+    refute_predicate pod, :deploy_failed?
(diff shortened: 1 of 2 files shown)


test/unit/kubernetes-deploy/kubernetes_resource/pod_test.rb
```diff
@@ -170,6 +170,21 @@ def test_deploy_failed_is_false_for_evicted
     assert_nil pod.failure_message
   end
 
+  def test_deploy_failed_is_false_for_preempting
+    container_state = pod_spec.merge(
+      "status" => {
```
dturn: Are you confident that this is what the status will look like (or at least the reason). Otherwise this PR wont work.

KnVerey: Looks like it gets set here 👍 

https://github.com/kubernetes/kubernetes/blob/817869e78af9d3250a36fce845336535dd29d0e4/pkg/kubelet/preemption/preemption.go#L106

DazWorrall: Thanks both!

dturn:
>When I run into this case it looks like progressDeadlineSeconds is not respected, kubernetes-deploy will watch and wait for the deployment of affected resources forever.
Is this intentional?

I suspect that kubernetes thinks progress is being made, progress to the controller is defined rather broadly, so it never actually hits `progressDeadlineSeconds` without making progress. We added the 
`--max-watch-seconds` so that the deploy will fail after a configurable amount of wall clock time for situations like this. If you think this is a bug and not something subtle with kuberentes please open a bug, ideally with the output of kubernetes-deploy and the result of `kubectl -o json` for the deployment/resource.

PR #318, opened 2018-07-16
[Caching-Queuing] Rename redis resources

tskorupa:
This commit follows the work performed in https://github.com/Shopify/cloudbuddies/pull/1115.

The redis resources were renamed. The kubernetes deployment script specifically looked for a prefix 'redis-' in the name. Whereas we wish to name resources according to their given names.

The current commit first looks for resources using the new names, otherwise it default to the 'legacy' naming. Legacy naming can be deployed once we migrate all redis instances.

@Shopify/caching-queuing

### DIFF
--- a/lib/kubernetes-deploy/kubernetes_resource/redis.rb
+++ b/lib/kubernetes-deploy/kubernetes_resource/redis.rb
@@ -7,8 +7,12 @@ module KubernetesDeploy
     SYNC_DEPENDENCIES = %w(Deployment Service)
     def sync(mediator)
       super
-      @deployment = mediator.get_instance(Deployment.kind, "redis-#{redis_resource_uuid}")
-      @service = mediator.get_instance(Service.kind, "redis-#{redis_resource_uuid}")
+
+      @deployment = mediator.get_instance(Deployment.kind, name)
+      @deployment = mediator.get_instance(Deployment.kind, deprecated_name) if @deployment.empty?
+
+      @service = mediator.get_instance(Service.kind, name)
+      @service = mediator.get_instance(Service.kind, deprecated_name) if @service.empty?
     end
 
     def status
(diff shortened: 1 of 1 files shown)


lib/kubernetes-deploy/kubernetes_resource/redis.rb
```diff
@@ -7,8 +7,10 @@ class Redis < KubernetesResource
     SYNC_DEPENDENCIES = %w(Deployment Service)
     def sync(mediator)
       super
-      @deployment = mediator.get_instance(Deployment.kind, "redis-#{redis_resource_uuid}")
-      @service = mediator.get_instance(Service.kind, "redis-#{redis_resource_uuid}")
+      @deployment = mediator.get_instance(Deployment.kind, name) ||
+        mediator.get_instance(Deployment.kind, deprecated_name)
```
shivnagarajan: why not deprecated_name || new_name? I guess you want the new name to have precedence?

lib/kubernetes-deploy/kubernetes_resource/redis.rb
```diff
@@ -41,6 +43,15 @@ def service_ready?
       @service.dig("spec", "clusterIP").present?
     end
 
+    def name
+      return unless @instance_data.present?
```
KnVerey: Since the name couldn't possibly have changed between the live copy (`@instance_data`) and the in-repo copy (`@definition`), I'd suggest using `@definition` for this instead.

lib/kubernetes-deploy/kubernetes_resource/redis.rb
```diff
@@ -41,6 +43,15 @@ def service_ready?
       @service.dig("spec", "clusterIP").present?
     end
 
+    def name
+      return unless @definition.present?
```
KnVerey: doesn't really matter, but this guard is unnecessary. `@definition` is the loaded YAML from their deploy directory; if it weren't present (minimally having `kind: Redis`), no instance of this class would be created.

Issue #304, opened 2018-06-22
kubernetes-run should watch the pod & dump its logs


KnVerey:
IMO this should actually be the default behaviour, and we should provide a `--skip-wait` flag like `kubernetes-deploy` itself has to override it. The primary user of `kubernetes-run` internally at Shopify is ActiveFailover, which is implementing log tailing itself. Other apps are using it in Shipit tasks, where waiting and seeing logs also strikes me as better behaviour in the average case.

cc @dturn @klautcomputing @fw42

dwradcliffe:
I came across another need for this too, so I'm glad we're in agreement that this should be a feature. 😄
Right now I'm hacking it together by faking a full deploy, but that's really clumsy.

fightingtheboss:
I don't know much about how this works, but happy to contribute if I can help move this effort forward. 

It would make the Database Snapshot Service usable from a Shipit task rather than requiring users to run the rake task on a production web pod.

dturn:
This is in progress. You can follow along in: https://github.com/Shopify/kubernetes-deploy/pull/320

PR #320, opened 2018-07-20
K8s run watch pod

dturn:
Make `kubernetes-run` watch the pod and dump its logs by default. `--skip-wait` will restore current behavior.

43802014-7cd469e0-9a49-11e8-9ed3-81e8aeb63c55

screen shot 2018-08-16 at 12 41 03 pm

Fixes: https://github.com/Shopify/kubernetes-deploy/issues/304

cc: @Shopify/cloudx


test/fixtures/hello-cloud/template-runner.yml
```diff
@@ -12,6 +12,7 @@ template:
       name: runner
       app: hello-cloud
   spec:
+    restartPolicy: Never
```
dturn: I had to add this to get `test_run_with_verify_result` to pass. Otherwise the deploy failed with: 
```
INFO][2018-07-20 08:55:11 -0700]	---------------------------------Phase 1: Validating configuration----------------------------------
[INFO][2018-07-20 08:55:11 -0700]	
[INFO][2018-07-20 08:55:11 -0700]	----------------------------------Phase 2: Fetching task template-----------------------------------
[INFO][2018-07-20 08:55:11 -0700]	Fetching task runner pod template: 'hello-cloud-template-runner' in namespace: 'k8sdeploy-test-run-with-verify-result-ec88a252fd86931b'
[INFO][2018-07-20 08:55:11 -0700]	
[INFO][2018-07-20 08:55:11 -0700]	---------------------------Phase 3: Constructing final pod specification----------------------------
[INFO][2018-07-20 08:55:11 -0700]	Rendering template for task runner pod
[INFO][2018-07-20 08:55:11 -0700]	
[INFO][2018-07-20 08:55:11 -0700]	---------------------------------------Phase 4: Creating pod----------------------------------------
[INFO][2018-07-20 08:55:11 -0700]	Starting task runner pod: 'task-runner-b9df5e758ecfb97d'
[ERROR][2018-07-20 08:55:21 -0700]	Pod/task-runner-b9df5e758ecfb97d failed to deploy after 9.2s
[INFO][2018-07-20 08:55:21 -0700]	
[INFO][2018-07-20 08:55:21 -0700]	------------------------------------------Result: FAILURE-------------------------------------------
[FATAL][2018-07-20 08:55:21 -0700]	Failed to deploy 1 resource
[FATAL][2018-07-20 08:55:21 -0700]	
[FATAL][2018-07-20 08:55:21 -0700]	Pod/task-runner-b9df5e758ecfb97d: FAILED
[FATAL][2018-07-20 08:55:21 -0700]	The following containers encountered errors:
[FATAL][2018-07-20 08:55:21 -0700]	> task-runner: Crashing repeatedly (exit 0). See logs for more information.
[FATAL][2018-07-20 08:55:21 -0700]	
[FATAL][2018-07-20 08:55:21 -0700]	  - Final status: Running
[FATAL][2018-07-20 08:55:21 -0700]	  - Events (common success events excluded):
[FATAL][2018-07-20 08:55:21 -0700]	      [Pod/task-runner-b9df5e758ecfb97d]	SuccessfulMountVolume: MountVolume.SetUp succeeded for volume "default-token-dg6j9"  (1 events)
[FATAL][2018-07-20 08:55:21 -0700]	      [Pod/task-runner-b9df5e758ecfb97d]	BackOff: Back-off restarting failed container (2 events)
[FATAL][2018-07-20 08:55:21 -0700]	  - Logs from container 'task-runner' (last 250 lines shown):
[FATAL][2018-07-20 08:55:21 -0700]	      KUBERNETES-DEPLOY
```

test/integration/runner_task_test.rb
```diff
@@ -86,10 +132,11 @@ def test_run_with_env_vars
   private
 
   def valid_run_params
-    { task_template: 'hello-cloud-template-runner', entrypoint: ['/bin/bash'], args: ["echo", "'KUBERNETES-DEPLOY'"] }
+    { task_template: 'hello-cloud-template-runner', entrypoint: ['/bin/sh', '-c'], args: ["echo 'KUBERNETES-DEPLOY'"] }
```
dturn: I don't think the busybox container has bash based on the error messages I got.

dturn:
> Therefore, we always need to print the logs, regardless of the final status. Minimally we can dump them on finalization, but I'd love for us to actually stream them if that is doable.

I think it makes sense to stream the logs (or at least dump them every x seconds). I'm going to refactor this to not use `ResourceWatcher`. Its not buying us much since there is only 1 resource and I don't think it makes sense for the `ResourceWatcher` to know how to dump logs every x seconds.

dturn:
screen shot 2018-08-06 at 11 16 49 am


latest logging

dturn:
screen shot 2018-08-07 at 1 53 29 pm

screen shot 2018-08-07 at 1 53 41 pm

I think this is ready for 👀

dturn:
@Shopify/pod-failovers I know` ActiveFailover` uses `kuberentes-run` and wanted to ask for 👀 on this change. Let me know if you have any questions.

lib/kubernetes-deploy/kubernetes_resource/pod.rb
```diff
@@ -85,14 +86,14 @@ def failure_message
...
-    def fetch_logs(kubectl)
+    def fetch_logs(kubectl, since: nil)
       return {} unless exists? && @containers.present?
       @containers.each_with_object({}) do |container, container_logs|
         cmd = [
           "logs",
           @name,
           "--container=#{container.name}",
-          "--since-time=#{@deploy_started_at.to_datetime.rfc3339}",
+          "--since-time=#{since || @deploy_started_at.to_datetime.rfc3339}",
```
fw42: Seems like you might want `(since || @deploy_started_at).to_datetime.rfc3339` instead?

lib/kubernetes-deploy/runner_task.rb
```diff
@@ -15,43 +15,115 @@ def initialize(task_template, namespace, context)
...
-      @logger.phase_heading("Validating configuration")
+      @logger.phase_heading("Initializing deploy")
       validate_configuration(task_template, args)
       if kubectl.server_version < Gem::Version.new(MIN_KUBE_VERSION)
         @logger.warn(KubernetesDeploy::Errors.server_version_warning(kubectl.server_version))
       end
-      @logger.phase_heading("Fetching task template")
       raw_template = get_template(task_template)
 
-      @logger.phase_heading("Constructing final pod specification")
```
fw42: Why remove all these loggings?

lib/kubernetes-deploy/runner_task.rb
```diff
@@ -15,43 +15,115 @@ def initialize(task_template, namespace, context)
...
-      @logger.fatal "#{error.class}: #{error.message}"
+      if error.message != error.class.to_s
+        @logger.summary.add_action(error.message)
+        @logger.print_summary(:failure)
+      end
       false
     end
 
-    def run!(task_template:, entrypoint:, args:, env_vars: [])
+    def run!(task_template:, entrypoint:, args:, env_vars: [], verify_result: true)
```
fw42: This method is pretty long now and does things on different levels of abstractions. Can you clean this up?

test/unit/kubernetes-deploy/runner_task_test.rb
```diff
@@ -16,9 +16,9 @@ def test_run_with_invalid_configuration
     )
 
     refute task_runner.run(task_template: nil, entrypoint: nil, args: nil)
-    assert_logs_match(/Task template name can't be nil/)
-    assert_logs_match(/Namespace can't be empty/)
-    assert_logs_match(/Args can't be nil/)
+    assert_logs_match(/Task template name can't be nil/i)
+    assert_logs_match(/Namespace can't be empty/i)
+    assert_logs_match(/Args can't be nil/i)
```
fw42: why are all those `/i` needed now?

lib/kubernetes-deploy/runner_task.rb
```diff
@@ -15,43 +15,115 @@ def initialize(task_template, namespace, context)
...
+        if (sleep_duration = sync_interval - (Time.now.utc - last_loop_time)) > 0
+          sleep(sleep_duration)
+        end
+        last_loop_time = Time.now.utc
+
+        return if deploy_ended?(pod, deploy_started_at)
+      end
+    end
+
+    def deploy_ended?(pod, deploy_started_at)
```
fw42: I dislike that this predicate method (`?`) has so many side-effects (it logs and it might raise an exception). That's unconventional. In Ruby, predicate-methods should either return `true` or `false` but not do anything else.

lib/kubernetes-deploy/runner_task.rb
```diff
@@ -15,43 +15,115 @@ def initialize(task_template, namespace, context)
...
+            @logger.info("\t...")
+          end
+        end
+
+        if (sleep_duration = sync_interval - (Time.now.utc - last_loop_time)) > 0
+          sleep(sleep_duration)
+        end
+        last_loop_time = Time.now.utc
+
+        return if deploy_ended?(pod, deploy_started_at)
```
fw42: What's the difference between `deploy_started_at` and `pod.deploy_started_at`? Why do we need both? Can this just say `deploy_ended?(pod)` instead?

lib/kubernetes-deploy/runner_task.rb
```diff
@@ -15,43 +15,115 @@ def initialize(task_template, namespace, context)
...
+          Result verification is disabled for this task.
+          This means the desired pod was successfully created, but the runner did not make sure it actually succeeded.
+        MSG
+        @logger.summary.add_paragraph(ColorizedString.new(warning).yellow)
+      end
     end
 
     private
 
+    def watch_and_report(pod, deploy_started_at, sync_mediator)
```
fw42: Same here, the `pod` already knows it's own `deploy_started_at`, so the additional `deploy_started_at` parameter seems unnecessary.

lib/kubernetes-deploy/runner_task.rb
```diff
@@ -15,43 +15,115 @@ def initialize(task_template, namespace, context)
...
+      sync_interval = 5.seconds
+      last_loop_time = last_log_time = deploy_started_at
+      @logger.info("Logs from #{pod.id}:")
+
+      loop do
+        pod.sync(sync_mediator)
+        logs = pod.fetch_logs(kubectl, since: last_log_time.to_datetime.rfc3339)
```
fw42: nitpick but it feels weird to me that the caller is responsible for the `to_datetime.rfc3339` stuff here. Shouldn't the `fetch_logs` method do that itself?

lib/kubernetes-deploy/runner_task.rb
fw42: I feel like `deploy_ended?` should be a method on `Pod`, not a method here. And it shouldn't do anything related to timeouts but rather that should be a separate method (`deploy_timed_out?` or `raise_if_deploy_timed_out`).

lib/kubernetes-deploy/runner_task.rb
```diff
@@ -89,7 +161,7 @@ def get_template(template_name)
       if error.error_code == 404
         raise TaskTemplateMissingError.new(template_name, @namespace, @context)
       else
-        raise
+        raise FatalDeploymentError, "Error communication with the API server"
```
fw42: `s/communication/communicating/`?

test/integration/runner_task_test.rb
```diff
@@ -7,18 +7,83 @@ def test_run_works
     deploy_fixtures("hello-cloud", subset: ["template-runner.yml"])
 
     task_runner = build_task_runner
-    assert task_runner.run(**valid_run_params)
+    refute task_runner.run(**valid_run_params)
```
fw42: why this change?

lib/kubernetes-deploy/runner_task.rb
```diff
@@ -15,43 +15,115 @@ def initialize(task_template, namespace, context)
...
-      @logger.phase_heading("Constructing final pod specification")
       rendered_template = build_pod_template(raw_template, entrypoint, args, env_vars)
-
-      validate_pod_spec(rendered_template)
-
+      validate_restart_policy(rendered_template, verify_result)
+      pod = Pod.new(namespace: @namespace, context: @context, logger: @logger, log_on_success: false,
+                    definition: rendered_template.to_hash.deep_stringify_keys, statsd_tags: [])
+      pod.validate_definition(kubectl)
+      @logger.info("Configuration valid")
```
fw42: super nitpick but I feel like this is unnecessary. I generally dislike "nothing is broken" logging. It tends to add a lot of noise when done in many places and suddenly you have 150 lines of logging that all basically just say "things went ok". If the configuration wasn't valid then the user would see an error, right? So I think the absence of error logs is enough indication of the fact that the configuration was valid.

dturn:
@fw42 Thanks for looking at this!

This PR ended up doing at least three things: improve logging, verifying the running pod completed & tailing the logs of the running pod. I think breaking each of these into its own PR will make something easier to review. I'm optimistic it will also help identify what responsibilities should be pulled into their own class. 

We're aware there is a level of abstraction missing w.r.t. states (failed, succeeded, timed-out). This project just hasn't had the resources to tackle it yet. 

One final question, does the goal of the PR make sense / seem valuable to you?

fw42:
> One final question, does the goal of the PR make sense / seem valuable to you?

I don't feel super strongly, but I'm leaning towards agreeing that the new default behaviour (tail the logs unless we tell it not to) is what I would expect from a tool like this.

test/unit/kubernetes-deploy/runner_task_test.rb
dturn: We changed logging which caused capitalization changes. I originally thought the diff would be easier to understand with `/i` but will flip it to just match the right capitalization

test/integration/runner_task_test.rb
dturn: Because the `run`method didn't actually verify success, it turns these tests didn't. I've fixed the tests so that they do actually run successfully

lib/kubernetes-deploy/runner_task.rb
dturn: `pod.deploy_started_at` wasn't exposed, I don't see any harm in exposing it, will def make this chunk of code better.

lib/kubernetes-deploy/runner_task.rb
dturn: `watch_and_report` has been refactored into its own class, also fixed up this method since it lives there now. `deploy_ended?` we know we have an issue with this and there is a seperate PR investigating better state tracking.

lib/kubernetes-deploy/kubernetes_resource/pod.rb
```diff
@@ -33,7 +34,13 @@ def sync(mediator)
         @containers.each(&:reset_status)
       end
 
-      display_logs(mediator) if unmanaged? && deploy_succeeded?
+      if unmanaged?
```
dturn: I think this block is ok for now. But I think a future refactor should have the resource watcher be responsible for triggering the logging.

dturn:
Ready for 👀 again

test/integration/runner_task_test.rb
```diff
@@ -14,22 +14,104 @@ def test_run_works
...
+      "start", # From pod logs
+      "finish",
+      "Result: SUCCESS",
+      "Successfully deployed 1 resource",
+    ])
+    pods = kubeclient.get_pods(namespace: @namespace)
+    assert_equal 1, pods.length, "Expected 1 pod to exist, found #{pods.length}"
+  end
+
+  def test_run_doesnt_miss_logs_across_pollings
```
dturn: So we don't miss logs, but due to the timestamp granularity we do double log:
```
...
	4999
[INFO][2018-09-18 09:37:47 -0700]		4806
	4807
...
```

lib/kubernetes-deploy/kubernetes_resource/pod.rb
```diff
@@ -85,16 +92,17 @@ def failure_message
...
-    def fetch_logs(kubectl)
+    def fetch_logs(kubectl, since: nil, timestamps: false)
       return {} unless exists? && @containers.present?
       @containers.each_with_object({}) do |container, container_logs|
         cmd = [
           "logs",
           @name,
           "--container=#{container.name}",
-          "--since-time=#{@deploy_started_at.to_datetime.rfc3339}",
+          "--since-time=#{(since || @deploy_started_at).to_datetime.rfc3339(9)}",
```
dturn: While the server accepts fractional seconds, its still gives a .2-.3 seconds worth of overlap so I manually filter on line 156.

test/integration/runner_task_test.rb
dturn: This is now fixed.

lib/kubernetes-deploy/pod_logger.rb
```diff
@@ -0,0 +1,68 @@
...
+            logs.each { |line| @logger.info("\t#{line}") }
+            @logger.blank_line
+          end
+        end
+
+        @already_displayed = true
+      end
+
+      def parse_date(dt)
+        DateTime.parse(dt)
```
dturn: We're using DateTime instead of `Time` because comparing fractions of seconds is important to prevent double logging and while `Time` stores them, it doesn't considering them in comparisons.

lib/kubernetes-deploy/kubernetes_resource/pod.rb
```diff
@@ -33,7 +36,7 @@ def sync(mediator)
         @containers.each(&:reset_status)
       end
 
-      display_logs(mediator) if unmanaged? && deploy_succeeded?
+      @pod_logger.log(mediator: mediator, pod: self) if unmanaged?
```
dturn: I'm passing the entire pod, since the PodLogger calls `fetch_logs` with different args depending on if we're streaming or not.

exe/kubernetes-run
```diff
@@ -23,10 +30,12 @@ logger = KubernetesDeploy::FormattedLogger.build(namespace, context)
 runner = KubernetesDeploy::RunnerTask.new(
   namespace: namespace,
   context: context,
-  logger: logger
+  logger: logger,
+  max_watch_seconds: max_watch_seconds
```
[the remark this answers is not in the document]
dturn: Here its to keep parity with [`kubernetes-deploy`](https://github.com/Shopify/kubernetes-deploy/blob/master/exe/kubernetes-deploy#L69). 

I added `max_watch_seconds` to new in `kubernetes-deploy` since is used like 3 levels deep in methods call from `run` and I didn't want to just pass it down when it could be an instance var, and it seemed like instance vars should be set by new. Can't speak to `verify_result `, but the split is probably an oversight / wasn't noticed till now.

exe/kubernetes-run
```diff
@@ -23,10 +30,12 @@ logger = KubernetesDeploy::FormattedLogger.build(namespace, context)
...
 )
 
 success = runner.run(
+  verify_result: !skip_wait,
```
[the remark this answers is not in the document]
dturn: Its like that for parity, which I think is important since we have a long term goal of trying of extracting the commonality between the 3 exe files.

exe/kubernetes-run
```diff
@@ -8,8 +8,15 @@ require 'optparse'
 template = "task-runner-template"
 entrypoint = nil
 env_vars = []
+skip_wait = nil
+max_watch_seconds = nil
 
 ARGV.options do |opts|
+  opts.on("--skip-wait", "Skip verification of pod success") { skip_wait = true }
+  opts.on("--max-watch-seconds=seconds",
+    "Timeout error is raised if the pod runs for longer than the specified number of seconds") do |t|
```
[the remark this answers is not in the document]
dturn: This wouldn't be hard to add as currently built, though it would be at the granularity of how often sync is called.

lib/kubernetes-deploy/kubernetes_resource/pod.rb
```diff
@@ -21,6 +24,7 @@ def initialize(namespace:, context:, definition:, logger:,
       @containers += definition["spec"].fetch("initContainers", []).map { |c| Container.new(c, init_container: true) }
       super(namespace: namespace, context: context, definition: definition,
             logger: logger, statsd_tags: statsd_tags)
+      @pod_logger = KubernetesDeploy::PodLogger.new(logger: logger, stream: stream_logs, pod_name: @name, pod_id: id)
```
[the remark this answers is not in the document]
dturn: `super` sets `@name`

exe/kubernetes-run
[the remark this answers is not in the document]
KnVerey: I don't think this a super big deal since an instance will rarely (if ever) get `run` multiple times in practice. I can't say I remember why I put each argument where I did. For now, I think consistency with the other commands is the most important thing, and if we want to change them in the future (a breaking change), we should do it for all three commands at once.

lib/kubernetes-deploy/pod_logger.rb
```diff
@@ -0,0 +1,89 @@
...
+module KubernetesDeploy
+  class PodLogger
+    def initialize(stream:, logger:, pod_name:, pod_id:)
+      @stream = stream
+      @logger = logger
+      @pod_name = pod_name
+      @pod_id = pod_id
+    end
+
+    def log(kubectl:, deploy_succeeded:, deploy_started_at:, unmanaged:, container_names:)
```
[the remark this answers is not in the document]
KnVerey: It is intentionally not available in `Pod#new` (where this is currently initialized) because it is really important that `kubectl` not be used outside of `sync` (which is given the kubectl instance) in the resource classes.

lib/kubernetes-deploy/pod_logger.rb
```diff
@@ -0,0 +1,89 @@
...
+      if unmanaged
```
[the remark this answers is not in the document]
KnVerey: The short version is that it means we're treating the pod in question as an inline-but-remote script runner (potentially one of many in the deploy). When it exits, we print its logs right away. A "managed" pod is one that was created by e.g. a deployment. We only dump logs from those pods if the parent's rollout fails and that pod is selected for debug information logging.

lib/kubernetes-deploy/pod_logger.rb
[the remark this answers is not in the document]
dturn: For performance reasons we only allow requests to the k8s api servers through the sync method. Adding this to the initializer would make it easy to violate that constraint.

lib/kubernetes-deploy/pod_logger.rb
```diff
@@ -0,0 +1,89 @@
...
+      logs = fetch_logs(kubectl, deploy_started_at: deploy_started_at, since: since, timestamps: true,
+        unmanaged: unmanaged, container_names: container_names)
+
+      logs.each do |_, log|
+        if log.present?
+          ts_logs = log.map do |line|
+            dt, message = line.split(" ", 2)
+            [parse_date(dt), message]
+          end
+          ts_logs.select! { |dt, _| dt.nil? || dt > since }
```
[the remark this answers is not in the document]
dturn: > can it return old logs even though we only asked for recent ones?

yes, about 200-300 milliseconds worth, which isn't a lot, but enough to be trouble. I'll add a comment explaining that.

lib/kubernetes-deploy/pod_logger.rb
```diff
@@ -0,0 +1,89 @@
...
+        container_logs[container_name] = out.split("\n")
+      end
+    end
+
+    private
+
+    def stream_logs(kubectl, deploy_started_at, unmanaged, container_names)
+      @logger.info("Logs from #{@pod_id}:") unless @last_log_fetch
+      since = @last_log_fetch || deploy_started_at
+      logs = fetch_logs(kubectl, deploy_started_at: deploy_started_at, since: since, timestamps: true,
```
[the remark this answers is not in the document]
dturn: `display_logs`: Is basically fetch everything and log them.

`stream_logs`: Is fetch them in batches and log them in a way that looks like its stream. @KnVerey was suggesting actually streaming from `kubectl`, though it has some corner cases and I think would make the timeout idea from https://github.com/Shopify/kubernetes-deploy/pull/320/files#r220508066 harder to implement.

I can re-name these methods to be more clear display_all_logs, stream_format_logs or something like that

lib/kubernetes-deploy/pod_logger.rb
```diff
@@ -0,0 +1,89 @@
...
+      @logger = logger
+      @pod_name = pod_name
+      @pod_id = pod_id
+    end
+
+    def log(kubectl:, deploy_succeeded:, deploy_started_at:, unmanaged:, container_names:)
+      if unmanaged
+        if @stream
+          stream_logs(kubectl, deploy_started_at, unmanaged, container_names)
+        elsif deploy_succeeded
```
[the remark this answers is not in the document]
KnVerey: > is that because if the deplay failed, there can't possibly be any logs? (is that true?)

No, not true. There are three times we might want to output pod logs:
- Inline dump: a script-runner ("unmanaged") pod terminates successfully and we're about to move on with the deploy, e.g. end of Phase 3 [here](https://shipit.shopify.io/shopify/cumulus-cat/production/deploys/624563)
- Inline stream: the new use case in kubernetes-run, where we're waiting on a single pod and want logs to appear asap
- Dump in summary: Either the pod itself or its parent (if it was the selected child) failed, and we want to dump debug information (which includes logs) in the deploy summary section, e.g. the summary section [here](https://shipit.shopify.io/shopify/cumulus-cat/configdemo/deploys/565200)

lib/kubernetes-deploy/pod_logger.rb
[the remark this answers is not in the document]
dturn: Basically that this `pod` was created directly by the user/tool as opposed to via a `Deployment`, `Job`, `DS`...

lib/kubernetes-deploy/runner_task.rb
```diff
@@ -15,44 +15,76 @@ def initialize(task_template, namespace, context)
...
-        @logger.warn(KubernetesDeploy::Errors.server_version_warning(kubectl.server_version))
+      pod = build_pod(task_template, entrypoint, args, env_vars, verify_result)
+
+      @logger.phase_heading("Running pod")
+      create_pod(pod)
+
+      if verify_result
+        sm = SyncMediator.new(namespace: @namespace, context: @context, logger: @logger)
+        rw = ResourceWatcher.new(resources: [pod], logger: @logger, timeout: @max_watch_seconds, sync_mediator: sm)
+        rw.run(delay_sync: 5.seconds, reminder_interval: 1.day) # never print reminders
```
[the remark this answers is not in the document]
KnVerey: On the other hand, even then the UX will be worse than a streaming approach.

lib/kubernetes-deploy/pod_logger.rb
```diff
@@ -0,0 +1,89 @@
...
+          @logger.blank_line
+        end
+      end
+
+      @already_displayed = true
+    end
+
+    def parse_date(dt)
+      DateTime.parse(dt)
+    rescue ArgumentError
```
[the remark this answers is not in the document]
dturn: The reason I added it was in 1.8 if the pod fails to be created k8s returns a log message that is basically "I can't open the log file" without a leading timestamp.

lib/kubernetes-deploy/pod_logger.rb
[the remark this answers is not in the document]
dturn: This is the existing behavior. There def can be logs, but the assumption is we want to see logs for unmanaged pods only on success. (In a normal deploy this would be the asset-uploader pod, but I can't speak to why its that way)

lib/kubernetes-deploy/pod_logger.rb
```diff
@@ -0,0 +1,89 @@
...
+          @logger.info("\t" + ts_logs.map(&:last).join("\n\t"))
+          @last_log_fetch = ts_logs.last.first if ts_logs.last&.first
+        else
+          @logger.info("\t...")
+        end
+      end
+    end
+
+    def display_logs(kubectl, deploy_started_at, unmanaged, container_names)
+      return if @already_displayed
```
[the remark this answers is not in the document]
dturn: Existing behavior, but its because this gets hit on every `sync` call.

lib/kubernetes-deploy/runner_task.rb
```diff
@@ -15,44 +15,76 @@ def initialize(task_template, namespace, context)
...
+      @logger.print_summary(:failure)
+      raise
+    end
+
+    private
 
-      @logger.phase_heading("Constructing final pod specification")
+    def create_pod(pod)
+      @logger.info("Starting task runner pod: '#{pod.name}'")
+      pod.deploy_started_at = Time.now.utc
```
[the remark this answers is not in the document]
KnVerey: This is now back to being used exclusively for client-side watch timeout purposes. It isn't compared to server-side times.

KnVerey:
Closing in favour of https://github.com/Shopify/kubernetes-deploy/pull/337. Description is WIP (will finish tomorrow) but existing reviewers probably have enough context already to dive into the code.

PR #323, opened 2018-08-03
Use k8s ci queue

stefanmb:
* Drop 1.7
* ~Add 1.11~
* Run all supported versions on k8s-ci
* Automatic retries (up to 2 tries per k8s version)

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -1,5 +1,19 @@
+shared: &shared
+  retry:
+     automatic:
+       - exit_status: "*"
+         limit: 1
 steps:
+  - name: 'Run Test Suite (:kubernetes: 1.11-latest)'
+    <<: *shared
+    command: bin/ci
+    agents:
+      queue: k8s-ci
+    env:
+      LOGGING_LEVEL: 4
+      KUBERNETES_VERSION: v1.11-latest
   - name: 'Run Test Suite (:kubernetes: 1.10-latest)'
+    <<: *shared
     command: bin/ci
     agents:
       queue: minikube-ci
@@ -7,6 +21,7 @@ steps:
       LOGGING_LEVEL: 4
       KUBERNETES_VERSION: v1.10-latest
   - name: 'Run Test Suite (:kubernetes: 1.9-latest)'
+    <<: *shared
     command: bin/ci
     agents:
       queue: minikube-ci
@@ -14,6 +29,7 @@ steps:
       LOGGING_LEVEL: 4
       KUBERNETES_VERSION: v1.9-latest
   - name: 'Run Test Suite (:kubernetes: 1.8-latest)'
+    <<: *shared
     command: bin/ci
     agents:
       queue: minikube-ci
--- a/lib/kubernetes-deploy/deploy_task.rb
+++ b/lib/kubernetes-deploy/deploy_task.rb
@@ -78,7 +78,6 @@ module KubernetesDeploy
         batch/v1/Job
         extensions/v1beta1/DaemonSet
         extensions/v1beta1/Deployment
-        apps/v1beta1/Deployment
         extensions/v1beta1/Ingress
         apps/v1beta1/StatefulSet
         autoscaling/v1/HorizontalPodAutoscaler
--- a/test/helpers/fixture_set.rb
+++ b/test/helpers/fixture_set.rb
@@ -29,8 +29,10 @@ module FixtureSetAssertions
 
     def refute_resource_exists(type, name, beta: false)
       client = beta ? v1beta1_kubeclient : kubeclient
-      client.public_send("get_#{type}", name, namespace) # 404s
-      flunk "#{type} #{name} unexpectedly existed"
+      res = client.public_send("get_#{type}", name, namespace)
+      if res.metadata.deletionTimestamp.blank?
+        flunk "#{type} #{name} unexpectedly existed and is not being deleted."
+      end
     rescue KubeException => e
       raise unless e.to_s.include?("not found")
     end
--- a/test/integration-serial/run_serial_test.rb
+++ b/test/integration-serial/run_serial_test.rb
@@ -172,9 +172,11 @@ class RunSerialTest < KubernetesDeploy::IntegrationTest
     assert_deploy_success(deploy_fixtures("crd", subset: %w(configmap-data.yml configmap2.yml)))
 
     assert_predicate build_kubectl.run("get", "mail.stable.example.io", "my-first-mail").last, :success?
-    refute_logs_match(/The following resources were pruned: mail(.stable.example.io)? "my-first-mail"/)
+    refute_logs_match(
+      /The following resources were pruned: #{prune_matcher("mail", "stable.example.io", "my-first-mail")}/
+    )
     assert_logs_match_all([
-      /The following resources were pruned: widget(.stable.example.io)? "my-first-widget"/,
+      /The following resources were pruned: #{prune_matcher("widget", "stable.example.io", "my-first-widget")}/,
       "Pruned 1 resource and successfully deployed 2 resource"
     ])
   ensure
--- a/test/integration/kubernetes_deploy_test.rb
+++ b/test/integration/kubernetes_deploy_test.rb
@@ -62,18 +62,17 @@ class KubernetesDeployTest < KubernetesDeploy::IntegrationTest
     hello_cloud.refute_configmap_data_exists
     hello_cloud.refute_unmanaged_pod_exists
     hello_cloud.refute_web_resources_exist
-
     expected_pruned = [
-      'configmap "hello-cloud-configmap-data"',
-      'pod "unmanaged-pod-',
-      'service "web"',
-      'resourcequota "resource-quotas"',
-      'deployment(\.extensions)? "web"',
-      'ingress(\.extensions)? "web"',
-      'daemonset(\.extensions)? "ds-app"',
-      'statefulset(\.apps)? "stateful-busybox"',
-      'job(\.batch)? "hello-job"',
-      'poddisruptionbudget(.policy)? "test"',
+      prune_matcher("configmap", "", "hello-cloud-configmap-data"),
+      prune_matcher("pod", "", "unmanaged-pod-"),
+      prune_matcher("service", "", "web"),
+      prune_matcher("resourcequota", "", "resource-quotas"),
+      prune_matcher("deployment", "extensions", "web"),
+      prune_matcher("ingress", "extensions", "web"),
+      prune_matcher("daemonset", "extensions", "ds-app"),
+      prune_matcher("statefulset", "apps", "stateful-busybox"),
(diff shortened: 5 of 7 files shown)


dturn:
We should 🔥 1.7 in this PR

dev.yml
```diff
@@ -8,7 +8,7 @@ up:
   - custom:
       name: Minikube Cluster
       met?: test $(minikube status | grep Running | wc -l) -eq 2 && $(minikube status | grep -q 'Correctly Configured')
-      meet: minikube start --kubernetes-version=v1.10.0 --vm-driver=hyperkit
+      meet: minikube start --kubernetes-version=v1.11.1 --vm-driver=hyperkit
```
dturn: Not officially supported by minikube yet, should we back it out?

test/integration/kubernetes_deploy_test.rb
```diff
@@ -619,7 +619,6 @@ def test_failure_logs_from_unmanaged_pod_appear_in_summary_section
 
     assert_logs_match_all([
       "Failed to deploy 1 priority resource",
-      "SuccessfulMountVolume", # from an event
```
dturn: 1.11 isn't generating any events on this failure.

dev.yml
stefanmb: Yeah, I don't see another option (while we're still using Minikube, sigh).

dturn:
Making CI pass is going to take some work. `test_pruning_works` is failing in 1.11, which at face value is rather scary.

stefanmb:
We can skip it for now, it was a bit of an omnibus PR move. But the sooner we're aware of these problems the better. Going forward all released (alpha/beta/stable) versions will be supported in CI so this should have less of a chance to creep up.

dturn:
I dropped support for 1.11 in hopes of not stalling this PR. Fixing up CI for 1.11 is going to be a larger undertaking than I originally expected.

dturn:
I had to run the 1.10 tests 4 times for them to pass. Each time was a different test or two that failed, that makes me rather nervous about merging this.

KnVerey:
Were there any commonalities in the parts of the tests that failed? For example, with a previous CI setup, we kept seeing tests fail on missing events because the event-throttling settings in the master configuration needed to be tuned.

dturn:
4 different tests failed across 3 runs failed runs (https://buildkite.com/shopify/kubernetes-deploy-gem/builds/964), not seeing much of a pattern, maybe deployments are taking longer and we're getting time outs instead of failures. But this will take some digging.

test_resource_watcher_reports_failed_after_timeout
test_deploy_result_logging_for_mixed_result_deploy

test_bad_container_image_on_deployment_pod_fails_quickly

test_roll_back_a_bad_deploy

test/integration-serial/run_serial_test.rb
```diff
@@ -160,9 +160,11 @@ def test_crd_pruning
     assert_deploy_success(deploy_fixtures("crd", subset: %w(configmap-data.yml configmap2.yml)))
 
     assert_predicate build_kubectl.run("get", "mail.stable.example.io", "my-first-mail").last, :success?
-    refute_logs_match(/The following resources were pruned: mail(.stable.example.io)? "my-first-mail"/)
+    mail_separator = '[ \/]"?my-first-mail"?'
+    widget_separator = '[ \/]"?my-first-widget"?'
+    refute_logs_match(/The following resources were pruned: mail(.stable.example.io)?#{mail_separator}/)
     assert_logs_match_all([
-      /The following resources were pruned: widget(.stable.example.io)? "my-first-widget"/,
+      /The following resources were pruned: widget(.stable.example.io)?#{widget_separator}/,
```
karanthukral: with k8s ci the res is printed as `widget.stable.example.io/my-first-widget` so I switched the regex but policial was yelling at me to use `%r{}` when I tried to use `[ \/]` but I couldn't figure it out so ended up using a variable for the separator section. If anyone is better at regex and knows how to please policial let me know

karanthukral:
Update: I kept k8s 1.8, 1.9 and 1.10 on minikube due to QPS issues and not being able to run the metrics server on 1.9. 

1.11 is now set to k8s-ci and future version should continue to use it. I seemed to have fixed the test failures which ranged from running the metrics server in k8s-ci for HPA deployments, to changing regex matches due to different formatting and updating the deleted check to check for the deletion timestamp instead of just a 404 error.

cc/ @dturn @KnVerey for thoughts on the changes

cc/ @stefanmb, incase you had any comments on the ci setup

lib/kubernetes-deploy/deploy_task.rb
```diff
@@ -78,7 +78,6 @@ def prune_whitelist
         batch/v1/Job
         extensions/v1beta1/DaemonSet
         extensions/v1beta1/Deployment
-        apps/v1beta1/Deployment
```
dturn: @KnVerey do you know why we have two versions of `Deployment` in the whitelist?

test/helpers/fixture_set.rb
```diff
@@ -29,8 +29,10 @@ def app_name
 
     def refute_resource_exists(type, name, beta: false)
       client = beta ? v1beta1_kubeclient : kubeclient
-      client.public_send("get_#{type}", name, namespace) # 404s
-      flunk "#{type} #{name} unexpectedly existed"
+      res = client.public_send("get_#{type}", name, namespace)
+      if res.metadata.deletionTimestamp.nil? || res.metadata.deletionTimestamp == ""
```
dturn: I think you can use `.blank?` in this project. Why was this change needed?

.buildkite/pipeline.yml
```diff
@@ -1,29 +1,38 @@
+shared: &shared
```
dturn: Don't forget to update `.buildkite/pipeline.nightly.yml `

test/integration-serial/run_serial_test.rb
dturn: `%r{The following resources were pruned: widget(.stable.example.io)?/"?my-first-widget"?}` didn't do what you wanted?

test/integration/kubernetes_deploy_test.rb
```diff
@@ -64,18 +64,18 @@ def test_pruning_works
...
+      'resourcequota[ \/]"?resource-quotas"?',
+      'deployment(\.extensions)?[ \/]"?web"?',
+      'ingress(\.extensions)?[ \/]"?web"?',
+      'daemonset(\.extensions)?[ \/]"?ds-app"?',
+      'statefulset(\.apps)?[ \/]"?stateful-busybox"?',
+      'job(\.batch)?[ \/]"?hello-job"?',
+      'poddisruptionbudget(.policy)?[ \/]"?test"?',
     ] # not necessarily listed in this order
-    expected_msgs = [/Pruned 10 resources and successfully deployed 6 resources/]
+    expected_msgs = [/Pruned 9 resources and successfully deployed 6 resources/]
```
dturn: Why did this drop back to 9?

test/helpers/fixture_set.rb
karanthukral: When testing for deletion, some objects on 1.11 were marked for deletion but weren't deleted when we ran this check. If the deletion timestamp is checked, k8s garbage collector will clean up the resources so it made sense to check this as well in the case the res wasn't deleted

lib/kubernetes-deploy/deploy_task.rb
KnVerey: Looks like I added it back in https://github.com/Shopify/kubernetes-deploy/pull/150, which was long enough ago that the reasoning could have been based on a misunderstanding of how group versions work. I can't think of a reason that it would make sense to have to include every supported GV. If you look [here](https://github.com/kubernetes/kubernetes/blob/release-1.11/pkg/kubectl/cmd/apply.go#L501) you can see the pruning code loops over all the group versions you pass it, and for each one [runs a List call](https://github.com/kubernetes/kubernetes/blob/release-1.11/pkg/kubectl/cmd/apply.go#L587) and deletes appropriate things from that list. My guess about what's happening here is that GC is struggling in our new CI (hence needing to modify the deletion checker to look at the timestamp rather than expecting a 404), so some of the deleted resources are still there when we do the list a second time through the extra GV.

TL;DR I'm quite confident this list should only include each kind once, and that we simply need to choose a GV that is available on all k8s versions we support.

test/integration/kubernetes_deploy_test.rb
KnVerey: Hm, I guess events aren't really the point of this test, so deleting is ok.

test/integration/kubernetes_deploy_test.rb
```diff
@@ -48,7 +48,7 @@ def test_service_account_predeployed_before_unmanaged_pod
     hello_cloud.assert_unmanaged_pod_statuses("Succeeded")
     assert_logs_match_all([
       %r{Successfully deployed in \d.\ds: ServiceAccount/build-robot},
-      %r{Successfully deployed in \d.\ds: Pod/unmanaged-pod-.*}
+      %r{Successfully deployed in \d.\ds: .*Pod/unmanaged-pod-.*}
```
KnVerey: what new text is appearing here?

test/integration/kubernetes_deploy_test.rb
```diff
@@ -64,16 +64,16 @@ def test_pruning_works
...
-      'pod "unmanaged-pod-',
-      'service "web"',
-      'resourcequota "resource-quotas"',
-      'deployment(\.extensions)? "web"',
-      'ingress(\.extensions)? "web"',
-      'daemonset(\.extensions)? "ds-app"',
-      'statefulset(\.apps)? "stateful-busybox"',
-      'job(\.batch)? "hello-job"',
-      'poddisruptionbudget(.policy)? "test"',
+      'configmap[ \/]"?hello-cloud-configmap-data"?',
```
KnVerey: Can you make a method like `prune_matcher(kind, group, name)` since we need to construct this format so many times?

test/integration/kubernetes_deploy_test.rb
karanthukral: This was a change before I took the PR on. Seems like it was for a fix in 1.9 but since 1.9 is back on minikube, I'll test to see if reverting this works

test/integration/kubernetes_deploy_test.rb
karanthukral: 👍

dev.yml
```diff
@@ -8,7 +8,7 @@ up:
   - custom:
       name: Minikube Cluster
       met?: test $(minikube status | grep Running | wc -l) -eq 2 && $(minikube status | grep -q 'Correctly Configured')
-      meet: minikube start --kubernetes-version=v1.10.6 --vm-driver=hyperkit
+      meet: minikube start --kubernetes-version=v1.10.0 --vm-driver=hyperkit
```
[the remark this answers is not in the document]
karanthukral: My bad! That was a mistake on my part. Will fix, squash and merge

PR #324, opened 2018-08-07
Check for presence of the `STATSD_IMPLEMENTATION' env:

Edouard-chin:
Check for presence of the `STATSD_IMPLEMENTATION' env:

- Calling `empty?` when ENV['STATSD_IMPLEMENTATION'] isn't set makes kubernetes-deploy to crash.

  Fixing the issue by calling ActiveSuport's `present?` instead

### DIFF
--- a/lib/kubernetes-deploy/statsd.rb
+++ b/lib/kubernetes-deploy/statsd.rb
@@ -15,7 +15,7 @@ module KubernetesDeploy
       if ENV['STATSD_DEV'].present?
         ::StatsD.backend = ::StatsD::Instrument::Backends::LoggerBackend.new(Logger.new($stderr))
       elsif ENV['STATSD_ADDR'].present?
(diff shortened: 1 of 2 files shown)


lib/kubernetes-deploy/statsd.rb
```diff
@@ -15,7 +15,7 @@ def self.build
       if ENV['STATSD_DEV'].present?
         ::StatsD.backend = ::StatsD::Instrument::Backends::LoggerBackend.new(Logger.new($stderr))
       elsif ENV['STATSD_ADDR'].present?
-        statsd_impl = ENV['STATSD_IMPLEMENTATION'].empty? ? "datadog" : ENV['STATSD_IMPLEMENTATION']
+        statsd_impl = ENV['STATSD_IMPLEMENTATION'].to_s.empty? ? "datadog" : ENV['STATSD_IMPLEMENTATION']
```
dturn: `ENV['STATSD_IMPLEMENTATION'].present?` is probably what you want.

PR #325, opened 2018-08-07
Compare generations when available

KnVerey:
## Problem

I just observed a instant false positive success on a web deployment. Relevant lines:

```yaml
[INFO][2018-08-07 19:27:25 +0000]	Successfully deployed in 1.9s: [...] Deployment/web
[INFO][2018-08-07 19:27:25 +0000]	Deployment/web                                    3 replicas, 3 updatedReplicas, 3 availableReplicas
```

The rollout definitely did not complete in less than 2s.

Checking the code made me notice that we aren't comparing `metadata.generation` to `status.observedGeneration` before using the status to determine success. This means we might be making that determination with state data, which could explain the above behaviour.

## Solution

Compare `metadata.Generation` to `status.observedGeneration` before `_succeeded?` and `_failed?` determinations whenever possible. I used the API documentation to figure out which controllers actually populate the observed generation.

Since this is a controller race, I can't think of a way to make it happen for integration tests. I have added regression tests at the unit test level. 

@Shopify/cloudx

### DIFF
--- a/lib/kubernetes-deploy/kubernetes_resource.rb
+++ b/lib/kubernetes-deploy/kubernetes_resource.rb
@@ -139,6 +139,17 @@ module KubernetesDeploy
       @instance_data.present?
     end
 
+    def current_generation
+      return -1 unless exists? # must be different default than observed_generation
+      @instance_data["metadata"]["generation"]
+    end
+
+    def observed_generation
+      return -2 unless exists?
+      # populating this is a best practice, but not all controllers actually do it
+      @instance_data["status"]["observedGeneration"]
+    end
+
     def status
       exists? ? "Exists" : "Unknown"
     end
--- a/lib/kubernetes-deploy/kubernetes_resource/daemon_set.rb
+++ b/lib/kubernetes-deploy/kubernetes_resource/daemon_set.rb
@@ -24,7 +24,8 @@ module KubernetesDeploy
     end
 
     def deploy_failed?
-      pods.present? && pods.any?(&:deploy_failed?)
+      pods.present? && pods.any?(&:deploy_failed?) &&
+      observed_generation == current_generation
     end
 
     def fetch_logs(kubectl)
@@ -34,16 +35,6 @@ module KubernetesDeploy
     end
 
     private
-
-    def current_generation
-      return -1 unless exists? # must be different default than observed_generation
-      @instance_data["metadata"]["generation"]
-    end
-
-    def observed_generation
-      return -2 unless exists?
-      @instance_data["status"]["observedGeneration"]
-    end
 
     def rollout_data
       return { "currentNumberScheduled" => 0 } unless exists?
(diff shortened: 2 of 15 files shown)


lib/kubernetes-deploy/kubernetes_resource/daemon_set.rb
```diff
@@ -24,7 +24,8 @@ def deploy_succeeded?
     end
 
     def deploy_failed?
-      pods.present? && pods.any?(&:deploy_failed?)
+      pods.present? && pods.any?(&:deploy_failed?) &&
+      observed_generation == current_generation
```
karanthukral: should this not be `!=`? Or maybe I just misunderstood it

lib/kubernetes-deploy/kubernetes_resource.rb
```diff
@@ -139,6 +139,17 @@ def exists?
...
     end
 
+    def current_generation
+      return -1 unless exists? # must be different default than observed_generation
+      @instance_data["metadata"]["generation"]
+    end
+
+    def observed_generation
+      return -2 unless exists?
+      # populating this is a best practice, but not all controllers actually do it
```
karanthukral: Out of curiosity, do you have an example of a controller that doesn't populate it?

lib/kubernetes-deploy/kubernetes_resource/daemon_set.rb
dturn: As I understand it, and @KnVerey will hopefully correct me if I'm wrong, `current_generation ` gets bumped when we `apply` the config. `observed_generation ` gets set by the controller reporting the status. When they don't match it means that we're looking at a status for the previous config not current. E.g. we don't want to say the deployment has failed or succeeded before the status reflects the current config's state.

karanthukral: oh! Thanks for the clarification

KnVerey:
> Is there a reasonable way to return early from sync when observed_generation != current_generation. Doesn't seem like we get anything useful when that happens.

What do you have in mind that we could skip? The basic sync is just `@instance_data = mediator.get_instance(type, name)`, and we don't know whether the generations match until we get that request. Maybe this is suggesting an additional state for our tentative state machine implementation though (names not to be taken seriously): `undeployed -> (outdated) -> pending -> succeeded/failed/timed out`

KnVerey:
Yeah, it definitely irked me to write that line in so many places. I could technically add that pseudo-state in our current model like this, but we might get unexpected results unless we also make everywhere that partitions resources into groups aware of the possibility:

image

WDYT?

dturn:
I think its better as is than adding a new state to unwind later.

KnVerey:
👍 I'll merge as-is then. I added a comment to your WIP so we don't forget about this extra state.

PR #326, opened 2018-08-08
Replace Apply, for pruning

dturn:
Add a new mode `replace_apply`  for `deploy_method`. In this mode a resource will first be replaced/created as if it were just `replace`. It will then be `kubectl apply` because this is done second it should be a noop, expect that it allows for pruning.

Testing this seems tricky...

### DIFF
--- a/lib/kubernetes-deploy/deploy_task.rb
+++ b/lib/kubernetes-deploy/deploy_task.rb
@@ -359,12 +359,14 @@ module KubernetesDeploy
 
       # Apply can be done in one large batch, the rest have to be done individually
       applyables, individuals = resources.partition { |r| r.deploy_method == :apply }
+      # For resources that need to be done individually, but should also be pruned
+      applyables += resources.select { |r| r.deploy_method == :replace_apply }
 
       individuals.each do |r|
         @logger.info("- #{r.id} (#{r.pretty_timeout_type})") if resources.length > 1
         r.deploy_started_at = Time.now.utc
         case r.deploy_method
-        when :replace
+        when :replace, :replace_apply
           _, _, replace_st = kubectl.run("replace", "-f", r.file_path, log_failure: false)
         when :replace_force
           _, _, replace_st = kubectl.run("replace", "--force", "--cascade", "-f", r.file_path,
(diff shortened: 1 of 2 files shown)


dturn:
It looks like the apply issue has been resolved: https://github.com/kubernetes/kubernetes/issues/53379
 
Think we should support pruning on replaceable resources?

KnVerey:
If I'm remembering correctly, the `:replace` method was add _because of_ that bug. I'd say if the fix was cherry-picked to all currently supported versions (looks like it was), we should go back to applying all CRs. At one point we were discussing relying on the deploy clearing the non-subresource status, but since we decided against that, I don't think we have a good reason to be replacing anymore.

Looks like we still need `replace_force` though, as [the PDB issue](https://github.com/kubernetes/kubernetes/issues/45398) is still open. I'd consider the fact that those aren't pruned to be a bug, but not relevant to your current project.

dturn:
@Shopify/cloudx 👀 please

dturn:
> Statefulservice still has replace in it

This was intentional. I was unsure why SS are `replace` since doesn't seem like they would be part of the CRD merge bug.

dturn:
🎩 with cloud-portal in the staging cluster by increasing & decreasing redis size. Let me know if you want to other resources.

KnVerey:
> This was intentional. I was unsure why SS are replace since doesn't seem like they would be part of the CRD merge bug.

Are you thinking StatefulSet rather than StatefulService, perhaps? StatefulSet are a core resource and are already applied; StatefulService are CRs and were replaced for the same reason as the others. Confusing naming for sure.

> with cloud-portal in the staging cluster by increasing & decreasing redis size. Let me know if you want to other resources

I'd suggest testing a change with a more complicated patch just in case. For example. a change inside an array or a hash (I'm actually curious which strategy it uses by default... for core resources it is field-specific). ES probably has the most elaborate schema, and cumulus-cat has one so could be a good target for testing.

dturn:
> Are you thinking StatefulSet rather than StatefulService, perhaps? StatefulSet are a core resource and are already applied; StatefulService are CRs and were replaced for the same reason as the others. Confusing naming for sure.

Yep I definitely got StatefulService confused with StatefulSet

I'll 🎩 something more complicated after fixing up StatefulSet

dturn:
🎩 a more complicated change went smoothly.

dturn:
Ready for final 👀

KnVerey:
🚢

PR #330, opened 2018-09-07
Relax the ejson depenencies constraints

Edouard-chin:
There is no reason to restrict app to use a specific version of `ejson`, there is nothing indicating that this gem would break if an app uses ejson `1.0.x` or even `1.x.x`.


dturn:
Can you please update the description to include the reason (e.g. bug fix / additional features) that caused you to open this pr?

Edouard-chin:
Done although I thought this was self explanatory /shrug

dturn:
> Done although I thought this was self explanatory /shrug

The code is self-explanatory about *what* its doing. But *why* you've opened this PR still isn't clear. 

- Are you just doing this out of the goodness of your heart.
- Is there a new feature in ejson that you want to use that other people might want to also know about
- Is this causing an incompatibility in a different repo
- Something I haven't thought about ...

PR #334, opened 2018-09-19
Formalize some feature guidelines

KnVerey:
This is something I've been meaning to propose for a while, both to help contributors and to sync up as maintainers. 
* Is this readme section the right place for this content?
* Anything I missed? 
* Anything you disagree with?

cc @Shopify/cloudx

### DIFF
--- a/README.md
+++ b/README.md
@@ -423,9 +423,18 @@ Bug reports and pull requests are welcome on GitHub at https://github.com/Shopif
 
 Contributions to help us support additional resource types or increase the sophistication of our success heuristics for an existing type are especially encouraged! (See tips below)
 
+## Feature acceptance guidelines
 
+- This project's mission is to make it easy to ship changes to a Kubernetes namespace and understand the result. Features that introduce new classes of responsibility to the tool are not usually accepted.
+  - Deploys can be a very tempting place to cram features. Imagine a proposed feature actually fits better elsewhere—where might that be? (Examples: validator in CI, custom controller, initializer, pre-processing step in the CD pipeline, or even Kubernetes core)
+  - The basic ERB renderer included with the tool is intended as a convenience feature for a better out-of-the box experience. Providing complex rendering capabilities is out of scope of this project's mission, and enhancements in this area may be rejected.
+  - The deploy command does not officially support non-namespaced resource types.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -423,9 +423,18 @@ Bug reports and pull requests are welcome on GitHub at https://github.com/Shopif
 
 Contributions to help us support additional resource types or increase the sophistication of our success heuristics for an existing type are especially encouraged! (See tips below)
 
+## Feature acceptance guidelines
 
+- This project's mission is to make it easy to ship changes to a Kubernetes namespace and understand the result. Features that introduce new classes of responsibility to the tool are not usually accepted.
+  - Deploys can be a very tempting place to cram features. Imagine a proposed feature actually fits better elsewhere—where might that be? (Examples: validator in CI, custom controller, initializer, pre-processing step in the CD pipeline, or even Kubernetes core)
```
kirs: Well put ❤️

README.md
```diff
@@ -423,9 +423,18 @@ Bug reports and pull requests are welcome on GitHub at https://github.com/Shopif
...
+  - The basic ERB renderer included with the tool is intended as a convenience feature for a better out-of-the box experience. Providing complex rendering capabilities is out of scope of this project's mission, and enhancements in this area may be rejected.
+  - The deploy command does not officially support non-namespaced resource types.
+- This project strives to be composable with other tools in the ecosystem, such as renderers and validators. The deploy command must work with any Kubernetes templates provided to it, no matter how they were generated.
+- This project is open-source. Features tied to any specific organization (including Shopify) will be rejected.
```
kirs: > Features tied to any specific organization (including Shopify) will be rejected.

While this is a good north start to have, we already do have some Shopify specific features, like ejson handling, `Memcache` and `Redis` resource types. What do we do about it? Do we want to mention that the rule only applies to new features?

KnVerey: That's a good point, although ejson is also open-source so I think it's ok, and we're ripping out the proprietary CR code "Soon" (by reviving https://github.com/Shopify/kubernetes-deploy/pull/188, more or less).

README.md
dturn: `Imagine a proposed feature actually fits better elsewhere—where might that be?` -> `Does the proposed feature actually fits better elsewhere?`

PR #337, opened 2018-10-03
Add result verification to task runner

KnVerey:
**PLEASE NOTE:** The refactoring piece of this has now been extracted to https://github.com/Shopify/kubernetes-deploy/pull/340

## Feature
The runner task now streams container logs and waits for the pod to succeed or fail **by default**
  - You can disable this using `--skip-wait` or `verify_result: false` (Note: named for consistency with the deploy task)
  - If you choose not to verify the result, we'll now take a look at the pod once and print its status in the summary (it will probably be pending, but seemed more helpful than nothing).
  - You can use `--max-watch-seconds=seconds`  or `max_watch_seconds: N` to set a time limit on the new result verification feature

Closes https://github.com/Shopify/kubernetes-deploy/issues/304

@Shopify/cloudx @fw42 @sirupsen 

## Detail

This proposes using the existing ResourceWatcher but moving the log handling for all pods into a new object with a traditional (for our resource models) sync-data/use-result separation. This makes it possible to change log output from being a side-effect of sync, to using a new post-sync hook that is run sequentially (sync is concurrent). I have not done so in this PR, but I expect this implementation to make it trivially possible to stream task-runner logs during deploys as well, which could be a nice improvement.

This PR also tweaks the UX (i.e. logging) for this task in general, to try to make it more useful and more consistent with the deploy task. See screenshots at the bottom of this description.

## Notes for reviewers

- Since this overhauls the task's logging, please tophat ALL the runner integration tests using `PRINT_LOGS=1`.
- Please also try to come up with test scenarios I may have missed.
- Since this PR is already pretty big, I'm deferring some potential improvements to future PRs. If you think any of these should go in before feature release, please speak up. List:
  - If a pod is stuck pending for more than X seconds, either emit a warning or time out (task runner pods in deploys too)
  - If a pod hasn't logged in the past X seconds, time out (not entirely convinced we should do this at all)
  - Add a context check to the validations (like we have for the deploy task)
  - Add a central retry mechanism and use it on many/most of the calls the task makes itself (as opposed to the ones in the watcher, which effectively retries every second anyway)
  - Stream task runner pods during deploys too (when/once there's only one pod to stream)
  - Try to make the watcher responsible for raising timeout and failure exceptions at the end of the watch (currently done by the calling tasks, with the exception of "global" timeouts)
  - Reconsider our treatment of `exists?` in all resource classes. If the API server has no record of it after the creation request, it isn't going to spontaneously appear (the user probably deleted it out of band?), so we should consider it failed.
  - Consider a "delete pod on timeout" option

If we're in agreement that these are ok to defer, I'll cut them into separate issues before merging.

## Alternatives considered

* https://github.com/Shopify/kubernetes-deploy/pull/320, in particular at: 
  * [this commit](https://github.com/Shopify/kubernetes-deploy/compare/50ddc9359b45c3a41cd770d46d71f9f10f762f24), which introduces a separate watcher type for streaming
  * [this commit](https://github.com/Shopify/kubernetes-deploy/compare/8e6c09cf9d65492efced24e7520967e44d19d2d2), which uses the normal watcher but doesn't have a separate class for logs yet
  * [this commit](https://github.com/Shopify/kubernetes-deploy/compare/a8180dbaefc93cbe7a83dea0862889da68dfd0a3), which uses the normal watcher and has a class for the logs, but changes less overall that I have her (logging still happens during sync, the logs class exposes both fetching and printing, and it does not store logs seen internally)
* I also tried using `kubeclient.watch_pod_log` [on this branch](https://github.com/Shopify/kubernetes-deploy/compare/k8s-run-stream-logs). I'm glad I tried it, but I liked it a lot less. It was more complex (e.g. after a few iterations, I ended up using the watcher in a side thread anyway, to be able to abort appropriately), is not compatible with the deploy part of the tool so duplicated/fought some of its functionality, and to boot the stream still wasn't working against my minikube by the end of the day I gave myself to try it.

## Screenshots

image

image

image

image


lib/kubernetes-deploy/resource_watcher.rb
```diff
@@ -17,19 +17,16 @@ def initialize(resources:, sync_mediator:, logger:, deploy_started_at: Time.now.
...
-        end
-        if (sleep_duration = delay_sync_until - Time.now.utc) > 0
-          sleep(sleep_duration)
-        end
-        delay_sync_until = Time.now.utc + delay_sync # don't pummel the API if the sync is fast
+        report_and_give_up(remainder) if global_timeout?(monitoring_started)
+        sleep_until_next_sync(delay_sync)
 
         @sync_mediator.sync(remainder)
+        remainder.each(&:post_sync)
```
dturn: As implemented `sync` is run all on all the resources then `post_sync` is called. In rails /  https://api.rubyonrails.org/classes/ActiveSupport/Callbacks.html `post_sync` would trigger immediately after each resource has `sync` called. Is the implemented behavior needed? If not I think it would be preferable to do it the more traditional way.

lib/kubernetes-deploy/runner_task.rb
```diff
@@ -8,51 +8,99 @@ module KubernetesDeploy
...
 
-      @logger.phase_heading("Constructing final pod specification")
-      rendered_template = build_pod_template(raw_template, entrypoint, args, env_vars)
+    private
 
-      validate_pod_spec(rendered_template)
+    def create_pod(pod)
+      @logger.info "Creating pod '#{pod.name}'"
+      pod.deploy_started_at = Time.now.utc
+      kubeclient.create_pod(Kubeclient::Resource.new(pod.definition))
```
dturn: Is this inverted? e.g `pod.to_kubeclient_resource`

lib/kubernetes-deploy/runner_task.rb
```diff
@@ -8,51 +8,99 @@ module KubernetesDeploy
...
+      @pod_name = pod.name
```
dturn: Why an instance var?

lib/kubernetes-deploy/pod_logs.rb
```diff
@@ -0,0 +1,95 @@
...
+        prefix = "[#{container_name}]  " if @container_names.length > 1
+        start_at = @last_printed_indexes[container_name] + 1
+        logs[start_at..-1].each do |line|
+          @logger.info "#{prefix}#{line}"
+        end
+        @last_printed_indexes[container_name] = logs.length - 1
+      end
+    end
+
+    def print_all(prevent_duplicate: true)
```
dturn: Plans for `prevent_duplicate` later?

lib/kubernetes-deploy/resource_watcher.rb
KnVerey: tl;dr The technical reason is that `sync` runs concurrently and `post_sync` shouldn't. I also don't see the value of the indirection (vs writing the code at the end of `sync` itself) if it were executed immediately after.

The idea here is that `sync` should exclusively collect state from the API server and should not have any side-effects. It is run in parallel, so weird things can happen if the instance associates additional behaviours with it (e.g. seems to me that if two pods complete in different sync groups during the same polling cycle today, their logs will get interleaved). The `post_sync` hook (naming suggestions appreciated) is a way to let instances do something every polling cycle if they need to, with the guarantee that (a) their instance data is already up to date; and (b) nothing else is logging at the same time. I considered making this `dump_logs` or something, but thought it would be better design to keep the watcher agnostic about the details of what instances might want to do.

lib/kubernetes-deploy/runner_task.rb
KnVerey: 👍 I like that idea

lib/kubernetes-deploy/pod_logs.rb
KnVerey: Not really. I had added this based on Flo's comment to the effect that having `print_all` no-op on the second call by default is unintuitive.

lib/kubernetes-deploy/runner_task.rb
KnVerey: Because today, the pod's name is exposed as the return value of `run`. I don't want to keep doing that, but there are legitimate use cases for wanting to know the name of the pod created so you can take follow-up action. For example, the caller might want to rescue DeploymentTimeoutError, delete the pod created (to make sure two aren't running at once), and retry the runner. Incidentally, I'm wondering if we should add a `--delete-pod-on-timeout` option to the runner itself (not in this PR).

lib/kubernetes-deploy/pod_logs.rb
```diff
@@ -0,0 +1,95 @@
...
+          @container_logs[container_name] << msg
+          @last_timestamp_seen_for_container[container_name] = timestamp
+        end
+      end
+    end
+
+    def print_latest
+      @last_printed_indexes ||= Hash.new { |hash, key| hash[key] = -1 }
+      @container_logs.each do |container_name, logs|
+        prefix = "[#{container_name}]  " if @container_names.length > 1
```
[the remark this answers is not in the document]
dturn: Part of me was thinking that we have to do some ugly things because a pod can have multiple containers. But is that something we actually need/want to allow for the runner?

lib/kubernetes-deploy/resource_watcher.rb
dturn: (b) seems like its just a result of running the hook serially. If the goal is to keep the watcher agnostic, which I agree with,  I'm think a more descriptive name would help `serialized_post_sync` (I'm terrible at naming :( )

KnVerey:
I split off the refactoring pieces of this into https://github.com/Shopify/kubernetes-deploy/pull/340 for ease of review. This branch is now based on that one. I will still merge them at the same time.

lib/kubernetes-deploy/pod_logs.rb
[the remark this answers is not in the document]
KnVerey: We definitely need to not care if they have extra containers, but they do already have to tell us which one is doing the main work (so we can inject the env vars/args/etc in that one). Maybe we could set `pod.main_container` or something and pick that one. If we use this in deploys though (which I'd like to), we won't have that information, so we'd probably need to keep the prefixing for that case. WDYT? 🤔

KnVerey:
> What do you think about adding a test for RemoteLogs where a pod has 2 containers.

Added a bunch of unit tests in https://github.com/Shopify/kubernetes-deploy/pull/340.  👍

lib/kubernetes-deploy/kubernetes_resource/pod.rb
```diff
@@ -8,9 +8,10 @@ class Pod < KubernetesResource
       Evicted
       Preempting
     )
+    attr_writer :stream_logs
```
dturn: do you have a use case in mind where you'd create the pod w/ `stream_logs: false` but change it before pod completed?

test/integration/runner_task_test.rb
```diff
@@ -3,93 +3,259 @@
...
+      refute missing_nums.present?, "Some lines were not streamed: #{missing_nums}"
+
+      num_lines_duplicated = nums_printed.length - nums_printed.uniq.length
+      assert num_lines_duplicated.zero?, "#{num_lines_duplicated} lines were duplicated"
+    end
   end
 
-  def test_run_with_template_runner_template_missing
+  def test_run_with_bad_restart_policy
+    deploy_task_template do |f|
```
fw42: what does `f` stand for?

test/integration/runner_task_test.rb
```diff
@@ -3,93 +3,259 @@
...
+      "Pod status: Failed"
+    ], in_order: true)
+
+    pods = kubeclient.get_pods(namespace: @namespace)
+    assert_equal 1, pods.length, "Expected 1 pod to exist, found #{pods.length}"
   end
 
-  def test_run_with_env_vars
-    deploy_fixtures("hello-cloud", subset: ["template-runner.yml"])
+  def test_run_with_missing_namespace
```
fw42: nitpick: `test_run_fails_if_namespace_is_missing`?

lib/kubernetes-deploy/runner_task.rb
```diff
@@ -8,51 +8,99 @@ module KubernetesDeploy
...
-      @logger.phase_heading("Creating pod")
-      @logger.info("Starting task runner pod: '#{rendered_template.metadata.name}'")
-      @kubeclient.create_pod(rendered_template)
+    def build_pod(template_name, entrypoint, args, env_vars, verify_result)
+      task_template = get_template(template_name)
+      @logger.info("Using template '#{template_name}'")
+      pod_template = build_pod_definition(task_template)
+      set_container_overrides!(pod_template, entrypoint, args, env_vars)
+      ensure_valid_restart_policy!(pod_template, verify_result)
+      Pod.new(namespace: @namespace, context: @context, logger: @logger, stream_logs: true,
```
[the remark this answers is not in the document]
KnVerey: This option is really about _how_ the logs are displayed when the pod is being watched, not _if_: "stream logs as we get them" vs. "dump them all when this succeeds/fails". If you don't start watching this pod at all (because `verify_result` is false), `stream_logs` is irrelevant. Currently `stream_logs` is always true in task runs and always false in deploys. As discussed with Danny in a comment above, I think it would be cool if we could intelligently set it to true during certain deploys too.

test/integration/runner_task_test.rb
```diff
@@ -3,93 +3,259 @@
...
+    task_runner = build_task_runner
+
+    # Sketchy, but stubbing the kubeclient doesn't work (and wouldn't be concurrency-friendly)
+    # Finding a way to reliably trigger a create failure would be much better, if possible
+    mock = mock()
+    mock.expects(:get_namespace)
+    template = kubeclient.get_pod_template('hello-cloud-template-runner', @namespace)
+    mock.expects(:get_pod_template).returns(template)
+    mock.expects(:create_pod).raises(Kubeclient::HttpError.new("409", "Pod with same name exists", {}))
+    task_runner.instance_variable_set(:@kubeclient, mock)
```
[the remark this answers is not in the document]
fw42: Oh sweet. I didn't even know about `stubbing_non_public_method`. TIL. I guess this is ok then.

test/integration/runner_task_test.rb
```diff
@@ -116,23 +116,28 @@ def test_run_with_verify_result_success
...
     deploy_task_template
     task_runner = build_task_runner
-    result = task_runner.run(run_params(log_lines: 5_000, log_interval: 0.0001))
+    result = task_runner.run(run_params(log_lines: 5_000, log_interval: 0.0005))
     assert_task_run_success(result)
 
     logging_assertion do |all_logs|
       nums_printed = all_logs.scan(/Line (\d+)$/).flatten
 
-      missing_nums = nums_printed - (1..5_000).map(&:to_s)
```
KnVerey: This was backwards, so the test wasn't working as expected.

PR #340, opened 2018-10-05
Introduce classes for remote logs

KnVerey:
Extracted from https://github.com/Shopify/kubernetes-deploy/pull/337, which is now based on top of this for ease of reviewing each piece. See that PR for context. 

@Shopify/cloudx

### DIFF
--- a/lib/kubernetes-deploy/container_logs.rb
+++ b/lib/kubernetes-deploy/container_logs.rb
@@ -10,7 +10,7 @@ module KubernetesDeploy
       @container_name = container_name
       @logger = logger
       @lines = []
-      @last_printed_index = -1
+      @next_print_index = 0
     end
 
     def sync(kubectl)
@@ -25,13 +25,12 @@ module KubernetesDeploy
 
     def print_latest(prefix: false)
       prefix_str = "[#{container_name}]  " if prefix
-      start_at = @last_printed_index + 1
 
-      lines[start_at..-1].each do |msg|
+      lines[@next_print_index..-1].each do |msg|
         @logger.info "#{prefix_str}#{msg}"
       end
 
-      @last_printed_index = lines.length - 1
+      @next_print_index = lines.length
     end
 
     def print_all
@@ -58,15 +57,17 @@ module KubernetesDeploy
     def deduplicate(logs)
       deduped = []
       timestamps = []
+      check_for_duplicate = true
 
       logs.each do |line|
         timestamp, msg = split_timestamped_line(line)
-        next if likely_duplicate?(timestamp)
+        next if check_for_duplicate && likely_duplicate?(timestamp)
+        check_for_duplicate = false # logs are ordered, so once we've seen a new one, assume all subsequent logs are new
         timestamps << timestamp if timestamp
         deduped << msg
       end
 
-      @last_timestamp = timestamps.max
+      @last_timestamp = timestamps.last
       deduped
     end
 
--- a/lib/kubernetes-deploy/container_logs.rb
+++ b/lib/kubernetes-deploy/container_logs.rb
@@ -1,0 +1,81 @@
+# frozen_string_literal: true
+module KubernetesDeploy
+  class ContainerLogs
+    attr_reader :lines, :container_name
+
+    DEFAULT_LINE_LIMIT = 250
+
+    def initialize(parent_id:, container_name:, logger:)
+      @parent_id = parent_id
+      @container_name = container_name
+      @logger = logger
+      @lines = []
+      @last_printed_index = -1
+    end
+
+    def sync(kubectl)
+      new_logs = fetch_latest(kubectl)
+      return unless new_logs.present?
+      @lines += deduplicate(new_logs)
+    end
+
+    def empty?
+      @lines.empty?
+    end
+
+    def print_latest(prefix: false)
+      prefix_str = "[#{container_name}]  " if prefix
+      start_at = @last_printed_index + 1
+
+      @lines[start_at..-1].each do |msg|
+        @logger.info "#{prefix_str}#{msg}"
+      end
+
+      @last_printed_index = @lines.length - 1
+    end
+
+    def print_all
+      @lines.each { |line| @logger.info("\t#{line}") }
+    end
+
+    private
+
+    def fetch_latest(kubectl)
+      cmd = ["logs", @parent_id, "--container=#{container_name}", "--timestamps"]
+      cmd << if @last_timestamp.present?
+        "--since-time=#{rfc3330_timestamp(@last_timestamp)}"
+      else
+        "--tail=#{DEFAULT_LINE_LIMIT}"
+      end
+      out, _err, _st = kubectl.run(*cmd, log_failure: false)
+      out.split("\n")
+    end
+
+    def rfc3330_timestamp(time)
+      time.strftime("%FT%T.%N%:z")
+    end
+
+    def deduplicate(logs)
+      logs.each_with_object([]) do |line, deduped|
+        timestamp, msg = split_timestamped_line(line)
+        next if likely_duplicate?(timestamp)
+        deduped << msg
+        @last_timestamp = timestamp
+      end
+    end
+
+    def split_timestamped_line(log_line)
+      timestamp, message = log_line.split(" ", 2)
+      [Time.parse(timestamp), message]
+    rescue ArgumentError
+      nil
+    end
+
+    def likely_duplicate?(timestamp)
+      return false unless @last_timestamp && timestamp
+      # The --since-time granularity the API server supports is not adequate to prevent duplicates
+      # This comparison takes the fractional seconds into account
+      timestamp <= @last_timestamp
+    end
+  end
+end
--- a/lib/kubernetes-deploy/container_logs.rb
+++ b/lib/kubernetes-deploy/container_logs.rb
@@ -20,22 +20,22 @@ module KubernetesDeploy
     end
 
     def empty?
-      @lines.empty?
+      lines.empty?
     end
 
     def print_latest(prefix: false)
       prefix_str = "[#{container_name}]  " if prefix
       start_at = @last_printed_index + 1
 
-      @lines[start_at..-1].each do |msg|
+      lines[start_at..-1].each do |msg|
         @logger.info "#{prefix_str}#{msg}"
       end
 
-      @last_printed_index = @lines.length - 1
+      @last_printed_index = lines.length - 1
     end
 
     def print_all
-      @lines.each { |line| @logger.info("\t#{line}") }
+      lines.each { |line| @logger.info("\t#{line}") }
     end
 
     private
@@ -43,7 +43,7 @@ module KubernetesDeploy
     def fetch_latest(kubectl)
       cmd = ["logs", @parent_id, "--container=#{container_name}", "--timestamps"]
       cmd << if @last_timestamp.present?
-        "--since-time=#{rfc3330_timestamp(@last_timestamp)}"
+        "--since-time=#{rfc3339_timestamp(@last_timestamp)}"
       else
         "--tail=#{DEFAULT_LINE_LIMIT}"
       end
@@ -51,7 +51,7 @@ module KubernetesDeploy
       out.split("\n")
     end
 
-    def rfc3330_timestamp(time)
+    def rfc3339_timestamp(time)
       time.strftime("%FT%T.%N%:z")
     end
 
--- a/lib/kubernetes-deploy/container_logs.rb
+++ b/lib/kubernetes-deploy/container_logs.rb
@@ -56,19 +56,26 @@ module KubernetesDeploy
     end
 
     def deduplicate(logs)
-      logs.each_with_object([]) do |line, deduped|
+      deduped = []
+      timestamps = []
+
+      logs.each do |line|
         timestamp, msg = split_timestamped_line(line)
         next if likely_duplicate?(timestamp)
+        timestamps << timestamp if timestamp
         deduped << msg
-        @last_timestamp = timestamp
       end
+
+      @last_timestamp = timestamps.max
+      deduped
     end
 
     def split_timestamped_line(log_line)
       timestamp, message = log_line.split(" ", 2)
       [Time.parse(timestamp), message]
     rescue ArgumentError
-      nil
+      # If the log file can't be opened, k8s 1.8 writes an error message without a timestamp to stdout
+      [nil, log_line]
     end
 
     def likely_duplicate?(timestamp)
--- a/lib/kubernetes-deploy/kubernetes_resource.rb
+++ b/lib/kubernetes-deploy/kubernetes_resource.rb
@@ -3,6 +3,8 @@ require 'json'
 require 'open3'
 require 'shellwords'
 
+require 'kubernetes-deploy/remote_logs'
+
 module KubernetesDeploy
   class KubernetesResource
     attr_reader :name, :namespace, :context
@@ -118,6 +120,9 @@ module KubernetesDeploy
       @instance_data = mediator.get_instance(kubectl_resource_type, name)
     end
 
+    def post_sync
+    end
+
     def deploy_failed?
       false
     end
@@ -173,9 +178,8 @@ module KubernetesDeploy
     end
 
     def sync_debug_info(kubectl)
-      @events = fetch_events(kubectl) unless ENV[DISABLE_FETCHING_EVENT_INFO]
-      @logs = fetch_logs(kubectl) if supports_logs? && !ENV[DISABLE_FETCHING_EVENT_INFO]
-      @debug_info_synced = true
+      @debug_events = fetch_events(kubectl) unless ENV[DISABLE_FETCHING_EVENT_INFO]
+      @debug_logs = fetch_debug_logs(kubectl) if supports_debug_logs? && !ENV[DISABLE_FETCHING_LOG_INFO]
     end
 
     def debug_message(cause = nil, info_hash = {})
@@ -198,9 +202,9 @@ module KubernetesDeploy
       end
       helpful_info << "  - Final status: #{status}"
 
-      if @events.present?
+      if @debug_events.present?
         helpful_info << "  - Events (common success events excluded):"
-        @events.each do |identifier, event_hashes|
+        @debug_events.each do |identifier, event_hashes|
           event_hashes.each { |event| helpful_info << "      [#{identifier}]\t#{event}" }
         end
       elsif ENV[DISABLE_FETCHING_EVENT_INFO]
@@ -209,21 +213,24 @@ module KubernetesDeploy
         helpful_info << "  - Events: #{DEBUG_RESOURCE_NOT_FOUND_MESSAGE}"
       end
 
-      if supports_logs?
+      if supports_debug_logs?
         if ENV[DISABLE_FETCHING_LOG_INFO]
           helpful_info << "  - Logs: #{DISABLED_LOG_INFO_MESSAGE}"
-        elsif @logs.blank? || @logs.values.all?(&:blank?)
+        elsif @debug_logs.blank?
           helpful_info << "  - Logs: #{DEBUG_RESOURCE_NOT_FOUND_MESSAGE}"
         else
-          sorted_logs = @logs.sort_by { |_, log_lines| log_lines.length }
-          sorted_logs.each do |identifier, log_lines|
-            if log_lines.empty?
-              helpful_info << "  - Logs from container '#{identifier}': #{DEBUG_RESOURCE_NOT_FOUND_MESSAGE}"
+          container_logs = @debug_logs.container_logs.sort_by { |c| c.lines.length }
+          container_logs.each do |c|
+            if c.empty?
+              helpful_info << "  - Logs from container '#{c.container_name}': #{DEBUG_RESOURCE_NOT_FOUND_MESSAGE}"
               next
             end
 
-            helpful_info << "  - Logs from container '#{identifier}' (last #{LOG_LINE_COUNT} lines shown):"
-            log_lines.each do |line|
+            if c.lines.length == ContainerLogs::DEFAULT_LINE_LIMIT
+              truncated = " (last #{ContainerLogs::DEFAULT_LINE_LIMIT} lines shown)"
+            end
+            helpful_info << "  - Logs from container '#{c.container_name}'#{truncated}:"
+            c.lines.each do |line|
               helpful_info << "      #{line}"
             end
           end
@@ -366,8 +373,8 @@ module KubernetesDeploy
       file&.close
     end
 
-    def supports_logs?
-      respond_to?(:fetch_logs)
+    def supports_debug_logs?
+      false
     end
 
     def statsd_tags
--- a/lib/kubernetes-deploy/kubernetes_resource.rb
+++ b/lib/kubernetes-deploy/kubernetes_resource.rb
@@ -120,7 +120,7 @@ module KubernetesDeploy
       @instance_data = mediator.get_instance(kubectl_resource_type, name)
     end
 
-    def post_sync
+    def after_sync
     end
 
     def deploy_failed?
@@ -179,7 +179,7 @@ module KubernetesDeploy
 
     def sync_debug_info(kubectl)
       @debug_events = fetch_events(kubectl) unless ENV[DISABLE_FETCHING_EVENT_INFO]
-      @debug_logs = fetch_debug_logs(kubectl) if supports_debug_logs? && !ENV[DISABLE_FETCHING_LOG_INFO]
+      @debug_logs = fetch_debug_logs(kubectl) if print_debug_logs? && !ENV[DISABLE_FETCHING_LOG_INFO]
     end
 
     def debug_message(cause = nil, info_hash = {})
@@ -213,7 +213,7 @@ module KubernetesDeploy
         helpful_info << "  - Events: #{DEBUG_RESOURCE_NOT_FOUND_MESSAGE}"
       end
 
-      if supports_debug_logs?
+      if print_debug_logs?
         if ENV[DISABLE_FETCHING_LOG_INFO]
           helpful_info << "  - Logs: #{DISABLED_LOG_INFO_MESSAGE}"
         elsif @debug_logs.blank?
@@ -373,7 +373,7 @@ module KubernetesDeploy
       file&.close
     end
 
-    def supports_debug_logs?
+    def print_debug_logs?
       false
     end
 
--- a/lib/kubernetes-deploy/kubernetes_resource.rb
+++ b/lib/kubernetes-deploy/kubernetes_resource.rb
@@ -220,17 +220,17 @@ module KubernetesDeploy
           helpful_info << "  - Logs: #{DEBUG_RESOURCE_NOT_FOUND_MESSAGE}"
         else
           container_logs = @debug_logs.container_logs.sort_by { |c| c.lines.length }
-          container_logs.each do |c|
-            if c.empty?
-              helpful_info << "  - Logs from container '#{c.container_name}': #{DEBUG_RESOURCE_NOT_FOUND_MESSAGE}"
+          container_logs.each do |logs|
+            if logs.empty?
+              helpful_info << "  - Logs from container '#{logs.container_name}': #{DEBUG_RESOURCE_NOT_FOUND_MESSAGE}"
               next
             end
 
-            if c.lines.length == ContainerLogs::DEFAULT_LINE_LIMIT
+            if logs.lines.length == ContainerLogs::DEFAULT_LINE_LIMIT
               truncated = " (last #{ContainerLogs::DEFAULT_LINE_LIMIT} lines shown)"
             end
-            helpful_info << "  - Logs from container '#{c.container_name}'#{truncated}:"
-            c.lines.each do |line|
+            helpful_info << "  - Logs from container '#{logs.container_name}'#{truncated}:"
+            logs.lines.each do |line|
               helpful_info << "      #{line}"
             end
           end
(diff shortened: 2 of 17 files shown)


lib/kubernetes-deploy/kubernetes_resource.rb
```diff
@@ -118,6 +120,9 @@ def sync(mediator)
       @instance_data = mediator.get_instance(kubectl_resource_type, name)
     end
 
+    def post_sync
```
KnVerey: Note: try to think of a better name for this

lib/kubernetes-deploy/container_logs.rb
```diff
@@ -0,0 +1,81 @@
...
+    private
+
+    def fetch_latest(kubectl)
+      cmd = ["logs", @parent_id, "--container=#{container_name}", "--timestamps"]
+      cmd << if @last_timestamp.present?
+        "--since-time=#{rfc3339_timestamp(@last_timestamp)}"
+      else
+        "--tail=#{DEFAULT_LINE_LIMIT}"
+      end
+      out, _err, _st = kubectl.run(*cmd, log_failure: false)
```
KnVerey: Repost from other PR:

This continues the existing behaviour of treating ALL failures to retrieve logs as "none available I guess". I'm thinking it would just be spammy to output anything about failures because:
(a) log output is for the benefit of the human operator only (in no way affects the control loop);
(b) this call already ends up getting "retried" every polling cycle, which for the runner is every second; and
(c) If Kubernetes is behind on consuming the logs from the container, there's nothing the operator can do about it.
Can you think of a case where it could be helpful to the operator, or a class of error that should actually be handled in some way?

lib/kubernetes-deploy/kubernetes_resource.rb
```diff
@@ -209,21 +213,24 @@ def debug_message(cause = nil, info_hash = {})
...
-        elsif @logs.blank? || @logs.values.all?(&:blank?)
+        elsif @debug_logs.blank?
           helpful_info << "  - Logs: #{DEBUG_RESOURCE_NOT_FOUND_MESSAGE}"
         else
-          sorted_logs = @logs.sort_by { |_, log_lines| log_lines.length }
-          sorted_logs.each do |identifier, log_lines|
-            if log_lines.empty?
-              helpful_info << "  - Logs from container '#{identifier}': #{DEBUG_RESOURCE_NOT_FOUND_MESSAGE}"
+          container_logs = @debug_logs.container_logs.sort_by { |c| c.lines.length }
+          container_logs.each do |c|
```
dturn: Can you use a more descriptive name than `c`

test/unit/container_logs_test.rb
```diff
@@ -0,0 +1,127 @@
...
+  def logs_part_1
+    # beginning of logs from second 1
+    <<~STRING
+      2018-10-04T19:40:30.997382362Z Line 1
+      2018-10-04T19:40:30.997941546Z Line 2
+      2018-10-04T19:40:30.998388739Z Line 3
+    STRING
+  end
+
+  def logs_part_2
```
dturn: I think this would be more clear if it called `logs_part_1`

test/unit/container_logs_test.rb
```diff
@@ -0,0 +1,127 @@
...
+    assert_logs_match_all(expected, in_order: true)
+  end
+
+  private
+
+  def expected_log_lines(range)
+    range.map { |i| "Line #{i}" }
+  end
+
+  def logs_part_1
```
dturn: `logs_1_through_3` might be more clear than part_1

test/unit/remotes_logs_test.rb
```diff
@@ -0,0 +1,95 @@
...
+    logs.container_logs.last.expects(:lines).returns(mock_lines('A'..'D')).at_least_once
+
+    logs.print_all
+    assert_logs_match_all([
+      "Logs from pod/pod-123-456 container 'Container1'",
+      "Line 1",
+      "Line 2",
+      "Line 3",
+      "Line 4",
+      "Logs from pod/pod-123-456 container 'ContainerA'",
```
dturn: Would this be better with a blank line printed between container logs?

lib/kubernetes-deploy/container_logs.rb
```diff
@@ -0,0 +1,81 @@
...
+      start_at = @last_printed_index + 1
+
+      lines[start_at..-1].each do |msg|
+        @logger.info "#{prefix_str}#{msg}"
+      end
+
+      @last_printed_index = lines.length - 1
+    end
+
+    def print_all
```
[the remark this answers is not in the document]
KnVerey: Do you see a use case for that? Right now it's assuming the caller is going to print any identifiers it wants (which isn't really workable for `print_latest` beacuse it is trying to look like a stream).

test/unit/remotes_logs_test.rb
KnVerey: It does, but that's shitty to assert on because the full lines have timestamps in them. You can run these tests with `PRINT_LOGS=1` too btw:

image

And hey, looks like I've actually got an _extra_ blank line in there!

lib/kubernetes-deploy/container_logs.rb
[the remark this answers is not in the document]
dturn: Good point.

test/unit/container_logs_test.rb
KnVerey: The chunking isn't based on number of lines though; these three methods are simulating three requests to the API server with sub-second log entries:
- Part 1: requests all logs so far, during second 1. Gets the first few logs from second 1.
- Part 2: requests all logs since second 1, during second 2. Gets all logs from second 1 and the first few from second 2.
- Part 3: requests all logs since second 2, during second 3. Gets all logs from second 2 and the first few from second 3.

Not sure how to find names that explain that succinctly. I could make a fancier helper like `dummy_log_response(start_at_second:, current_second:)` but not sure it's worth the effort.

dturn: My concern was about the disconnect between `kubectl.stubs(:run).returns([logs_part_1, "", ""])`  and `expected_log_lines(1..3)`

lib/kubernetes-deploy/container_logs.rb
```diff
@@ -0,0 +1,81 @@
...
+        "--since-time=#{rfc3339_timestamp(@last_timestamp)}"
+      else
+        "--tail=#{DEFAULT_LINE_LIMIT}"
+      end
+      out, _err, _st = kubectl.run(*cmd, log_failure: false)
+      out.split("\n")
+    end
+
+    def rfc3339_timestamp(time)
+      time.strftime("%FT%T.%N%:z")
```
fw42: Looks like Ruby has a method for this. Should we use that?

https://ruby-doc.org/stdlib-2.3.1/libdoc/date/rdoc/DateTime.html#method-i-rfc3339

lib/kubernetes-deploy/container_logs.rb
```diff
@@ -0,0 +1,81 @@
...
+        next if likely_duplicate?(timestamp)
+        deduped << msg
+        @last_timestamp = timestamp
+      end
+    end
+
+    def split_timestamped_line(log_line)
+      timestamp, message = log_line.split(" ", 2)
+      [Time.parse(timestamp), message]
+    rescue ArgumentError
```
fw42: Does this actually happen under normal circumstances? What does it mean?

lib/kubernetes-deploy/container_logs.rb
```diff
@@ -0,0 +1,81 @@
...
+    end
+
+    def rfc3339_timestamp(time)
+      time.strftime("%FT%T.%N%:z")
+    end
+
+    def deduplicate(logs)
+      logs.each_with_object([]) do |line, deduped|
+        timestamp, msg = split_timestamped_line(line)
+        next if likely_duplicate?(timestamp)
```
fw42: `split_timestamped_line` can return `nil`, in which case `timestamp` will be nil, which means `likely_duplicate?` will return `false`, and then this line here will not remove the line and the next one will call `deduped << nil`.

Is that intentional?

lib/kubernetes-deploy/container_logs.rb
```diff
@@ -0,0 +1,81 @@
...
+        deduped << msg
+        @last_timestamp = timestamp
```
fw42: This line doesn't run if `likely_duplicate?` returned `true`. Is that intentional? If so why?

lib/kubernetes-deploy/container_logs.rb
```diff
@@ -0,0 +1,81 @@
...
+      end
+      out, _err, _st = kubectl.run(*cmd, log_failure: false)
+      out.split("\n")
+    end
+
+    def rfc3339_timestamp(time)
+      time.strftime("%FT%T.%N%:z")
+    end
+
+    def deduplicate(logs)
```
fw42: Just out of curiosity: Are these logs guaranteed to be returned in order? If so then as soon as we find one line that's not a duplicate, we know that all subsequent lines (from the same API call result) are also not duplicates, right? Seems like knowing this fact would allow us to simplify the code (although I'm not sure yet how).

lib/kubernetes-deploy/container_logs.rb
fw42: Does this line actually need to be in the loop body? Couldn't this run after the filtering? I think that would make the code a tiny bit easier to understand (since the block that's being passed to `each_with_object` would have fewer side-effects).

lib/kubernetes-deploy/resource_watcher.rb
```diff
@@ -48,6 +45,18 @@ def run(delay_sync: 3.seconds, reminder_interval: 30.seconds, record_summary: tr
...
+    def global_timeout?(started_at)
+      @timeout && (Time.now.utc - started_at > @timeout)
+    end
+
+    def sleep_until_next_sync(min_interval)
+      @next_sync_time ||= Time.now.utc
+      if (sleep_duration = @next_sync_time - Time.now.utc) > 0
+        sleep(sleep_duration)
+      end
+      @next_sync_time = Time.now.utc + min_interval
```
fw42: I'm scratching my head trying to understand whether or not this line should be in the `if` branch above, only in the `else` branch, or here where it is. Do we want to bump the `next_sync_time` even if we didn't actually sleep? It's a bit too late in the day here for me to have enough mental clarity to say, but maybe this can be expressed more clearly somehow?

test/integration/kubernetes_deploy_test.rb
```diff
@@ -619,7 +619,7 @@ def test_failure_logs_from_unmanaged_pod_appear_in_summary_section
 
     assert_logs_match_all([
       "Failed to deploy 1 priority resource",
-      "Logs from container 'hello-cloud' (last 250 lines shown):",
+      "Logs from container 'hello-cloud':",
```
fw42: why did this change?

lib/kubernetes-deploy/resource_watcher.rb
fw42: I just saw that's the previous behaviour already, so don't let my comment block you...

lib/kubernetes-deploy/container_logs.rb
KnVerey: The original code did, but [Policial didn't like it](https://policial.shopify.io/Shopify/kubernetes-deploy/builds/695493). TBH I got annoyed trying to figure out whether the justification for that blanket prohibition actually applies here, and made this method to avoid going through DateTime in any case. Looking again, the [PR](https://github.com/rubocop-hq/rubocop/issues/1952) description doesn't explain anything, but further down there's a discussion about DateTime not handling some timezones properly. I don't think that does matter here since the timestamps I'm getting back (locally and in prod) are all formatted like `2018-10-08T02:04:02.851080236Z`.

lib/kubernetes-deploy/container_logs.rb
KnVerey: You asked Danny the same thing earlier, so maybe I should add a comment!

image

OTOH maybe a better default value would be `[nil, log_line]` in that case. 🤔

lib/kubernetes-deploy/container_logs.rb
KnVerey: Nope! Good catch. I think if there is no timestamp, we should err on the side of duplicating the line. In the one known case causing the `nil`, it's an error message that is presumably only printed once.

lib/kubernetes-deploy/container_logs.rb
KnVerey: That's a good question. AFAIK k8s is just streaming us the log file written by Docker from the container's stdout/stderr, so it would be super weird for them to get out of order, right? In any case, this code already assumes they're in order, since every iteration overwrites the `@last_timestamp` (assumes ending up with the last timestamp value is correct).

[Logging architecture docs](https://kubernetes.io/docs/concepts/cluster-administration/logging/#logging-at-the-node-level)

lib/kubernetes-deploy/container_logs.rb
KnVerey: > This line doesn't run if likely_duplicate? returned true. Is that intentional? If so why?

Yes, if it is likely a duplicate, then its timestamp is necessarily <= the value we currently have in `@last_timestamp`. 

> Does this line actually need to be in the loop body?

Now that I think about it, it technically needs to move outside it. If it actually happened that two lines were logged at the "same time" by this precision (same 0.000000001 seconds), having this inside the loop would cause us to skip the second one.

lib/kubernetes-deploy/resource_watcher.rb
KnVerey: This code is pretty much a really dumb throttle. It assumes it is called before every sync and its job is to make sure the syncs begin at least `min_interval` seconds apart. So yes, we need to set the `next_sync_time` every time we're about to leave the method, i.e start a sync. Any suggestions on naming? Were the previous names clearer?

test/integration/kubernetes_deploy_test.rb
KnVerey: It was irrelevant in cases like this (there's only one line here), and in other cases the message was lying: [here](https://github.com/Shopify/kubernetes-deploy/pull/340/files#diff-e6b2b670631bf178ac256ed1780cf5a4L97) it conditionally truncated the number of lines requested, but [here](https://github.com/Shopify/kubernetes-deploy/pull/340/files#diff-d2654750695bbb60cf39d64df9b868eeL225) it unconditionally printed that they were truncated.

Keeping the code like that made even less sense now that the possible truncation is happening in another class entirely, and by slightly different logic. The new logic is that if this is our first request, tail 250 lines; otherwise, ask for all lines since our last request (based on timestamp). So I only print this message if there are exactly 250 lines now, [here](https://github.com/Shopify/kubernetes-deploy/pull/340/files#diff-d2654750695bbb60cf39d64df9b868eeR229).

Incidentally I think 250 is way too high (we picked it kinda arbitrarily to begin with), but that's not really related to this PR.

lib/kubernetes-deploy/container_logs.rb
```diff
@@ -0,0 +1,88 @@
...
+    attr_reader :lines, :container_name
+
+    DEFAULT_LINE_LIMIT = 250
+
+    def initialize(parent_id:, container_name:, logger:)
+      @parent_id = parent_id
+      @container_name = container_name
+      @logger = logger
+      @lines = []
+      @last_printed_index = -1
```
fw42: Nitpick: If we call this something like `@next_print_index`, then you might be able to simplify things a tiny bit. This could be initialized to `0` here and below the `+1` and`-1` logic would be unnecessary, you could just do `lines[@next_print_index .. -1]` and avoid the extra `start_at` variable.

lib/kubernetes-deploy/container_logs.rb
fw42: > log output is for the benefit of the human operator only

Sadly, this has historically not been true for activefailover at least, where we would sometimes have logic that depends on what is being logged. But I agree with you here in principle and that we should move away from that kind of behaviour.

> thinking it would just be spammy to output anything about failures

Spammy because it might print this over and over in each iteration? I'm imagining a situation where there is not just a "flaky failure" that "fixes itself" but something more permanently wrong. In that case it would be frustrating for the operator to have no idea why nothing is progressing. But I'm ok with this for now.

Is there a timeout for this `kubectl.run` command? Can this theoretically block indefinitely?

lib/kubernetes-deploy/container_logs.rb
```diff
@@ -0,0 +1,88 @@
...
+    end
+
+    def sync(kubectl)
```
fw42: just out of curiosity, why is `kubectl` not an instance variable passed via the constructor?

lib/kubernetes-deploy/container_logs.rb
```diff
@@ -0,0 +1,88 @@
...
+      timestamps = []
+
+      logs.each do |line|
+        timestamp, msg = split_timestamped_line(line)
+        next if likely_duplicate?(timestamp)
+        timestamps << timestamp if timestamp
+        deduped << msg
+      end
+
+      @last_timestamp = timestamps.max
```
fw42: Since it's an array and you said the logs are ordered anyway, I think maybe `.last` instead of `.max` might be a tiny bit simpler?

lib/kubernetes-deploy/container_logs.rb
fw42: Thanks for the reminder

lib/kubernetes-deploy/kubernetes_resource.rb
```diff
@@ -209,21 +213,24 @@ def debug_message(cause = nil, info_hash = {})
...
           helpful_info << "  - Logs: #{DISABLED_LOG_INFO_MESSAGE}"
-        elsif @logs.blank? || @logs.values.all?(&:blank?)
+        elsif @debug_logs.blank?
           helpful_info << "  - Logs: #{DEBUG_RESOURCE_NOT_FOUND_MESSAGE}"
         else
-          sorted_logs = @logs.sort_by { |_, log_lines| log_lines.length }
-          sorted_logs.each do |identifier, log_lines|
-            if log_lines.empty?
-              helpful_info << "  - Logs from container '#{identifier}': #{DEBUG_RESOURCE_NOT_FOUND_MESSAGE}"
+          container_logs = @debug_logs.container_logs.sort_by { |c| c.lines.length }
```
fw42: just out of curiosity, why do we want to sort like this?

lib/kubernetes-deploy/kubernetes_resource.rb
```diff
@@ -209,21 +213,24 @@ def debug_message(cause = nil, info_hash = {})
...
+          container_logs.each do |logs|
+            if logs.empty?
+              helpful_info << "  - Logs from container '#{logs.container_name}': #{DEBUG_RESOURCE_NOT_FOUND_MESSAGE}"
```
fw42: Couldn't "empty logs" just mean that the container actually did not log anything? Looks like we are assuming that "no logs" always means "resource not found"?

lib/kubernetes-deploy/kubernetes_resource.rb
fw42: maybe `after_sync`? that's maybe a bit more in line with how other popular Ruby frameworks call lifecycle hooks like this (`after_action` in Rails, `after_enqueue` in Resque, ...).

lib/kubernetes-deploy/kubernetes_resource/pod.rb
```diff
@@ -80,28 +86,25 @@ def failure_message
...
-        out, _err, _st = kubectl.run(*cmd, log_failure: false)
-        container_logs[container.name] = out.split("\n")
-      end
+    def fetch_debug_logs(kubectl)
+      logs.sync(kubectl)
+      logs
+    end
+
+    def supports_debug_logs?
+      exists? && !@stream_logs # don't print them a second time
```
fw42: Nitpick: This seems like a bit of a hack to me.. Just because the logs have already been printed doesn't mean that this thing doesn't support logging anymore. Seems like this check should either be in a different place or this method should have a different name (`print_debug_logs?` maybe)?

lib/kubernetes-deploy/remote_logs.rb
```diff
@@ -0,0 +1,49 @@
...
+
+    def sync(kubectl)
+      @container_logs.each { |cl| cl.sync(kubectl) }
+    end
+
+    def print_latest
+      @container_logs.each { |cl| cl.print_latest(prefix: @container_logs.length > 1) }
+    end
+
+    def print_all(prevent_duplicate: true)
```
fw42: What's a use case where someone would want to set `prevent_duplicate` to `false`? (doesn't seem like that's actually being used anywhere (yet?)).

lib/kubernetes-deploy/container_logs.rb
KnVerey: Because the objects instantiating this only get `kubectl` instances passed to them in certain methods, to make it a lot harder to do the wrong thing and make API calls / change internal state at unexpected points in the polling cycle (which could cause really confusing bugs). We _could_ make each of those methods do an `@logs ||=` itself instead of sharing the instance via an accessor, but that feels weirder/messier to me. And it would let them sync their logs at arbitrary times.

lib/kubernetes-deploy/container_logs.rb
KnVerey: This comment from your review summary is related to this code:
> For activefailover, this [raising if the pod doesn't log for X seconds] would have saved our ass a few times in the past. Not saying we need this necessarily, but one thing we definitely want to be resilient against is "pod dies and therefore will never log anything, but kubernetes-deploy already started tailing the logs and will never finish". I assume your code here is already better than the code we used to have in activefailover, but having a test specifically for this scenario would be great (not necessarily in this PR, just saying).

With logs outside the control loop, that test case doesn't really make sense. Kubernetes-deploy depends on the exit status of the container to determine its result (although there are some cases where we'll bail even earlier because we've detected configuration errors), so if the pod dies, the task will for sure exit. It will grab and dump any logs that have been available up to that point, which could be all or none of them--it doesn't care.

> Sadly, this has historically not been true for activefailover at least, where we would sometimes have logic that depends on what is being logged. But I agree with you here in principle and that we should move away from that kind of behaviour.

If AF picks up this code though, won't we delete that behaviour? Why would we still need it? And even if we did, I don't see how AF could be using _this copy_ of the log stream.

> Spammy because it might print this over and over in each iteration? I'm imagining a situation where there is not just a "flaky failure" that "fixes itself" but something more permanently wrong.

Yes, and the type of errors I'm anticipating are along the lines of "logs not available yet" or "server not reachable, try again". Can you think of an example of a failure from your experience that would actually be useful to see? (keeping in mind that failures of the pod itself are tracked and reported independently of this)

> Is there a timeout for this kubectl.run command? Can this theoretically block indefinitely?

The timeout is 30 seconds: https://github.com/Shopify/kubernetes-deploy/blob/classes_for_remote_logs/lib/kubernetes-deploy/kubectl.rb#L5. Do you think we should tighten that in general, or maybe for this specific less important command?

That reminds me that kubernetes-run is a heavy user of `Kubeclient` too, and it looks like we haven't configured its timeouts on this repo (I think that means they're 60s). I'll change that too based on your answer.

lib/kubernetes-deploy/container_logs.rb
KnVerey: Did you have an idea of how this could be used to simplify the code? I can certainly see how it could make it more efficient (skip parsing a lot of timestamps), but it seems to me adding awareness of whether the loop had already passed the "outdated" threshold would be more complex, not less. Am I missing something?

lib/kubernetes-deploy/kubernetes_resource.rb
KnVerey: IIRC we added this sorting for cases where containerA dumps 250 lines of useless crap (e.g. services ping logs) and containerB dumps a shorter number of useful lines, which users didn't even notice. The idea is that even if containerB's logs are less useful, you'll at least notice you have both. I don't feel strongly about this.

lib/kubernetes-deploy/kubernetes_resource.rb
KnVerey: It specifically says `None found. Please check your usual logging service (e.g. Splunk).`. Intended sentiment is "I didn't see anything but ¯\\\_(ツ)_/¯ don't take my word for it". Maybe the container logged nothing, maybe docker is behind in writing logs to disk, maybe the kubelet was unreachable, maybe the API server erred, who knows. This is another symptom of the project overall treating logs as an unreliably available nice-to-have.

lib/kubernetes-deploy/kubernetes_resource/pod.rb
KnVerey: `print_debug_logs?` works for me

lib/kubernetes-deploy/remote_logs.rb
KnVerey: There isn't one right now. I added this because of one of your comments on Danny's original PR, something to the effect of preventing duplicate being unexpected default behaviour. I obviously misunderstood your ask though. 😆

lib/kubernetes-deploy/container_logs.rb
fw42: makes sense 👍

lib/kubernetes-deploy/container_logs.rb
fw42: > if the pod dies, the task will for sure exit

👍 

> If AF picks up this code though, won't we delete that behaviour?

yep!

> Can you think of an example of a failure from your experience that would actually be useful to see? (keeping in mind that failures of the pod itself are tracked and reported independently of this)

What we used to do is try to fetch the logs, and on error (any error), just silently try again. I do remember two failure cases specifically: (1) we retry indefinitely because we didn't realize that the reason the logs never appeared is because the container never successfully started in the first place, (2) the kubeclient `watch_pod_logs` method (which uses some kind of HTTP streaming API) never returned and blocked indefinitely (possibly because the container was actually still alive but legitimately stuck and wasn't logging anything? or maybe because of some network fuckery and kubeclient never realized that the connection was already closed?).

(2) doesn't really apply to what we are talking about here. And sounds like you have (1) covered already. So 👍.

> The timeout is 30 seconds

That's the request timeout though, right? So the time after which we give up waiting for the connection to be established? What if the connection is established successfully but then no data is returned for hours (not even a response that says "no logs yet, try again in a bit")? (Somewhat related to my case (2) above).

I'm not sure if this PR (or even this project) is the right level of abstraction to solve this kind of potential problem. So feel free to ignore for now 😄

KnVerey: The doc string for that option actually implies it is talking about the whole request, not just connect: `--request-timeout='0': The length of time to wait before giving up on a single server request.` 

I dug in and found the PR that introduced the option, [here](https://github.com/kubernetes/kubernetes/pull/33958/files). As you can see, it is setting `httpClient.Timeout = config.Timeout`, so I looked up that option in the [golang docs](https://golang.org/pkg/net/http/#Client):

```go
        // Timeout specifies a time limit for requests made by this
        // Client. The timeout includes connection time, any
        // redirects, and reading the response body. The timer remains
        // running after Get, Head, Post, or Do return and will
        // interrupt reading of the Response.Body.
```

So I think it really is a timeout on the request completing in its entirety. Do you think 30s is too generous?

lib/kubernetes-deploy/container_logs.rb
KnVerey: I added the optimization in my last commit

lib/kubernetes-deploy/container_logs.rb
```diff
@@ -0,0 +1,89 @@
...
+    def deduplicate(logs)
+      deduped = []
+      timestamps = []
+      check_for_duplicate = true
+
+      logs.each do |line|
+        timestamp, msg = split_timestamped_line(line)
+        next if check_for_duplicate && likely_duplicate?(timestamp)
+        check_for_duplicate = false # logs are ordered, so once we've seen a new one, assume all subsequent logs are new
+        timestamps << timestamp if timestamp
```
dturn: nit: I think you could just track the lastest timestamp, instead of all of them

test/unit/container_logs_test.rb
```diff
@@ -0,0 +1,141 @@
...
+  private
+
+  def generate_log_messages(range)
+    range.map { |i| "Line #{i}" }
+  end
+
+  def logs_response_1
+    # beginning of logs from second 1
+    <<~STRING
+      2018-10-04T19:40:30.997382362Z Line 1
```
dturn: These timestamps aren't hard to think about. How would you feel about the fractional seconds starting at something like: `30.00000000z` and incrementing by `0.01` or something that's much easier to compare.

KnVerey: I don't think it's worth the effort tbh. As you may have guessed, these lines are from real logs, and I kinda liked that idea. I also don't expect people to spend a lot of time reading these fixtures, and even so the subsecond details aren't really important--if you believe they're ordered correctly (which they are, because it's real data), the message acts as a clear identifier and beyond that it's the overlap at the whole-second level that's important.

lib/kubernetes-deploy/container_logs.rb
KnVerey: I just spent a couple hours debugging a weird sporadic failure on my other branch in the integration test checking for duplicates. Digging in I noticed that it was often (but not always) the last lines that were duplicated, and often (but not always) exactly 250 duplicates. Turns out that if we saw an entire batch of duplicates, the corresponding `timestamps.last` line would set `@last_timestamp` to `nil`, foiling the deduplication system. The simplest fix was to set the timestamp directly here instead of collecting them, as you suggested. (Note re: [this comment](https://github.com/Shopify/kubernetes-deploy/pull/340#discussion_r223246008):  the `check_for_duplicate` optimization makes it safe to put it back inside the loop)

KnVerey: I put the fix and regression test in the other PR btw, since that one needed it to pass CI: https://github.com/Shopify/kubernetes-deploy/pull/337/commits/6d971000b91d3cf677ad0bca5ffa8b517ca266e2. I'll merge both back-to-back tomorrow.

PR #341, opened 2018-10-08
CI for 1.12 and drop 1.8

dturn:
- 1.12 is out so we should support it even if its not on GKE. 
- Drop 1.8

### DIFF
--- a/README.md
+++ b/README.md
@@ -75,7 +75,7 @@ This repo also includes related tools for [running tasks](#kubernetes-run) and [
 * Each app managed by `kubernetes-deploy` must have its own exclusive Kubernetes namespace.
 
 <sup>1</sup> We run integration tests against these Kubernetes versions.
-Kubernetes v1.8 was officially supported in gem versions < 0.21. Kubernetes v1.7 was
+Kubernetes v1.8 was officially supported in gem versions < 0.22. Kubernetes v1.7 was
 officially supported in gem versions < 0.21. Kubernetes v1.6 was officially supported in gem
 versions < 0.16. Kubernetes v1.5 was officially supported in gem versions < 0.12.
 
--- a/README.md
+++ b/README.md
@@ -69,14 +69,15 @@ This repo also includes related tools for [running tasks](#kubernetes-run) and [
 ## Prerequisites
 
 * Ruby 2.3+
-* Your cluster must be running Kubernetes v1.8.0 or higher<sup>1</sup>
+* Your cluster must be running Kubernetes v1.9.0 or higher<sup>1</sup>
 * Each app must have a deploy directory containing its Kubernetes templates (see [Templates](#using-templates-and-variables))
 * You must remove the` kubectl.kubernetes.io/last-applied-configuration` annotation from any resources in the namespace that are not included in your deploy directory. This annotation is added automatically when you create resources with `kubectl apply`. `kubernetes-deploy` will prune any resources that have this annotation and are not in the deploy directory.<sup>2</sup>
 * Each app managed by `kubernetes-deploy` must have its own exclusive Kubernetes namespace.
 
-<sup>1</sup> We run integration tests against these Kubernetes versions. Kubernetes v1.7 was officially
-supported in gem versions < 0.21. Kubernetes v1.6 was officially supported in gem versions < 0.16. Kubernetes
-v1.5 was officially supported in gem versions < 0.12.
+<sup>1</sup> We run integration tests against these Kubernetes versions.
+Kubernetes v1.8 was officially supported in gem versions < 0.21. Kubernetes v1.7 was
+officially supported in gem versions < 0.21. Kubernetes v1.6 was officially supported in gem
+versions < 0.16. Kubernetes v1.5 was officially supported in gem versions < 0.12.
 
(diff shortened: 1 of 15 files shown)


README.md
```diff
@@ -69,14 +69,15 @@ This repo also includes related tools for [running tasks](#kubernetes-run) and [
...
+* Your cluster must be running Kubernetes v1.9.0 or higher<sup>1</sup>
 * Each app must have a deploy directory containing its Kubernetes templates (see [Templates](#using-templates-and-variables))
 * You must remove the` kubectl.kubernetes.io/last-applied-configuration` annotation from any resources in the namespace that are not included in your deploy directory. This annotation is added automatically when you create resources with `kubectl apply`. `kubernetes-deploy` will prune any resources that have this annotation and are not in the deploy directory.<sup>2</sup>
 * Each app managed by `kubernetes-deploy` must have its own exclusive Kubernetes namespace.
 
-<sup>1</sup> We run integration tests against these Kubernetes versions. Kubernetes v1.7 was officially
-supported in gem versions < 0.21. Kubernetes v1.6 was officially supported in gem versions < 0.16. Kubernetes
-v1.5 was officially supported in gem versions < 0.12.
+<sup>1</sup> We run integration tests against these Kubernetes versions.
+Kubernetes v1.8 was officially supported in gem versions < 0.21. Kubernetes v1.7 was
```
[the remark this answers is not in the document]
dturn: ```<sup>1</sup> We run integration tests against these Kubernetes versions. If you need to use an older Kubernetes version you will need to find the last supported version from the changelog```?

lib/kubernetes-deploy/container_logs.rb
```diff
@@ -72,9 +72,6 @@ def deduplicate(logs)
     def split_timestamped_line(log_line)
       timestamp, message = log_line.split(" ", 2)
       [Time.parse(timestamp), message]
-    rescue ArgumentError
```
dturn: I'm tempted to put this back. I don't like the idea of this blowing up just because of a parsing error...

README.md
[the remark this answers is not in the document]
KnVerey: You're right that the info is in the changelog, but I do think it's nicer of us to maintain a compatibility list in the readme for easy reference. I'm thinking:


image


If you like that, here's the markdown: 

| Kubernetes version | Last officially supported in gem version |
| :----------------: | :-------------------: |
|        1.5         |        0.11.2         |
|        1.6         |        0.15.2         |
|        1.7         |        0.20.6         |
|        1.8         |        0.21.1         |

test/unit/kubernetes-deploy/kubernetes_resource/stateful_set_test.rb
```diff
@@ -40,7 +40,7 @@ def build_synced_ss(template:)
     sync_mediator.kubectl.expects(:run).with("get", "Pod", "-a", "--output=json", anything).returns(
       ['{ "items": [] }', "", SystemExit.new(0)]
     )
-    sync_mediator.kubectl.expects(:server_version).returns(Gem::Version.new("1.8"))
+    sync_mediator.kubectl.expects(:server_version).returns(Gem::Version.new(KubernetesDeploy::MIN_KUBE_VERSION))
```
KnVerey: I think neither Deployment nor StatefulSet need to be actually grabbing the `version` during sync anymore. Looks like that should have been removed when we deprecated 1.7.

lib/kubernetes-deploy/container_logs.rb
```diff
@@ -73,7 +73,7 @@ def split_timestamped_line(log_line)
       timestamp, message = log_line.split(" ", 2)
       [Time.parse(timestamp), message]
     rescue ArgumentError
-      # If the log file can't be opened, k8s 1.8 writes an error message without a timestamp to stdout
+      # Don't fail on unparsable timestamp
```
KnVerey: 👍 I agree with you

README.md
```diff
@@ -69,14 +69,15 @@ This repo also includes related tools for [running tasks](#kubernetes-run) and [
...
 
-<sup>1</sup> We run integration tests against these Kubernetes versions. Kubernetes v1.7 was officially
-supported in gem versions < 0.21. Kubernetes v1.6 was officially supported in gem versions < 0.16. Kubernetes
-v1.5 was officially supported in gem versions < 0.12.
+<sup>1</sup> We run integration tests against these Kubernetes versions.
+Kubernetes v1.8 was officially supported in gem versions < 0.22. Kubernetes v1.7 was
+officially supported in gem versions < 0.21. Kubernetes v1.6 was officially supported in gem
+versions < 0.16. Kubernetes v1.5 was officially supported in gem versions < 0.12.
 
 <sup>2</sup> This requirement can be bypassed with the `--no-prune` option, but it is not recommended.
```
KnVerey: Maybe you're not done updating the readme, but in case you missed them, there are additional 1.8 references still further down.

dturn: No, I missed them. I assumed the grep hits were all coming from this paragraph.

README.md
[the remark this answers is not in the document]
dturn: I think the table is way better than the paragraph, but my concern was a table that grew forever.

PR #349, opened 2018-10-10
Mocha is not thread-safe, and unit tests should never invoke kubectl

KnVerey:
The reason for the intermittent mock failure I was seeing in CI ([example](https://buildkite.com/shopify/kubernetes-deploy-gem/builds/1050#5653d9e2-29d1-4aa6-af42-cd3e50109f6b)) is that mocha is not thread-safe: https://github.com/freerange/mocha#thread-safety

> Mocha is currently not thread-safe. There are two main reasons for this: (a) in multi-threaded code Mocha exceptions may be raised in a thread other than the one which is running the test and thus a Mocha exception may not be correctly intercepted by Mocha exception handling code; and (b) partial mocking changes the state of objects in the ObjectSpace which is shared across all threads in the Ruby process and this access to what is effectively global state is not synchronized.

This moves the integration tests that uses mocks to a new serially-run integration test file for the task runner (which required extracting shared helpers). **The test and helpers are unmodified.**

I also noticed that my PR caused a real `kubectl` call to happen during unit tests, which isn't obvious until you try to run them when your local cluster is unavailable. I added an `expects.never` on all kubectl instances in the same place where we disable net connect for unit tests. Previously many but not all suites were doing this in their individual setup.

### DIFF
--- a/test/unit/kubernetes-deploy/kubernetes_resource/pod_test.rb
+++ b/test/unit/kubernetes-deploy/kubernetes_resource/pod_test.rb
@@ -196,6 +196,7 @@ class PodTest < KubernetesDeploy::TestCase
       logger: @logger, deploy_started_at: Time.now.utc)
     mediator = KubernetesDeploy::SyncMediator.new(namespace: 'test', context: 'minikube', logger: logger)
     mediator.expects(:get_instance).with('Pod', anything).returns(template)
+    KubernetesDeploy::ContainerLogs.any_instance.stubs(:sync)
     pod.sync(mediator)
     pod
   end
--- a/test/helpers/env_test_helper.rb
+++ b/test/helpers/env_test_helper.rb
@@ -1,5 +1,7 @@
 # frozen_string_literal: true
 module EnvTestHelper
+  extend self
+
   def with_env(key, value)
     old_env_value = ENV[key]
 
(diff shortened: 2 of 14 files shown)


test/unit/kubernetes-deploy/kubernetes_resource/pod_test.rb
```diff
@@ -196,6 +196,7 @@ def build_synced_pod(template)
       logger: @logger, deploy_started_at: Time.now.utc)
     mediator = KubernetesDeploy::SyncMediator.new(namespace: 'test', context: 'minikube', logger: logger)
     mediator.expects(:get_instance).with('Pod', anything).returns(template)
+    KubernetesDeploy::ContainerLogs.any_instance.stubs(:sync)
```
karanthukral: was this the test that was invoking `kubectl`?

KnVerey: Yep! Every test using this helper (so most of them in this suite) were actually trying to sync logs.

KnVerey:
> Should we be enforcing 10m timeouts in CI?

I think 10 minutes would be too strict (if we didn't let that test complete, it would be super hard to debug), but maybe a 20min timeout would be good to have in place to guard against someone pushing a WIP branch that makes _many_ tests hang.

PR #354, opened 2018-10-18
Predeploy RoleBinding before unmanaged pods

lautis:
RoleBindings should be deployed before unmanaged pods. This matters when RoleBindings define Pod Security Policies. This might prevent unmanaged pods from starting unless RoleBinding has been set up and the deployment fails.

For normal deployments only scenario it already works. Even if there is a race condition, it works as the pods retry.

Should there be a separate resource type? Other similar resource types seemed to have also the corresponding class defined.

### DIFF
--- a/lib/kubernetes-deploy/kubernetes_resource/role_binding.rb
+++ b/lib/kubernetes-deploy/kubernetes_resource/role_binding.rb
@@ -1,0 +1,22 @@
+# frozen_string_literal: true
+module KubernetesDeploy
+  class RoleBinding < KubernetesResource
+    TIMEOUT = 30.seconds
+
+    def status
+      exists? ? "Created" : "Unknown"
+    end
+
+    def deploy_succeeded?
+      exists?
+    end
+
+    def deploy_failed?
+      false
+    end
+
+    def timeout_message
+      UNUSUAL_FAILURE_MESSAGE
+    end
+  end
+end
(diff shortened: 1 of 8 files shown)


lib/kubernetes-deploy/kubernetes_resource/role_binding.rb
```diff
@@ -0,0 +1,22 @@
...
+module KubernetesDeploy
+  class RoleBinding < KubernetesResource
+    TIMEOUT = 30.seconds
+
+    def status
+      exists? ? "Created" : "Unknown"
+    end
+
+    def deploy_succeeded?
+      exists?
```
dturn: As far as I can tell there is no status field for this resource so the success condition seems reasonable

test/integration/kubernetes_deploy_test.rb
```diff
@@ -52,6 +52,22 @@ def test_service_account_predeployed_before_unmanaged_pod
...
 
+  def test_role_binding_predeployed_before_unmanaged_pod
+    result = deploy_fixtures("hello-cloud",
+      subset: ["configmap-data.yml", "unmanaged-pod.yml.erb", "role-binding.yml", "service-account.yml"])
+
+    # Expect that role binding account is deployed before the unmanaged pod
+    assert_deploy_success(result)
+    hello_cloud = FixtureSetAssertions::HelloCloud.new(@namespace)
+    hello_cloud.assert_configmap_data_present
+    hello_cloud.assert_all_service_accounts_up
```
KnVerey: Can you please create the equivalent of this method for role bindings in `test/helpers/fixture_sets/hello_cloud.rb`, call it after this line, and also call it in `HelloCloud#assert_all_up` (that's used by `test_full_hello_cloud_set_deploy_succeeds`)?

lautis:
Rebased. I used Docker for Mac for testing from #356, which almost worked as it didn't see to like context name "docker-for-desktop" rather than "docker-for-mac". Didn't realise that it also changed the constant name.

Issue #350, opened 2018-10-10
test_jobs_can_fail is flakey and may point to a real problem


KnVerey:
I've seen the same failure multiple times in the past week. Example run: https://buildkite.com/shopify/kubernetes-deploy-gem/builds/1051#4c3290a9-6ab6-4ce6-86bf-40f44c885a0a

Notes the events at the bottom of the output below. This seems potentially indicative of a real problem. Do we hang forever if the job's deadline is exceeded? We should dig into this. @Shopify/cloudx 

```
KubernetesDeployTest#test_jobs_can_fail [/usr/src/app/test/integration/kubernetes_deploy_test.rb:944]
--
  | Minitest::Assertion: 'Result: FAILURE' not found in the following logs:
  | [INFO][2018-10-10 20:52:22 +0000]
  | [INFO][2018-10-10 20:52:22 +0000]	------------------------------------Phase 1: Initializing deploy------------------------------------
  | [INFO][2018-10-10 20:52:22 +0000]	All required parameters and files are present
  | [INFO][2018-10-10 20:52:22 +0000]	Context minikube found
  | [INFO][2018-10-10 20:52:23 +0000]	Namespace k8sdeploy-test-jobs-can-fail-6e34dddc0ed62388 found
  | [INFO][2018-10-10 20:52:23 +0000]	Discovering templates:
  | [INFO][2018-10-10 20:52:23 +0000]	  - Job/hello-job
  | [INFO][2018-10-10 20:52:23 +0000]
  | [INFO][2018-10-10 20:52:23 +0000]	----------------------------Phase 2: Checking initial resource statuses-----------------------------
  | [INFO][2018-10-10 20:52:23 +0000]	Job/hello-job                                     Unknown
  | [INFO][2018-10-10 20:52:24 +0000]
  | [INFO][2018-10-10 20:52:24 +0000]	----------------------------------Phase 3: Deploying all resources----------------------------------
  | [INFO][2018-10-10 20:52:24 +0000]	Deploying Job/hello-job (timeout: 600s)
  | [INFO][2018-10-10 20:52:54 +0000]	Still waiting for: Job/hello-job
  | [INFO][2018-10-10 20:53:25 +0000]	Still waiting for: Job/hello-job
  | [INFO][2018-10-10 20:53:55 +0000]	Still waiting for: Job/hello-job
  | [INFO][2018-10-10 20:54:25 +0000]	Still waiting for: Job/hello-job
[...]

PR #355, opened 2018-10-23
Check jobs for failure status condition

dturn:
test_jobs_can_fail is flakey due to not checking the status condition of jobs. So let's check it.

fixes: https://github.com/Shopify/kubernetes-deploy/issues/350


lib/kubernetes-deploy/kubernetes_resource/job.rb
```diff
@@ -13,6 +13,9 @@ def deploy_succeeded?
     def deploy_failed?
       return false unless deploy_started?
       return false unless @instance_data.dig("spec", "backoffLimit").present?
+      (@instance_data.dig("status", "conditions") || []).each do |condition|
+        return true if condition["type"] == 'Failed' && condition['status'] == "True"
+      end
```
KnVerey: Nit: can you extract this into a method?

This condition also has richer information about the reason for the failure in its reason/message fields. Let's adjust `failure_message` to include that information if we have it.

Can you also add tests? Maybe unit tests that show `deploy_failed?` is true with either of the statuses I discovered are possible on the issue?

lib/kubernetes-deploy/kubernetes_resource/job.rb
```diff
@@ -30,8 +31,20 @@ def status
...
+    def failure_message
+      if (condition = failed_status_condition.presence)
+        "#{condition['reason']} (#{condition['message']})"
+      end
+    end
+
     private
 
+    def failed_status_condition
+      (@instance_data.dig("status", "conditions") || []).detect do |condition|
```
KnVerey: nit: this would be easier to read if it used a guard clause instead of the `|| []` trick. I personally prefer `find` over `detect`, but I see we already use both in this repo.

lib/kubernetes-deploy/kubernetes_resource/job.rb
```diff
@@ -30,8 +31,20 @@ def status
       end
     end
 
+    def failure_message
+      if (condition = failed_status_condition.presence)
```
KnVerey: Why the parentheses?

test/unit/kubernetes-deploy/kubernetes_resource/job_test.rb
```diff
@@ -0,0 +1,58 @@
...
+    job = build_synced_job(job_spec.merge(status).deep_stringify_keys)
+
+    assert_predicate job, :deploy_failed?
+
+    expected_msg = "BackoffLimitExceeded (Job has reached the specified backoff limit)"
+    assert_equal expected_msg, job.failure_message
+  end
+
+  def test_job_fails_without_failed_status_condition
+    status = {
```
KnVerey: nit: the data this contains isn't a status in this test

test/unit/kubernetes-deploy/kubernetes_resource/job_test.rb
```diff
@@ -0,0 +1,58 @@
...
+
+  def job_spec
+    { metadata: { name: 'test-job' } }
+  end
+
+  def build_synced_job(template)
+    job = KubernetesDeploy::Job.new(namespace: 'test', context: 'nope', definition: template,
+      logger: @logger)
+    job.deploy_started_at = Time.now.utc
+    mediator = KubernetesDeploy::SyncMediator.new(namespace: 'test', context: 'minikube', logger: logger)
```
KnVerey: nit: I don't think this could change anything, but you're using two different fake context names here and two lines above this (and accessing the logger differently)

lib/kubernetes-deploy/kubernetes_resource/job.rb
dturn: Its a habit from a style checker, but basically saying I'm intentionally using `=`  and didn't mean to use `==`

test/unit/kubernetes-deploy/kubernetes_resource/job_test.rb
dturn: This was actually copied from the pod test, I'll fix both.

dturn:
Screen shot of the error message:
screen shot 2018-10-24 at 11 02 58 am

PR #356, opened 2018-10-24
Support test clusters other than minikube

KnVerey:
Makes it possible to run the tests against a docker-in-docker cluster for example (which I've been using lately). To do that, you create a `.local-context` file (which is git ignored) containing the name of the context you want to use. This is similar to what we did for Cloud Portal.

@Shopify/cloudx for review


test/integration/kubernetes_deploy_test.rb
```diff
@@ -1078,4 +1080,14 @@ def test_not_apply_resource_can_be_pruned
...
     ])
   end
+
+  private
+
+  def expected_daemonset_pod_count
+    nodes = kubeclient.get_nodes
+    return 1 if nodes.one?
+    nodes.count do |node|
+      !node.metadata.labels.to_h.keys.include?(:"node-role.kubernetes.io/master")
```
dturn: dind gives you a multi node setup?

KnVerey: Yep! Two nodes by default. Here's the repo for it FYI: https://github.com/kubernetes-sigs/kubeadm-dind-cluster

PR #357, opened 2018-10-24
Fix problems with unit tests

KnVerey:
- Makes it possible to run the unit tests without a local k8s cluster running
- Fixes sporadic UDP errors that get logged from unit tests
- Makes unit test 4x faster by tuning some unnecessarily long sleeps used in the slowest tests (based on the list we print on every run)
 - Fixes two naming errors (`docker-for-mac` should have been `docker-for-desktop`, and the rbac client we're using is v1 not v1beta1)
- Removes the "valid_config.yml" fixture, which isn't really needed and furthermore isn't a valid config!

### DIFF
--- a/test/helpers/test_provisioner.rb
+++ b/test/helpers/test_provisioner.rb
@@ -1,0 +1,73 @@
+# frozen_string_literal: true
+
+require_relative './kubeclient_helper'
+
+class TestProvisioner
+  extend KubeclientHelper
+
+  class << self
+    def prepare_cluster
+      WebMock.allow_net_connect!
+      $stderr.print "Preparing test cluster... "
+      prepare_pv("pv0001")
+      prepare_pv("pv0002")
+      deploy_metric_server
+      $stderr.puts "Done."
+      WebMock.disable_net_connect!
+    end
+
+    def claim_namespace(test_name)
+      test_name = test_name.gsub(/[^-a-z0-9]/, '-').slice(0, 36) # namespace name length must be <= 63 chars
+      ns = "k8sdeploy-#{test_name}-#{SecureRandom.hex(8)}"
+      create_namespace(ns)
+      ns
+    rescue KubeException => e
+      retry if e.to_s.include?("already exists")
+      raise
+    end
+
+    def delete_namespace(namespace)
+      kubeclient.delete_namespace(namespace) if namespace.present?
+    rescue KubeException => e
+      raise unless e.to_s.include?("not found")
+    end
+
+    private
+
+    def create_namespace(namespace)
+      ns = Kubeclient::Resource.new(kind: 'Namespace')
+      ns.metadata = { name: namespace }
+      kubeclient.create_namespace(ns)
+    end
+
+    def prepare_pv(name)
+      existing_pvs = kubeclient.get_persistent_volumes(label_selector: "name=#{name}")
+      return if existing_pvs.present?
+
+      pv = Kubeclient::Resource.new(kind: 'PersistentVolume')
+      pv.metadata = {
+        name: name,
+        labels: { name: name }
+      }
+      pv.spec = {
+        accessModes: %w(ReadWriteOnce),
+        capacity: { storage: "150Mi" },
+        hostPath: { path: "/data/#{name}" },
+        persistentVolumeReclaimPolicy: "Recycle"
+      }
+      kubeclient.create_persistent_volume(pv)
+    end
+
+    def deploy_metric_server
+      # Set-up the metric server that the HPA needs https://github.com/kubernetes-incubator/metrics-server
+      logger = KubernetesDeploy::FormattedLogger.build("default", KubeclientHelper::TEST_CONTEXT, $stderr)
+      kubectl = KubernetesDeploy::Kubectl.new(namespace: "kube-system", context: KubeclientHelper::TEST_CONTEXT,
+        logger: logger, log_failure_by_default: true, default_timeout: '5s')
+
+      Dir.glob("test/setup/metrics-server/*.{yml,yaml}*").map do |resource|
+        found = kubectl.run("get", "-f", resource, log_failure: false).last.success?
+        kubectl.run("create", "-f", resource) unless found
+      end
+    end
+  end
+end
--- a/test/helpers/test_provisioner.rb
+++ b/test/helpers/test_provisioner.rb
@@ -17,13 +17,13 @@ class TestProvisioner
     end
 
     def claim_namespace(test_name)
-      test_name = test_name.gsub(/[^-a-z0-9]/, '-').slice(0, 36) # namespace name length must be <= 63 chars
-      ns = "k8sdeploy-#{test_name}-#{SecureRandom.hex(8)}"
-      create_namespace(ns)
-      ns
-    rescue KubeException => e
-      retry if e.to_s.include?("already exists")
-      raise
+      prefix = "k8sdeploy-"
+      suffix = "-#{SecureRandom.hex(8)}"
+      max_base_length = 63 - (prefix + suffix).length # namespace name length must be <= 63 chars
+      ns_name = prefix + test_name.gsub(/[^-a-z0-9]/, '-').slice(0, max_base_length) + suffix
+
+      create_namespace(ns_name)
+      ns_name
     end
 
     def delete_namespace(namespace)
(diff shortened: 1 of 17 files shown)


test/unit/kubernetes-deploy/statsd_test.rb
```diff
@@ -18,5 +18,6 @@ def test_build_when_statsd_addr_env_present_but_statsd_implementation_is_not
     ENV['STATSD_ADDR'] = original_addr
     ENV['STATSD_IMPLEMENTATION'] = original_impl
     ENV['STATSD_DEV'] = original_dev
+    KubernetesDeploy::StatsD.build
```
KnVerey: This sets global state on `::StatsD`, so if we don't redo it after resetting the env vars, the other tests will try to use the UDP backend and throw errors (because this test doesn't actually fully configure it).

test/unit/kubernetes-deploy/kubeclient_builder_test.rb
```diff
@@ -13,17 +13,17 @@ def setup
...
+    ENV['KUBECONFIG'] = "#{old_config}:#{dummy_config}"
     # Build kubeclient for an unknown context fails
     context_name = "unknown_context"
     assert_raises_message(ContextMissingError,
       "`#{context_name}` context must be configured in your KUBECONFIG file(s) " \
       "(#{ENV['KUBECONFIG']}).") do
       build_v1_kubeclient(context_name)
     end
-    # Build kubeclient for an existing context success
-    context_name = KubeclientHelper::TEST_CONTEXT
```
KnVerey: This test would fail if the real test context exists in your kubeconfig but points to a cluster that is not fully configured (e.g. it failed for me trying to get auth credentials for minikube). By using the dummy config, we can ensure the context will exist and will not require additional external calls when you create a client for it.

test/integration_test_helper.rb
```diff
@@ -0,0 +1,46 @@
+# frozen_string_literal: true
+
+require 'test_helper'
+
+module KubernetesDeploy
+  class IntegrationTest < KubernetesDeploy::TestCase
+    include KubeclientHelper
+    include FixtureDeployHelper
+
+    TestProvisioner.prepare_cluster
```
KnVerey: This is the most important part. With this call moved into a file we only load when running the integration suite, we no longer make unnecessary API calls during unit tests. I'm still doing it on file load because it should happen only once even if we're running the parallel integration suite. Can you think of a nicer way?

fw42:
> Removes the "valid_config.yml" fixture, which isn't (..) a valid config

😆

test/helpers/test_provisioner.rb
```diff
@@ -0,0 +1,73 @@
...
+      $stderr.print "Preparing test cluster... "
+      prepare_pv("pv0001")
+      prepare_pv("pv0002")
+      deploy_metric_server
+      $stderr.puts "Done."
+      WebMock.disable_net_connect!
+    end
+
+    def claim_namespace(test_name)
+      test_name = test_name.gsub(/[^-a-z0-9]/, '-').slice(0, 36) # namespace name length must be <= 63 chars
```
fw42: nitpick: Took me a second to understand why 36 is correct here, could be a tiny bit clearer

test/helpers/test_provisioner.rb
```diff
@@ -0,0 +1,73 @@
...
+      ns = "k8sdeploy-#{test_name}-#{SecureRandom.hex(8)}"
+      create_namespace(ns)
+      ns
+    rescue KubeException => e
+      retry if e.to_s.include?("already exists")
```
fw42: Is this correct? Did you mean `unless` instead of `if`? Also, it seems like this might retry indefinitely and the user might wonder why it's stuck if it never resolves. Is that a concern?

test/helpers/test_provisioner.rb
```diff
@@ -0,0 +1,73 @@
...
+    end
+
+    def deploy_metric_server
+      # Set-up the metric server that the HPA needs https://github.com/kubernetes-incubator/metrics-server
+      logger = KubernetesDeploy::FormattedLogger.build("default", KubeclientHelper::TEST_CONTEXT, $stderr)
+      kubectl = KubernetesDeploy::Kubectl.new(namespace: "kube-system", context: KubeclientHelper::TEST_CONTEXT,
+        logger: logger, log_failure_by_default: true, default_timeout: '5s')
+
+      Dir.glob("test/setup/metrics-server/*.{yml,yaml}*").map do |resource|
+        found = kubectl.run("get", "-f", resource, log_failure: false).last.success?
```
fw42: how come you sometimes use `kubectl` and sometimes `kubeclient`? How do you decide when to use which?

test/integration_test_helper.rb
fw42: I'd probably have put this into some kind of `IntegrationTest#before_setup` method (and then memoize that it run to make sure it doesn't run more than once). Maybe unlikely but in theory you can load the code for the integration tests but still only run the unit tests, and in that case your strategy here wouldn't work (because it assumes that integration tests run if and only if they were loaded).

test/integration_test_helper.rb
```diff
@@ -0,0 +1,46 @@
...
+
+    if ENV["PARALLELIZE_ME"]
+      puts "Running tests in parallel!"
+      parallelize_me!
+    end
+
+    def run
+      WebMock.allow_net_connect!
```
fw42: This is never turned off again? If you always want this on, I'd probably move this to `test_helper` or something.

test/helpers/test_provisioner.rb
```diff
@@ -0,0 +1,73 @@
...
+
+require_relative './kubeclient_helper'
+
+class TestProvisioner
+  extend KubeclientHelper
+
+  class << self
+    def prepare_cluster
+      WebMock.allow_net_connect!
+      $stderr.print "Preparing test cluster... "
```
dturn: Why stderr?

KnVerey: STDERR is the normal descriptor for diagnostic messages. Kir and Simon taught me about this when we started the gem: https://www.jstorimer.com/blogs/workingwithcode/7766119-when-to-use-stderr-instead-of-stdout. Most of this gem's logging has been done to STDERR since https://github.com/Shopify/kubernetes-deploy/pull/5/files#r97137080; our "result" is pretty much just the exit code.

test/helpers/test_provisioner.rb
KnVerey: Some things we use are features of `kubectl` specifically rather than Kubernetes (most notably the `apply` verb, for now anyway), which forces the decision. Originally everything used kubectl because of that. However, kubeclient is decent now, and it is often cleaner to use it for the things it does support. So I slightly favour Kubeclient unless using it would make the code a lot more complicated, which is the case here (we'd need to load/parse the file instead of just passing its name as an arg).

That said, this was a copy-paste and we have an open issue to improve this code by paying more attention to the resources it creates. That will likely require us to parse the files anyway, so perhaps we'll switch it to Kubeclient after all.

test/integration_test_helper.rb
KnVerey: Isn't that memoization ineffective if we're running tests in parallel, given that the `prepare_cluster` operation can take some time? I guess I can use a mutex to address that, but seems a bit complicated compared to the assumption that this file won't be loaded by unit tests (which as you say is probably pretty safe in practice).

I checked whether minitest has a `before_all`, and apparently the maintainers refused to provide such a hook because you can do what we're doing here instead: https://github.com/seattlerb/minitest/issues/61

test/integration_test_helper.rb
KnVerey: I want it on for all integration tests and off for all unit tests. Moving to the superclass's `run` in test_helper to make it more obvious what is happening.

test/test_helper.rb
```diff
@@ -57,45 +57,48 @@
...
 
-    def logger
-      @logger ||= begin
-        device = log_to_stderr? ? $stderr : @logger_stream
-        KubernetesDeploy::FormattedLogger.build(@namespace, KubeclientHelper::TEST_CONTEXT, device)
-      end
-    end
-
-    def teardown
-      @logger_stream.close
```
KnVerey: This doesn't actually do anything useful. It's just a string and we don't try to use it after teardown.

PR #374, opened 2018-11-02
Add statsd to k8s-run

dturn:
Add statsd calls to aid in tracking useage.

### DIFF
--- a/lib/kubernetes-deploy/runner_task.rb
+++ b/lib/kubernetes-deploy/runner_task.rb
@@ -28,6 +28,7 @@ module KubernetesDeploy
     end
 
     def run!(task_template:, entrypoint:, args:, env_vars: [], verify_result: true)
+      start = Time.now.utc
       @logger.reset
 
       @logger.phase_heading("Initializing task")
@@ -44,11 +45,14 @@ module KubernetesDeploy
       else
         record_status_once(pod)
       end
+      ::StatsD.distribution('task_runner.duration', StatsD.duration(start), tags: statsd_tags('success'))
       @logger.print_summary(:success)
     rescue DeploymentTimeoutError
+      ::StatsD.distribution('task_runner.duration', StatsD.duration(start), tags: statsd_tags('timeout'))
       @logger.print_summary(:timed_out)
       raise
     rescue FatalDeploymentError
+      ::StatsD.distribution('task_runner.duration', StatsD.duration(start), tags: statsd_tags('failure'))
       @logger.print_summary(:failure)
       raise
     end
@@ -199,6 +203,10 @@ module KubernetesDeploy
 
     def kubeclient
       @kubeclient ||= build_v1_kubeclient(@context)
+    end
+
+    def statsd_tags(status)
+      %W(namespace:#{@namespace} context:#{@context} status:#{status})
     end
   end
 end
--- a/test/integration-serial/serial_task_run_test.rb
+++ b/test/integration-serial/serial_task_run_test.rb
@@ -19,7 +19,6 @@ class SerialTaskRunTest < KubernetesDeploy::IntegrationTest
     task_runner.instance_variable_set(:@kubeclient, mock)
 
     result = task_runner.run(run_params(verify_result: false))
-
     assert_task_run_failure(result)
 
     assert_logs_match_all([
@@ -28,5 +27,58 @@ class SerialTaskRunTest < KubernetesDeploy::IntegrationTest
       "Failed to create pod",
       "Kubeclient::HttpError: Pod with same name exists"
     ], in_order: true)
+  end
+
+  # We can't ensure that all the metrics are specific test when they are running in parallel
+  # so we run them serially
+  def test_failure_statsd_metric_emitted
+    deploy_task_template
+    task_runner = build_task_runner
+
+    result = false
+    metrics = capture_statsd_calls do
+      result = task_runner.run(run_params.merge(args: ["/not/a/command"]))
+    end
+
+    assert_task_run_failure(result)
+
+    metric_tags = metrics.detect { |m| m.tags.include? "namespace:#{@namespace}" }.tags
+    assert_includes metric_tags, "namespace:#{@namespace}"
+    assert_includes metric_tags, "context:#{KubeclientHelper::TEST_CONTEXT}"
+    assert_includes metric_tags, "status:failure"
+  end
+
+  def test_success_statsd_metric_emitted
+    deploy_task_template
+    task_runner = build_task_runner
+
+    result = false
+    metrics = capture_statsd_calls do
+      result = task_runner.run(run_params.merge(args: %w(ls)))
+    end
+
+    assert_task_run_success(result)
+
+    metric_tags = metrics.detect { |m| m.tags.include? "namespace:#{@namespace}" }.tags
+    assert_includes metric_tags, "namespace:#{@namespace}"
+    assert_includes metric_tags, "context:#{KubeclientHelper::TEST_CONTEXT}"
+    assert_includes metric_tags, "status:success"
+  end
+
+  def test_timedout_statsd_metric_emitted
+    deploy_task_template
+    task_runner = build_task_runner(max_watch_seconds: 5)
+
+    result = false
+    metrics = capture_statsd_calls do
+      result = task_runner.run(run_params.merge(args: ["sleep 600"]))
+    end
+
+    assert_task_run_failure(result, :timed_out)
+
+    metric_tags = metrics.detect { |m| m.tags.include? "namespace:#{@namespace}" }.tags
+    assert_includes metric_tags, "namespace:#{@namespace}"
+    assert_includes metric_tags, "context:#{KubeclientHelper::TEST_CONTEXT}"
+    assert_includes metric_tags, "status:timeout"
   end
 end
--- a/test/integration-serial/serial_task_run_test.rb
+++ b/test/integration-serial/serial_task_run_test.rb
@@ -29,21 +29,22 @@ class SerialTaskRunTest < KubernetesDeploy::IntegrationTest
     ], in_order: true)
   end
 
-  # We can't ensure that all the metrics are specific test when they are running in parallel
-  # so we run them serially
+  # Run statsd tests in serial because capture_statsd_calls modifies global state in a way
+  # that makes capturing metrics across parrallel runs unreliable
   def test_failure_statsd_metric_emitted
-    deploy_task_template
-    task_runner = build_task_runner
+    bad_ns = "missing"
+    task_runner = build_task_runner(ns: bad_ns)
 
     result = false
     metrics = capture_statsd_calls do
-      result = task_runner.run(run_params.merge(args: ["/not/a/command"]))
+      result = task_runner.run(run_params)
     end
 
     assert_task_run_failure(result)
 
-    metric_tags = metrics.detect { |m| m.tags.include? "namespace:#{@namespace}" }.tags
-    assert_includes metric_tags, "namespace:#{@namespace}"
+    metric = metrics.detect { |m| m.tags.include? "namespace:#{bad_ns}" }
+    assert metric, "No metrics found for this test"
+    metric_tags = metric.tags
     assert_includes metric_tags, "context:#{KubeclientHelper::TEST_CONTEXT}"
     assert_includes metric_tags, "status:failure"
   end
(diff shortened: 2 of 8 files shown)


lib/kubernetes-deploy/runner_task.rb
```diff
@@ -44,11 +45,14 @@ def run!(task_template:, entrypoint:, args:, env_vars: [], verify_result: true)
...
         record_status_once(pod)
       end
+      ::StatsD.distribution('run.duration', StatsD.duration(start), tags: tags('success'))
       @logger.print_summary(:success)
     rescue DeploymentTimeoutError
+      ::StatsD.distribution('run.duration', StatsD.duration(start), tags: tags('timeout'))
       @logger.print_summary(:timed_out)
       raise
     rescue FatalDeploymentError
+      ::StatsD.distribution('restart.duration', StatsD.duration(start), tags: tags('failure'))
```
KnVerey: `run.duration` to match the others

Or maybe we should use `task_runner` or something to be more explicit that it wasn't a deploy or a restart running?

lib/kubernetes-deploy/runner_task.rb
```diff
@@ -44,11 +45,14 @@ def run!(task_template:, entrypoint:, args:, env_vars: [], verify_result: true)
       else
         record_status_once(pod)
       end
+      ::StatsD.distribution('run.duration', StatsD.duration(start), tags: tags('success'))
```
KnVerey: why `distribution` instead of `measure`?

lib/kubernetes-deploy/runner_task.rb
```diff
@@ -200,5 +204,9 @@ def sync_mediator
     def kubeclient
       @kubeclient ||= build_v1_kubeclient(@context)
     end
+
+    def tags(status)
```
KnVerey: nit: can we call this `statsd_tags` like in DeployTask? It's very much a statsd thing.

test/integration-serial/serial_task_run_test.rb
```diff
@@ -18,7 +18,10 @@ def test_run_without_verify_result_fails_if_pod_was_not_created
     mock.expects(:create_pod).raises(Kubeclient::HttpError.new("409", "Pod with same name exists", {}))
     task_runner.instance_variable_set(:@kubeclient, mock)
 
-    result = task_runner.run(run_params(verify_result: false))
+    result = false
+    metrics = capture_statsd_calls do
+      result = task_runner.run(run_params(verify_result: false))
+    end
```
KnVerey: Why in this serialized, stubbed test instead of one of the basic ones in `test/integration/runner_task_test.rb`? Shouldn't we add similar assertions to a timeout and a success case too?

lib/kubernetes-deploy/runner_task.rb
dturn: Good question, I only learned this when policial yelled at me. I believe its because https://github.com/Shopify/statsd-instrument#statsddistribution is more efficient (/cost effective)

test/integration-serial/serial_task_run_test.rb
dturn: `capture_statsd_calls` isn't thread safe

lib/kubernetes-deploy/runner_task.rb
KnVerey: Hm, those docs say `This is only supported by the beta datadog implementatation.`--is that going to work for everyone? What was the Policial error exactly? Reading the [DD docs](https://docs.datadoghq.com/developers/metrics/distributions/) about this, I don't really see why Distribution would be the right fit for this kind of metric. These will be really sparse, and we can't do any sampling of them. If we _do_ need to switch, then shouldn't we switch all of our equivalent metrics too?

test/integration-serial/serial_task_run_test.rb
KnVerey: I see we're doing it in the main file though, and using the namespace tag to identify metrics that were actually emitted by the test in question (since each test gets its own namespace). Can't we do that for these tests too? The fact that we're lacking coverage in the other suite isn't a good reason not to add good coverage here.

lib/kubernetes-deploy/runner_task.rb
dturn: > Shopify/Datadog: The measure and histogram APIs are no longer the accepted way to submit metrics to Datadog with statsd-instrument. Please prefer StatsD.distribution - a modified guage that submits a distribution of values of a sample period. All arithmetic and calculation happens server side and saves Shopify tons of money on Datadog metric submission counts. See https://github.com/Shopify/statsd-instrument#statsddistribution and https://www.datadoghq.com/blog/summary-graphs-metric-graphs-101/#distributions

dturn: > Hm, those docs say This is only supported by the beta datadog implementatation.--is that going to work for everyone? 

I think so, though other people will  have to talk to DD to get it turned on while it's in beta. 

>  Reading the DD docs about this, I don't really see why Distribution would be the right fit for this kind of metric. These will be really sparse, and we can't do any sampling of them. If we do need to switch,

I didn't see anything about them being sparse. Can you link to where you saw that? What I read (https://docs.datadoghq.com/graphing/metrics/distributions/) makes it look like you have to be explicit in DD itself about which tags from built-in integrations we care about.  

>  then shouldn't we switch all of our equivalent metrics too?

Probably, there are only 9 of them.

test/integration-serial/serial_task_run_test.rb
dturn: Yeah, I misunderstood the point of that comment. 

> The fact that we're lacking coverage in the other suite isn't a good reason not to add good coverage here.

Agree, but given the high quality of the test coverage in the project when I lack context I usually assume these things are intentional. e.g. that the level of coverage was  intentional and reflected that amount of testing needed here.

lib/kubernetes-deploy/runner_task.rb
KnVerey: > I think so, though other people will have to talk to DD to get it turned on while it's in beta.

I watched the internal video about distributions, and now don't see any internal reason not to switch. But forcing other companies to enable a beta feature on their Datadog account to be able to emit metrics from our gem feels wrong to me.

test/integration-serial/serial_task_run_test.rb
KnVerey: Ok, so you were definitely right that `capture_statsd_calls` _itself_ isn't threadsafe. It's actually [modifying the global statsd configuration](https://github.com/Shopify/statsd-instrument/blob/273017153aa915ac3f02558d1ed691d2a8d68e2f/lib/statsd/instrument/helpers.rb). I'd suggest we make a statsD test in the serial suite (where you had it originally) that covers the three possible results using configurations that succeed/fail/time out as quickly as possible.

lib/kubernetes-deploy/runner_task.rb
KnVerey: Ah, ok. In that case I'm ok with switching them all. This might still count as a breaking change though, since if I understand correctly it will break dashboards until the distribution is configured in Datadog to produce the aggregations we need (and there's no backfill, so we need to be sure to do that right away).

test/integration/kubernetes_deploy_test.rb
```diff
@@ -232,8 +232,7 @@ def test_invalid_k8s_spec_that_is_valid_yaml_fails_fast_and_prints_template
       "Template validation failed",
       /Invalid template: ConfigMap-hello-cloud-configmap-data.*yml/,
       "> Error message:",
-      "error validating data: ValidationError(ConfigMap.metadata): \
-unknown field \"myKey\" in io.k8s.apimachinery.pkg.apis.meta.v1.ObjectMeta",
+      /error validating data: .*myKey.* v1.ObjectMeta/,
```
dturn: My error message locally was just enough different to be annoying:
``` error: error validating "/var/folders/pk/7tjdz4295zl0t0_h8dlp99z40000gp/T/ConfigMap-hello-cloud-configmap-data20181107-23607-xe3tdb.yml": error validating data: found invalid field myKey for v1.ObjectMeta; if you choose to ignore these errors, turn validation off with --validate=false```

Comment (KnVerey) [reply test/integration/kubernetes_deploy_test.rb]: Which k8s version is that from? I'm surprise our CI hasn't caught that possibility. The `v1.ObjectMeta` part isn't so important as the "unknown field myKey" / "invalid field myKey"

File: test/integration-serial/serial_task_run_test.rb [hunk 6]
```diff
@@ -28,4 +28,57 @@ def test_run_without_verify_result_fails_if_pod_was_not_created
...
+
+  # We can't ensure that all the metrics are specific test when they are running in parallel
+  # so we run them serially
+  def test_failure_statsd_metric_emitted
+    deploy_task_template
+    task_runner = build_task_runner
+
+    result = false
+    metrics = capture_statsd_calls do
+      result = task_runner.run(run_params.merge(args: ["/not/a/command"]))
```
KnVerey: Can we use a failure scenario that fails validations and doesn't launch the pod at all? The idea is to keep these serial tests as quick as possible, and this one doesn't really need to launch anything to be valid.

test/integration-serial/serial_task_run_test.rb
```diff
@@ -28,4 +28,57 @@ def test_run_without_verify_result_fails_if_pod_was_not_created
...
+    end
+
+    assert_task_run_failure(result)
+
+    metric_tags = metrics.detect { |m| m.tags.include? "namespace:#{@namespace}" }.tags
+    assert_includes metric_tags, "namespace:#{@namespace}"
```
KnVerey: This is redundant given the filter above, no?  (same in the other tests)

test/integration/kubernetes_deploy_test.rb
dturn:
``` ➜ k version                                                                                                                 
Client Version: version.Info{Major:"1", Minor:"7", GitVersion:"v1.7.8", GitCommit:"bc6162cc70b4a39a7f39391564e0dd0be60b39e9", GitTreeState:"clean", BuildDate:"2017-10-05T06:54:19Z", GoVersion:"go1.8.3", Compiler:"gc", Platform:"darwin/amd64"}
Server Version: version.Info{Major:"1", Minor:"10", GitVersion:"v1.10.6", GitCommit:"a21fdbd78dde8f5447f5f6c331f7eb6f80bd684e", GitTreeState:"clean", BuildDate:"2018-07-26T10:04:08Z", GoVersion:"go1.9.3", Compiler:"gc", Platform:"linux/amd64"}```

File: test/integration-serial/serial_task_run_test.rb [hunk 8]
```diff
@@ -28,4 +28,57 @@ def test_run_without_verify_result_fails_if_pod_was_not_created
...
+    task_runner = build_task_runner
+
+    result = false
+    metrics = capture_statsd_calls do
+      result = task_runner.run(run_params.merge(args: ["/not/a/command"]))
+    end
+
+    assert_task_run_failure(result)
+
+    metric_tags = metrics.detect { |m| m.tags.include? "namespace:#{@namespace}" }.tags
```
KnVerey: I noticed in your previous version that when this fails, it fails with a no method error because of `nil.tags`. Can we change it to fail with a nice "No metrics found for this test" message?

test/integration-serial/serial_task_run_test.rb
```diff
@@ -28,4 +28,57 @@ def test_run_without_verify_result_fails_if_pod_was_not_created
...
+    assert_includes metric_tags, "status:failure"
+  end
+
+  def test_success_statsd_metric_emitted
+    deploy_task_template
+    task_runner = build_task_runner
+
+    result = false
+    metrics = capture_statsd_calls do
+      result = task_runner.run(run_params.merge(args: %w(ls)))
```
KnVerey: similar to my comment above, can we run this with `verify_result: false` to speed it up? We don't really care how it succeeds, just that it does.

test/integration-serial/serial_task_run_test.rb
```diff
@@ -28,4 +28,57 @@ def test_run_without_verify_result_fails_if_pod_was_not_created
...
+
+    metric_tags = metrics.detect { |m| m.tags.include? "namespace:#{@namespace}" }.tags
+    assert_includes metric_tags, "namespace:#{@namespace}"
+    assert_includes metric_tags, "context:#{KubeclientHelper::TEST_CONTEXT}"
+    assert_includes metric_tags, "status:success"
+  end
+
+  def test_timedout_statsd_metric_emitted
+    deploy_task_template
+    task_runner = build_task_runner(max_watch_seconds: 5)
```
KnVerey: Similar to other comments, why not make this 1 second, or even shorter?

test/integration-serial/serial_task_run_test.rb
```diff
@@ -28,4 +28,57 @@ def test_run_without_verify_result_fails_if_pod_was_not_created
...
+
+    result = false
+    metrics = capture_statsd_calls do
+      result = task_runner.run(run_params.merge(args: ["sleep 600"]))
```
KnVerey: Let's make this shorter. We wouldn't ever want it to hang around for five minutes.

test/integration/kubernetes_deploy_test.rb
KnVerey: Can you try upgrading your client? We don't support 1.7 anymore.

dturn: Yep that fixed it. I expect that error to be from the server not client.

KnVerey: Yeah, dry-run is currently client-side. There's an alpha in 1.12 for server-side dry run: https://kubernetes.io/docs/reference/using-api/api-concepts/#dry-run.

dturn:
I've limited the changes to k8s restart & run. Will do deploy later. Last chance for 👀

KnVerey:
👍

Issue #182, opened 2017-09-29
Dry-run?


alexcurtin:
Can `kubernetes-deploy` gem support a `dry-run` flag? so before it runs it would test the dry-run as a way to lint your yamls for kubernetes before you deploy for real?

KnVerey:
We actually already run `kubectl create --dry-run` on all the templates before starting the deploy, [here](https://github.com/Shopify/kubernetes-deploy/blob/7da5ab3a76586cc775f6f86d1f9610eb44712c8b/lib/kubernetes-deploy/runner.rb#L186-L195). If that identifies any template as invalid, the deploy will not proceed. Here's an example of what that looks like:

![image](https://user-images.githubusercontent.com/4789493/31032554-71631a08-a52a-11e7-9f7f-0ef9cece6653.png)

Is that sufficient for your use case, or are you hoping to use this dry run flag at a separate stage of your CI/CD?

alexcurtin:
That should be sufficient thanks!

airhorns:
*peep peep* I think that this feature might still be useful for just validating templates before actually triggering a deploy, for example in branches that shouldn't be deployed but will eventually be deployed. My specific use case is that I am porting a bunch of other k8s yaml to `kubernetes-deploy` and I want to know how close I got to what is in production and what would actually change when I run `kubernetes-deploy` for the first time. 

Not a huge deal because it is already run, but I think it would be useful for sure!

KnVerey:
> for just validating templates before actually triggering a deploy

For validating the templates, I'd suggest it would be more effective to use our renderer to generate them and then put them through a purpose-built tool like [kubeval](https://github.com/garethr/kubeval), which can work without access to a cluster (much better for branch CI). We extracted the rendering functionality to its own class a while ago to make things like this possible from Ruby, and I'm definitely open to exposing a `kubernetes-render` command that prints the templates to stdout or something if that would be useful.

> My specific use case is that I am porting a bunch of other k8s yaml to kubernetes-deploy and I want to know how close I got to what is in production and what would actually change when I run kubernetes-deploy for the first time.

Doing a dry run of the deploy could give you information about which resources will change, but still wouldn't tell you whether the details of the resources would change, no? For that you'd need something like `kubectl diff`, which is in alpha in 1.12, entering beta in 1.13: https://github.com/kubernetes/enhancements/issues/491

[...]

Issue #351, opened 2018-10-12
Make setting ENV["REVISION"] optional


KnVerey:
We are using this to:
- Set the value of `current_sha` in the ERB renderer
- Generate a `deployment_id`, also in the renderer
- Tag statsD metrics with this value

None of that is really necessary to deploy, and indeed some projects do not use revision-based deploys so this is forcing them to set a dummy value. Let's make this optional.

benlangfeld:
What do you think of additionally making it possible to pass this as an option rather than an environment variable? My use case is running kubernetes-deploy inside a generic container full of tools with a wrapper script which invokes `docker run`, but I don't want to infect the wrapper script with knowledge of kubernetes-deploy by doing `docker run --env REVISION`.

KnVerey:
@benlangfeld in that scenario wouldn't it work to inline the env var at the invocation site, like `REVISION=abc kubernetes-deploy ...`?

benlangfeld:
@KnVerey No. I have a wrapper script something like this:

```bash
#!/bin/bash

set -e

VERSION=${VERSION:-"latest"}
TILLER_NAMESPACE=${TILLER_NAMESPACE:-"app-tiller"}
IMAGE="some-image:$VERSION"

if [ -t 0 ]; then
  INTERACTIVE="--interactive"
fi

docker run \
  $INTERACTIVE --tty --rm \
  --env CONTEXT \
  --env TILLER_NAMESPACE=$TILLER_NAMESPACE \
  --mount type=bind,source="$HOME/.aws/credentials",destination=/root/.aws/credentials,readonly \
  --mount type=bind,source="$HOME/.kube",destination=/root/.kube \
  --mount type=bind,source="$(pwd)",destination=/app \
  $IMAGE bash --login -c "$*"
```

The container image referenced here contains tools such as kubectl, helm, stern, kubernetes-deploy etc. It is a way for our platform team to deliver an approved set of tools with a controlled upgrade procedure to our application teams.

Application developers invoke kubernetes-deploy as: `app kubernetes-deploy ...`, much as they invoke kubectl as `app kubectl --namespace foo get pods`. In order to pass `REVISION` through, I would need to do `--env REVISION` here, but that is putting knowledge of kubernetes-deploy into this wrapper script, which I would like to avoid. I would like to be able to invoke it as `app kubernetes-deploy --revision abc ...` instead.

PR #375, opened 2018-11-03
Add a kubernetes-render task 

airhorns:
See https://github.com/Shopify/kubernetes-deploy/issues/182 for inspiration

I am in the process of converting an existing kubernetes deployment over to `kubernetes-deploy`, and I wanted to be able to at least do a little verification of my Kubernetes templates & partials before running the production deploy for the first time. Moreover, I would love to be able to do at least some verification of the YAML `kubernetes-deploy` would deploy to production on branches where it is changed before it's actually deployed to catch dumb stuff like syntax errors or binding typos. 

@KnVerey informed me about https://github.com/garethr/kubeval, which is dope, and serves both those use cases. However, in order to get YAML to pass to kubeval, I needed a way to get the "raw" k8s YAML to then pass on.

This introduces `kubernetes-render`, which is a UNIX-philosphy abiding simple tool that can pipe said raw YAML along to the next place anyone wants to send it. It renders one template in a `template_dir` with bindings.

It does this by:
 - Adding a new `RenderTask` that invokes a `Renderer` on a passed template
 - Adding a new `exe` file that invokes `RenderTask`

Not too complicated. Also, note this is on top of #357 so my tests run fast too. Here's the CI validation script I want to run for reference:

```python
import tempfile
import subprocess
import glob
import os

deploy_dir = os.path.dirname(os.path.realpath(__file__))

for environment in ('production', 'staging'):
    tmpdir = tempfile.mkdtemp()

    deploy_dir_glob = os.path.join(deploy_dir, environment, '*.*')

    for absolute_template in glob.glob(deploy_dir_glob):
        template = os.path.basename(absolute_template)
        tmppath = os.path.join(tmpdir, template)

        with open(tmppath, 'w') as f:
            f.write(subprocess.check_output([
                "kubernetes-render",
                "--template-dir", os.path.join(deploy_dir, environment),
                "--bindings=container_registry=test-registry.docker.io,current_version=v0.2.3-fellow-gdeadbeef",
                template
            ]).decode("utf-8"))

        subprocess.check_call(["kubeval", tmppath])
```


lib/kubernetes-deploy/render_task.rb
```diff
@@ -0,0 +1,64 @@
...
+    end
+
+    def run(*args)
+      run!(*args)
+      true
+    rescue KubernetesDeploy::InvalidTemplateError, KubernetesDeploy::TaskConfigurationError
+      false
+    end
+
+    def run!(filename, stream)
```
KnVerey: What do you think about adding a bit more logging to make the UX of this command more similar to the others? Example:

image

Not a requirement for this PR, but I think it would be great if this ultimately gets an option to set the output dir to a file or maybe a directory instead of stdout if you want.

test/integration/render_task_test.rb
```diff
@@ -0,0 +1,102 @@
+# frozen_string_literal: true
+require 'integration_test_helper'
+require 'kubernetes-deploy/render_task'
+
+class RenderTaskTest < KubernetesDeploy::IntegrationTest
+  def test_render_task
+    render = build_render_task(fixture_path('hello-cloud'))
+    fixture = 'configmap-data.yml'
+
+    stream = StringIO.new
```
KnVerey: I'd like to keep it possible to run any test with `PRINT_LOGS=1` to see exactly how it would look to the user. This is the first time we've had as second output stream to deal with, so our current helpers (see in particular `configure_logger`, `logging_assertion` and `assert_logs_match`) aren't equipped to deal with it. One option would be to make new helpers just for this suite that similarly flip the output device between this new stream and stdout (and disable log assertions) as appropriate. However, if you see a nice way to adapt the existing helpers to handle this additional stream, that would be great. I'll keep thinking about this too.

exe/kubernetes-render
```diff
@@ -0,0 +1,38 @@
...
+  opts.on("--template-dir=DIR", "Set the template dir (default: config/deploy/$ENVIRONMENT)") { |v| template_dir = v }
+  opts.parse!
+end
+
+template = ARGV[0]
+logger = KubernetesDeploy::FormattedLogger.build('test', 'test', verbose_prefix: false)
+revision = KubernetesDeploy::OptionsHelper.revision_from_environment
+
+if !template || template.empty?
+  puts "A template must be provided as an argument to the program."
```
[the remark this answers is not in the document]
airhorns: I was thinking about this too and I think that a lot of folks will indeed have that use case of rendering a lot of stuff in a directory, but, I think the main use cases for this new program are 
generally oriented towards rendering one file, not a whole directory. The use cases in my mind are rendering in a CI pipeline to validate, and rendering in development to understand how the heck the thing works. For my CI pipeline example above, I actually don't want to render the whole group at once so that if rendering one file fails, or kubeval fails, I can output a rich error message to the user about which file it actually was that caused the error. That's all outer-program logic that composes kubernetes-render of course but I think that kubernetes-render would be best used in the same way.

That said, it'd be convenient to render thole whole template dir sometimes I imagine if you don't care about file-wise error messages from downstream tooling, so I could add it. `cat` supports catting multiple files, I don't see why I should be so much more persnickety. 

Do we do that `template-dir` is the parent assumption anywhere else? I feel like we could definitely do what you described with the mutually exclusive options, or adopt a `cat` style interface where all of `ARGV` after options parsing is treated as templates that we loop over and render. That puts the onus of globbing on the user but is a bit more unix-y an interface.

test/integration/render_task_test.rb
airhorns: Ok, updated to add a `stdout_assertion` that works the same way with a different `StringIO` instance.

lib/kubernetes-deploy/render_task.rb
airhorns: Added some success logging

test/integration/render_task_test.rb
```diff
@@ -0,0 +1,102 @@
+# frozen_string_literal: true
+require 'integration_test_helper'
+require 'kubernetes-deploy/render_task'
+
+class RenderTaskTest < KubernetesDeploy::IntegrationTest
```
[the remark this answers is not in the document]
airhorns: Switched to using `TestCase` and including the `FixtureDeployHelper`

test/integration/render_task_test.rb
```diff
@@ -0,0 +1,102 @@
...
+          name: hello-cloud-configmap-data
+          app: hello-cloud
+      data:
+        datapoint1: value1
+        datapoint2: value2
+    RENDERED
+  end
+
+  def test_render_task_with_partials_and_bindings
+    render = build_render_task(fixture_path('for_unit_tests'), 'a': 'binding-a', 'b': 'binding-b')
```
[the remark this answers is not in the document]
airhorns: ✅

exe/kubernetes-render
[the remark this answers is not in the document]
KnVerey: I discussed this with the team today, and we're thinking it would make sense to support using `--template-dir=/path/to/dir` on its own to render all the files in the directory, or `--template-dir=/path/to/dir FILENAME` to render a specific file only (that file must be in the template dir, per your current implementation). We also agreed that your `cat` idea to support rendering multiple specific files in the directory would be great, but we wouldn't consider it a requirement for this PR. Does that sound reasonable to you?

lib/kubernetes-deploy/options_helper.rb
```diff
@@ -0,0 +1,27 @@
...
+        + "as a default path."
+        exit 1
+      end
+
+      template_dir
+    end
+
+    def self.revision_from_environment
+      ENV.fetch('REVISION') do
+        puts "ENV['REVISION'] is missing. Please specify the commit SHA"
```
KnVerey: I'm fine with requiring this in the new command for consistency right now, but FYI we intend to make it optional in the future: https://github.com/Shopify/kubernetes-deploy/issues/351

exe/kubernetes-render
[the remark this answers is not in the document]
airhorns: Ok, updated to do this.

lib/kubernetes-deploy/options_helper.rb
airhorns: Roger

airhorns:
Ok, this is almost good to go, but I am getting a peculiar failing unit test locally after my changes to the test helper:

```
Error:
KubeClientBuilderTest#test_output_stream:
Mocha::StubbingError: stubbing method unnecessarily: #<AnyInstance:Kubeclient::Client>.discover(any_parameters)
    /Users/airhorns/Code/kubernetes-deploy/test/unit/kubernetes-deploy/kubeclient_builder_test.rb:9:in `setup'
```

After a little investigation I am pretty confused, if I disable `Mocha::Configuration.prevent(:stubbing_method_unnecessarily)` the tests pass, but if I disable the stub in question, the tests fail because they try to invoke the real implementation:

```
Error:
KubeClientBuilderTest#test_build_client_from_multiple_config_files:
WebMock::NetConnectNotAllowedError: Real HTTP connections are disabled. Unregistered request: GET https://192.168.5.1/api/v1 with headers {'Accept'=>'*/*', 'Accept-Encoding'=>'gzip, deflate', 'Authorization'=>'Basic dXNlcm5hbWU6cGFzc3dvcmQ=', 'Host'=>'192.168.5.1', 'User-Agent'=>'rest-client/2.0.2 (darwin18.0.0 x86_64) ruby/2.5.1p57'}

You can stub this request with the following snippet:

stub_request(:get, "https://192.168.5.1/api/v1").
  with(
    headers: {
	  'Accept'=>'*/*',
	  'Accept-Encoding'=>'gzip, deflate',
	  'Authorization'=>'Basic dXNlcm5hbWU6cGFzc3dvcmQ=',
	  'Host'=>'192.168.5.1',
	  'User-Agent'=>'rest-client/2.0.2 (darwin18.0.0 x86_64) ruby/2.5.1p57'
    }).
  to_return(status: 200, body: "", headers: {})

============================================================
    /Users/airhorns/.rbenv/versions/2.5.1/lib/ruby/gems/2.5.0/gems/webmock-3.4.2/lib/webmock/http_lib_adapters/net_http.rb:114:in `request'
    /Users/airhorns/.rbenv/versions/2.5.1/lib/ruby/gems/2.5.0/gems/rest-client-2.0.2/lib/restclient/request.rb:448:in `net_http_do_request'
    /Users/airhorns/.rbenv/versions/2.5.1/lib/ruby/gems/2.5.0/gems/rest-client-2.0.2/lib/restclient/request.rb:721:in `block in transmit'
    /Users/airhorns/.rbenv/versions/2.5.1/lib/ruby/gems/2.5.0/gems/webmock-3.4.2/lib/webmock/http_lib_adapters/net_http.rb:123:in `start_without_connect'
    /Users/airhorns/.rbenv/versions/2.5.1/lib/ruby/gems/2.5.0/gems/webmock-3.4.2/lib/webmock/http_lib_adapters/net_http.rb:150:in `start'
    /Users/airhorns/.rbenv/versions/2.5.1/lib/ruby/gems/2.5.0/gems/rest-client-2.0.2/lib/restclient/request.rb:715:in `transmit'
    /Users/airhorns/.rbenv/versions/2.5.1/lib/ruby/gems/2.5.0/gems/rest-client-2.0.2/lib/restclient/request.rb:145:in `execute'
    /Users/airhorns/.rbenv/versions/2.5.1/lib/ruby/gems/2.5.0/gems/rest-client-2.0.2/lib/restclient/request.rb:52:in `execute'
    /Users/airhorns/.rbenv/versions/2.5.1/lib/ruby/gems/2.5.0/gems/rest-client-2.0.2/lib/restclient/resource.rb:51:in `get'
    /Users/airhorns/.rbenv/versions/2.5.1/lib/ruby/gems/2.5.0/gems/kubeclient-3.1.2/lib/kubeclient/common.rb:488:in `block in fetch_entities'
    /Users/airhorns/.rbenv/versions/2.5.1/lib/ruby/gems/2.5.0/gems/kubeclient-3.1.2/lib/kubeclient/common.rb:110:in `handle_exception'
    /Users/airhorns/.rbenv/versions/2.5.1/lib/ruby/gems/2.5.0/gems/kubeclient-3.1.2/lib/kubeclient/common.rb:488:in `fetch_entities'
    /Users/airhorns/.rbenv/versions/2.5.1/lib/ruby/gems/2.5.0/gems/kubeclient-3.1.2/lib/kubeclient/common.rb:478:in `load_entities'
    /Users/airhorns/.rbenv/versions/2.5.1/lib/ruby/gems/2.5.0/gems/kubeclient-3.1.2/lib/kubeclient/common.rb:123:in `discover'
    /Users/airhorns/Code/kubernetes-deploy/lib/kubernetes-deploy/kubeclient_builder.rb:106:in `_build_kubeclient'
    /Users/airhorns/Code/kubernetes-deploy/lib/kubernetes-deploy/kubeclient_builder.rb:17:in `build_v1_kubeclient'
    /Users/airhorns/Code/kubernetes-deploy/test/unit/kubernetes-deploy/kubeclient_builder_test.rb:27:in `test_build_client_from_multiple_config_files'
```

I dug in a bit but it isn't immediately obvious to me what the heck is going on. Y'all seen that before?

lib/kubernetes-deploy/render_task.rb
```diff
@@ -0,0 +1,92 @@
...
+      end
+      @logger.info("Rendered #{File.basename(filename)}")
+
+    rescue KubernetesDeploy::InvalidTemplateError => exception
+      log_invalid_template(filename, exception)
+      raise
+    rescue Psych::SyntaxError => exception
+      log_invalid_template(filename, exception)
+      raise InvalidTemplateError.new("Template is not valid YAML", filename: filename)
+    rescue KubernetesDeploy::FatalDeploymentError
```
KnVerey: why rescue and re-raise this one?

lib/kubernetes-deploy/render_task.rb
```diff
@@ -0,0 +1,92 @@
...
+          @logger.summary.add_action("Configuration invalid: #{message}")
+          raise KubernetesDeploy::TaskConfigurationError, message
+        end
+      end
+
+      validate_configuration(filenames)
+
+      filenames.each { |filename| render_filename(filename, stream) }
+
+      @logger.print_summary(:success)
```
KnVerey: Can you add a short description of what we did via `@logger.summary.add_action` before this line? Generally we say something like "Successfully [action]ed X [resources]"

lib/kubernetes-deploy/render_task.rb
```diff
@@ -0,0 +1,92 @@
...
+    end
+
+    private
+
+    def render_filename(filename, stream)
+      file_content = File.read(File.join(@template_dir, filename))
+      rendered_content = @renderer.render_template(filename, file_content)
+      YAML.load_stream(rendered_content) do |doc|
+        stream.puts YAML.dump(doc)
```
KnVerey: Currently, if you are rendering multiple templates and one is invalid, we're going to output results until we hit the bad one, and if you're running this manually, those templates will be visually interspersed with our stderr logging. Is that what we want, or should we save up the results and dump them all to stdout on successful completion? I'm thinking the latter would be better UX.

test/integration/render_task_test.rb
```diff
@@ -0,0 +1,191 @@
...
+          labels:
+            name: hello-cloud-configmap-data
+            app: hello-cloud
+        data:
+          datapoint1: value1
+          datapoint2: value2
+      RENDERED
+    end
+
+    assert_render_success(render.run([fixture, fixture], test_output_stream))
```
KnVerey: This is probably rendering the same thing twice, right? Should we be de-duping the file names? 

Let's also make a test that renders two different fixtures, and one that attempts to specify a filename that is not inside the deploy dir.

test/integration/render_task_test.rb
```diff
@@ -0,0 +1,191 @@
...
+      assert_match(/name: unmanaged-pod-\w+/, output)
+      assert_match(/name: web/, output)
+    end
+  end
+
+  def test_render_invalid_binding
+    render = build_render_task(fixture_path('test-partials'), 'a': 'binding-a', 'b': 'binding-b')
+    fixture = 'deployment.yaml.erb'
+
+    assert !render.run([fixture], test_output_stream)
```
KnVerey: I'd like to have an `assert_render_failure` for this that also checks that the stderr logging reports the failure correctly.

test/integration/render_task_test.rb
```diff
@@ -0,0 +1,191 @@
...
+    assert_logs_match_all([
+      /Invalid template: .*deployment.yaml.erb/,
+      /undefined local variable or method `supports_partials'/
+    ])
+  end
+
+  def test_render_invalid_arguments
+    render = build_render_task(fixture_path('test-partials'), 'a': 'binding-a')
+
+    assert !render.run(["../"], test_output_stream)
```
KnVerey: Can you please add some assertions about the error message for this one too?

test/integration/render_task_test.rb
```diff
@@ -0,0 +1,191 @@
...
+        data:
+          supports_partials: yep
+      RENDERED
+    end
+  end
+
+  def test_render_task_rendering_all_files
+    render = build_render_task(fixture_path('hello-cloud'))
+
+    assert render.run([], test_output_stream)
```
KnVerey: this looks a little weird--I think I'd expect it to render nothing. Maybe this should be `render.run(test_output_stream)`.

test/test_helper.rb
```diff
@@ -136,6 +137,10 @@ def assert_deploy_success(result)
     alias_method :assert_restart_success, :assert_deploy_success
     alias_method :assert_task_run_success, :assert_deploy_success
 
+    def assert_render_success(result)
+      assert_equal true, result, "Render failed when it was expected to succeed"
```
KnVerey: Let's also add an assertion that we've printed the correct result heading, like we have in the other `assert_[action]_success` helpers

lib/kubernetes-deploy/render_task.rb
airhorns: Oops, left over from logging shennanigans, thanks

lib/kubernetes-deploy/render_task.rb
```diff
@@ -0,0 +1,92 @@
...
+      true
+    rescue KubernetesDeploy::InvalidTemplateError, KubernetesDeploy::TaskConfigurationError
+      @logger.print_summary(:failure)
+      false
+    end
+
+    def run!(filenames, stream)
+      if filenames.empty?
+        filenames = TemplateDiscovery.new(@template_dir).templates
+        if filenames.empty?
```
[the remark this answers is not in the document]
airhorns: Yep, this was silly of me.

test/integration/render_task_test.rb
airhorns: > This is probably rendering the same thing twice, right? Should we be de-duping the file names?

Yeah. In line with `cat` behaviour, I think we should actually render the same file twice. Things taking the output of `kubernetes-render` are likely going to expect a 1-1 mapping between input filenames and rendered content, and I think silently dropping arguments (even if they resolve to be duplicates) could be pretty unexpected and cause downstream complexity. I would vote for keeping it simple and doing what the invoker said even if it seems weird.

Added those tests, thanks.

airhorns:
> Currently, if you are rendering multiple templates and one is invalid, we're going to output results until we hit the bad one, and if you're running this manually, those templates will be visually interspersed with our stderr logging. Is that what we want, or should we save up the results and dump them all to stdout on successful completion? I'm thinking the latter would be better UX.

Correct, that's how it works now. I think we should keep it that way however -- those ones that are rendered before an error are useful IMO. It results in a more responsive user experience in that you get output sooner, it means if you're testing something structural like a command line argument you don't have to wait until the end to know if it worked, and it lets downstream tooling start to do the work before the whole batch is completed. Again, in the spirit of unix tooling, I think streaming it as it is available is good. In this scenario it's going to fail regardless and showing any clues in the existing output seems wise as well.

airhorns:
Also thanks for helping me with that test fail, duh, nice catch! Hawkeye over here my oh my.

test/integration/render_task_test.rb
```diff
@@ -0,0 +1,266 @@
...
+  end
+
+  def test_render_invalid_yaml
+    render = build_render_task(fixture_path('invalid'))
+    fixture = 'yaml-error.yml'
+
+    assert_render_failure render.run(mock_output_stream, [fixture])
+    assert_logs_match_all([
+      /Invalid template: .*yaml-error.yml/,
+      /mapping values are not allowed/
```
dturn: Is ` (<unknown>):` intentional in the error message:
`Error message: (<unknown>): mapping values are not allowed in this context at line 9 column 21`

test/integration/render_task_test.rb
```diff
@@ -0,0 +1,266 @@
...
+  end
+
+  def test_render_invalid_binding
+    render = build_render_task(fixture_path('test-partials'), 'a': 'binding-a', 'b': 'binding-b')
+    fixture = 'deployment.yaml.erb'
+
+    assert_render_failure render.run(mock_output_stream, [fixture])
+    assert_logs_match_all([
+      /Invalid template: .*deployment.yaml.erb/,
+      /undefined local variable or method `supports_partials'/
```
dturn: The ruby object info doesn't really add anything here.
`Error message: undefined local variable or method `supports_partials' for #<KubernetesDeploy::Renderer::TemplateContext:0x007fbf4252bb70>`

lib/kubernetes-deploy/render_task.rb
dturn: I agree w/ @KnVerey here. Its really frustrating to have to fix items 1 at a time.

lib/kubernetes-deploy/formatted_logger.rb
```diff
@@ -6,12 +6,23 @@ module KubernetesDeploy
...
-    def self.build(namespace, context, stream = $stderr, verbose_prefix: false)
+    def self.build(namespace = nil, context = nil, stream = $stderr, verbose_prefix: false)
       l = new(stream)
       l.level = level_from_env
 
       l.formatter = proc do |severity, datetime, _progname, msg|
-        middle = verbose_prefix ? "[#{context}][#{namespace}]" : ""
+        middle = if verbose_prefix
+          if namespace.blank?
+            raise ArgumentError, 'Must pass a namespace if logging verbosely'
```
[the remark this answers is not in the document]
airhorns: Ah good point

lib/kubernetes-deploy/render_task.rb
```diff
@@ -0,0 +1,100 @@
...
+      absolute_template_dir = File.expand_path(@template_dir)
+
+      filenames.each do |filename|
+        absolute_file = File.expand_path(File.join(@template_dir, filename))
+        if !File.exist?(absolute_file)
+          errors << "Filename \"#{absolute_file}\" could not be found"
+        elsif !File.file?(absolute_file)
+          errors << "Filename \"#{absolute_file}\" is not a file"
+        elsif !absolute_file.start_with?(absolute_template_dir)
+          errors << "Filename \"#{absolute_file}\" is outside the template directory"
```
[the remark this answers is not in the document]
airhorns: I did that on purpose because I think that the interaction between `--template-dir` and templates passed as ARGV is not altogether clear to the uninitiated. This tool doesn't allow subfolders other than partials I think, so it's not like you can actually pass any old relative path as an ARGV, especially something with like `../` in it, so I think it's worth being extra clear. For bonus points you can copy paste that absolute path directly and `cat` it or whatever to understand exactly what the tool is doing.

lib/kubernetes-deploy/template_discovery.rb
```diff
@@ -0,0 +1,15 @@
+# frozen_string_literal: true
+
+module KubernetesDeploy
+  class TemplateDiscovery
+    def initialize(template_dir)
+      @template_dir = template_dir
+    end
+
+    def templates
+      Dir.foreach(@template_dir).to_a.keep_if do |filename|
```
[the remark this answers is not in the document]
airhorns: Derp

test/integration/render_task_test.rb
airhorns: It is not, but that's from Psych. I added a commit to address this, it changes the output from looking like `(<unknown>)` to `(<rendered> the-file-in-question.yaml.erb)`

lib/kubernetes-deploy/render_task.rb
```diff
@@ -0,0 +1,100 @@
...
+      @logger.phase_heading("Initializing render task")
+
+      filenames = if only_filenames.empty?
+        TemplateDiscovery.new(@template_dir).templates
+      else
+        only_filenames
+      end
+
+      @logger.phase_heading("Rendering template(s)")
+      validate_configuration(filenames)
```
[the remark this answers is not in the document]
airhorns: Oops, fixed

test/integration/render_task_test.rb
airhorns: True, however, that's coming right out of ERB land, occurs in the other tasks as well I think, and personally I think getting into the business of parsing exception messages is not going to yield all that much gold. Ok if I leave it as is?

airhorns:
So with regards to rendering batches where some are broken, I think we could do the best of both worlds and render templates as they are available to STDOUT, progress through the whole list, but catch any errors and reraise at the end of the batch. `cat` does the same and exits 1 if anything went wrong:

```
pilot ~/C/kubernetes-deploy (render-task) ➜  echo "foo" > foo
pilot ~/C/kubernetes-deploy (render-task) ➜  echo "baz" > baz
pilot ~/C/kubernetes-deploy (render-task) ➜  file bar
bar: cannot open `bar' (No such file or directory)
pilot ~/C/kubernetes-deploy (render-task) ➜  cat foo bar baz
foo
cat: bar: No such file or directory
baz
```

Sound ok?

lib/kubernetes-deploy/render_task.rb
```diff
@@ -0,0 +1,101 @@
...
+
+    private
+
+    def render_filename(filename, stream)
+      file_content = File.read(File.join(@template_dir, filename))
+      rendered_content = @renderer.render_template(filename, file_content)
+      YAML.load_stream(rendered_content, "<rendered> #{filename}") do |doc|
+        stream.puts YAML.dump(doc)
+      end
+      @logger.info("Rendered #{File.basename(filename)}")
```
dturn: Not sure how I missed this in the last review, but I think we change this to `Rendering  ...` and put it at the top. Having an empty phase 2 here isn't awesome:

```
Begin test: test_render_invalid_yaml
***************************************************************************

[INFO][2018-11-16 10:47:50 -0800]	
[INFO][2018-11-16 10:47:50 -0800]	---------------------------------Phase 1: Initializing render task----------------------------------
[INFO][2018-11-16 10:47:50 -0800]	Validating configuration
[INFO][2018-11-16 10:47:50 -0800]	
[INFO][2018-11-16 10:47:50 -0800]	-----------------------------------Phase 2: Rendering template(s)-----------------------------------
[INFO][2018-11-16 10:47:50 -0800]	
[INFO][2018-11-16 10:47:50 -0800]	------------------------------------------Result: FAILURE-------------------------------------------
[FATAL][2018-11-16 10:47:50 -0800]	Invalid template: yaml-error.yml
[FATAL][2018-11-16 10:47:50 -0800]	Error message: (<rendered> yaml-error.yml): mapping values are not allowed in this context at line 9 column 21
```

KnVerey:
Hey @airhorns ! Is there anything we can do to unblock you on this PR? I would love to have it merged soon. 😄

airhorns:
Ok, updated to group exceptions and only fail at the very end!

lib/kubernetes-deploy/render_task.rb
airhorns: Added that statement, output looks like this now:

```
[INFO][2018-12-11 11:36:42 -0500]
[INFO][2018-12-11 11:36:42 -0500]	---------------------------------Phase 1: Initializing render task----------------------------------
[INFO][2018-12-11 11:36:42 -0500]	Validating configuration
[INFO][2018-12-11 11:36:42 -0500]
[INFO][2018-12-11 11:36:42 -0500]	-----------------------------------Phase 2: Rendering template(s)-----------------------------------
[INFO][2018-12-11 11:36:42 -0500]	Rendering yaml-error.yml ...
[INFO][2018-12-11 11:36:42 -0500]
[INFO][2018-12-11 11:36:42 -0500]	------------------------------------------Result: FAILURE-------------------------------------------
[FATAL][2018-12-11 11:36:42 -0500]	Invalid template: yaml-error.yml
[FATAL][2018-12-11 11:36:42 -0500]	Error message: Template is not valid YAML. (<rendered> yaml-error.yml): mapping values are not allowed in this context at line 9 column 21
```

test/integration/render_task_test.rb
```diff
@@ -0,0 +1,284 @@
...
+    ])
+  end
+
+  def test_render_invalid_arguments
+    render = build_render_task(fixture_path('test-partials'), 'a': 'binding-a')
+
+    assert_render_failure render.run(mock_output_stream, ["../"])
+    assert_logs_match_all([
```
KnVerey: There's also `assert_logs_match` you can use for cases like this where you're only looking for one thing. That said, I often like to use `_match_all` instead and include surrounding headings in the assertion, to confirm it's printed _when_ I expected (right away vs in the summary, for example).

test/test_helper.rb
```diff
@@ -214,19 +205,41 @@ def obj.run(*)
...
         @logger_stream.rewind
         yield @logger_stream.read
       end
     end
+
+    def stdout_assertion
+      if log_to_real_fds?
+        $stderr.puts("\033[0;33mWARNING: Skipping stream assertions while logs are redirected to stderr\033[0m")
+      else
+        yield @mock_output_stream.string
```
KnVerey: This does seem simpler than the rewind/read we do in the original method above. If you're already going to make another commit, would you mind making this improvement there too?

lib/kubernetes-deploy/render_task.rb
```diff
@@ -0,0 +1,118 @@
...
+        @logger.summary.add_action("Configuration invalid")
+        @logger.summary.add_paragraph(errors.map { |err| "- #{err}" }.join("\n"))
+        raise KubernetesDeploy::TaskConfigurationError, "Configuration invalid: #{errors.join(', ')}"
+      end
+    end
+
+    def log_render_failure(filename, exception)
+      debug_msg = ColorizedString.new("Render failure for #{filename}\n").red
+      debug_msg += "Error message: #{exception}"
+      @logger.error(debug_msg)
```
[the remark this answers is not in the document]
KnVerey: Actually, maybe we should do both. Right away, we could do `@logger.error("Failed to render #{filename}")` to mirror the "Rendered #{filename}" message in the success case, and then this more detailed message can go in the summary.

lib/kubernetes-deploy/render_task.rb
```diff
@@ -0,0 +1,118 @@
...
+
+      filenames.each do |filename|
+        begin
+          render_filename(filename, stream)
+        rescue KubernetesDeploy::InvalidTemplateError => exception
+          exceptions << exception
+          log_invalid_template(filename, exception)
+        rescue KubernetesDeploy::FatalDeploymentError => exception
+          exceptions << exception
+          log_render_failure(exception)
```
KnVerey: Why have a separate case for invalid template error above? It would be possible to take advantage of the augmented information on `InvalidTemplateError` to improve the message, `log_invalid_template` doesn't actually do that. More concerning, I don't think this case has test coverage (this line is missing the `filename` arg).

airhorns: After reviewing this was just me being silly, I forgot that InvalidTemplateError was actually a subclass of the other one and was just assuming I had to rescue it as well. Fixing.

lib/kubernetes-deploy/render_task.rb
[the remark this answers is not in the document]
airhorns: I am honestly a little confused about what the right way to use the summary logger is so I was just trying to mimic what I saw elsewhere. Updated with your suggestion.

lib/kubernetes-deploy/render_task.rb
```diff
@@ -0,0 +1,119 @@
...
+      end
+    end
+
+    def log_invalid_template(exception)
+      @logger.error("Failed to render #{exception.filename}")
+
+      debug_msg = ColorizedString.new("Invalid template: #{exception.filename}\n").red
+      debug_msg += "> Error message: #{exception}"
+      if exception.content
+        debug_msg += "> Template content:\n#{exception.content}"
```
KnVerey: I think we need a newline at the beginning of this as well. It's getting stuck on the end of the exception:

image

PR #376, opened 2018-11-12
Dynamic Custom Resource Deployments

timothysmith0609:
## Motivation and Goals

required for #229 
see #128 

- [Relevant CloudPortal issue](https://github.com/Shopify/cloud-portal/issues/427)
- [Cloudbuddies CR status conventions](https://github.com/Shopify/cloudbuddies/pull/1381)

This PR represents the kubernetes-deploy side of our [custom resource status implementation](https://shopify.kanbanize.com/ctrl_board/1/cards/468/details). The goal of this PR is to create a backwards compatible means of meaningfully monitoring the rollouts of custom resources. It has the following goals:

- Hardcoded custom resources (e.g. [redis.rb](https://github.com/Shopify/kubernetes-deploy/blob/master/lib/kubernetes-deploy/kubernetes_resource/redis.rb), [cloudsql.rb](https://github.com/Shopify/kubernetes-deploy/blob/master/lib/kubernetes-deploy/kubernetes_resource/cloudsql.rb)) are deployed using the custom logic that exists for them.
- Custom resources **with no** hardcoded logic **and no** annotations notifying kubernetes-deploy to treat the CR as a generic custom resource using our Pass/Fail convention will be deployed as before. E.g. warn the user that we don't know how to monitor the deployment of such a resource and assume it has passed
- _**New case:**_ A custom resource **with no** hardcoded logic **but with an annotation declaring it implements our Pass/Fail status convention** will be deployed using the generic CR watcher. This watcher observes the following states:
  - `deploy_succeeded? == true` if the `Ready` condition on the CR status is true
  - `deploy_failed? == true` if the `Failed` condition on the CR status is true
  - The deploy is progressing if both `Ready` and `Failed` are false
  - Custom timeouts for CRs can be placed on the owning CRD or, for more granularity, on specific instances of a CR spec

### (OUTDATED) Changes to `PREDEPLOY_SEQUENCE`

Since we cannot know, _a priori_, the types of custom resources in a cluster, we must dynamically find them during the discovery phase. As an additional concern, I argue that the common case is that custom resources must be deployed before other kubernetes objects. That is, we need a working `cloudsql` before we can think of running a `db-migrate` pod, e.g. As it stands, we hardcode this priority inside the `PREDEPLOY_SEQUENCE` constant in [deploy_task.rb](https://github.com/Shopify/kubernetes-deploy/blob/master/lib/kubernetes-deploy/deploy_task.rb#L47). In order to maintain the rough ordering of the `PREDEPLOY_SEQUENCE` const while also handling the case of dynamic custom resource discovery, I have moved the creation of `PREDEPLOY_SEQUENCE` into 2 separate phases.
  - In the first phase, we hardcode the core kubernetes resources that we know ante deploy and place them in `BASE_PREDEPLOY_SEQUENCE`.
  - During discovery, we find all the CRDs and union them with `BASE_PREDEPLOY_SEQUENCE`. Using the result of this union, we then set the `PREDEPLOY_SEQUENCE` constant. [See here](https://github.com/Shopify/kubernetes-deploy/compare/dynamic_cr_capturing?expand=1#diff-4c33bbb177fdf882c41d892f3bde9570R153)
  - @stefanmb has proposed using a dependency graph to model deployment priority, but unless an explicit case can be made against the proposed implementation I think we can defer that issue for now.

### (OUTDATED) Caching discovered CRDs

As an implementation detail, I have opted to cache the value of `ResourceDiscovery.crds` in an instance variable. Linking together CRs with their parent CRDs requires passing around the list of CRDs in a number of places, and it doesn't seem risky to avoid the extra work of rediscovering them for every call.

## TODO

- [x] Tests
- [x] Annotation scheme for CRs -> e.g. annotation for declaring monitorability? Do we want additional annotations to declare which conditions map to ready/failed or should we enforce `Ready` and `Failed`?

cc @Shopify/cloudx


lib/kubernetes-deploy/deploy_task.rb
```diff
@@ -204,7 +203,18 @@ def run!(verify_result: true, allow_protected_ns: false, prune: true)
...
+        context: @context,
+        logger: @logger,
+        namespace_tags: @namespace_tags
+      )
+    end
+
+    def add_dynamic_custom_resources_to_predeploy_sequence(resources)
+      crds = cluster_resource_discoverer.crds(@sync_mediator).map(&:kind)
+      discovered_crs = resources.select { |resource| crds.include? resource.type }
+      DeployTask.const_set("PREDEPLOY_SEQUENCE", (discovered_crs.map(&:type) + BASE_PREDEPLOY_SEQUENCE).uniq)
```
mkobetic: Do we expect a case where the base crds would not be in the discovered set?

timothysmith0609: On deploys that contain both the CRD and an instance of it as a resource. AFAICT we don't do that ourselves, but am unsure if the wider community that uses Kubernetes-deploy uses that pattern.

karanthukral: we are only adding the CRs here not the CRD right?

lib/kubernetes-deploy/deploy_task.rb
```diff
@@ -244,14 +254,15 @@ def validate_definitions(resources)
...
+      crds = cluster_resource_discoverer.crds(@sync_mediator)
       @logger.info("Discovering templates:")
 
       Dir.foreach(@template_dir) do |filename|
         next unless filename.end_with?(".yml.erb", ".yml", ".yaml", ".yaml.erb")
 
         split_templates(filename) do |r_def|
           r = KubernetesResource.build(namespace: @namespace, context: @context, logger: @logger,
-                                       definition: r_def, statsd_tags: @namespace_tags)
+                                       definition: r_def, statsd_tags: @namespace_tags, crds: crds)
```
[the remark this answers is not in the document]
karanthukral: I think passing CRDs in favour of not having to duplicate the build logic makes sense since you would have to still know which resources in build to create a generic k8s resource for (non crs and not implemented) and then the rest used for crs or hardcoded crs. Passing a variable for a check seems a valid approach to me

lib/kubernetes-deploy/kubernetes_resource.rb
```diff
@@ -31,14 +31,17 @@ class KubernetesResource
...
         opts = { namespace: namespace, context: context, definition: definition, logger: logger,
                  statsd_tags: statsd_tags }
         if definition["kind"].blank?
           raise InvalidTemplateError.new("Template missing 'Kind'", content: definition.to_yaml)
-        elsif KubernetesDeploy.const_defined?(definition["kind"])
+        elsif KubernetesDeploy.const_defined?(definition["kind"]) # Hardcoded CRs
           klass = KubernetesDeploy.const_get(definition["kind"])
           klass.new(**opts)
+        elsif crds.map(&:kind).include? definition["kind"]        # Dynamic CRs
+          crd = crds.find { |crd| crd.kind == definition["kind"] }
```
mkobetic: nit: bugs me a little to see the same search performed twice, it seems to me the whole if would be better written as a series of guardclauses with explicit returns.

lib/kubernetes-deploy/deploy_task.rb
```diff
@@ -204,7 +203,18 @@ def run!(verify_result: true, allow_protected_ns: false, prune: true)
...
+        namespace: @namespace,
+        context: @context,
+        logger: @logger,
+        namespace_tags: @namespace_tags
+      )
+    end
+
+    def add_custom_resources_to_predeploy_sequence(resources)
+      crds = cluster_resource_discoverer.crds(@sync_mediator).map(&:kind)
+      discovered_crs = resources.select { |resource| crds.include? resource.type }
```
mkobetic: nit: looks like you could do the `.map(&type).uniq` before the `select` as well to reduce the potentially large set of resources into just their unique types

lib/kubernetes-deploy/resource_discovery.rb
```diff
@@ -10,7 +10,7 @@ def initialize(namespace:, context:, logger:, namespace_tags:)
     end
 
     def crds(sync_mediator)
-      sync_mediator.get_all(CustomResourceDefinition.kind).map do |r_def|
+      @crds ||= sync_mediator.get_all(CustomResourceDefinition.kind).map do |r_def|
```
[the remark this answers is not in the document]
timothysmith0609: That's one shortcoming of the current approach, unfortunately. There's a few solutions available to us here:
- Add discovered CRD specs to the top of the priority list. This seems risk-free since they should have no external dependencies
- In the future, think about using a dependency graph to produce the priority list (e.g. a CloudSQL has something that says `iDependOn: cloudsqls.stable.shopify.io`. I'd say this is out of scope for this PR.

It's unclear to me whether this issue is handled right now, anyway. Not saying we shouldn't fix it here, just a note

lib/kubernetes-deploy/deploy_task.rb
```diff
@@ -204,7 +203,18 @@ def run!(verify_result: true, allow_protected_ns: false, prune: true)
...
+      @cluster_resource_discoverer ||= ResourceDiscovery.new(
+        namespace: @namespace,
+        context: @context,
+        logger: @logger,
+        namespace_tags: @namespace_tags
+      )
+    end
+
+    def add_custom_resources_to_predeploy_sequence(resources)
+      crds = cluster_resource_discoverer.crds(@sync_mediator).map(&:kind)
```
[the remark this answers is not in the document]
timothysmith0609: I'm wary of annotation bloat so trying to minimize the number I'm adding for this feature, though I agree that would provide nice flexibility.

lib/kubernetes-deploy/kubernetes_resource.rb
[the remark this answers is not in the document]
timothysmith0609: Changed to just assign it in the conditional itself

lib/kubernetes-deploy/kubernetes_resource/custom_resource.rb
```diff
@@ -0,0 +1,57 @@
+# frozen_string_literal: true
+module KubernetesDeploy
+  class CustomResource < KubernetesResource
+    def initialize(namespace:, context:, definition:, logger:, statsd_tags: [], crd:)
+      @crd = crd
```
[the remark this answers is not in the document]
timothysmith0609: Since some of the methods defined on the class have a hard dependency on the CRD, I would prefer to declare it at the point of instantiation

lib/kubernetes-deploy/deploy_task.rb
timothysmith0609: Updated to use the `prune_whitelist` approach`

lib/kubernetes-deploy/kubernetes_resource.rb
```diff
@@ -31,14 +31,16 @@ class KubernetesResource
...
-      def build(namespace:, context:, definition:, logger:, statsd_tags:)
+      def build(namespace:, context:, definition:, logger:, statsd_tags:, crds: [])
         opts = { namespace: namespace, context: context, definition: definition, logger: logger,
                  statsd_tags: statsd_tags }
         if definition["kind"].blank?
           raise InvalidTemplateError.new("Template missing 'Kind'", content: definition.to_yaml)
         elsif KubernetesDeploy.const_defined?(definition["kind"])
           klass = KubernetesDeploy.const_get(definition["kind"])
           klass.new(**opts)
+        elsif crd = crds.find { |crd| crd.kind == definition["kind"] } # Dynamic CRs
```
dturn: We could avoid changing this file by creating the CR classes with the right names dynamically. And then finding the parent CRD on the first sync. I can't tell if its a bad idea though. Its safe as is, but would require, that the CR classes never have class ivars

timothysmith0609: Do you have an example of us using class ivars in any other resources, or a use-case where class-ivars might be worthwhile for custom resources?

I'm uneasy about dynamically creating classes, particularly since the implementation I've chosen is very much black-box and doesn't rely on the particulars of a given resource, only the contract it exposes via its metadata and status.

lib/kubernetes-deploy/deploy_task.rb
```diff
@@ -244,14 +250,15 @@ def validate_definitions(resources)
...
+      crds = cluster_resource_discoverer.crds(@sync_mediator)
       @logger.info("Discovering templates:")
 
       Dir.foreach(@template_dir) do |filename|
         next unless filename.end_with?(".yml.erb", ".yml", ".yaml", ".yaml.erb")
 
         split_templates(filename) do |r_def|
           r = KubernetesResource.build(namespace: @namespace, context: @context, logger: @logger,
-                                       definition: r_def, statsd_tags: @namespace_tags)
+                                       definition: r_def, statsd_tags: @namespace_tags, crds: crds)
```
timothysmith0609: We could capture the fact that a resource is a CRD here instead of inside `KubernetesResource.build`. This would avoid having to pass through the `crds` array, which for most resources is a useless argument. We would have to deal with the case where the deploy logic is hardcoded (e.g. `redis.rb`, `cloud_sql.rb`), but that won't be too much trouble.

This would necessitate adding a `KubernetesResource.build_cr` method or something to that effect

timothysmith0609:
## Configurable success/failure conditions
I've decided to add configurable success/failure statuses as part of this PR.  Users can supply a JSON string to the `kubernetes.shopify.io/cr-rollout-params` that takes:
- A JSON array of success conditions (JsonPath/expected value pairs)
- A JSON array of failure conditions
- Custom timeout
- ... we can add whatever other configurable fields we desire

For convenience, default values (which conform to our buddies Status implementation) are used if such fields are missing.

### Limitations
Currently, resources, such as CloudSQL, reference other Kubernetes objects to discern their readiness (in CloudSQLs case, its deployment + service). In the new implementation, deploying resources are only able to observe themselves.

lib/kubernetes-deploy/kubernetes_resource/custom_resource_definition.rb
```diff
@@ -41,8 +43,28 @@ def prunable?
       prunable == "true"
     end
 
+    def timeout_for_children
+      @definition.dig("metadata", "annotations", CHILD_CR_TIMEOUT_ANNOTATION)&.to_i
+    end
+
+    def rollout_params
+      return nil unless rollout_params_string
```
mkobetic: So the annotation has to be present, possibly empty for the deploy to look at the status?

lib/kubernetes-deploy/kubernetes_resource/custom_resource_definition.rb
```diff
@@ -2,6 +2,8 @@
 module KubernetesDeploy
   class CustomResourceDefinition < KubernetesResource
     TIMEOUT = 2.minutes
+    CHILD_CR_TIMEOUT_ANNOTATION = "kubernetes-deploy.shopify.io/cr-timeout-override"
```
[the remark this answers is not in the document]
mkobetic: Why `CHILD_` ?

lib/kubernetes-deploy/deploy_task.rb
```diff
@@ -244,14 +250,21 @@ def validate_definitions(resources)
...
       @logger.info("Discovering templates:")
 
       Dir.foreach(@template_dir) do |filename|
         next unless filename.end_with?(".yml.erb", ".yml", ".yaml", ".yaml.erb")
 
         split_templates(filename) do |r_def|
-          r = KubernetesResource.build(namespace: @namespace, context: @context, logger: @logger,
-                                       definition: r_def, statsd_tags: @namespace_tags)
+          kind = r_def["kind"]
+          r = if crds[kind] && !KubernetesDeploy.const_defined?(r_def["kind"])
```
mkobetic: s/r_def["kind"]/kind/ ?

lib/kubernetes-deploy/kubernetes_resource/custom_resource_definition.rb
timothysmith0609: Yes, the presence of an empty object will default to using the preset values for success/failure. These can be overridden piecemeal as needed. The defaults set now are the base conditions we decided on in cloudbuddies

mkobetic: OK, would be good to make that clear in the docs.

lib/kubernetes-deploy/kubernetes_resource/custom_resource_definition.rb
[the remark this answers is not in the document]
timothysmith0609: Open to suggestions. I'm mainly trying to avoid the perennial issue of confusing CRDs with CRs by being more explicit

lib/kubernetes-deploy/deploy_task.rb
```diff
@@ -244,14 +250,21 @@ def validate_definitions(resources)
 
     def discover_resources
       resources = []
+      crds = cluster_resource_discoverer.crds(@sync_mediator).each_with_object({}) { |crd, memo| memo[crd.kind] = crd }
```
dturn: Can we give this a slightly more descriptive name: `crd_by_kid` or something...

lib/kubernetes-deploy/deploy_task.rb
```diff
@@ -244,14 +250,21 @@ def validate_definitions(resources)
...
       @logger.info("Discovering templates:")
 
       Dir.foreach(@template_dir) do |filename|
         next unless filename.end_with?(".yml.erb", ".yml", ".yaml", ".yaml.erb")
 
         split_templates(filename) do |r_def|
-          r = KubernetesResource.build(namespace: @namespace, context: @context, logger: @logger,
-                                       definition: r_def, statsd_tags: @namespace_tags)
+          kind = r_def["kind"]
+          r = if crds[kind] && !KubernetesDeploy.const_defined?(kind)
```
dturn: What made you decide to move this here instead of ` KubernetesResource.build`?

timothysmith0609: I wasn't happy with having to pass a `crd` object into all calls to `KubernetesResource.build`, so wanted a solution where I could avoid that. For this reason, I moved the logic to check for a generic CR outside of `KubernetesResource.build`.

Initially I was going to make a method like `KubernetesResource.build_cr`, but the logic for it is quite sparse, so the same thing can be tersely accomplished here in `deploy_task.rb`. My underlying motivation is that instances of `CustomResource` aren't as concrete as completely specified resources (e.g. `Pod`, `Deployment`), and we can infer when we have one with the simple conditional `if crds[kind] && !KubernetesDeploy.const_defined?(kind)`. To that end, `CustomResource.new` is semantically similar enough to `KubernetesResource.build` that I was ok placing them together. Another way of looking at it is that both `CustomResource` and `KubernetesResource` infer properties of generic objects and return a fully specified resource, which puts them on equal footing from an abstraction point of view.

Said that, I'm ok moving things around if there's disagreement on my approach. At the end of the day the only thing that's important here is that we create `CustomResource` objects when appropriate.

lib/kubernetes-deploy/resource_discovery.rb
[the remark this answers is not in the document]
KnVerey: > That's one shortcoming of the current approach, unfortunately.

What is the exact behaviour though? It still falls back on a KubernetesResource with the "dunno what to do here" message, right?

I'm on the fence about introducing an annotation-driven dependency graph, but FWIW we did have someone request customized sequencing for another reason earlier this week. Regardless, I agree it doesn't need to be handled in this PR.

lib/kubernetes-deploy/kubernetes_resource/custom_resource_definition.rb
KnVerey: Can you please provide a sample of the complete annotation for reviewers?

lib/kubernetes-deploy/kubernetes_resource/custom_resource_definition.rb
```diff
@@ -41,6 +43,25 @@ def prunable?
...
+      raw_params = JSON.parse(rollout_params_string)
+      params = {
+        success_queries: raw_params["success_queries"] || default_success_query,
+        failure_queries: raw_params["failure_queries"] || default_failure_query,
+      }.deep_symbolize_keys
+      # Preemptively check if the supplied JsonPath expressions are valid
+      params if validate_params(params)
+
+    rescue JSON::ParserError
+      raise FatalDeploymentError, "custom rollout params are not valid JSON: '#{rollout_params_string}'"
```
KnVerey: Should we actually fail the whole deploy on this, or is there a more graceful fallback behaviour we could adopt?

lib/kubernetes-deploy/kubernetes_resource/custom_resource_definition.rb
```diff
@@ -51,5 +72,37 @@ def names_accepted_condition
...
+      }]
+    end
+
+    def validate_params(params)
+      params[:success_queries].each { |query| JsonPath.new(query[:path]) }
+      params[:failure_queries].each do |query|
+        JsonPath.new(query[:path])
+        JsonPath.new(query[:error_msg_path]) if query[:error_msg_path]
+        if query[:custom_error_msg] && !query[:custom_error_msg].is_a?(String)
+          raise RuntimeError, "custom_error_msg must be a string, but found #{query[:custom_error_msg]}"
```
KnVerey: Why not raise FatalDeploymentError here to begin with?

lib/kubernetes-deploy/kubernetes_resource/custom_resource.rb
```diff
@@ -0,0 +1,59 @@
...
+    end
+
+    def timeout
+      timeout_override || @crd.timeout_for_children || super
+    end
+
+    def deploy_succeeded?
+      return super unless rollout_params
+      rollout_params[:success_queries].all? do |query|
+        JsonPath.new(query[:path]).first(@instance_data) == query[:value]
```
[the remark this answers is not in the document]
KnVerey: I see you've validated that ahead of time. Could the `success_queries` we're getting here reuse the pre-parsed JsonPath instances instead?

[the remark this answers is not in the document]
timothysmith0609: Aggravatingly, jsonpath does not raise any custom exceptions, only `ArgumentError` :( 
Agree that reusing the pre-parsed paths is prefereable

kubernetes-deploy.gemspec
```diff
@@ -29,6 +29,7 @@ Gem::Specification.new do |spec|
   spec.add_dependency "ejson", "~> 1.0"
   spec.add_dependency "colorize", "~> 0.8"
   spec.add_dependency "statsd-instrument", "~> 2.2"
+  spec.add_dependency "jsonpath", "~> 0.9.6"
```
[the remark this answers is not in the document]
timothysmith0609: ## Pros
- Checked for jsonpath CVE vulnerabilities, found none.
- Examined source code for cases of `eval`. `eval` not used in the gem since version `0.8.4` (currently using `0.9.8`).
- Rubygems records 4 million+ downloads of gem, ~200k for current version.

## Cons
- Repo not very active
- Poor error handling

## Neutral
- No obvious alternative jsonpath implementation (`jq` maybe?)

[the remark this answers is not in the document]
stefanmb: `jq` has other problems, it requires a native library and then proceeds to pollute your `stdout` since it wasn't really designed to work as a library.

lib/kubernetes-deploy/kubernetes_resource/custom_resource_definition.rb
timothysmith0609: This is a value-judgement we'll need to make. My opinion is that, if users are opting-in to use this feature, we should consider it critical and fail fast if something goes wrong. 

On the other hand, aborting a deploy because of some bad JSON might cause too much friction.

timothysmith0609:
Example of custom query parameters:
```
'{
      "success_queries": [
        {
          "path": "$.status.conditions[?(@.type == "Ready")].status",
          "value":"success_value"
        },
        {
          "path":"$.spec.test_field",
          "value":"success_value"
        }
      ],
      "failure_queries": [
        {
          "path":"$.status.condition",
          "value":"failure_value",
          "custom_error_msg":"test custom error message"
        },
        {
          "path":"$.spec.test_field",
          "value":"failure_value",
          "error_msg_path":"$.spec.error_msg"
        }
      ]
    }'
```

KnVerey:
Two questions about the example:
* How do multiple queries combine? Is it the same for success queries and failure queries?
* Is the inclusion of `custom_error_msg` based on a real use case? This is not always true of course, but in cases where the CRD is the company's own, the error messages are already their own too.

timothysmith0609:
>How do multiple queries combine? Is it the same for success queries and failure queries?

For success, _all_ conditions must evaluate to true before a deploy is considered suceeded
For failures, _any_ condition evaluating to true causes the deploy to fail

>Is the inclusion of custom_error_msg based on a real use case? This is not always true of course, but in cases where the CRD is the company's own, the error messages are already their own too.

`error_msg_path` is a jsonpath expression that will yield the contents of part of the CR definition. If we use the typical `Conditions` structure for statuses, this is usually in the `message` field. I decided to add `custom_error_msg` for cases where the conditions structure isn't used and no suitable message appears in the actual spec. I'm ok with removing it if it simplifies things

timothysmith0609:
> Maybe CRD has a private ChildConfiguration object that it exposes, and that has methods like error_message_path and failure_status_path and such?

Per @KnVerey 's suggestion, I've opted to move the specifics of the rollout config into a nested class of `CustomResourceDefinition`. This decouples `CustomResource` from the internals of monitoring its rollout and centralizes the logic for rollout configuration/implementation.

lib/kubernetes-deploy/kubernetes_resource/custom_resource.rb
```diff
@@ -0,0 +1,52 @@
...
+
+      rollout_config.deploy_failed?(@instance_data)
+    end
+
+    def failure_message
+      messages = rollout_config.failure_messages(@instance_data)
+      messages.present? ? messages.join("\n") : "error deploying #{id}"
+    end
+
+    def id
```
dturn: Is this needed given the super class is defined as `"#{type}/#{name}"`?

lib/kubernetes-deploy/kubernetes_resource.rb
```diff
@@ -31,11 +31,14 @@ class KubernetesResource
...
     class << self
-      def build(namespace:, context:, definition:, logger:, statsd_tags:)
+      def build(namespace:, context:, definition:, logger:, statsd_tags:, crd: nil)
         opts = { namespace: namespace, context: context, definition: definition, logger: logger,
                  statsd_tags: statsd_tags }
+
         if definition["kind"].blank?
           raise InvalidTemplateError.new("Template missing 'Kind'", content: definition.to_yaml)
+        elsif crd && !KubernetesDeploy.const_defined?(definition["kind"])
+          CustomResource.new(crd: crd, **opts)
```
dturn: should this come after line 42 so you can skip the second check

lib/kubernetes-deploy/kubernetes_resource/custom_resource.rb
```diff
@@ -0,0 +1,52 @@
...
+require 'jsonpath'
+module KubernetesDeploy
+  class CustomResource < KubernetesResource
+    def initialize(namespace:, context:, definition:, logger:, statsd_tags: [], crd:)
+      super(namespace: namespace, context: context, definition: definition,
+        logger: logger, statsd_tags: statsd_tags)
+      @crd = crd
+    end
+
+    def timeout
```
dturn: How is this different from the super class behavior?

lib/kubernetes-deploy/kubernetes_resource/custom_resource.rb
```diff
@@ -0,0 +1,52 @@
...
+    end
+
+    private
+
+    def kind
+      @definition["kind"]
+    end
+
+    def rollout_config
+      @rollout_config ||= @crd.rollout_config
```
dturn: why not set this in the constructor?

test/integration-serial/serial_deploy_test.rb
```diff
@@ -235,6 +235,109 @@ def test_all_expected_statsd_metrics_emitted_with_essential_tags
...
+            "type": "Ready",
+            "reason": "test",
+            "message": "test",
+            "status": "True",
+          },
+        ],
+      },
+    }.to_json
+
+    thread = Thread.new { get_and_patch_cr(kind: "parameterized", name: "with-default-params", patch: success_patch) }
```
dturn: What happens if you don't do this in a separate thread?

lib/kubernetes-deploy/kubernetes_resource/custom_resource_definition.rb
```diff
@@ -2,8 +2,86 @@
 module KubernetesDeploy
   class CustomResourceDefinition < KubernetesResource
     TIMEOUT = 2.minutes
+    ROLLOUT_CONFIG_ANNOTATION = "kubernetes-deploy.shopify.io/monitor-instance-rollout"
     GLOBAL = true
 
+    class RolloutConfig
```
dturn: Let's stick this in its own file.

lib/kubernetes-deploy/kubernetes_resource/custom_resource_definition.rb
```diff
@@ -2,8 +2,86 @@
...
+      attr_reader :success_queries, :failure_queries
+
+      def initialize(rollout_config)
```
dturn: If I'm reading this right the annotation value has to be a hash. Can we support `true` as well for just wanting the default behavior, seems more obvious than `{}`

timothysmith0609: Fair point, I'll add that in

lib/kubernetes-deploy/kubernetes_resource/custom_resource.rb
timothysmith0609: Ah, it isn't. Leftover from my attempts at adding some more configurability (decided it's outside of this PR's scope)

lib/kubernetes-deploy/kubernetes_resource/custom_resource_definition.rb
timothysmith0609: Looking around I'm not sure I'd like to move this elsewhere, as it belongs squarely as a nested class of CustomResourceDefinition, even if it does bloat the file a little bit. Having it all in one place seems more advantageous to me in the end. If someone comes in with a tie-breaker I'll respect that decision, though

test/integration-serial/serial_deploy_test.rb
timothysmith0609: It turns out `deploy_fixtures` is more powerful than I thought at first. I've removed the crappy threaded code and used the optional block capability of `deploy_fixtures` instead :+1:

lib/kubernetes-deploy/kubernetes_resource/custom_resource_definition.rb
```diff
@@ -2,8 +2,86 @@
...
+
+        # Preemptively create JsonPath objects
+        config[:success_queries].map! do |query|
+          query.update(query) { |k, v| k == :path ? JsonPath.new(v) : v }
+        end
+        config[:failure_queries].map! do |query|
+          query.update(query) { |k, v| k == :path || k == :error_msg_path ? JsonPath.new(v) : v }
+        end
+
+        validate_config(config)
```
KnVerey: Generally speaking I've found it is a best practice to keep initializers lean and limit their ability to raise exceptions, so I'm a bit concerned about the fact that this one runs validations. What do you think about requiring the JSON to be parsed before this object is created (e.g. the signature might be `RolloutConfig.new(success_queries: [], failure_queries: [])`) and then having an explicit `valid?` method on it? We could call that from `validate_definition` on both CRDs and CR instances. That would also force us to think about other possible invalid states, e.g. if you have failure conditions but no success conditions or vice-versa, is that ok?

lib/kubernetes-deploy/kubernetes_resource/custom_resource_definition.rb
```diff
@@ -2,8 +2,86 @@
...
+        @failure_queries = config[:failure_queries]
+      rescue JSON::ParserError
+        raise FatalDeploymentError, "custom rollout params are not valid JSON: '#{rollout_config}'"
+      rescue RuntimeError => e
+        raise FatalDeploymentError, "error creating jsonpath objects, failed with: #{e}"
+      end
+
+      def deploy_succeeded?(instance_data)
+        @success_queries.all? do |query|
+          query[:path].on(instance_data).first == query[:value]
```
KnVerey: Why `.first`? When might we get multiple matches?

lib/kubernetes-deploy/kubernetes_resource/custom_resource.rb
```diff
@@ -0,0 +1,40 @@
...
+    def deploy_succeeded?
+      return super unless @rollout_config
+      return false unless observed_generation == current_generation
+
+      @rollout_config.deploy_succeeded?(@instance_data)
+    end
+
+    def deploy_failed?
+      return super unless @rollout_config
+      return false unless observed_generation == current_generation
```
KnVerey: Custom resource controllers are not required to implement the observed generation field, though it is a best practice. We should only make this comparison if the observed generation field is present.

lib/kubernetes-deploy/kubernetes_resource/custom_resource_definition.rb
```diff
@@ -2,8 +2,86 @@
...
+        validate_config(config)
+        @success_queries = config[:success_queries]
+        @failure_queries = config[:failure_queries]
+      rescue JSON::ParserError
+        raise FatalDeploymentError, "custom rollout params are not valid JSON: '#{rollout_config}'"
+      rescue RuntimeError => e
+        raise FatalDeploymentError, "error creating jsonpath objects, failed with: #{e}"
+      end
+
+      def deploy_succeeded?(instance_data)
```
KnVerey: Nit: maybe `rollout_succeeded?` or `successful_status?`? This object doesn't have to do with deploys, strictly speaking... it's kinda a status evaluator?

lib/kubernetes-deploy/kubernetes_resource/custom_resource.rb
timothysmith0609: I'm more inclined to make `observedGeneration` be a necessary field for those taking advantage of this feature. Without it, there's no way of knowing what generation of the resource the status is referring to. In the worst case, we'll deploy breaking changes to a previously healthy resource and, since the old status will indicate success, the deploy will pass instantaneously before the controller has a chance to update the status and indicate something has gone wrong.

A possible workaround is to measure against `lastTransitionTime`, but we can't really guarantee that field will be there, either. I'm open to suggestions but I don't see another way around it.

FWIW I intend to update all the cloudbuddies `UpdateStatus` functions to transparently populate the `observedGeneration` and `lastTransitionTime` fields (example [here](https://github.com/Shopify/cloudbuddies/pull/1522/files#diff-c06febb409f662c2c8f104b739e4888cR18))

lib/kubernetes-deploy/kubernetes_resource/custom_resource_definition.rb
timothysmith0609: I was thinking them more along the lines of delegated methods but I see your point. I'll make the change

lib/kubernetes-deploy/kubernetes_resource/custom_resource_definition.rb
timothysmith0609: This is a peculiarity of `jsonpath`

lib/kubernetes-deploy/kubernetes_resource/custom_resource_definition.rb
timothysmith0609: I like that idea and had the same reservations about raising inside the constructor

lib/kubernetes-deploy/kubernetes_resource/custom_resource_definition.rb
KnVerey: I guess when we will get multiple matches depends entirely on the path they give us and the structure it is matching... it's very easy to build queries that will match multiple results given a multi-level object, and [the gem's docs](https://github.com/joshbuddy/jsonpath) have some examples. If we did get multiple results, do you think the correct behaviour be `.first` or `.all?`?
 I also noticed in the docs that you can do `query.first(instance_data)` directly if that's what we want.

lib/kubernetes-deploy/kubernetes_resource/custom_resource.rb
KnVerey: You're definitely right that there are race conditions without that field--we saw them in the past, when even some core controllers didn't have it (and before we knew how important it is)! I don't think we should use any time-based logic. How do we require observed generation to be implemented though? Assume a user gets into the situation where they are correctly set up in every respect except this field--what do they see? One option would be to time out (which I think is the current behaviour with this PR), but make sure our error message explains that the observed generation field was missing / didn't match.

timothysmith0609:
I've pulled out `RolloutConfig` into its own file and added a number of class methods related to parsing and validating the initial config. I've also

- Setting the rollout annotation to '"true"' uses the default config ('{}' is now invalid)
- If you are specifying your own rollout config, you need to explicitly set _all_ of them (previous behaviour was that either failure or success queries could be defaulted if the config was only partially specified)
- the CRD attempts to parse the config at validation time and will fail with meaningful errors.

timothysmith0609:
>I like where this is going! I don't see any way to customize CR instance timeouts anymore; did you remove that from this PR on purpose?

I think we can use the timeout override annotation instead of specifying it again in the rollout config

lib/kubernetes-deploy/kubernetes_resource/custom_resource_definition.rb
timothysmith0609: I'm imagining that most jsonpaths will be fully specified (from root of document to target field) so am not too concerned about dealing with multiple matches. I think that would cause too much confusion in the implementation without a corresponding increase in usability; users should be able to get by with only single field matches

KnVerey:
> I think we can use the timeout override annotation instead of specifying it again in the rollout config

Typically there are three possible levels of timeouts:
(1) the gem's global default (`KubernetesResource::TIMEOUT #5 minutes`)
(2) the default for the resource type (`Deployment::TIMEOUT #7 minutes`)
(3) the timeout override for a resource instance (`kubernetes-deploy.shopify.io/timeout-override`)

It's (2) that we'd be missing for CRs unless it is settable via the rollout config. Users shouldn't have to put (3) on every single CR instance to avoid inheriting (1), which in cases like cloudsql is a really bad default.

lib/kubernetes-deploy/kubernetes_resource/custom_resource_definition.rb
```diff
@@ -2,8 +2,86 @@
 module KubernetesDeploy
   class CustomResourceDefinition < KubernetesResource
     TIMEOUT = 2.minutes
+    ROLLOUT_CONFIG_ANNOTATION = "kubernetes-deploy.shopify.io/monitor-instance-rollout"
     GLOBAL = true
 
+    class RolloutConfig
+      attr_reader :success_queries, :failure_queries
```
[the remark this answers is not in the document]
timothysmith0609: They were. I've changed the tests around and removed the accessors

test/fixtures/crd/with_custom_params.yml
```diff
@@ -0,0 +1,38 @@
...
+  name: customizeds.stable.example.io
+  annotations:
+    kubernetes-deploy.shopify.io/monitor-instance-rollout: '{
+      "success_queries": [
+        {
+          "path": "$.status.condition",
+          "value":"success_value"
+        },
+        {
+          "path":"$.spec.test_field",
```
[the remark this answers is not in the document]
timothysmith0609: I don't want to artificially limit what people may or may not want to do but will remove paths to `.spec` in the tests so as not to encourage it.

test/integration-serial/serial_deploy_test.rb
```diff
@@ -235,6 +235,116 @@ def test_all_expected_statsd_metrics_emitted_with_essential_tags
...
+            "message": "test",
+            "status": "True",
+          },
+        ],
+      },
+    }
+
+    result = deploy_fixtures("crd", subset: ["with_default_params_cr.yml"]) do |resource|
+      cr = resource["with_default_params_cr.yml"]["Parameterized"].first
+      resource["with_default_params_cr.yml"]["Parameterized"][0] = cr.merge!(success_conditions).deep_stringify_keys!
```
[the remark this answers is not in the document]
timothysmith0609: Ah, using a string-based hash works with `merge!` :+1:

test/integration-serial/serial_deploy_test.rb
```diff
@@ -235,6 +235,116 @@ def test_all_expected_statsd_metrics_emitted_with_essential_tags
...
+      resource["with_custom_params_cr.yml"]["Customized"][0] = cr.merge!(failure_conditions).deep_stringify_keys!
+    end
+    assert_deploy_failure(result)
+    assert_logs_match_all([
+      /test error message jsonpath/,
+      /test custom error message/,
+    ])
+  ensure
+    wait_for_all_crd_deletion
+  end
```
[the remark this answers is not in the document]
timothysmith0609: A CRD with invalid rollout annotation will fail validation.
A CR will attempt to validate it's CRD's rollout config and will fail if it is not valid.

To answer your question, then: deploying multiple CRs of a kind whose CRD has invalid rollout specs will fail the deploys of the CR (irrespective of whether or not the CRD is deployed concurrently with the CR)

timothysmith0609:
@Shopify/cloudx This is ready for :eyes: again

README.md
```diff
@@ -312,7 +313,93 @@ Since their data is only base64 encoded, Kubernetes secrets should not be commit
...
+
+*Requirements:*
+
+* The custom resource must expose a `status` field with an `observedGeneration` field.
+* The `kubernetes-deploy.shopify.io/cr-instance-rollout-conditions` annotation must be present on the *CRD* that defines the custom resource.
+* (optional) The `kubernetes-deploy.shopify.io/cr-instance-timeout` annotation can be added to the *CRD* that defines the custom resource to specify a default timeout for all instances of the CRD (otherwise the global default is used).
+
+### Specifying pass/fail conditions
+
+The presence of a valid `kubernetes-deploy.shopify.io/cr-instance-rollout-conditions` annotation on a CRD will cause kubernetes-deploy to monitor the rollout of any of its instances with the supplied annotation value. This value must be a valid JSON string with the following format:
```
dturn: Does this need to be updated to say we also support `True`

lib/kubernetes-deploy/rollout_conditions.rb
```diff
@@ -0,0 +1,115 @@
+# frozen_string_literal: true
+module KubernetesDeploy
+  class RolloutConditionsError < StandardError
+  end
+
+  class RolloutConditions
+    class << self
+      def parse_conditions(conditions_string)
```
dturn: Why did you decide to have parsing be a class method that returns a hash, instead of `from_conditions` method ?

test/unit/kubernetes-deploy/kubernetes_resource/custom_resource_definition_test.rb
```diff
@@ -0,0 +1,200 @@
...
+        },
+      }
+    ))
+    cr = KubernetesDeploy::KubernetesResource.build(namespace: "test", context: "test",
+      logger: @logger, statsd_tags: @statsd_tags, crd: crd,
+      definition: { "kind" => "UnitTest", "metadata" => { "name" => "test" } })
+    assert_equal(cr.timeout, 60)
+  end
+
+  def test_instance_timeout_messages_with_rollout_conditions
```
dturn: Do the timeout tests actually take the full timeout duration? If so should we do something to speed that up?

lib/kubernetes-deploy/rollout_conditions.rb
KnVerey: I think I agree with Danny's suggestion (`from_annotation` that returns `RolloutConditions` instance?), and I have a couple other ideas about this class I'd like to run by to you tomorrow if you have time. :)

lib/kubernetes-deploy/rollout_conditions.rb
```diff
@@ -0,0 +1,115 @@
...
+
+      private
+
+      def validate_conditions(conditions)
+        errors = []
+
+        top_level_keys = [:success_conditions, :failure_conditions]
+        missing = top_level_keys.reject { |k| conditions.key?(k) }
+        unless missing.blank?
+          errors << "Missing required top-level key(s): #{missing}"
```
KnVerey: Wouldn't it be valid for them to provide success conditions but not failure conditions? That's the case for a lot of core resource types we've manually modelled.

README.md
timothysmith0609: As this is an annotation, we can only support strings, not bools. Said that, kubernetes-deploy will downcase the value so things like `"True", "TRUE"`, etc. will work, though I don't think that needs to be documented

lib/kubernetes-deploy/kubernetes_resource/custom_resource_definition.rb
```diff
@@ -1,7 +1,11 @@
 # frozen_string_literal: true
+require 'kubernetes-deploy/rollout_conditions'
+
 module KubernetesDeploy
   class CustomResourceDefinition < KubernetesResource
     TIMEOUT = 2.minutes
+    ROLLOUT_CONDITIONS_ANNOTATION = "kubernetes-deploy.shopify.io/cr-instance-rollout-conditions"
```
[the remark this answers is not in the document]
timothysmith0609: I'm ok with that. Originally I was just worried that seeing the annotation out of context may be confusing and that's why I opted to put the `cr` prefix in the first place

test/integration-serial/serial_deploy_test.rb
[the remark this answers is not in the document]
timothysmith0609: Added a test that uses a CRD in-cluster with invalid rollout conditions. As expected, this fails the deploy of the CR during validation

timothysmith0609:
- memoized `@rollout_conditions` using `defined?`
- add test for deploying CR with already-present CRD that has bad rollout_conditions (CR should fail on validation)
- README updates / PR feedback
- `predeploy_sequence` has `before_crs` and `after_crs` sections for resources that should be deployed first
- added `CustomResourceDefinition` to `predeploy_sequence`

@Shopify/cloudx ready for another pass

lib/kubernetes-deploy/deploy_task.rb
```diff
@@ -73,6 +60,23 @@ class DeployTask
     # extensions/v1beta1/ReplicaSet -- managed by deployments
     # core/v1/Secret -- should not committed / managed by shipit
 
+    def predeploy_sequence
+      before_crs = %w(
+        ResourceQuota
+        CustomResourceDefinition
```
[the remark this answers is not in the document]
dturn: I believe can't deploy the CRD for the first time and CR at the same time. The CR will fail validation since the cluster doesn't know about the crd.

lib/kubernetes-deploy/kubernetes_resource/custom_resource.rb
```diff
@@ -0,0 +1,70 @@
+# frozen_string_literal: true
+require 'jsonpath'
+module KubernetesDeploy
+  class CustomResource < KubernetesResource
+    TIMEOUT_MESSAGE_DIFFERENT_GENERATIONS = <<~MSG
+      This resource's status could not be used to determine rollout success because it was still out of date
+      (.metadata.generation != .status.observedGeneration) after #{timeout}s.
```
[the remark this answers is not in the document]
timothysmith0609: It's referencing the `timeout` _class_ method on `KubernetesResource` :( 
From what I see `debug_message` will already show the timeout for a given resource so we can just drop the reference here but it will still be visible in the logs

test/unit/kubernetes-deploy/kubernetes_resource/custom_resource_definition_test.rb
```diff
@@ -0,0 +1,225 @@
...
+        "kind" => "UnitTest",
+        "metadata" => {
+          "name" => "test",
+          "generation" => 1,
+        },
+        "status" => { "observedGeneration" => 1 },
+      })
+
+    cr.expects(:current_generation).returns(1)
+    cr.expects(:observed_generation).returns(1)
```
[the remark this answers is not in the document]
timothysmith0609: those methods observe the value from `@instance_data`, not `@definition`. Since `@instance_data` isn' instantiated in this test, I need the stub

test/unit/kubernetes-deploy/kubernetes_resource/custom_resource_definition_test.rb
```diff
@@ -0,0 +1,225 @@
...
+    crd = build_crd(crd_spec.merge(
+      "metadata" => {
+        "name" => "unittests.stable.example.io",
+        "annotations" => {
+          KubernetesDeploy::CustomResourceDefinition::ROLLOUT_CONDITIONS_ANNOTATION => "true",
+        },
+      },
+    ))
+    cr = KubernetesDeploy::KubernetesResource.build(namespace: "test", context: "test",
+      logger: @logger, statsd_tags: @statsd_tags, crd: crd,
```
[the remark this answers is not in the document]
timothysmith0609: Nowhere (it's just a nil instance var). I can use explicit nil if you prefer

test/integration-serial/serial_deploy_test.rb
```diff
@@ -235,6 +235,176 @@ def test_all_expected_statsd_metrics_emitted_with_essential_tags
...
+    crd_result = deploy_fixtures("crd", subset: ["with_custom_conditions.yml"]) do |resource|
+      crd = resource["with_custom_conditions.yml"]["CustomResourceDefinition"].first
+      crd["metadata"]["annotations"].merge!(
+        KubernetesDeploy::CustomResourceDefinition::ROLLOUT_CONDITIONS_ANNOTATION => "blah"
+      )
+    end
+    assert_deploy_success(crd_result)
+    KubernetesDeploy::DeployTask.any_instance.unstub(:validate_resources)
+
+    cr_result = deploy_fixtures("crd", subset: ["with_custom_conditions_cr.yml"])
```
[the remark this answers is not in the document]
timothysmith0609: As it stands this seems quite tricky to do and I'm not convinced it's the desired behaviour. Currently, we will print the error for each CR associated with a poorly specced CRD. I'm wondering how to get around this since
- All CRs contain a reference to their parent CRD
- In `validate_definition` of the CRD, `@validation_errors` will contain some errors.
- I'm not sure how/if to short-circuit this so that if the CRD is already invalidated we suppress the errors. We _could_ set `@validation_errors` to `nil` _if_ `@rollout_conditions_validated?`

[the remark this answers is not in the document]
timothysmith0609: On further reflection I feel strongly that we should raise the error per CR. We should definitely make the error more explicit that the CR is failing because of the CRD as you suggest below.

test/unit/kubernetes-deploy/kubernetes_resource/custom_resource_definition_test.rb
```diff
@@ -0,0 +1,225 @@
...
+    cr.expects(:current_generation).returns(1)
+    cr.expects(:observed_generation).returns(1)
+    assert_equal(cr.timeout_message, KubernetesDeploy::KubernetesResource::STANDARD_TIMEOUT_MESSAGE)
+
+    cr.expects(:current_generation).returns(1)
+    cr.expects(:observed_generation).returns(2)
+    assert_equal(cr.timeout_message, KubernetesDeploy::CustomResource::TIMEOUT_MESSAGE_DIFFERENT_GENERATIONS)
+  end
+
+  def test_instance_timeout_messages_without_rollout_conditions
```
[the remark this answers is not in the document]
timothysmith0609: I asked myself the same question and my answer was to keep it in `custom_resource_definition_test.rb` due to the overlap of the `RolloutConditions, CustomResourceDefinition, and CustomResource` classes.

test/integration-serial/serial_deploy_test.rb
```diff
@@ -235,6 +235,176 @@ def test_all_expected_statsd_metrics_emitted_with_essential_tags
...
+    end
+    assert_deploy_success(result)
+    assert_logs_match_all([
+      %r{Successfully deployed in .*: Customized\/with-customized-params},
+    ])
+  ensure
+    wait_for_all_crd_deletion
+  end
+
+  def test_cr_success_with_only_success_rollout_conditions
```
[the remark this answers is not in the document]
timothysmith0609: I'll go over these and eliminate the redundant tests. I had the same feeling that I was testing the same thing twice

test/integration-serial/serial_deploy_test.rb
```diff
@@ -235,6 +235,176 @@ def test_all_expected_statsd_metrics_emitted_with_essential_tags
...
+        KubernetesDeploy::CustomResourceDefinition::ROLLOUT_CONDITIONS_ANNOTATION => "blah"
+      )
+    end
+    assert_deploy_success(crd_result)
+    KubernetesDeploy::DeployTask.any_instance.unstub(:validate_resources)
+
+    cr_result = deploy_fixtures("crd", subset: ["with_custom_conditions_cr.yml"])
+    assert_deploy_failure(cr_result)
+    assert_logs_match_all([
+      /Invalid template: Customized-with-customized-params/,
```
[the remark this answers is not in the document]
timothysmith0609: What do you think about this?

screen shot 2019-01-18 at 12 44 19 pm

lib/kubernetes-deploy/kubernetes_resource/custom_resource.rb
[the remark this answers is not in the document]
timothysmith0609: For ref:
screen shot 2019-01-18 at 1 10 36 pm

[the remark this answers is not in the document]
KnVerey: maybe "it was still out of date when the timeout was reached", or rephrase in the present tense? I think you're totally right that we don't need to repeat the actual timeout, but the sentence sounds a little odd to me with the end removed.

test/integration-serial/serial_deploy_test.rb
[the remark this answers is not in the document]
KnVerey: ok, I agree with you.

test/integration-serial/serial_deploy_test.rb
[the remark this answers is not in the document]
KnVerey: Excellent! Very helpful. Just lowercase the "U" in ", Under the annotation".

test/unit/kubernetes-deploy/kubernetes_resource/custom_resource_definition_test.rb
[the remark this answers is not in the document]
KnVerey: Ah right, you'd need to call `sync` (i.e. use `ResourceCacheTestHelper`)

test/unit/kubernetes-deploy/kubernetes_resource/custom_resource_definition_test.rb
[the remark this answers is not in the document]
KnVerey: I would use an explicit `[]` actually since that's `initialize`'s default.

lib/kubernetes-deploy/kubernetes_resource/custom_resource.rb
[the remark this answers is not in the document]
KnVerey: Also `Final status: <blank>` needs to be fixed there

test/integration-serial/serial_deploy_test.rb
```diff
@@ -235,6 +235,147 @@ def test_all_expected_statsd_metrics_emitted_with_essential_tags
...
+    }
+
+    result = deploy_fixtures("crd", subset: ["with_default_conditions_cr.yml"]) do |resource|
+      cr = resource["with_default_conditions_cr.yml"]["Parameterized"].first
+      cr.merge!(failure_conditions)
+    end
+    assert_deploy_failure(result)
+    assert_logs_match_all([
+      "Parameterized/with-default-params: FAILED",
+      "custom resource rollout failed",
```
KnVerey: here too

lib/kubernetes-deploy/rollout_conditions.rb
```diff
@@ -0,0 +1,102 @@
...
+      errors += validate_conditions(@failure_conditions, 'failure_conditions', required: false)
+      raise RolloutConditionsError, errors.join(", ") unless errors.empty?
+    end
+
+    private
+
+    def validate_conditions(conditions, source_key, required: true)
+      return [] unless conditions.present? || required
+      errors = []
+      errors << "#{source_key} should be Array but found #{conditions.class}" unless conditions.is_a?(Array)
```
KnVerey: I think we're still missing a unit test for this case

PR #377, opened 2018-11-14
[skip ci] Update change log

dturn:
Update change log for v0.22.0 release


dturn:
I'll add both. Couldn't decide if that bug fix was big enough to include.

PR #379, opened 2018-11-16
Fix lint issues on master

airhorns:
Totally take it or leave it, but I like to be able to run `bx rubocop -a` to make sure I am not going to hit policial issues when I push. `-a` fixes everything it can, which is awesome because it means less precious human time spent wrapping lines, but, if there are outstanding lint errors on master, it fixes those too, and gives you a dirtier git diff that you have to selectively stage to avoid fixing those master lint issues by accident.

This just fixes those master lint issues in one go. There are not very many, so I think the risk to the git blame being diluted is low, and it means that everyone can turn on `Format on Save` in their editor or run the command line auto correct without prejudice. Horray?


.rubocop.yml
```diff
@@ -9,3 +9,8 @@ Style/TrailingCommaInArrayLiteral:
 
 Style/TrailingCommaInHashLiteral:
   Enabled: false
+
+Naming/FileName:
+  Enabled: true
+  Exclude:
+    - lib/kubernetes-deploy.rb
```
dturn: Should we just fix the file name?

airhorns: Ya, I think it needs to match for ultimately silly reasons.

lib/kubernetes-deploy/kubeclient_builder/google_friendly_config.rb
```diff
@@ -34,8 +34,8 @@ def new_token
       def json_error_message(body)
         json_error_msg = begin
           JSON.parse(body || '') || {}
-        rescue JSON::ParserError
-          {}
+                         rescue JSON::ParserError
```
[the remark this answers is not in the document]
airhorns: Switched to Katrina's suggestion

PR #380, opened 2018-11-24
Predeploy roles before RoleBindings

lautis:
RoleBindings were added to predeploy in #354. If role bindings rely on a role, deployment fails as role is not available. Fixed by predeploying roles.

### DIFF
--- a/test/helpers/fixture_sets/hello_cloud.rb
+++ b/test/helpers/fixture_sets/hello_cloud.rb
@@ -89,6 +89,10 @@ module FixtureSetAssertions
       assert_service_account_present("build-robot")
     end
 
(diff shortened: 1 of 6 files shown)


lautis:
Fixed the rubocop error.

test/helpers/fixture_sets/hello_cloud.rb
```diff
@@ -89,6 +89,10 @@ def assert_all_service_accounts_up
       assert_service_account_present("build-robot")
     end
 
+    def assert_all_roles_up
```
KnVerey: Can you please call this from `assert_all_up` (L9 of this file) too?

PR #381, opened 2018-11-26
Performance visbility

KnVerey:
## Purpose

Add debugging tools to the deploy task and make a few small changes that _might_ help performance.

## How
* Adds duration metrics for all the major steps in a deploy
* Changes the timeout used by kubectl to 15s (down from 30s)
* Adds retries to the cache-population step. The only way I can think of to make multiple kubectl processes request all resources of a given kind in parallel is for this cache to be empty, so let's try harder to make sure it isn't.
* Caches the list of CRDs. The call likely isn't very expensive, but I noticed in my debugging efforts that we're requesting them over and over for no good reason.
* Makes debug logging more usable
* Reverts the batch size optimization. I don't think that it is causing the problem, but it doesn't hurt to eliminate this as a reason for the cache being unpopulated. I also now think there might be better solutions to underlying problem that was trying to solve.

The functional diff is actually quite small. Most of the changes are moving code around and fixing unit tests. The changes are fully tested and all tests are passing.


lib/kubernetes-deploy/statsd.rb
```diff
@@ -22,5 +22,30 @@ def self.build
       end
       ::StatsD.backend
     end
+
+    module MeasureMethods
+      def measure_method(method_name, metric = nil)
```
KnVerey: This is based on a [Shopify core PR](https://github.com/Shopify/shopify/pull/152598/files) that I found in looking up how to use dynamic tags in statd-instrument (which isn't possible).

lib/kubernetes-deploy/deploy_task.rb
```diff
@@ -43,6 +43,7 @@
 module KubernetesDeploy
   class DeployTask
     include KubeclientBuilder
+    extend KubernetesDeploy::StatsD::MeasureMethods
```
KnVerey: The only real changes in this file are the new instrumentations via `measure_method`. The rest is just moving code around to package it into the methods I want to measure.

lib/kubernetes-deploy/kubectl.rb
```diff
@@ -20,26 +21,35 @@ def initialize(namespace:, context:, logger:, log_failure_by_default:, default_t
...
-        if log_failure
-          @logger.warn("The following command failed: #{Shellwords.join(args)}")
-          @logger.warn(err) unless output_is_sensitive?
-        end
-
-        if raise_if_not_found && err.match(NOT_FOUND_ERROR_TEXT)
-          raise ResourceNotFoundError, err
+      out, err, st = nil
+
+      (1..attempts).to_a.each do |attempt|
```
fw42: `attempts.times`?

lib/kubernetes-deploy/statsd.rb
```diff
@@ -22,5 +22,30 @@ def self.build
...
+        unless method_defined?(method_name) || private_method_defined?(method_name)
+          raise NotImplementedError, "Cannot instrument undefined method #{method_name}"
+        end
+
+        unless const_defined?("InstrumentationProxy")
+          const_set("InstrumentationProxy", Module.new)
+          should_prepend = true
+        end
+        instrumentation_proxy = const_get("InstrumentationProxy")
```
fw42: why isn't this just saying `instrumentation_proxy = InstrumentationProxy`?

lib/kubernetes-deploy/statsd.rb
```diff
@@ -22,5 +22,30 @@ def self.build
...
     end
+
+    module MeasureMethods
+      def measure_method(method_name, metric = nil)
+        unless method_defined?(method_name) || private_method_defined?(method_name)
+          raise NotImplementedError, "Cannot instrument undefined method #{method_name}"
+        end
+
+        unless const_defined?("InstrumentationProxy")
+          const_set("InstrumentationProxy", Module.new)
```
fw42: why not just `InstrumentationProxy = Module.new`?

lib/kubernetes-deploy/statsd.rb
```diff
@@ -22,5 +22,30 @@ def self.build
...
+          should_prepend = true
+        end
+        instrumentation_proxy = const_get("InstrumentationProxy")
+        metric ||= "#{method_name}.duration"
+
+        instrumentation_proxy.send(:define_method, method_name) do |*args, &block|
+          start_time = Time.now.utc
+          result = super(*args, &block)
```
fw42: If the `super` raises an error then you will never send the metric. Is that on purpose? Maybe you want to send it with a `error: true` tag?

lib/kubernetes-deploy/sync_mediator.rb
```diff
@@ -36,25 +43,29 @@ def get_all(kind, selector = nil)
...
       end
+      kinds = (resources.map(&:kubectl_resource_type) + dependencies).compact.uniq
+      @logger.debug("Populating cache for kinds: #{kinds.join(', ')}")
+      kinds.each { |kind| fetch_by_kind(kind, attempts: 5) }
 
       KubernetesDeploy::Concurrency.split_across_threads(resources) do |r|
         r.sync(dup)
       end
     end
+    measure_method(:sync, "sync.duration")
```
fw42: nitpick: the `"sync.duration"` is unnecessary, that's already the default name

lib/kubernetes-deploy/statsd.rb
fw42: The gem already supports [tags](https://github.com/Shopify/statsd-instrument#tags) as well as [dynamic metric names](https://github.com/Shopify/statsd-instrument#dynamic-metric-names). Maybe we should just add this feature to the gem (which is owned by Shopify)?

lib/kubernetes-deploy/kubectl.rb
KnVerey: That would yield 0 to 4 instead of 1 to 5, so there'd be a tradeoff in less elegant code below where `attempt` is used.

lib/kubernetes-deploy/statsd.rb
KnVerey: Good question. When I did it that way, the constant seemed to get looked up in the module's namespace instead of the extending class's, whereas `self` when we do `const_set` is the extending class. Error is:`uninitialized constantKubernetesDeploy::StatsD::MeasureMethods::InstrumentationProxy`. I can't say I understand why. `self::InstrumetationProxy` actually works though if that seems better.

lib/kubernetes-deploy/statsd.rb
KnVerey: `dynamic constant assignment (SyntaxError)`

lib/kubernetes-deploy/statsd.rb
KnVerey: Not on purpose. Added that, with tests.

lib/kubernetes-deploy/deploy_task.rb
```diff
@@ -227,8 +214,9 @@ def predeploy_priority_resources(resource_list)
         @logger.blank_line
       end
     end
+    measure_method(:predeploy_priority_resources, 'priority_resources.duration')
```
dturn: The lack of decorators in ruby makes me sad

lib/kubernetes-deploy/kubectl.rb
```diff
@@ -20,26 +21,35 @@ def initialize(namespace:, context:, logger:, log_failure_by_default:, default_t
...
+            @logger.warn(err) unless output_is_sensitive?
+          end
+
+          if err.match(NOT_FOUND_ERROR_TEXT)
+            raise(ResourceNotFoundError, err) if raise_if_not_found
+          else
+            @logger.debug("Kubectl err: #{err}") unless output_is_sensitive?
+            ::StatsD.increment('kubectl.error', 1, tags: { context: @context, namespace: @namespace, cmd: args[1] })
+          end
+          sleep RETRY_DELAY unless attempt == attempts
```
dturn: Any reason why you decided not to have increasing back off on each attempt?

lib/kubernetes-deploy/resource_discovery.rb
```diff
@@ -10,7 +10,7 @@ def initialize(namespace:, context:, logger:, namespace_tags:)
     end
 
     def crds(sync_mediator)
-      sync_mediator.get_all(CustomResourceDefinition.kind).map do |r_def|
+      @crds ||= sync_mediator.get_all(CustomResourceDefinition.kind).map do |r_def|
```
dturn: Should we save the `ResourceDiscovery` instance into a ivar is deploy_task.rb?

test/unit/kubernetes-deploy/kubectl_test.rb
```diff
@@ -21,7 +21,7 @@ def test_raises_if_initialized_with_null_namespace
   end
 
   def test_run_constructs_the_expected_command_and_returns_the_correct_values
-    stub_open3(%w(kubectl get pods -a --output=json --namespace=testn --context=testc --request-timeout=30),
+    stub_open3(%w(kubectl get pods -a --output=json --namespace=testn --context=testc --request-timeout=15),
```
dturn: Should this just reference the actual constant?

lib/kubernetes-deploy/kubectl.rb
```diff
@@ -20,26 +21,35 @@ def initialize(namespace:, context:, logger:, log_failure_by_default:, default_t
...
+      (1..attempts).to_a.each do |attempt|
+        @logger.debug "Running command (attempt #{attempt}): #{args.join(' ')}"
+        out, err, st = Open3.capture3(*args)
+        @logger.debug("Kubectl out: " + out.gsub(/\s+/, ' ')) unless output_is_sensitive?
+
+        if st.success?
+          break
+        else
+          if log_failure
+            @logger.warn("The following command failed: #{Shellwords.join(args)}")
```
dturn: I think we should add the something like `attempt: #{attempt}/#{attempts}` to this message.

lib/kubernetes-deploy/kubectl.rb
KnVerey: Not really. Just did the simplest thing to start. Another simple option could be `sleep attempt` though (so an additional second each time).

lib/kubernetes-deploy/resource_discovery.rb
KnVerey: This PR does do that (the method is `cluster_resource_discoverer`)

lib/kubernetes-deploy/kubectl.rb
KnVerey: The attempt number is already recorded at the beginning of each attempt, so you see it whether the command ultimately succeeds or fails: `@logger.debug "Running command (attempt #{attempt}): #{args.join(' ')}"`

lib/kubernetes-deploy/kubectl.rb
dturn: `sleep attempt` works for me.

lib/kubernetes-deploy/kubectl.rb
dturn: It just seemed odd to me that the test in `test_run_with_multiple_attempts_retries_and_emits_failure_metrics` doesn't actually have the attempts in the output

KnVerey: It does though:

image

dturn:
Was I looking at an older version?
[image: Screen Shot 2018-11-26 at 2.47.54 PM.png]

lib/kubernetes-deploy/kubectl.rb
```diff
@@ -20,32 +20,43 @@ def initialize(namespace:, context:, logger:, log_failure_by_default:, default_t
...
 
-    def run(*args, log_failure: nil, use_context: true, use_namespace: true, raise_if_not_found: false)
+    def run(*args, log_failure: nil, use_context: true, use_namespace: true, raise_if_not_found: false, attempts: 1)
       log_failure = @log_failure_by_default if log_failure.nil?
 
       args = args.unshift("kubectl")
       args.push("--namespace=#{@namespace}") if use_namespace
       args.push("--context=#{@context}")     if use_context
       args.push("--request-timeout=#{@default_timeout}") if @default_timeout
+      out, err, st = nil
```
dturn: til

PR #383, opened 2018-11-26
Use distribution everywhere

dturn:
This is a follow-up #374  to use `distribution` over `measure`. See  https://github.com/Shopify/statsd-instrument#statsddistribution for more details. But at a high level `distribution` has all arithmetic and calculation happening server side which is more efficient, correct, and cost effective.

### DIFF
--- a/lib/kubernetes-deploy/deploy_task.rb
+++ b/lib/kubernetes-deploy/deploy_task.rb
@@ -150,7 +150,8 @@ module KubernetesDeploy
         @logger.phase_heading("Predeploying priority resources")
         start_priority_resource = Time.now.utc
         predeploy_priority_resources(resources)
-        ::StatsD.measure('priority_resources.duration', StatsD.duration(start_priority_resource), tags: statsd_tags)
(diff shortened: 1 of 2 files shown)


lib/kubernetes-deploy/deploy_task.rb
```diff
@@ -161,7 +162,7 @@ def run!(verify_result: true, allow_protected_ns: false, prune: true)
       if verify_result
         start_normal_resource = Time.now.utc
         deploy_resources(resources, prune: prune, verify: true)
-        ::StatsD.measure('normal_resources.duration', StatsD.duration(start_normal_resource), tags: statsd_tags)
+        ::StatsD.distribution('normal_resources.duration', StatsD.duration(start_normal_resource), tags: statsd_tags)
```
KnVerey: These first two were converted to use `measure_method` so they are already distributions on master. Please rebase. 🙏

dturn:
#386 (rather than fight rebase conflicts)

Issue #336, opened 2018-10-01
Don't hijack global StatsD configuration


KnVerey:
# Bug report

Kubernetes-deploy is making global changes to StatsD configuration, which makes it a bad citizen when used as a gem. In particular, we saw an app require it in production and start emitting metrics like `KubernetesDeploy.request_time` because of this line: https://github.com/Shopify/kubernetes-deploy/blob/master/lib/kubernetes-deploy/statsd.rb#L13

**Expected behavior:** 
Kubernetes-deploy uses the app's existing StatsD configuration if applicable, and prefixes only its own metrics.

**Actual behavior:** 
Kubernetes-deploy hijacks the StatsD configuration for the whole app and prefixes all metrics.

We should fix this, but note that the offending app is being shut down anyway, so there is no urgency internally. @Shopify/cloudx

PR #384, opened 2018-11-26
Don't hijack statsd

timothysmith0609:
## Problem

Currently, if KubernetesDeploy is used as part of a library, it will hijack the global StatsD configuration and, among other things, will prefix all metrics with `KubernetesDeploy`. Ideally, KubernetesDeploy should not mutate global config like this and its StatsD configuration should be scoped more tightly to itself.

Closes [#336](https://github.com/Shopify/kubernetes-deploy/issues/336).

## Proposed solution

Instead of `KubernetesDeploy::StatsD` simply acting as an interface to the global StatsD configuration, it will instead extend it and become a singleton in its own right. This allows us to specify config parameters for `KubernetesDeploy::StatsD` without overriding the global configuration.

## Issues

Unfortunately, some methods in the StatsD library make statically scoped calls to the [StatsD gem itself](https://github.com/Shopify/statsd-instrument/blob/master/lib/statsd/instrument/metric.rb#L50). Thankfully this only seems to occur when initializing new metrics. After perusing the statsd-instrument repo, I noticed that we aren't the first to come across this and the proposed solution was pretty much what I had in mind. The PR is [here](https://github.com/Shopify/statsd-instrument/pull/148). Testing locally with the changes to the gem yields successful test passes **Edit: this has been shipped**


lib/kubernetes-deploy/deploy_task.rb
```diff
@@ -150,7 +150,7 @@ def run!(verify_result: true, allow_protected_ns: false, prune: true)
         @logger.phase_heading("Predeploying priority resources")
         start_priority_resource = Time.now.utc
         predeploy_priority_resources(resources)
-        ::StatsD.measure('priority_resources.duration', StatsD.duration(start_priority_resource), tags: statsd_tags)
+        StatsD.measure('priority_resources.duration', StatsD.duration(start_priority_resource), tags: statsd_tags)
```
dturn: I've got a PR for this as an fyi https://github.com/Shopify/kubernetes-deploy/pull/383

lib/kubernetes-deploy/statsd.rb
```diff
@@ -4,23 +4,36 @@
...
+      mock_backend = ::StatsD::Instrument::Backends::CaptureBackend.new
+      old_backend, self.backend = self.backend, mock_backend
+      block.call
+      mock_backend.collected_metrics
+    ensure
+      if old_backend.kind_of?(::StatsD::Instrument::Backends::CaptureBackend)
+        old_backend.collected_metrics.concat(mock_backend.collected_metrics)
+      end
+
+      self.backend = old_backend
```
timothysmith0609: Unfortunately we can't use the actual library version of this since it scopes to the global `::StatsD` object. While ugly, overriding it here accomplishes the goal of mutating only the KubernetesDeploy StatsD settings.

lib/kubernetes-deploy/statsd.rb
```diff
@@ -4,23 +4,36 @@
...
-        ::StatsD.backend = ::StatsD::Instrument::Backends::UDPBackend.new(ENV['STATSD_ADDR'], statsd_impl)
+        self.backend = ::StatsD::Instrument::Backends::UDPBackend.new(ENV['STATSD_ADDR'], statsd_impl)
       else
-        ::StatsD.backend = ::StatsD::Instrument::Backends::NullBackend.new
+        self.backend = ::StatsD::Instrument::Backends::NullBackend.new
       end
-      ::StatsD.backend
+    end
+
+    def self.capture_statsd_calls(&block)
```
dturn: This should probably get moved to a file in `test/`

timothysmith0609:
This PR is currently blocked by [statsd-instrument #148](https://github.com/Shopify/statsd-instrument/pull/148), which allows for multiple prefixes.

timothysmith0609:
[statsd-instrument #148](https://github.com/Shopify/statsd-instrument/pull/148) has been shipped and released as `v2.3.2`. This lets us use the `:prefix` option when creating metrics.

lib/kubernetes-deploy/statsd.rb
```diff
@@ -49,7 +50,13 @@ def measure_method(method_name, metric = nil)
               dynamic_tags[:error] = error if dynamic_tags.is_a?(Hash)
               dynamic_tags << "error:#{error}" if dynamic_tags.is_a?(Array)
             end
-            ::StatsD.distribution(metric, KubernetesDeploy::StatsD.duration(start_time), tags: dynamic_tags)
+
+            StatsD.distribution(
+              metric,
+              KubernetesDeploy::StatsD.duration(start_time),
+              tags: dynamic_tags,
+              prefix: "KubernetesDeploy"
```
KnVerey: let's make this a constant

lib/kubernetes-deploy/runner_task.rb
```diff
@@ -45,14 +45,17 @@ def run!(task_template:, entrypoint:, args:, env_vars: [], verify_result: true)
...
       @logger.print_summary(:success)
     rescue DeploymentTimeoutError
-      ::StatsD.distribution('task_runner.duration', StatsD.duration(start), tags: statsd_tags('timeout'))
+      StatsD.distribution('task_runner.duration', StatsD.duration(start),
+        tags: statsd_tags('timeout'), prefix: "KubernetesDeploy")
       @logger.print_summary(:timed_out)
       raise
     rescue FatalDeploymentError
-      ::StatsD.distribution('task_runner.duration', StatsD.duration(start), tags: statsd_tags('failure'))
+      StatsD.distribution('task_runner.duration', StatsD.duration(start),
```
KnVerey: could we overwrite this (and the other similar methods) in our class to call `super` with the prefix so that we don't have to write out the prefix every time?

test/helpers/statsd_helper.rb
```diff
@@ -0,0 +1,19 @@
+# frozen_string_literal: true
+module StatsDHelper
+  extend self
```
KnVerey: Why make this a class method? I think our other helpers use the model of making the test classes that need them `include` the helper's class.

test/integration-serial/serial_task_run_test.rb
```diff
@@ -29,13 +29,13 @@ def test_run_without_verify_result_fails_if_pod_was_not_created
     ], in_order: true)
   end
 
-  # Run statsd tests in serial because capture_statsd_calls modifies global state in a way
+  # Run statsd tests in serial because StatsDHelper.StatsDHelper.capture_statsd_calls modifies global state in a way
```
KnVerey: unintentional change?

test/test_helper.rb
```diff
@@ -67,6 +67,7 @@ def run
     end
 
     def setup
+      KubernetesDeploy::StatsD.build
```
KnVerey: Why do we need to do this now?

lib/kubernetes-deploy/runner_task.rb
timothysmith0609: I was thinking the same thing and agree with you. I'll make the change

test/test_helper.rb
timothysmith0609: Ah, we don't. My mistake

timothysmith0609:
Rebased on master. This is ready for a final round of 👀

lib/kubernetes-deploy/deploy_task.rb
```diff
@@ -159,25 +159,25 @@ def run!(verify_result: true, allow_protected_ns: false, prune: true)
         MSG
         @logger.summary.add_paragraph(ColorizedString.new(warning).yellow)
       end
-      ::StatsD.event("Deployment of #{@namespace} succeeded",
+      StatsD.event("Deployment of #{@namespace} succeeded",
         "Successfully deployed all #{@namespace} resources to #{@context}",
         alert_type: "success", tags: statsd_tags << "status:success")
-      ::StatsD.measure('all_resources.duration', StatsD.duration(start), tags: statsd_tags << "status:success")
+      StatsD.measure('all_resources.duration', StatsD.duration(start), tags: statsd_tags << "status:success")
```
KnVerey: The Policial complaint is already addressed in this PR: https://github.com/Shopify/kubernetes-deploy/pull/386. That's going to be a conflict, so I'd say let's just make those changes in this PR and close that one.

lib/kubernetes-deploy/statsd.rb
```diff
@@ -4,23 +4,41 @@
...
+
     def self.duration(start_time)
       (Time.now.utc - start_time).round(1)
     end
 
     def self.build
-      ::StatsD.default_sample_rate = 1.0
-      ::StatsD.prefix = "KubernetesDeploy"
+      self.default_sample_rate = 1.0
+      self.prefix = "KubernetesDeploy"
```
KnVerey: Use the constant 🙂

lib/kubernetes-deploy/statsd.rb
```diff
@@ -4,23 +4,41 @@
...
+        self.backend = ::StatsD::Instrument::Backends::UDPBackend.new(ENV['STATSD_ADDR'], statsd_impl)
       else
-        ::StatsD.backend = ::StatsD::Instrument::Backends::NullBackend.new
+        self.backend = ::StatsD::Instrument::Backends::NullBackend.new
       end
-      ::StatsD.backend
+    end
+
+    def self.measure(key, value = nil, *metric_options, &block)
+      metric_options.first[:prefix] = PREFIX if metric_options.first.is_a?(Hash)
```
KnVerey: I see you're copying the signature from the superclass, but I'm confused about why they're collecting the options as an array rather than a hash. It doesn't reflect what we actually expect, so it makes it much more difficult to deal with the possibilities. Can't we make the signature `(key, value=nil, **metric_options, &block)` instead? If that isn't possible for some reason, we'd need to check `value` as well because it might actually contain the options.

lib/kubernetes-deploy/statsd.rb
```diff
@@ -4,23 +4,41 @@
 
 module KubernetesDeploy
   class StatsD
+    extend ::StatsD
+
+    PREFIX = "KubernetesDeploy"
+
     def self.duration(start_time)
```
KnVerey: Isn't `duration` an instrumentation method in the superclass? I think we need to rename this.

test/unit/kubernetes-deploy/statsd_test.rb
```diff
@@ -40,14 +41,23 @@ def test_build_when_statsd_addr_env_present_but_statsd_implementation_is_not
...
     ENV['STATSD_ADDR'] = original_addr
     ENV['STATSD_IMPLEMENTATION'] = original_impl
     ENV['STATSD_DEV'] = original_dev
     KubernetesDeploy::StatsD.build
   end
 
+  def test_kubernetes_statsd_does_not_override_global_config
+    KubernetesDeploy::StatsD.build
+    ::StatsD.prefix = "test"
+    ::StatsD.default_sample_rate = 2.0
```
KnVerey: Even though it's not that likely other tests use the `::StatsD` class now, we should reset these in an `ensure` since they are global state.

test/unit/kubernetes-deploy/statsd_test.rb
```diff
@@ -40,14 +41,23 @@ def test_build_when_statsd_addr_env_present_but_statsd_implementation_is_not
...
+    assert_equal(:datadog, KubernetesDeploy::StatsD.backend.implementation)
   ensure
     ENV['STATSD_ADDR'] = original_addr
     ENV['STATSD_IMPLEMENTATION'] = original_impl
     ENV['STATSD_DEV'] = original_dev
     KubernetesDeploy::StatsD.build
   end
 
+  def test_kubernetes_statsd_does_not_override_global_config
+    KubernetesDeploy::StatsD.build
```
KnVerey: Nit: The test would offer stronger proof that we don't overwrite the global config if this line were after L54-55 instead of before.

lib/kubernetes-deploy/deploy_task.rb
timothysmith0609: 🤦‍♂️

lib/kubernetes-deploy/statsd.rb
timothysmith0609: 🤦‍♂️

lib/kubernetes-deploy/statsd.rb
timothysmith0609: Thankfully **args works here and forwards the arguments nicely on the call to super

lib/kubernetes-deploy/statsd.rb
timothysmith0609: It appears that `duration` is indeed a private method on `StatsD::Instrument`. Since it's a private method and statically scoped to the statsd-instrument package, I don't foresee this being a problem.

lib/kubernetes-deploy/statsd.rb
```diff
@@ -4,23 +4,40 @@
...
       (Time.now.utc - start_time).round(1)
     end
 
     def self.build
-      ::StatsD.default_sample_rate = 1.0
-      ::StatsD.prefix = "KubernetesDeploy"
```
timothysmith0609: There's no point in setting `StatsD.prefix` (_not_ `::StatsD.prefix`) here, since `::StatsD.prefix` is what's used inside the library (hence the need to add the `:prefix` option in the first place). I think leaving it here would be misleading so I'm opting to just be rid of it

lib/kubernetes-deploy/statsd.rb
```diff
@@ -4,23 +4,40 @@
...
+      self.default_sample_rate = 1.0
```
dturn: Isn't this the default behavior?

lib/kubernetes-deploy/statsd.rb
dturn: Can you leave this comment above line 28. Since its important but not obvious

lib/kubernetes-deploy/statsd.rb
timothysmith0609: Actually this suffers from the same problem as `StatsD.prefix` in that it's statically referenced in the underlying library so won't have an effect here (`options[:sample_rate] is used to override if ever we need). I'll remove this line as well.

timothysmith0609:
>Looks good, the dashboards will need to be updated before we deploy this.
Perhaps it's my unfamiliarity with DataDog, but what exactly do we need to change, exactly?

PR #385, opened 2018-11-26
Replace SyncMediator with ResourceCache

KnVerey:
Based on https://github.com/Shopify/kubernetes-deploy/pull/381

Replaces the `SyncMediator` abstraction with a new `ResourceCache` that I think makes a lot more intuitive sense, is less complicated, and will be more efficient overall (in terms of requests made and memory usage).


lib/kubernetes-deploy/container_logs.rb
```diff
@@ -39,17 +41,21 @@ def print_all
...
-      end
+               "--since-time=#{rfc3339_timestamp(@last_timestamp)}"
+             else
+               "--tail=#{DEFAULT_LINE_LIMIT}"
+             end
       out, _err, _st = kubectl.run(*cmd, log_failure: false)
       out.split("\n")
     end
 
+    def kubectl
```
KnVerey: It isn't the greatest to have to give this class permanent `kubectl` access, but I think it's acceptable (unlike doing so in `KubernetesResource` for instance).

lib/kubernetes-deploy/kubernetes_resource/service.rb
```diff
@@ -3,11 +3,15 @@ module KubernetesDeploy
   class Service < KubernetesResource
     TIMEOUT = 7.minutes
 
-    SYNC_DEPENDENCIES = %w(Pod Deployment)
-    def sync(mediator)
+    def sync(cache)
       super
-      @related_deployments = selector.present? ? mediator.get_all(Deployment.kind, selector) : []
-      @related_pods = selector.present? ? mediator.get_all(Pod.kind, selector) : []
+      if exists? && selector.present?
```
KnVerey: This is an unrelated optimization: don't bother grabbing the deployments and pods if we're not going to use them.

lib/kubernetes-deploy/resource_watcher.rb
```diff
@@ -118,9 +127,12 @@ def record_failed_statuses(failed_resources, global_timeouts = [])
             "failed to #{@operation_name} #{failures.length} #{'resource'.pluralize(failures.length)}"
           )
         end
+
+        kubectl = Kubectl.new(namespace: @namespace, context: @context, logger: @logger, log_failure_by_default: false)
```
KnVerey: This isn't pretty either, but I think I'm ok with it.

lib/kubernetes-deploy/resource_cache.rb
```diff
@@ -0,0 +1,66 @@
...
+    end
+
+    private
+
+    def statsd_tags
+      { namespace: @namespace, context: @context }
+    end
+
+    def use_or_populate_cache(kind)
+      @mutexes[kind].synchronize do
```
dturn: Is there a race condition in creating the mutex?

lib/kubernetes-deploy/resource_cache.rb
```diff
@@ -0,0 +1,66 @@
...
+        return @data.fetch(kind, {}) if @data.key?(kind)
+        ::StatsD.increment("sync.cache_miss", tags: statsd_tags.merge(type: kind))
```
dturn: I think we should limit the code in the mutex to what has to be protected.

test/unit/kubernetes-deploy/kubernetes_resource/daemon_set_test.rb
```diff
@@ -50,20 +52,9 @@ def build_ds_template(status: {})
...
-    sync_mediator.kubectl.expects(:run)
-      .with("get", "DaemonSet", "ds-app", "-a", "--output=json", raise_if_not_found: true)
-      .returns([template.to_json, "", SystemExit.new(0)])
-
-    sync_mediator.kubectl.expects(:run).with("get", "Pod", "-a", "--output=json", anything).returns(
-      ['{ "items": [] }', "", SystemExit.new(0)]
-    )
-
-    ds.sync(sync_mediator)
+    stub_kind_get("DaemonSet", items: [template])
```
dturn: This is a lot nicer to read 👍

lib/kubernetes-deploy/deploy_task.rb
```diff
@@ -86,7 +86,7 @@ def prune_whitelist
         policy/v1beta1/PodDisruptionBudget
         batch/v1beta1/CronJob
       )
-      wl + cluster_resource_discoverer.crds(@sync_mediator).select(&:prunable?).map(&:group_version_kind)
+      wl + cluster_resource_discoverer.crds.select(&:prunable?).map(&:group_version_kind)
```
dturn: I can't put my finger on why, but it seems odd to me that ResourceDiscovery has its own cache. Do you have a guiding light for what should share a cache and what shouldn't?

lib/kubernetes-deploy/resource_cache.rb
KnVerey: Hm, yeah, probably a really really tiny one? Not sure how I'd write tests for that, or whether it's likely enough to warrant handling with another mutex. 😕

lib/kubernetes-deploy/resource_cache.rb
KnVerey: I need to protect the line before _and_ the line after (which is the return value) together though 🤔

lib/kubernetes-deploy/resource_cache.rb
dturn: I actually think its worth protecting against it. But agree its probably too hard to capture with a test.

lib/kubernetes-deploy/deploy_task.rb
KnVerey: Everything involved in a sync cycle of some kind (so checking initial resource statuses or the watcher) should share a cache. ResourceDiscovery's cache is valid for the lifetime of the instance (and of the deploy really) and there's nothing else that shares this lifecycle. It doesn't really need to use a ResourceCache at all. Maybe it would be better to give it a `Kubeclient`.

lib/kubernetes-deploy/kubernetes_resource/cloudsql.rb
```diff
@@ -3,11 +3,10 @@ module KubernetesDeploy
   class Cloudsql < KubernetesResource
     TIMEOUT = 10.minutes
 
-    SYNC_DEPENDENCIES = %w(Deployment Service)
```
fw42: all those `SYNC_DEPENDENCIES` constants weren't used?

lib/kubernetes-deploy/resource_cache.rb
```diff
@@ -0,0 +1,66 @@
+# frozen_string_literal: true
+module KubernetesDeploy
+  class ResourceCache
+    def initialize(namespace, context, logger)
+      @namespace = namespace
+      @context = context
+      @logger = logger
+
+      @mutexes = Hash.new { |hash, key| hash[key] = Mutex.new }
```
fw42: I don't understand what those mutexes are for. Can you explain?

lib/kubernetes-deploy/resource_cache.rb
```diff
@@ -0,0 +1,66 @@
...
+    def statsd_tags
+      { namespace: @namespace, context: @context }
+    end
+
+    def use_or_populate_cache(kind)
+      @mutexes[kind].synchronize do
+        return @data.fetch(kind, {}) if @data.key?(kind)
+        ::StatsD.increment("sync.cache_miss", tags: statsd_tags.merge(type: kind))
+        @logger.debug("Cache miss for kind #{kind}.")
+        @data[kind] = fetch_by_kind(kind)
```
fw42: I'm assuming that multiple threads can trigger a call to `use_or_populate_cache` in parallel? If so, then this line here is problematic since two calls could modify the `@data` hash in parallel here if they are using different `kind` values (and hashes in Ruby aren't thread-safe).

lib/kubernetes-deploy/resource_cache.rb
fw42: Yea, Danny is right I think. There's actually two potential problems here as far as I can tell:
* Two threads with different `kind` values might access the `@mutexes` hash in parallel and hash access in Ruby isn't thread-safe.
* Two threads with the same `kind` values might race and both trigger the creation of a new mutex and one thread's mutex can overwrite the other thread's mutex.

Would it make sense to have one ResourceCache instance per kind? Rather than one instance that's shared by all kinds. That would simplify this mutex stuff a lot I think.

lib/kubernetes-deploy/container_logs.rb
```diff
@@ -50,6 +52,10 @@ def fetch_latest(kubectl)
...
+      @kubectl ||= Kubectl.new(namespace: @namespace, context: @context, logger: @logger, log_failure_by_default: false)
```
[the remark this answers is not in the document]
fw42: `sync` can be called from multiple threads at once, right? and `sync` calls `fetch_latest` which calls `kubectl`, which writes an instance variable here. Probably totally harmless, but I think in theory there's a race condition here where two threads both run `kubectl` while `@kubectl` is still `nil` and then the second one will overwrite the `@kubectl` that the first one created.

lib/kubernetes-deploy/resource_cache.rb
KnVerey: > Would it make sense to have one ResourceCache instance per kind? Rather than one instance that's shared by all kinds. That would simplify this mutex stuff a lot I think.

I'm having trouble imagining how that would work. The list of kinds changes dynamically on every polling cycle, so we'd still need a higher-level thing generating/managing the mutexes. Even the list of possible kinds can't be anticipated, because we don't explicitly support all of the possibilities (and custom resources are a thing, though we aren't handling them dynamically _yet_).

lib/kubernetes-deploy/container_logs.rb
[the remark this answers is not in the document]
KnVerey: > just curious: why do we not want to log failures by default?

Because logs are a nicety as far as this gem is concerned, and they can be unavailable for reasons other than kubectl commands failing. Regardless of why we don't have them, we just print `None found. Please check your usual logging service (e.g. Splunk).`.  

There are many other cases where we also don't log kubectl errors by default, because we assume they are transient and retry on the next polling cycle. You can use `DEBUG=1` to turn on error logging regardless of which setting is used bye the `Kubectl` instance.

lib/kubernetes-deploy/resource_cache.rb
```diff
@@ -0,0 +1,66 @@
...
+      instances = {}
+      JSON.parse(raw_json)["items"].each do |resource|
+        resource_name = resource.dig("metadata", "name")
+        instances[resource_name] = resource
+      end
+      instances
+    end
+
+    def kubectl
+      @kubectl ||= Kubectl.new(namespace: @namespace, context: @context, logger: @logger, log_failure_by_default: false)
```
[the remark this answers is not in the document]
KnVerey: You're right, but this one definitely doesn't matter, so I'm going to ignore it.

lib/kubernetes-deploy/resource_cache.rb
KnVerey: Actually I think we should remove both the metric and the log line. Previously, we had a code path that was supposed to make these calls and a code path that could make them as a fallback, so the latter were clearly cash misses. Now, unless something is broken, these are all cache fills.

lib/kubernetes-deploy/kubernetes_resource/cloudsql.rb
KnVerey: They were used in the deleted file, which Github is hiding in the diff by default: https://github.com/Shopify/kubernetes-deploy/pull/385/files#diff-3dfdf53f121528a357792ba9ddc1825eL47

KnVerey:
Updated to use `Concurrent::Hash` since both of the outstanding concurrency issues were hash-access related. Note that on MRI this is still using `::Hash` anyway though: 

https://github.com/ruby-concurrency/concurrent-ruby/blob/master/lib/concurrent/hash.rb#L16-L20

And that's ok for MRI because of the GIL: https://www.jstorimer.com/pages/ruby-core-classes-arent-thread-safe

I also moved `@kubectl` setting to the initializer for good measure, even though it doesn't actually matter if that race happens.

Note that if `Concurrent::Hash` has a bug resulting in the race condition getting hit on another ruby implementation, the worse case is still just that we'll make extra requests. That was already the worst case before this PR (i.e. via the cache population failing for a given kind) but now it is way less likely.

lib/kubernetes-deploy/resource_cache.rb
```diff
@@ -0,0 +1,64 @@
...
+require 'concurrent/hash'
+
+module KubernetesDeploy
+  class ResourceCache
+    def initialize(namespace, context, logger)
+      @namespace = namespace
+      @context = context
+      @logger = logger
+
+      @kind_fetcher_locks = Concurrent::Hash.new { |hash, key| hash[key] = Mutex.new }
```
fw42: Hmmm. I'm not sure this is now fine. There were two race conditions before: Two threads accessing different keys of the hash in parallel. And the other one was about two threads racing to evaluate the "cache miss" block here. I think the `Concurrent::Hash` only protects against the former, but not the latter.

KnVerey: I'm not sure for non-MRI, and I'm not sure how to find out either. For MRI I think the GIL should still have us covered, and earlier today I ran my only-makes-one-call test with 500 resources of 3 different kind over 5,000 times in a row and it passed every time. That doesn't 100% prove anything, but it seems more than good enough for sure.

KnVerey:
I rebased my watcher branch on this and watched a core deploy, and this branch currently performs _worse_ than production because the mutex creation is not working as expected. It turns out the block is only evaluated once, so all kinds are using the same mutex.

I also noticed that kubectl get pods still takes way longer than 15s to return, gradually increasing in memory over time. That reminds me that kubectl paginates by default as of a a few versions ago. I want to look into whether I can add a parameter to increase its batch size.

KnVerey:
Ignore my last comment -- it is wrong. What I actually saw was only one kubectl command ever being run at a time, but that's because the deploy I was watching only had one type of resource in it by that point: Deployment, which requests Deployment, then ReplicaSet then Pod in that order. We can try to be smarter about that in the future, but those operations aren't parallelized on master either. I just ran the watcher again during a deploy and the sync performance I saw was similar to the averages we're seeing for core today.

My other point about pagination is correct though. The default batch size seems to be 100, and it was done to improve perceived latency when used interactively, which clearly isn't helping us. I tried running some simple benchmarks, and saw that batch size 500 actually makes things a little worse on average but removing the limit altogether makes things a little better (5-15%). The `--chunk-size` option has been available since 1.9, so we can use it--I'll make another PR for that and do a watch from it to see if it makes a difference in practice.. 

What's even more striking though in my testing is the performance difference between `get pods` and `get pods --output=json`. Is `kubectl` requesting a different format and being really slow at turning it into json? TBD.

dturn:
> What's even more striking though in my testing is the performance difference between get pods and get pods --output=json. Is kubectl requesting a different format and being really slow at turning it into json? TBD.

Have we investigated limiting the response with something like `--output=jsonpath` or do we feel like we want / need all of it?

fw42:
> Is kubectl requesting a different format and being really slow at turning it into json? TBD.

The JSON output contains a lot more information than just the regular text output, doesn't it? For example, the text output for `get pods` in `shopify-core` is only 150 KB, but the JSON output is 44 MB (300 times as large). Maybe for the server to find and the client to download all that extra information is what takes extra time? Can we somehow disable the information that we don't need? Maybe using the `--field-selector` option or the `-o=jsonpath` option or something like that? Is there a way to request only the pieces of the JSON that we actually care about?

PR #386, opened 2018-11-27
Replace measure with distribution

dturn:
This is a follow-up #374 to use distribution over measure. See https://github.com/Shopify/statsd-instrument#statsddistribution for more details. But at a high level distribution has all arithmetic and calculation happening server side which is more efficient, correct, and cost effective.

Note: this replaces #383 rather than deal with rebase conflicts


dturn:
wip dash https://github.com/Shopify/dog/pull/5631

timothysmith0609:
This branch has been merged into #384 so we should be able to close this PR now.

PR #390, opened 2018-11-28
Use Oj instead of JSON to speed up json parsing

dturn:
Speed up large deploys.

**How is this accomplished?**
Use the [OJ gem](https://github.com/ohler55/oj) which is almost a drop in replacement for the JSON module but 2-3x faster. We get large json blobs back from k8s, so parsing them faster should be an easy win.

**What could go wrong?**
OJ isn't an exact replacement, we could hit a corner case.

### DIFF
--- a/kubernetes-deploy.gemspec
+++ b/kubernetes-deploy.gemspec
@@ -29,6 +29,7 @@ Gem::Specification.new do |spec|
   spec.add_dependency "ejson", "~> 1.0"
   spec.add_dependency "colorize", "~> 0.8"
   spec.add_dependency "statsd-instrument", "~> 2.3"
+  spec.add_dependency "oj", "~> 3.7"
 
   spec.add_development_dependency "bundler"
   spec.add_development_dependency "rake", "~> 10.0"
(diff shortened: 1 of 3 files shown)


timothysmith0609:
Have we done any comparison testing to validate the hypothesis that this indeed speeds up large deploys?

dturn:
So this is probably going to save us seconds not minutes :( 

```
require 'json'
require 'benchmark'

json = nil
n = 10

Benchmark.bm do |x|
  x.report("kubectl get pods: ") { json = `kubectl --context tier1-us-east-4 -n shopify-core get pods -o json` }
  x.report("JSON: ") { n.times { JSON.parse(json) } }
  require 'oj'
  Oj.mimic_JSON
  x.report("OJ: ") { n.times { JSON.parse(json) } }
end
```

```
       user     system      total        real
kubectl get pods:   0.120000   0.100000  23.230000 ( 22.338402)
JSON:   6.890000   0.170000   7.060000 (  7.082341)
OJ:   4.640000   0.030000   4.670000 (  4.682790)
```

(Note that the json & oj lines need are for 10 runs, not 1)

KnVerey:
Given that we re-acquire and parse the JSON on every polling cycle, seconds is still worthwhile IMO.

PR #393, opened 2018-11-30
Disable rule exemptions and autocorrect

KnVerey:
Follow the Shopify styleguide more closely. I recently added a new exception for the `MethodCallWithArgsParentheses` because it first came into effect on an already-approved PR. We could commit the rubocop cache file to prevent that in the future, but then we'd need to periodically delete that file to pick up updates.

**How is this accomplished?** 
Disabling rule exemptions and running `rubocop -a`.

**What could go wrong?**
The new parentheses-heavy style recently adopted could annoy external contributors?

@Shopify/cloudx

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -4,15 +4,6 @@ inherit_from:
 AllCops:
   TargetRubyVersion: 2.3
 
-Style/TrailingCommaInArrayLiteral:
-  Enabled: false
-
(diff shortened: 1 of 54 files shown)


KnVerey:
@dturn I redid this branch from latest master rather than rebasing. I'll merge it assuming CI passes.

PR #398, opened 2018-12-07
Do not fetch pods when we do not need them

KnVerey:
On large clusters, `get pods` is really expensive. Try to do it less often.

**How is this accomplished?**
The main time we have a problem with this is while watching a large set of huge Deployment objects that are steadily rolling out. The reason we fetch pods in that watch is for eager failure detection, i.e. `pods.all?(&:deploy_failed?)` in `ReplicaSet`. The thing is, once the rollout has progressed a little, there's no real possibility of that condition becoming true. This PR check ReplicaSet's status, and if it shows that some pods are ready, we infer that `pods.all?(&:deploy_failed?)` would not be true if we fetched the pods, and bypass that fetch.

The same cannot be done for DS or SS because they use a `.any?` condition.

**What could go wrong?**
This is added complexity that could cause trouble in the future, e.g. if ReplicaSet starts doing something new with `@pods` that expects it to always be populated.

### DIFF
--- a/lib/kubernetes-deploy/kubernetes_resource/replica_set.rb
+++ b/lib/kubernetes-deploy/kubernetes_resource/replica_set.rb
@@ -60,8 +60,10 @@ module KubernetesDeploy
     def check_pods?
       # We're only using pods in deploy_failed? to check that they aren't ALL bad,
       # so if we can already tell that from the RS data, don't bother looking at (or expensively fetching) them
-      return false if !exists? || stale_status?
-      ready_replicas < [2, desired_replicas].min
+      return false unless exists?
+      return false if stale_status?
+      return false if desired_replicas == 0
+      ready_replicas < 2
     end
 
     def rollout_data
--- a/lib/kubernetes-deploy/kubernetes_resource/replica_set.rb
+++ b/lib/kubernetes-deploy/kubernetes_resource/replica_set.rb
@@ -16,7 +16,7 @@ module KubernetesDeploy
 
     def sync(cache)
       super
-      @pods = exists? ? find_pods(cache) : []
+      @pods = check_pods? ? find_pods(cache) : []
     end
 
     def status
@@ -25,7 +25,7 @@ module KubernetesDeploy
     end
 
     def deploy_succeeded?
-      observed_generation == current_generation &&
+      return false if stale_status?
       desired_replicas == rollout_data["availableReplicas"].to_i &&
       desired_replicas == rollout_data["readyReplicas"].to_i
     end
@@ -33,7 +33,7 @@ module KubernetesDeploy
     def deploy_failed?
       pods.present? &&
       pods.all?(&:deploy_failed?) &&
-      observed_generation == current_generation
+      !stale_status?
     end
 
     def desired_replicas
@@ -52,6 +52,17 @@ module KubernetesDeploy
     end
 
     private
+
+    def stale_status?
+      observed_generation != current_generation
+    end
+
+    def check_pods?
+      # We're only using pods in deploy_failed? to check that they aren't ALL bad,
+      # so if we can already tell that from the RS data, don't bother looking at (or expensively fetching) them
+      return false if !exists? || stale_status?
+      ready_replicas < [2, desired_replicas].min
+    end
 
     def rollout_data
       return { "replicas" => 0 } unless exists?
--- a/lib/kubernetes-deploy/kubernetes_resource/replica_set.rb
+++ b/lib/kubernetes-deploy/kubernetes_resource/replica_set.rb
@@ -16,7 +16,7 @@ module KubernetesDeploy
 
     def sync(cache)
       super
-      @pods = check_pods? ? find_pods(cache) : []
+      @pods = fetch_pods_if_needed(cache) || []
     end
 
     def status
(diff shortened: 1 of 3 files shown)


lib/kubernetes-deploy/kubernetes_resource/replica_set.rb
```diff
@@ -53,6 +53,17 @@ def available_replicas
 
     private
 
+    def stale_status?
+      observed_generation != current_generation
+    end
+
+    def check_pods?
```
dturn: `check_pods_for_fast_failure?`

test/unit/kubernetes-deploy/kubernetes_resource/deployment_test.rb
```diff
@@ -366,12 +367,13 @@ def build_rs_template(status: { 'replicas' => 3 })
     result
   end
 
-  def build_synced_deployment(template:, replica_sets:)
+  def build_synced_deployment(template:, replica_sets:, expect_pod_get: nil)
```
dturn: why `nil` seems like `true` is a better fit here

lib/kubernetes-deploy/kubernetes_resource/replica_set.rb
```diff
@@ -53,6 +53,17 @@ def available_replicas
...
+      # We're only using pods in deploy_failed? to check that they aren't ALL bad,
+      # so if we can already tell that from the RS data, don't bother looking at (or expensively fetching) them
+      return false if !exists? || stale_status?
+      ready_replicas < [2, desired_replicas].min
```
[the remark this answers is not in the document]
KnVerey: > because we only need to find a single one that didn't fail in order to know that not all of them failed

Technically true, but workloads that do not have proper health checks (e.g. a lot of apps' jobs deployments) will very briefly appear to come up as healthy even if they crash very soon afterwards. Requiring us to see two healthy rather than one before making this assumption is intended as a small safeguard against that scenario. Since the bypass this PR adds is most useful at the tail end of large rollouts, we could choose an even higher number and still get the benefit when it matters most. But the sooner we stop needing pods for _any_ deployment in the set, the more effective it will be overall.

> Why do we care about desired_replicas here?

Because deployments can have less than 2 replicas. If that happens, `ready_replicas < 2` will always be true even when the deployment is completely rolled out, and we'll be fetching the pods unconditionally. We definitely don't want to do that for zero-replica deployments, but maybe this would be more sensible:

```ruby
return false unless exists?
return false if stale_status?
return false if desired_replicas == 0
ready_replicas > 1
```

i.e. if the deployment only has one replica, that replica is pretty important and worth checking unconditionally.

> Also, is a test going to fail if I change this to just ready_replicas < 2

The `test_sync_does_not_request_pods_if_desired_replicas_is_zero` was supposed to fail if you ignore `desired_replicas`, but I forgot to set the generation in the status, so it wasn't. Will push a fix.

>  (or even < 1)?

Both new tests fail if you do this (and one still fails if you just change the 2 to a 1 without ignoring desired replicas).

lib/kubernetes-deploy/kubernetes_resource/replica_set.rb
```diff
@@ -16,7 +16,7 @@ def initialize(namespace:, context:, definition:, logger:, statsd_tags: nil,
 
     def sync(cache)
       super
-      @pods = exists? ? find_pods(cache) : []
+      @pods = check_pods? ? find_pods(cache) : []
```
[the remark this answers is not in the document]
KnVerey: This is called repeatedly until the resource is "done" (succeeded/failed/timed out), and the data retrieved on the previous cycle is no longer valid in subsequent cycles. For example, maybe on the first cycle we see 2 exist / 0 available, so we fetch them, and the look fine, so we continue. On the next cycle, we see 4 exist / 2 available, so we now don't bother fetching them. If we kept the stale data, anything that didn't respect `check_pods?` and actually used the data would draw incorrect conclusions. `@pods == []` already has to be handled as "dunno, try again later" in case of fetch errors, but the possibility that data is present but stale does not.

KnVerey:
> I'm wondering now if this filtering ("get all pods that failed") could be done server-side (by sending an API request that tells the server only to return those) rather than client-side (fetch ALL pods and then filter in Ruby)? 

Unfortunately not. You can filter on some status fields like phase (and that _is_ done server-side), but nothing granular enough to detect all the failure scenarios we want to (see the private `Container` class in `Pod.rb`). Another problem is that we'd lose the ability to safely centralize the pods cache if we do something more clever. Right now, the worst case is that we request all pods once, no matter how many different pod manager types (deployment, job, stateful set...) we've got in our set. If we stopped requesting all the data, that sharing across pod manager classes would no longer be safe.

> Is pods.any?(&:deploy_failed?) equivalent to something like pods.all?(&:deploy_succeeded?)? Does that help (since it would allow you to turn the any? into a all?)?

No, it is possible for pods to neither have succeeded nor have failed (they may be starting or stalled for example). If that condition persists until the end of the deploy, you get a timeout rather than a failure.

lib/kubernetes-deploy/kubernetes_resource/replica_set.rb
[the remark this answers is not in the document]
dturn: >  will very briefly appear to come up as healthy even if they crash very soon afterwards. Requiring us to see two healthy rather than one before making this assumption is intended as a small safeguard against that scenario.

My concern with this PR was `2` seemed like a magic number. Given the above did you consider also making this time based? We've done something like that in `Job#running?`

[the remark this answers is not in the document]
KnVerey: > Given the above did you consider also making this time based? 

I don't think we have a relevant timestamp to use in this case. ReplicaSet supports status conditions, but it doesn't have any that are always (or even usually) present. The only timestamp I'm seeing is the creation time, which isn't useful because RS can be reused (e.g. during rollbacks).

test/unit/kubernetes-deploy/kubernetes_resource/deployment_test.rb
KnVerey: The real default value I want is `replica_sets.present?`, and for this to be an override. That's not possible if the override's default value is `true`. The alternative is to default it to true, remove L375, and make all the callers do `expect_pod_get: replica_sets.present?` themselves.

KnVerey:
@fw42 did my answers make sense to you? Are you ok with this PR?

lib/kubernetes-deploy/kubernetes_resource/replica_set.rb
fw42: I still feel like this method is a bit "dense" with context and ideas that I will immediately forget and then look at this again in 2 weeks and have no idea what each of those lines means. Do you think that's only because I'm not familiar enough with this project or do you think others might have the same problem? If so, maybe clarifying with comments a bit might help?
```ruby
def check_pods?
  # No point checking for pods if the replica set itself doesn't exist
  return false unless exists?

  # dunno I don't really understand this. I guess don't check pods if the replica set we have is already stale?
  return false if stale_status?

  # if we don't care about how many pods succeed, we don't need to look for them
  return false if desired_replicas == 0

  # 2 instead of 1 because of some magical safe-guard for some edge case. imho worth explaining so that
  # the context isn't lost
  ready_replicas < 2
end
```

KnVerey: I think anyone might be confused. I added more comments and changed the method to `fetch_pods_if_necessary` which I think makes this easier to follow overall.

PR #400, opened 2018-12-12
Add CustomResourceDefinition to predeploy sequence

zrdaley:
Apps such as cert-manager-private use custom resources. Kubernetes-deploy should deploy the custom resource definitions as early as possible.

### DIFF
--- a/lib/kubernetes-deploy/deploy_task.rb
+++ b/lib/kubernetes-deploy/deploy_task.rb
@@ -46,6 +46,7 @@ module KubernetesDeploy
     extend KubernetesDeploy::StatsD::MeasureMethods
 
     PREDEPLOY_SEQUENCE = %w(
+      CustomResourceDefinition
       ResourceQuota
       Cloudsql
       Redis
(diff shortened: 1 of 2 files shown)


KnVerey:
Hey @zrdaley, kubernetes-deploy is meant to manage a namespace, so although we have in fact modelled the CRD class out of internal necessity, we don't officially support deploying non-namespaced (aka global) resources of any kind. For CRDs, possible workarounds include creating them on their own as an earlier step in the deploy pipeline, or having your controllers themselves register the CRDs for the resources they manage (that's what we generally prefer for our internal controllers). The issue with the change in this PR is that it implies you could deploy CRDs and corresponding CRs in the same batch, which is not going to work with our generic CR implementation (and I currently don't think we should prioritize making it work, because we don't support deploying global resources). I'm happy to discuss this more in Slack if you disagree. 🙂 

cc @dturn @timothysmith0609

zrdaley:
That makes sense. Thank you for clearing that up @KnVerey :) I'll close this PR and use a workaround for cert-manager!

PR #402, opened 2018-12-14
Bump statsd-instrument dep for prefix

KnVerey:
* Fix our statsd-instrument dependency. We need 2.3.2 or higher now because we're relying on the prefix. I noticed this because my tests failed locally (prefix missing). 
* Fix the `sync.duration` metric to be tagged with the namespace, context and sha. I noticed this was missing by deploying my test app using kubernetes-deploy metrics to test our new distribution-based dash.

**How is this accomplished?**
Pretty straightforward. 

**What could go wrong?**
Not much.


test/integration-serial/serial_deploy_test.rb
```diff
@@ -209,15 +209,11 @@ def test_stage_related_metrics_include_custom_tags_from_namespace
 
   def test_all_expected_statsd_metrics_emitted_with_essential_tags
     metrics = capture_statsd_calls do
-      result = deploy_fixtures('hello-cloud', subset: ['configmap-data.yml'], wait: false)
+      result = deploy_fixtures('hello-cloud', subset: ['configmap-data.yml'], wait: false, sha: 'test-sha')
       assert_deploy_success(result)
     end
 
     assert_equal 1, metrics.count { |m| m.type == :_e }, "Expected to find one event metric"
-    assert_predicate metrics, :all? do |metric|
```
KnVerey: This isn't doing what it appears to be doing. If it were, it would have caught the missing tags. Better to check on the explicit list of metrics below anyway.

kubernetes-deploy.gemspec
```diff
@@ -28,7 +28,7 @@ Gem::Specification.new do |spec|
   spec.add_dependency "googleauth", "~> 0.6.6" # https://github.com/google/google-auth-library-ruby/issues/153
   spec.add_dependency "ejson", "~> 1.0"
   spec.add_dependency "colorize", "~> 0.8"
-  spec.add_dependency "statsd-instrument", "~> 2.3"
+  spec.add_dependency "statsd-instrument", ">= 2.3.2"
```
timothysmith0609: This will automatically install a version `3.x` gem if it becomes available (or even `2.4.x`), no? `~> 2.3.2` might be safer in that respect.

KnVerey: Yes, you're right. I think what we really want is `'~> 2.3', '>= 2.3.2'` (i.e. same flexibility as before on minor versions, but mininum 2.3.2)

PR #404, opened 2018-12-14
Fix formatting in renderer errors

KnVerey:
Fix minor formatting issues from external contribution to expedite its release.

### DIFF
--- a/lib/kubernetes-deploy/formatted_logger.rb
+++ b/lib/kubernetes-deploy/formatted_logger.rb
@@ -7,7 +7,7 @@ module KubernetesDeploy
     include DeferredSummaryLogging
 
     def self.indent_four(str)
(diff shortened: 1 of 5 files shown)


lib/kubernetes-deploy/formatted_logger.rb
```diff
@@ -6,6 +6,10 @@ module KubernetesDeploy
   class FormattedLogger < Logger
     include DeferredSummaryLogging
 
+    def self.indent_four(str)
+      "    " + str.gsub("\n", "\n    ")
```
dturn: Can we fix the indet_four bug. e.g. `str&.gsub` here?

KnVerey:
Example showing that the newline is fixed and the indentation now matches that of DeployTask:

image

Issue #373, opened 2018-10-31
HPA setup flakey on k8s-ci clusters


KnVerey:
This seems relatively common, though it is being hidden by automatic retries (which I'd prefer to disable). This will only get worse now that we have two versions (1.11 and 1.12) using this queue.

Example build where 1.11 fails: https://buildkite.com/shopify/kubernetes-deploy-gem/builds/1098#5d2a9ff0-5181-4816-a7b8-1e4ff6781bac
Example build where 1.12 fails: https://buildkite.com/shopify/kubernetes-deploy-gem/builds/1097#ed4ac1f1-2046-4236-9108-e13689038b3b

@Shopify/cloudx

KnVerey:
We're doing a blind `create` of the metric server resources, so one good thing to check would be whether there's something we could look at to make sure that setup was successful (and retry it).

randy-abson:
Adding this to Kanbanize

PR #408, opened 2019-01-02
Bump metric server

dturn:
The HPA tests often fail due to the metric server not being avilable
fix: https://github.com/Shopify/kubernetes-deploy/issues/373

**How is this accomplished?**
Bump the metric server version. The original version of this pr 
checked if the service is up and if not slept a few times. However, that didn't actually fix the problem. It appeared the issue was the metric server didn't have the requested metrics, see a comment further down in this pr.

**What could go wrong?**
Its still flakey, though I've run it a few times and hasn't failed yet.

### DIFF
--- a/test/helpers/test_provisioner.rb
+++ b/test/helpers/test_provisioner.rb
@@ -71,9 +71,9 @@ class TestProvisioner
 
       # The service is often slow to start leading to randomly failed HPA tests
       3.times do |round|
-        service = kubectl.run("-n", "kube-system", "get", "APIService", "v1beta1.metrics.k8s.io", "-o", "json")
-        if service.last.success?
-          service = JSON.parse(service.first)
+        service, _, status = kubectl.run("-n", "kube-system", "get", "APIService", "v1beta1.metrics.k8s.io", "-o", "json")
+        if status.success?
+          service = JSON.parse(service)
           available = service.dig("status", "conditions")&.detect { |s| s["type"] == "Available" }
           break if available["status"] == "True"
         end
--- a/test/helpers/test_provisioner.rb
+++ b/test/helpers/test_provisioner.rb
@@ -71,13 +71,14 @@ class TestProvisioner
 
       # The service is often slow to start leading to randomly failed HPA tests
       3.times do |round|
-        service, _, status = kubectl.run("-n", "kube-system", "get", "APIService", "v1beta1.metrics.k8s.io", "-o", "json")
+        service, _, status = kubectl.run("-n", "kube-system", "-o", "json",
+          "get", "APIService", "v1beta1.metrics.k8s.io")
         if status.success?
           service = JSON.parse(service)
           available = service.dig("status", "conditions")&.detect { |s| s["type"] == "Available" }
           break if available["status"] == "True"
         end
-        sleep(2**(round + 1))
+        sleep(3**(round + 1))
       end
     end
   end
--- a/test/helpers/test_provisioner.rb
+++ b/test/helpers/test_provisioner.rb
@@ -65,6 +65,11 @@ class TestProvisioner
         logger: logger, log_failure_by_default: true, default_timeout: '5s')
 
       Dir.glob("test/setup/metrics-server/*.{yml,yaml}*").map do |resource|
+        if kubectl.server_version < Gem::Version.new('1.11.0') && resource =~ /metrics-server-deployment.yaml/
+          next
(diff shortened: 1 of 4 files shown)


test/helpers/test_provisioner.rb
```diff
@@ -68,6 +68,17 @@ def deploy_metric_server
         found = kubectl.run("get", "-f", resource, log_failure: false).last.success?
         kubectl.run("create", "-f", resource) unless found
       end
+
+      # The service is often slow to start leading to randomly failed HPA tests
+      3.times do |round|
+        service = kubectl.run("-n", "kube-system", "get", "APIService", "v1beta1.metrics.k8s.io", "-o", "json")
```
timothysmith0609: Is there a reason not to use the `attempts` parameter for `kubectl.run` here?

dturn: I expect the resource to have been created by line 69 so probably being a bit more defensive than needed, but its being ready that I would normally expect to fail and that attempts wont help with

timothysmith0609: Minor nit but I would prefer if the L-values here would be split into separate variables instead of just one. I had to dig through the `kubectl.run` code to see what `service.first/last` were.

dturn:
This doesn't appear to have actually helped. I'll keep digging

dturn:
```
> k -n kube-system logs metrics-server-6fbfb84cdd-qvxnw
...
I0102 22:26:53.376439       1 reststorage.go:93] No metrics for pod cloud-portal-test/happy-job-7c62x
I0102 22:26:53.376443       1 reststorage.go:140] No metrics for container my-cronjob in pod cloud-portal-test/delete-job-1-tkxq5
I0102 22:26:53.376445       1 reststorage.go:93] No metrics for pod cloud-portal-test/delete-job-1-tkxq5
```

Issue is likely that the metrics server doesn't actually have metrics for the pods https://github.com/kubernetes-incubator/metrics-server/blob/v0.2.1/metrics/storage/nodemetrics/reststorage.go#L112 .I bumped the version since  debugging an old version isn't fun.  But this seems to actually make things worse...

dturn:
Bumping the metric server seems to make the HPA tests reliable on non-minikube CI. I don't love excluding 1.10 (we're in the processes of dropping 1.9 support) but I think this worth the trade-off.

dturn:
🤞 using the old version on minikube ci and the new version everywhere else looks to be more stable

test/helpers/test_provisioner.rb
```diff
@@ -65,6 +65,11 @@ def deploy_metric_server
         logger: logger, log_failure_by_default: true, default_timeout: '5s')
 
       Dir.glob("test/setup/metrics-server/*.{yml,yaml}*").map do |resource|
+        if kubectl.server_version < Gem::Version.new('1.11.0') && resource =~ /metrics-server-deployment.yaml/
+          next
+        elsif kubectl.server_version >= Gem::Version.new('1.11.0') && resource =~ /metrics-server-deployment_021.yaml/
+          next
+        end
         found = kubectl.run("get", "-f", resource, log_failure: false).last.success?
```
KnVerey: not from this PR, but shouldn't we log failure here and below?

test/helpers/test_provisioner.rb
```diff
@@ -65,6 +65,11 @@ def deploy_metric_server
         logger: logger, log_failure_by_default: true, default_timeout: '5s')
 
       Dir.glob("test/setup/metrics-server/*.{yml,yaml}*").map do |resource|
```
KnVerey: not from this PR, but `map` -> `each`

test/setup/metrics-server/metrics-server-deployment.yaml
```diff
@@ -23,10 +23,15 @@ spec:
         k8s-app: metrics-server
     spec:
       serviceAccountName: metrics-server
+      volumes:
+      # mount in tmp so we can safely use from-scratch images and/or read-only containers
```
KnVerey: should we do this in the 0.2.1 version as well?

test/helpers/test_provisioner.rb
dturn: We don't on this line since its just a check to see if the next line needs to run and the resource not being present isn't an error. But seems reasonable to turn it on for line below.

test/setup/metrics-server/metrics-server-deployment.yaml
dturn: This wasn't supported till 0.3 https://github.com/kubernetes-incubator/metrics-server/commit/a823af80d438d642c29e038ca5336004b2a8b97e

PR #409, opened 2019-01-04
add k8s v13, remove v9

jessie-JNing:
add k8s v1.13 to pipeline

**How is this accomplished?**
update .buildkite.yml

**What could go wrong?**
...

### DIFF
--- a/.buildkite/pipeline.nightly.yml
+++ b/.buildkite/pipeline.nightly.yml
@@ -4,6 +4,14 @@ shared: &shared
        - exit_status: "*"
          limit: 1
 steps:
+  - name: 'Run Test Suite (:kubernetes: 1.13-latest)'
+    <<: *shared
+    command: bin/ci
+    agents:
+      queue: k8s-ci
+    env:
+      LOGGING_LEVEL: 4
+      KUBERNETES_VERSION: v1.13-latest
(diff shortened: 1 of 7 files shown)


jessie-JNing:
> I'll close out my version of this https://github.com/Shopify/kubernetes-deploy/pull/405/files. But there is more work besides making CI pass, we'll also need to remove the references to 1.9 in the code and updates the read me. See #341 as an example

Thanks Danny for the example PR. 
I'll clean 1.9 related code after finishing the change in k8s-ci

karanthukral:
we will have to update the [changelog](https://github.com/Shopify/kubernetes-deploy/blob/master/CHANGELOG.md) when we release the update to the gem

karanthukral:
> Looks good. I don't think we need to cut a release at this time as an fyi.

Fair enough, we could have an changelog update with an unreleased note to the latest version

jessie-JNing:
> > Looks good. I don't think we need to cut a release at this time as an fyi.
> 
> Fair enough, we could have an changelog update with an unreleased note to the latest version

sure, will not give a release at the moment. 
I update the changelog though, please check [this commit](https://github.com/Shopify/kubernetes-deploy/pull/409/commits/cf4d587aebfc8616bf30c38a66947833a3a5830a)

PR #413, opened 2019-01-15
prevent invalid const errors from lowercase resource kind

dwradcliffe:
Fix errors that occur if one attempts to deploy a resource that has a kind with a lowercase first letter. This is not the case for any of the standard k8s kinds, but custom resources may use lowercase kinds.

```
NameError: wrong constant name foobar
    lib/kubernetes-deploy/kubernetes_resource.rb:39:in `const_defined?'
    lib/kubernetes-deploy/kubernetes_resource.rb:39:in `build'
```

**How is this accomplished?**
~Capitalize the string before checking for a const. This seemed like the simplest change that wouldn't impact the existing logic.
This would also allow these resource types to have custom resources classes added, in the capitalized form.~

Rescue the NameError and fallback to the generic class. This is safer than changing the string and catches any other edge cases we haven't thought about.

**What could go wrong?**
I can't really think of anything.

### DIFF
--- a/test/unit/kubernetes-deploy/kubernetes_resource_test.rb
+++ b/test/unit/kubernetes-deploy/kubernetes_resource_test.rb
@@ -288,13 +288,13 @@ class KubernetesResourceTest < KubernetesDeploy::TestCase
     refute_predicate(dummy, :disappeared?)
   end
 
-  def test_custom_resource_kind
+  def test_lowercase_custom_resource_kind_does_not_raise
     definition = { "kind" => "foobar", "metadata" => { "name" => "test" } }
     KubernetesDeploy::KubernetesResource.build(
       namespace: 'test',
       context: 'test',
       definition: definition,
-      logger: ::Logger.new($stderr),
+      logger: logger,
       statsd_tags: []
     )
   end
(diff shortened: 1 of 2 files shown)


test/unit/kubernetes-deploy/kubernetes_resource_test.rb
```diff
@@ -288,6 +288,17 @@ def test_disappeared_is_false_if_resource_has_been_deployed_and_we_get_a_server_
     refute_predicate(dummy, :disappeared?)
   end
 
+  def test_custom_resource_kind
```
KnVerey: Can you make this name more descriptive and/or add an `assert_nothing_raised` and/or assert that `build` returned an instance of the expected class? I'm thinking it won't be very obvious in the future what this test is covering / what the bug was.

test/unit/kubernetes-deploy/kubernetes_resource_test.rb
```diff
@@ -288,6 +288,17 @@ def test_disappeared_is_false_if_resource_has_been_deployed_and_we_get_a_server_
...
+    definition = { "kind" => "foobar", "metadata" => { "name" => "test" } }
+    KubernetesDeploy::KubernetesResource.build(
+      namespace: 'test',
+      context: 'test',
+      definition: definition,
+      logger: ::Logger.new($stderr),
```
KnVerey: Not that it makes any difference here, but FYI there's a `logger` accessor all tests can use.

dwradcliffe:
Released in `v0.24.0`.

PR #414, opened 2019-01-15
Support timeout override for deployments

### LINKED ISSUE #315: Deployment should support `timeout-override` annotation
# Feature request

If the resource has the kubernetes-deploy.shopify.io/timeout-override annotation on it, use that. This is currently supported for all resources except deployment.

timothysmith0609:
Closes #315 
Required for #229 

Support timeout override annotation for deployments

**How is this accomplished?**
Have the timeout override annotation override considerations of `progressDeadlineSeconds` and the hard timeout.

**What could go wrong?**
Do we want to use the standard timeout message when using the override, or surface the fact that the timeout being used is in fact from the override annotation?

cc @Shopify/cloudx

### DIFF
--- a/test/integration/kubernetes_deploy_test.rb
+++ b/test/integration/kubernetes_deploy_test.rb
@@ -458,11 +458,7 @@ unknown field \"myKey\" in io.k8s.apimachinery.pkg.apis.meta.v1.ObjectMeta",
     end
     assert_deploy_failure(result, :timed_out)
 
-    assert_logs_match_all([
-      "Kubernetes will continue to attempt to deploy this resource in the cluster, but at this point it is" \
-        " considered unlikely that it will succeed.",
-      "If you have reason to believe it will succeed, retry the deploy to continue to monitor the rollout.",
-    ])
+    assert_logs_match_all(KubernetesDeploy::KubernetesResource::STANDARD_TIMEOUT_MESSAGE.split("\n"))
   end
 
   def test_wait_false_ignores_non_priority_resource_failures
--- a/test/integration/kubernetes_deploy_test.rb
+++ b/test/integration/kubernetes_deploy_test.rb
@@ -446,6 +446,24 @@ unknown field \"myKey\" in io.k8s.apimachinery.pkg.apis.meta.v1.ObjectMeta",
     ])
   end
 
+  def test_deployment_with_timeout_override
+    result = deploy_fixtures("long-running", subset: ['undying-deployment.yml.erb']) do |fixtures|
+      deployment = fixtures['undying-deployment.yml.erb']['Deployment'].first
+      deployment['spec']['progressDeadlineSeconds'] = 5
+      deployment["metadata"]["annotations"] = {
+        KubernetesDeploy::KubernetesResource::TIMEOUT_OVERRIDE_ANNOTATION => "10S"
+      }
+      container = deployment['spec']['template']['spec']['containers'].first
+      container['readinessProbe'] = { "exec" => { "command" => ['- ls'] } }
+    end
+    assert_deploy_failure(result, :timed_out)
+
+    assert_logs_match_all([
+      "Kubernetes will continue to attempt to deploy this resource in the cluster, but at this point it is considered unlikely that it will succeed.",
+      "If you have reason to believe it will succeed, retry the deploy to continue to monitor the rollout."
+    ])
+  end
+
   def test_wait_false_ignores_non_priority_resource_failures
     # web depends on configmap so will not succeed deployed alone
     assert_deploy_success(deploy_fixtures("hello-cloud", subset: ["web.yml.erb"], wait: false))
--- a/test/integration/kubernetes_deploy_test.rb
+++ b/test/integration/kubernetes_deploy_test.rb
@@ -451,7 +451,7 @@ unknown field \"myKey\" in io.k8s.apimachinery.pkg.apis.meta.v1.ObjectMeta",
       deployment = fixtures['undying-deployment.yml.erb']['Deployment'].first
       deployment['spec']['progressDeadlineSeconds'] = 5
       deployment["metadata"]["annotations"] = {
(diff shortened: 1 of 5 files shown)


test/integration/kubernetes_deploy_test.rb
```diff
@@ -446,6 +446,25 @@ def test_deployment_with_progress_times_out_for_short_duration
...
+      deployment['spec']['progressDeadlineSeconds'] = 5
+      deployment["metadata"]["annotations"] = {
+        KubernetesDeploy::KubernetesResource::TIMEOUT_OVERRIDE_ANNOTATION => "10S",
+      }
+      container = deployment['spec']['template']['spec']['containers'].first
+      container['readinessProbe'] = { "exec" => { "command" => ['- ls'] } }
+    end
+    assert_deploy_failure(result, :timed_out)
+
+    assert_logs_match_all([
```
dturn: `KubernetesDeploy::KubernetesResource::STANDARD_TIMEOUT_MESSAGE` instead of copying and pasting.

test/unit/kubernetes-deploy/kubernetes_resource/deployment_test.rb
```diff
@@ -307,6 +307,24 @@ def test_deploy_timed_out_based_on_progress_deadline
...
+      template = build_deployment_template
+      template["metadata"]["annotations"][KubernetesDeploy::KubernetesResource::TIMEOUT_OVERRIDE_ANNOTATION] = "10S"
+      deploy = build_synced_deployment(
+        template: template,
+        replica_sets: [build_rs_template(status: { "replica" => 1 })]
+      )
+      refute deploy.deploy_timed_out?, "Deploy not started shouldn't have timed out"
+      deploy.deploy_started_at = Time.now.utc - 3.minutes
+
+      assert_equal("Kubernetes will continue to attempt to deploy this resource in the cluster, but at this point it" \
```
dturn: same as above

test/integration/kubernetes_deploy_test.rb
timothysmith0609: Unfortunately that doesn't appear to work due to line splitting in the logs :(

timothysmith0609: Ah, nm I figured out why it wasn't working before

test/unit/kubernetes-deploy/kubernetes_resource/deployment_test.rb
```diff
@@ -307,6 +307,22 @@ def test_deploy_timed_out_based_on_progress_deadline
...
+  def test_deploy_timed_out_based_on_timeout_override
+    Timecop.freeze do
+      template = build_deployment_template
+      template["metadata"]["annotations"][KubernetesDeploy::KubernetesResource::TIMEOUT_OVERRIDE_ANNOTATION] = "10S"
+      deploy = build_synced_deployment(
+        template: template,
+        replica_sets: [build_rs_template(status: { "replica" => 1 })]
+      )
+      refute deploy.deploy_timed_out?, "Deploy not started shouldn't have timed out"
+      deploy.deploy_started_at = Time.now.utc - 3.minutes
```
KnVerey: This doesn't necessarily show that the annotation took precedence over the default and the progress deadline. IMO we should add an expired progress condition, an assertion as to what the default is, and an assertion on each edge of the time that the annotation specifies to prove it is the one causing the timeout.

test/integration/kubernetes_deploy_test.rb
```diff
@@ -446,6 +446,21 @@ def test_deployment_with_progress_times_out_for_short_duration
...
 
+  def test_deployment_with_timeout_override
+    result = deploy_fixtures("long-running", subset: ['undying-deployment.yml.erb']) do |fixtures|
+      deployment = fixtures['undying-deployment.yml.erb']['Deployment'].first
+      deployment['spec']['progressDeadlineSeconds'] = 5
+      deployment["metadata"]["annotations"] = {
+        KubernetesDeploy::KubernetesResource::TIMEOUT_OVERRIDE_ANNOTATION => "10S",
+      }
+      container = deployment['spec']['template']['spec']['containers'].first
+      container['readinessProbe'] = { "exec" => { "command" => ['- ls'] } }
```
KnVerey: Why this patch? Doesn't this change mean the pod could actually come up successfully?

timothysmith0609: I don't believe so, since `- ls` will always fail and the container will remain unready. Is there another way to ensure a timeout that I should be using?

test/unit/kubernetes-deploy/kubernetes_resource/deployment_test.rb
timothysmith0609: I've added assertions on each edge and changed `pretty_timeout_type` to check for timeout override

test/integration/kubernetes_deploy_test.rb
KnVerey: Oh, I misread this as being an array of commands one of which is `ls`. The bad-probe template uses `["test", "0", "-eq", "1"]` but it doesn't really matter.

dturn:
> It seems that Deployment#timeout doesn't behave the way one might expect. If spec.progressDeadlineSeconds is set, Deployment#timeout will just return the default KubernetesDeploy timeout (7 minutes). This doesn't seem like the correct behaviour. I would have assumed that Deployment#timeout would return progressDeadlineSeconds if it is present on the definition (overridden by timeout_override as necessary), but that doesn't appear to be the case. In fact, deployment check for progress deadlines seems more hidden than maybe it should be. While I'm making changes to the class, do you feel I should add a timeout method to the Deploy class so that we surface that progressDeadlineSeconds does indeed specify a timeout on the resource?

I think this behavior is ok. We probably could have given this a more specific name since this is the wall clock timeout which may or may not be relevant. But returning the PDS value from #timeout doesn't seem any more accurate as does about anything else I can think of...

PR #415, opened 2019-01-22
Support stdin

### LINKED ISSUE #410: Accept stdin as template source
For better compatibility with other tooling and to support separate invocation of our renderer (e.g. `kubernetes-render | kubernetes-deploy`) we should accept templates from STDIN instead of requiring them to be persisted to a directory.

timothysmith0609:
Closed #410 
Required for #229 

Allow `kubernetes-deploy` and `kubernetes-render` to accept input from STDIN instead of only accepting a directory path.

**How is this accomplished?**
~~Add a new flag to deploy and render: `--use-stdin` (cannot be combined with `--template-dir`)~~
~~If `--use-stdin` is set, create a temporary directory and write a `yml.erb` file to it (just one file for the entire input for simplicity's sake)~~

~~The tricky part of this PR was figuring out how to get automatic cleanup (e.g. Using the block form of `Dir.mktmpdir`) without having to deal too much with the other case (using `--template-dir`). I did this by wrapping the tasks inside lambdas. If `--use-stdin` is set, an additional wrapper lambda is used to take care of setup and teardown of the temporary files/directories. I think this is preferable to trying to maintain a reference to a tempdir that we have to then manually cleanup afterwards (particularly in the face of errors and unexpected exits).~~

_Edit: Instead of adding a new flag, `--template-dir` can now be set multiple times in an invocation. When set to `-`, it will read the contents of STDIN. In the case of multiple template dirs or reading from STDIN, the templates will all be written to a temporary directory to consolidate all the templates in a single location (keeping the logic in each task runner simple)._

### Incidental bugfix
I noticed that the docs for `kubernetes-render` were not accurate with regards to defaults for `template-dir`. As it stands, `kubernetes-render` _will not_ use the default `config/deploy/$ENVIRONMENT` path if `--template-dir` is not set. This PR corrects that deficiency.

**What could go wrong?**
Unexpected IO errors, possible issues with keeping all templates in a single file (maybe??)


timothysmith0609:
>given the implicit behavior of using config/deploy/$ENVIRONMENT not sure we want to muddy the waters.

We can probably revisit this when we're ready to switch the major version and can introduce such a breaking change.

There's two commits for this PR now: With Lambdas and Without Lambdas. My preference is still to use the initial version (lambdas) since proper cleanup is guaranteed. Without them, I feel that the error handling is a little clumsier. For instance, `KubernetesDeploy::OptionsHelper.default_and_check_template_dir` could raise and that has the potential to leak directories. Conversely, not ensuring `template_dir` in the ensure clause might yield confusing `ENOENT` errors. Maybe I'm worrying too much about rare edge cases, though

exe/kubernetes-deploy
```diff
@@ -40,23 +44,29 @@ ARGV.options do |opts|
...
-  context: context,
-  current_sha: revision,
-  template_dir: template_dir,
-  bindings: bindings,
-  logger: logger,
-  max_watch_seconds: max_watch_seconds
-)
-
 begin
+  template_dir = KubernetesDeploy::OptionsHelper.default_and_check_template_dir(template_dir, use_stdin: use_stdin)
```
KnVerey: Of the two options, I think this one is a lot easier to follow overall. However, I agree that it is a better design to make the place that creates the temp dir responsible for cleaning it up--callers should not have to know what the options helper did! Maybe we could get both advantages by modifying this version to yield the validated path instead of returning it:

```ruby
begin
  KubernetesDeploy::OptionsHelper.with_validated_template_path(template_dir, use_stdin: use_stdin) do |valid_path|
    runner = KubernetesDeploy::DeployTask.new(template_dir: valid_path, ...)
    runner.run!
  end
rescue KubernetesDeploy::DeploymentTimeoutError
  exit(70)
rescue KubernetesDeploy::FatalDeploymentError
  exit(1)
end
```

```ruby
    def self.with_validated_template_dir(template_dir, use_stdin: false)
      if use_stdin
        # ...validate and make temp dir
        yield temp_dir
        return
      end

      if !template_dir && ENV.key?("ENVIRONMENT")
        yield "config/deploy/#{ENV['ENVIRONMENT']}"
      elsif !template_dir || template_dir.empty?
        raise FatalDeploymentError, "Template directory is unknown. " \
          "Either specify --template-dir argument or set $ENVIRONMENT to use config/deploy/$ENVIRONMENT " \
        + "as a default path."
      else
        yield template_dir
      end

      nil
    ensure
      FileUtils.remove_entry(temp_dir) if temp_dir
    end
```

exe/kubernetes-deploy
```diff
@@ -23,6 +26,7 @@ ARGV.options do |opts|
   opts.on("--allow-protected-ns", "Enable deploys to #{prot_ns}; requires --no-prune") { allow_protected_ns = true }
   opts.on("--no-prune", "Disable deletion of resources that do not appear in the template dir") { prune = false }
   opts.on("--template-dir=DIR", "Set the template dir (default: config/deploy/$ENVIRONMENT)") { |v| template_dir = v }
+  opts.on("--use-stdin", "Use $stdin instead of a template dir") { use_stdin = true }
```
KnVerey: I agree with Danny that it isn't ideal to have an additional flag if we can help it. Could we use the unix convention of `-` representing stdin? Ultimately, it could be nice to make this arg feel more like kubectl's, which uses `--filename` for both files and directories, can combine multiple instances of the option into a list, and accepts `-` to mean stdin.

exe/kubernetes-deploy
```diff
@@ -40,23 +44,29 @@ ARGV.options do |opts|
   opts.parse!
 end
 
+if use_stdin && template_dir
```
KnVerey: Should we do this validation inside the OptionsHelper method?

lib/kubernetes-deploy/options_helper.rb
```diff
@@ -2,11 +2,22 @@
 
 module KubernetesDeploy
   module OptionsHelper
-    def self.default_and_check_template_dir(template_dir)
+    def self.default_and_check_template_dir(template_dir, use_stdin: false)
```
KnVerey: Let's add some unit tests for this method

lib/kubernetes-deploy/options_helper.rb
```diff
@@ -2,11 +2,22 @@
...
+      if use_stdin
+        if $stdin.tty?
```
KnVerey: Can you explain this check to me? Intuitively I'd expect `$stdin` to be a tty under normal usage, when this command is being invoked from the command line.

lib/kubernetes-deploy/options_helper.rb
```diff
@@ -2,11 +2,22 @@
...
+          puts "Nothing to read from $stdin"
+          exit(1)
```
KnVerey: I know this pattern exists elsewhere in this file, but conceptually, I find it is better to reserve `exit` calls for the code that is for sure being invoked from the command line, i.e. our `exe` files. Helpers like this should instead raise exceptions.

lib/kubernetes-deploy/options_helper.rb
```diff
@@ -2,11 +2,22 @@
...
+        end
+
+        template_dir = Dir.mktmpdir("kubernetes-deploy")
+        input = $stdin.readlines.join("")
```
KnVerey: Why `readlines.join` instead of `read`?
Similar question for below: Couldn't that be `File.write(File.join(template_dir, 'templates.yml.erb'), input)`? The template dir returned in this case should already be a full path.

exe/kubernetes-deploy
timothysmith0609: That sounds reasonable. So you'd like to be able to use `--template-dir` to be used multiple times (or comma-separated values, I suppose). WDYT about the following 2 options:

1- Serially call kubernetes-deploy for each `template-dir`
2- When multiple template-dirs exist and/or stdin is used, place the contents of _all_ the template dirs + stdin in a temp directory and call `deploy_task` just once.

I'm leaning towards 2, myself.

KnVerey: Option 2 👍 
Pruning would cause problems for option 1.

lib/kubernetes-deploy/options_helper.rb
timothysmith0609: `$stdin` will be a tty when _not_ reading from a pipe. So in our case, if we specify to use `stdin` but there's no pipe to read from, we'll return an error

exe/kubernetes-deploy
timothysmith0609: >...files and directories, can combine multiple instances of the option into a list, and accepts `-` to mean stdin.

It turns out `ARGF` supports this out of the box 👍

exe/kubernetes-deploy
timothysmith0609: This turned out simpler than I thought it would be and should avoid any manual cleanup outside of a bit of error handling since we can encompass everything in the `mktmpdir` block, it will clean everything up for us on exit.

lib/kubernetes-deploy/options_helper.rb
timothysmith0609: Changed this to check for length of `ARGV` instead. This gives us more flexibility for extra filename arguments as well as STDIN support via `-`

timothysmith0609:
stdin_deploy

^^ Example of deploying the hello-cloud fixture reading from stdin

Added unit tests and took @KnVerey 's suggestion to yield the template directory to the runner. 

cc @Shopify/cloudx this is ready for 👀

benlangfeld:
Would someone be so kind as to paste the test failures that were encountered in buildkite here so that I may attempt to resolve them?

randy-abson:
Test failures

```
Finished tests in 1.276781s, 158.9936 tests/s, 773.0377 assertions/s.
  |  
  |  
  | Failure:
  | OptionsHelperTest#test_missing_template_dir_no_extra_input [/usr/src/app/test/unit/kubernetes-deploy/options_helper_test.rb:26]
  | Minitest::Assertion: KubernetesDeploy::OptionsHelper::OptionsError expected but nothing was raised.
  |  
  | Failure:
  | OptionsHelperTest#test_multiple_template_dirs [/usr/src/app/test/unit/kubernetes-deploy/options_helper_test.rb:18]
  | Minitest::Assertion: Expected: 18
  | Actual: 19
  |  
  | Failure:
  | OptionsHelperTest#test_single_template_dir_only [/usr/src/app/test/unit/kubernetes-deploy/options_helper_test.rb:8]
  | Minitest::Assertion: --- expected
  | actual | 0s
  | @@ -1 +1 @@
  | -"/usr/src/app/test/fixtures/hello-cloud"
  | +"/tmp/kubernetes-deploy20190222-761-1teyzrb"


```

benlangfeld:
Thank you @randy-abson. Those tests pass for me locally; is there something else significantly different about the CI build environment, other than presumably being some Linux variant?

lib/kubernetes-deploy/options_helper.rb
```diff
@@ -2,25 +2,58 @@
...
+        File.open("#{temp_dir}/#{STDIN_TEMP_FILE}", 'w+') { |f| f.print(ARGF.read) } unless $stdin.tty?
 
-    def self.revision_from_environment
-      ENV.fetch('REVISION') do
-        puts "ENV['REVISION'] is missing. Please specify the commit SHA"
-        exit 1
+        template_dirs.each do |template_dir|
+          template_dir = File.expand_path(template_dir)
+          templates = Dir.entries(template_dir).reject { |f| File.directory?("#{template_dir}/#{f}") }
+          templates.each do |template|
```
KnVerey: could we use `Fileutils.cp` instead of individually opening and writing them?

lib/kubernetes-deploy/options_helper.rb
```diff
@@ -2,25 +2,58 @@
...
       end
 
-      if !template_dir || template_dir.empty?
-        puts "Template directory is unknown. " \
-          "Either specify --template-dir argument or set $ENVIRONMENT to use config/deploy/$ENVIRONMENT " \
-        + "as a default path."
-        exit(1)
+      def revision_from_environment
+        ENV.fetch('REVISION') do
+          puts "ENV['REVISION'] is missing. Please specify the commit SHA"
```
KnVerey: Let's raise this as an exception instead of exiting outside the executable

timothysmith0609: Handled by #429

timothysmith0609:
>As discussed on chat, I think we should stick to flag-based directory input (not positional)

Changed to only accept the `--template-dir` arg, setting `-` to specify STDIN. STDIN is no longer inferred nor are positional args supported to specify filenames.

>I like that we can now handle multiple directories
:+1:

>kubernetes-render needs to handle stdin too

Done

>other exes probably need to handle your new OptionsError too

Deploy and render are the only 2 exes that need to worry about this; they both now rescue the exception

lib/kubernetes-deploy/options_helper.rb
```diff
@@ -2,25 +2,58 @@
...
+      end
 
-    def self.revision_from_environment
-      ENV.fetch('REVISION') do
-        puts "ENV['REVISION'] is missing. Please specify the commit SHA"
-        exit 1
+      def populate_temp_dir(temp_dir:, template_dirs:)
+        template_dirs.each do |template_dir|
+          if template_dir == '-'
+            File.open("#{temp_dir}/#{STDIN_TEMP_FILE}", 'w+') { |f| f.print($stdin.read) }
```
KnVerey: Everywhere you're hardcoding a `/`  in this file we should change to `File.join(path, pieces)` for compatibility reasons

lib/kubernetes-deploy/options_helper.rb
```diff
@@ -2,25 +2,58 @@
...
+          else
+            template_dir = File.expand_path(template_dir)
+            templates = Dir.entries(template_dir).reject { |f| File.directory?("#{template_dir}/#{f}") }
+            templates.each do |template|
+              FileUtils.cp("#{template_dir}/#{template}", "#{temp_dir}/#{template_dir.tr('/', '_')}_#{template}")
```
KnVerey: > template_dir.tr('/', '_')

I think you want `File::SEPARATOR` here--it's not guaranteed to be a slash

test/unit/kubernetes-deploy/options_helper_test.rb
```diff
@@ -0,0 +1,104 @@
+# frozen_string_literal: true
+require 'test_helper'
+require 'tempfile'
+
+class OptionsHelperTest < KubernetesDeploy::TestCase
+  def test_single_template_dir_only
```
KnVerey: let's add one for when the template dir is defaulted based on `$ENVIRONMENT`--that case should also return a non-temp path

test/unit/kubernetes-deploy/options_helper_test.rb
```diff
@@ -0,0 +1,104 @@
...
+
+  def test_multiple_template_dirs
+    template_dirs = [fixture_path('hello-cloud'), fixture_path('partials')]
+
+    KubernetesDeploy::OptionsHelper.with_consolidated_template_dir(template_dirs) do |template_dir|
+      fixture_path_entries = template_dirs.collect { |dir| Dir.entries(dir) }.flatten.uniq
+      template_dir_entries = Dir.entries(template_dir)
+      assert_equal(fixture_path_entries.length, template_dir_entries.length)
+      fixture_path_entries.each do |fixture|
+        assert(template_dir_entries.select { |s| s.include?(fixture) })
```
KnVerey: I think this will always pass, because `select` will return `[]` if it never matches, which is truthy. I suspect that's what's happening, because the two files will have different full paths, one being in a tempdir and one being in the original location.

test/unit/kubernetes-deploy/options_helper_test.rb
```diff
@@ -0,0 +1,104 @@
...
+      KubernetesDeploy::OptionsHelper.with_consolidated_template_dir([]) do
+      end
+    end
+  end
+
+  def test_template_dir_with_stdin
+    old_stdin = $stdin
+
+    input = Tempfile.open("kubernetes_deploy_test")
+    File.open("#{fixture_path('for_unit_tests')}/service_test.yml", 'r') do |f|
```
KnVerey: Same comment about using `File.join`.

KnVerey: OTOH would it be simpler to use `StringIO` instead of actually manipulating files?

lib/kubernetes-deploy/options_helper.rb
KnVerey: Should we still check `$stdin.tty?` like you were before?

README.md
```diff
@@ -117,11 +117,12 @@ offical compatibility chart below.
...
 
-- `--template-dir=DIR`: Used to set the deploy directory. Set `$ENVIRONMENT` instead to use `config/deploy/$ENVIRONMENT`.
+- `--template-dir=DIR`: Used to set the deploy directory. Set `$ENVIRONMENT` instead to use `config/deploy/$ENVIRONMENT`. You can set multiple template directories by repeating the `--template-dir` flag.
 - `--bindings=BINDINGS`: Makes additional variables available to your ERB templates. For example, `kubernetes-deploy my-app cluster1 --bindings=color=blue,size=large` will expose `color` and `size`.
 - `--no-prune`: Skips pruning of resources that are no longer in your Kubernetes template set. Not recommended, as it allows your namespace to accumulate cruft that is not reflected in your deploy directory.
 - `--max-watch-seconds=seconds`: Raise a timeout error if it takes longer than _seconds_ for any
 resource to deploy.
+- `kubernetes-deploy` also supports reading from STDIN. You can do this by using `--template-dir=-`. Example: `cat templates_from_stdin/*.yml | kubernetes-deploy ns ctx --template-dir=- --template-dir=another_template_dir `. Ensure the templates you are reading in are separated by `---` when using this feature.
```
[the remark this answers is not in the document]
timothysmith0609: >Is this different than if you have multiple templates in the same directory? Why?

It happened to me while testing that, while reading in multiple files, a closing `---` was missing between the yamls of some files. On reflection the comment is more specific to YAML and not kubernetes-deploy so I'm in favor of dropping the reference.

lib/kubernetes-deploy/options_helper.rb
```diff
@@ -2,25 +2,58 @@
...
-      ENV.fetch('REVISION') do
-        puts "ENV['REVISION'] is missing. Please specify the commit SHA"
-        exit 1
+      def populate_temp_dir(temp_dir:, template_dirs:)
+        template_dirs.each do |template_dir|
+          if template_dir == '-'
+            File.open("#{temp_dir}/#{STDIN_TEMP_FILE}", 'w+') { |f| f.print($stdin.read) }
+          else
+            template_dir = File.expand_path(template_dir)
+            templates = Dir.entries(template_dir).reject { |f| File.directory?("#{template_dir}/#{f}") }
```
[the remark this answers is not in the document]
timothysmith0609: :+1: That's more revealing of the intent

lib/kubernetes-deploy/options_helper.rb
timothysmith0609: We _could_. I removed it here since that's the convention I see adopted by most command line tools. On the other hand, it's not great UX if the task hangs waiting for input if the user makes an error in the pipeline. 

I'm leaning towards raising an exception if STDIN is specified but attached to a tty

test/unit/kubernetes-deploy/options_helper_test.rb
timothysmith0609: `refute(template_dir_entries.select { |template| template.include?(fixture) }.empty?)` passes the test, so the logic is working even though the test was initially flawed.

If you feel it's worth massaging this test some more I'll do so but I'm satisfied with it for now.

test/unit/kubernetes-deploy/options_helper_test.rb
timothysmith0609: Seeing as the tests are already written I'm inclined to leave them be.

timothysmith0609:
PR comments have been addressed, rebased on master, and removed old references to `revision` that were caught up in this branch.

exe/kubernetes-deploy
```diff
@@ -42,27 +46,31 @@ end
...
+      allow_protected_ns: allow_protected_ns,
+      prune: prune
+    )
+  end
 rescue KubernetesDeploy::DeploymentTimeoutError
   exit(70)
 rescue KubernetesDeploy::FatalDeploymentError
   exit(1)
+rescue KubernetesDeploy::OptionsHelper::OptionsError => e
+  puts e.message
```
dturn: Do we want to do the logging here or in `with_consolidated_template_dir`. If here, should this be using `logger` instead of `puts`

test/unit/kubernetes-deploy/options_helper_test.rb
```diff
@@ -0,0 +1,114 @@
...
+require 'tempfile'
+
+class OptionsHelperTest < KubernetesDeploy::TestCase
+  def test_single_template_dir_only
+    KubernetesDeploy::OptionsHelper.with_consolidated_template_dir([fixture_path('hello-cloud')]) do |template_dir|
+      assert_equal(fixture_path('hello-cloud'), template_dir)
+    end
+  end
+
+  def test_with_default_env_var
```
dturn: do these tests get run in parallel?

test/unit/kubernetes-deploy/options_helper_test.rb
```diff
@@ -0,0 +1,114 @@
...
+
+  def test_template_dir_with_stdin
+    old_stdin = $stdin
+
+    input = Tempfile.open("kubernetes_deploy_test")
+    File.open(File.join(fixture_path('for_unit_tests'), "service_test.yml"), 'r') do |f|
+      input.print(f.read)
+    end
+    input.rewind
+    $stdin = input
```
dturn: Can you write to `$stdin` instead of abusing a global?

test/unit/kubernetes-deploy/options_helper_test.rb
timothysmith0609: From my observations, no.

See [this PR description](https://github.com/Shopify/kubernetes-deploy/pull/233)

KnVerey: no, unit tests don't (because a huge portion of them use mocha)

test/unit/kubernetes-deploy/options_helper_test.rb
timothysmith0609: From what I gather, no. The best I can do is ensure that `old_stdin` is reset at the end of the test. I'm not sure how I could even write to `$stdin` without changing the descriptor itself, regardless.

exe/kubernetes-deploy
timothysmith0609: Makes sense for me to use the logger. I'll update

test/unit/kubernetes-deploy/options_helper_test.rb
KnVerey: We should use `with_env(key, value)` from `EnvTestHelper` instead though

benlangfeld: Addressed in https://github.com/Shopify/kubernetes-deploy/pull/444

CHANGELOG.md
```diff
@@ -4,7 +4,9 @@
 - Setting the REVISION environment variable is now optional ([#429](https://github.com/Shopify/kubernetes-deploy/pull/429))
 - Defaults KUBECONFIG to `~/.kube/config` ([#429](https://github.com/Shopify/kubernetes-deploy/pull/429))
 - Uses `TASK_ID` environment variable as the `deployment_id` when rendering resource templates for better [Shipit](https://github.com/Shopify/shipit) integration. ([#430](https://github.com/Shopify/kubernetes-deploy/pull/430))
-- Arguments to `--bindings` will now be deep merged. ([#419](https://github.com/Shopify/kubernetes-deploy/pull/419))
+- Arguments to `--bindings` will now be deep merged. ([#430](https://github.com/Shopify/kubernetes-deploy/pull/430))
```
benlangfeld: This is re-instating an error. Needs rebasing.

https://github.com/Shopify/kubernetes-deploy/commit/d0343e7571471fc8d3c1b4d05296698fd70ac06f

benlangfeld:
```suggestion
- Arguments to `--bindings` will now be deep merged. ([#419](https://github.com/Shopify/kubernetes-deploy/pull/419))
```

PR #416, opened 2019-01-22
Support for deploying only secrets.ejson 

### LINKED ISSUE #412: Deploying ejson only fails
# Bug report

kubernetes-deploy always expects at least a valid yml|yml.erb file within the template directory, although a valid `secrets.ejson` file exists and could be applied to the cluster.

**Expected behavior:** 
Secret created successfully in targeted namespace.

**Actual behavior:** 
`kubernetes-deploy` failes the deploy with the message: `<given-template-dir> doesn't contain valid templates (postfix .yml or .yml.erb)`

**Version(s) affected:**
0.23.0

### Steps to Reproduce
run `kubernetes-deploy` in a template directory where there is only a `secrets.ejson` file for a namespace which exists.

As soon as a valid yaml is placed file into the template directory besides `secret.ejson` the ejson file is decrypted and being applied correctly along side the other yaml file(s).

timothysmith0609:
Closes #412 

Let users deploy directories that only contain `secrets.ejson`

**How is this accomplished?**
Add `secrets.ejson` to the regex that checks for valid templates

**What could go wrong?**
Not too much

### DIFF
--- a/lib/kubernetes-deploy/deploy_task.rb
+++ b/lib/kubernetes-deploy/deploy_task.rb
@@ -324,7 +324,7 @@ module KubernetesDeploy
 
       if !File.directory?(@template_dir)
         errors << "Template directory `#{@template_dir}` doesn't exist"
-      elsif Dir.entries(@template_dir).none? { |file| file =~ /(\.ya?ml(\.erb)?)|(secrets\.ejson)$/ }
+      elsif Dir.entries(@template_dir).none? { |file| file =~ /(\.ya?ml(\.erb)?)$|(secrets\.ejson)$/ }
         errors << "`#{@template_dir}` doesn't contain valid templates (secrets.ejson or postfix .yml, .yml.erb)"
       end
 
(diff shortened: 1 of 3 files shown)


lib/kubernetes-deploy/deploy_task.rb
```diff
@@ -324,8 +324,8 @@ def validate_configuration(allow_protected_ns:, prune:)
 
       if !File.directory?(@template_dir)
         errors << "Template directory `#{@template_dir}` doesn't exist"
-      elsif Dir.entries(@template_dir).none? { |file| file =~ /\.ya?ml(\.erb)?$/ }
-        errors << "`#{@template_dir}` doesn't contain valid templates (postfix .yml or .yml.erb)"
+      elsif Dir.entries(@template_dir).none? { |file| file =~ /(\.ya?ml(\.erb)?)|(secrets\.ejson)$/ }
```
KnVerey:
```suggestion
      elsif Dir.entries(@template_dir).none? { |file| file =~ /(\.ya?ml(\.erb)?)$|(secrets\.ejson)$/ }
```

timothysmith0609:
>Can you please add a test?

I had misread how `deploy_fixtures` worked at first so was struggling to build the test. Once I figured it out it was easy. For my sanity I tried the test without the rest of the diff and it fails as expected so 👍

PR #418, opened 2019-01-24
Kubeclient 4.2.2

### LINKED ISSUE #406: Update kubeclient
Related to #396 -- the most recent release should be compatible with AWS.

Our "google-friendly" patch might also be obsolete at this point. We should read the changelog and confirm.

timothysmith0609:
Closes #406 
Update Kubeclient to 4.2.2. This will let AWS users properly auth via their kubeconfig

**How is this accomplished?**
- Bump Gem
- Remove GoogleFriendlyConfig (still need a shim for GCP, but less involved)
- Go through Changelog

**Response to changelog**
- `KubeException` is deprecated (it already was, before). Replaced with `Kubeclient::HttpError`, which is just a [subclass of `KubeException`](https://github.com/abonas/kubeclient/blob/master/lib/kubeclient/http_error.rb#L23).

**What could go wrong?**
New, strange, behaviour not covered by our tests


lib/kubernetes-deploy/restart_task.rb
```diff
@@ -112,7 +112,7 @@ def build_watchables(kubeclient_resources, started)
     def verify_namespace
       kubeclient.get_namespace(@namespace)
       @logger.info("Namespace #{@namespace} found in context #{@context}")
-    rescue KubeException => error
+    rescue Kubeclient::HttpError => error
```
dturn: Might be out of scope of this PR but should we be differentiating between `HttpError` and `ResourceNotFoundError`

lib/kubernetes-deploy/kubeclient_builder/kube_config.rb
```diff
@@ -0,0 +1,21 @@
+# frozen_string_literal: true
+
+require 'googleauth'
+module KubernetesDeploy
+  module KubeclientBuilder
+    class KubeConfig < Kubeclient::Config
+      def self.read(filename)
+        parsed = YAML.safe_load(File.read(filename), [Date, Time])
```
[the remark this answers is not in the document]
timothysmith0609: See the [Kubeclient PR](https://github.com/abonas/kubeclient/pull/338)

lib/kubernetes-deploy/restart_task.rb
timothysmith0609: In [Kubeclient](https://github.com/abonas/kubeclient/blob/master/lib/kubeclient/resource_not_found_error.rb), `ResourceNotFoundError` is a subclass of HttpError so we will be catching it regardless. From what I see, Kubeclient always provides response code with `HttpError`, so our current checks in kubernetes-deploy will still work all the same.

Also worth noting that `KubeException` is [going to be removed on next release](https://github.com/abonas/kubeclient/pull/293) so this is a requirement for future kubeclient updates.

timothysmith0609:
As @KnVerey pointed out, it looks like [ActiveFailover](https://github.com/Shopify/activefailover/blob/master/lib/activefailover/kubernetes/kube_config.rb#L30) makes a reference to the to-be-removed `GoogleFriendlyConfig`. It looks like all we need to do is change the reference from `GoogleFriendlyConfig` to `KubernetesDeploy::KubeclientBuilder::Kubeconfig` and their client should retain the correct behaviour. 

I'm tempted to change the reference to just use the base KubeClient class (`Kubeclient::Config`) since it's passing the tests, but I think that will cause issues when it actually has to auth against GCP in production.

KnVerey:
Another note: we should definitely test this branch manually using both user credentials and application default credentials before merging. My test app has a Shipit task that can be used for the latter.

timothysmith0609:
>`auth_options: { bearer_token: Kubeclient::GoogleApplicationDefaultCredentials.token }`. Does that not work?

This would work up until the point we're not targeting a GCP cluster, which due to the open-source nature of this gem is not a valid assumption to make. I'm wondering if it's worth patching kubeclient upstream. Our simple fix (checking for `auth-provider == "gcp") seems sufficient.

timothysmith0609:
>Another note: we should definitely test this branch manually using both user credentials and application default credentials before merging. My test app has a Shipit task that can be used for the latter.

Running with user credentials locally is working fine. Is the [Cumulus Cat Task](https://shipit.shopify.io/shopify/cumulus-cat/production/tasks/k8s-deploy-branch/new) the Shipit task you're talking about?

timothysmith0609:
Created upstream [kubeclient issue](https://github.com/abonas/kubeclient/issues/392) for automatically loading default application credentials.

CHANGELOG.md
```diff
@@ -1,5 +1,8 @@
 ## next
 
+*Bug fixes*
```
dturn: Rebase issue?

PR #419, opened 2019-02-19
Do a deep merge on bindings rather than a shallow

n1koo:
Make it possible to have multiple overlapping configurations that will be deep merged to form one sane configuration.

Eg. lets say you have following configurations:

common.yaml:
```
common:
  registry: private-thingy
...
app:
  port: 4200
  name: cat-service
  lb_id: nonprod
....
```

and then you have environment specific one:

production.yaml:
```
app:
  lb_id: prod
```

Currently we only merge top level entries with `merge` so eg. if `lb_id` was on top this would be fine but for more complex configurations people tend to separate things in to nested entries. Another example could be eg. setting `resources` but overriding limits in certain env, think of:

```
  resources:
    limits:
      cpu: "3"
      memory: "1Gi"
    requests:
      cpu: "3"
      memory: "1Gi"
```

and then in `nonprod` you wanna override the `requests` for example.

**How is this accomplished?**
- Pull the bindings loop in to the parser
- do a `deep_merge!` instead of `merge!`
- profit

tl;dr `s/merge/deep_merge` but I thought its cleaner to pull the loop in to the parser too. Makes testing lot more easier and harder to screw up if doing changes in both deployer and renderer

**What could go wrong?**
Things dont get parsed, world blows up. 

Seriously though, if someone is relying on the wrong behaviour (shallow merge) this might cause a surprise, but technically doing so is wrong .. so :)

### DIFF
--- a/exe/kubernetes-deploy
+++ b/exe/kubernetes-deploy
@@ -13,10 +13,9 @@ verbose_log_prefix = false
 max_watch_seconds = nil
 
 ARGV.options do |opts|
+  parser = KubernetesDeploy::BindingsParser.new
   opts.on("--bindings=BINDINGS", "Expose additional variables to ERB templates " \
-    "(format: k1=v1,k2=v2, JSON string or file (JSON or YAML) path prefixed by '@')") do |binds|
-    bindings.merge!(KubernetesDeploy::BindingsParser.parse(binds))
-  end
+    "(format: k1=v1,k2=v2, JSON string or file (JSON or YAML) path prefixed by '@')") { |b| parser.add(b) }
 
   opts.on("--skip-wait", "Skip verification of non-priority-resource success (not recommended)") { skip_wait = true }
   prot_ns = KubernetesDeploy::DeployTask::PROTECTED_NAMESPACES.join(', ')
@@ -38,6 +37,7 @@ ARGV.options do |opts|
     exit
   end
   opts.parse!
+  bindings = parser.parse
 end
 
 namespace = ARGV[0]
--- a/lib/kubernetes-deploy/bindings_parser.rb
+++ b/lib/kubernetes-deploy/bindings_parser.rb
@@ -4,17 +4,29 @@ require 'yaml'
 require 'csv'
 
 module KubernetesDeploy
-  module BindingsParser
-    extend self
+  class BindingsParser
+    def self.parse(string)
+      new(string).parse
+    end
 
-    def parse(string)
-      bindings = parse_file(string) || parse_json(string) || parse_csv(string)
+    def initialize(initial_string = nil)
+      @raw_bindings = Array(initial_string)
+    end
 
-      unless bindings
-        raise ArgumentError, "Failed to parse bindings."
+    def add(string)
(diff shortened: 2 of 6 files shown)


exe/kubernetes-deploy
```diff
@@ -13,10 +13,9 @@ verbose_log_prefix = false
 max_watch_seconds = nil
 
 ARGV.options do |opts|
+  binds = []
   opts.on("--bindings=BINDINGS", "Expose additional variables to ERB templates " \
-    "(format: k1=v1,k2=v2, JSON string or file (JSON or YAML) path prefixed by '@')") do |binds|
-    bindings.merge!(KubernetesDeploy::BindingsParser.parse(binds))
```
dturn: any reason not to use `deep_merge!` here?

n1koo: That was my original change but after realising its done in two files and testing would be super funky it seemed to make sense extract. I'm fine either way but not sure how clean we can make tests if its just changed in the exes

n1koo:
Rebase done

lib/kubernetes-deploy/bindings_parser.rb
```diff
@@ -7,7 +7,15 @@ module KubernetesDeploy
   module BindingsParser
     extend self
 
-    def parse(string)
+    def parse(binds)
```
KnVerey: I think this PR is 👍 functionality-wise, but I feel like this class's API is a bit weird now--it's not intuitive what the difference is between `BindingsParser::parse` and `BindingsParser::parse_binding` would be.

Here's a suggestion:

```ruby
# If you need multiple, you do:
bindings = KubernetesDeploy::BindingsParser.new
# ...
  opts.on("--bindings=BINDINGS", "...") { |b| bindings.add(b) }
# ...
bindings.parse

# But you can still also use the exact same API as before:
KubernetesDeploy::BindingsParser.parse(bindings_string)

module KubernetesDeploy
  class BindingsParser
    def self.parse(string)
      new(string).parse
    end

    def initialize(initial_string=nil)
      @raw_bindings = Array(initial_string)
    end

    def add(string)
      @raw_bindings << string
    end

    def parse
      result = {}
      @raw_bindings.each do |string|
        # parse/merge
      end
      # etc
      result
    end
```

n1koo: Makes sense, refactored

exe/kubernetes-deploy
```diff
@@ -38,6 +37,7 @@ ARGV.options do |opts|
     exit
   end
   opts.parse!
+  bindings = bindings.parse
```
KnVerey: This should be `parser.parse`, which reminds me that someone needs to tophat all affected executables before we merge this.

n1koo: Yeah, I did my tophat with `render` 🤦‍♂️

KnVerey:
I was able to confirm successful 🎩  using the `collection-with-erb` fixtures with both `kubernetes-render` and `kubernetes-deploy`. We'll :shipit: once CI passes.

Issue #209, opened 2017-11-22
Secrets should be predeployed before tasks


agilgur5:
I have a task / bare pod that has part of its environment set from a Secret. These pods fail right now because Secrets are deployed after them

KnVerey:
The supported way to manage secrets with kubernetes-deploy is to use EJSON to store them in encrypted form: https://github.com/Shopify/kubernetes-deploy#deploying-kubernetes-secrets-from-ejson. Secrets managed this way will be correctly created/updated before anything else. I would not recommend keeping kubernetes secrets unencrypted in your source.

agilgur5:
I'm currently inserting Secrets at deploy time only, so they're of course not in source (and if I had to choose, I would prefer to use Vault or Bitnami's Sealed Secrets so they're not so easily accessible in Kubernetes). I wouldn't think it would be difficult to support regular Secrets as you'd just skip the decryption part

KnVerey:
> I'm assuming that was an intentional design decision

Correct, we intentionally added ejson-to-secret support instead to encourage safe secret management. It wouldn't be difficult to predeploy regular secrets, but I'm hesitant to support that for best-practice reasons as you said. We can revisit this if there is more demand for it.

agilgur5:
Er, I said that I _didn't_ believe that's a best practice. Unless there are plans to support all sorts of secrets providers and mechanisms (I don't believe there are), regular Secrets should be supported -- it's up to the developer to handle them correctly. The current approach creates lock-in to EJSON and does not allow other secret mechanisms.

A good alternative that would give developers freedom would be to hide regular Secrets behind some fla, thereby requiring explicit confirmation of potential pitfalls.

agilgur5:
The current approach is: "You don't know what you're doing, so you must use our preferred method, EJSON" vs. a flag would be: "This is not recommended, only use this if you know what you're doing".

The underlying assumptions are quite different between the two.

KnVerey:
Sorry for my misreading of your initial comment. We're open to your idea of adding a flag that enables predeployment of raw secrets. Then, if the flag isn't provided and we discover a raw secret among the resources, we can fail validation early in the process with a helpful message about the risks and the flag instead of just printing the unsupported resource warning as we do today. Implementing this is not a priority for us, but we would accept a PR.

asabourin:
We're in a similar situation where we've got a secrets.yml.erb with the following:
```ruby
apiVersion: v1
data:
[...]

PR #421, opened 2019-02-20
Support label based namespace partitioning

DazWorrall:
Closes #411 

This is little more than a rebase and refresh of #235 at the moment, from here we can figure out the next steps required to get this over the line. My only substantial change was a tweak to one test to fix a race where a pruned resource is still discoverable for a brief amount of time - is this ok?

https://github.com/Shopify/kubernetes-deploy/blob/13763f81bb742013c9cffa2b9f9143c112cf21f2/test/integration/kubernetes_deploy_test.rb#L138-L139

@KnVerey I saw you mention some potential work around secrets which might be a prerequisite for this, do you have any more context on that?

Issue #411, opened 2019-01-09
Support label-based sub-namespace resource management


KnVerey:
i.e. revive and merge https://github.com/Shopify/kubernetes-deploy/pull/235

benlangfeld:
What remains to be done on #235 to get it merged? A skim of the review comments suggests they were all either addressed or withdrawn.

KnVerey:
IIRC there was an unresolved gotcha related to selector-based pruning of secrets derived from EJSON.  At this point, the PR is a year old and will need a major rebase. But you're right that it was very close to the finish line.

PR #424, opened 2019-02-25
Deploy secrets in DeployTask like other resources

DazWorrall:
Ref: #421 #411 
Closes: #209

Have ejson secrets provisioned like any other resource, rather than separately/specially in `EjsonSecretProvisioner`. This PR also adds support for deploying secrets from yaml templates.

**How is this accomplished?**

The individual commits tell the story, it might be easier to review this PR by stepping through them individually.

**What could go wrong?**

A scary-ish change in this PR is adding secrets to the prune whitelist on deploy task. We were pruning secrets before of course, but in a more targeted way. Some questions I have:

* Do we still need/want the 'managed secret' annotation?
* What additional testing would add confidence this isn't going to break anything in the wild?
* Would anyone prefer I separate out the 'sensitive output' refactor?
* I flagged secrets for predeployment, on the basis that pods deployed later may need the new versions. Is that sound?

Still tbd after this: support for adding labels from ejson, and resolving how to 'template' ejson secrets. This PR is big enough already.


DazWorrall:
Test failures on CI are due to small output differences across k8s versions, I'll fix those.

benlangfeld:
I had been working on exactly the same change this morning, was nearly done with it too 😅

> I flagged secrets for predeployment, on the basis that pods deployed later may need the new versions. Is that sound?

This is a requirement for my use case.

lib/kubernetes-deploy/ejson_secret_provisioner.rb
```diff
@@ -17,28 +17,22 @@ class EjsonSecretProvisioner
...
 
-    def secret_changes_required?
-      File.exist?(@ejson_file) || managed_secrets_exist?
-    end
-
-    def run
-      create_secrets
-      prune_managed_secrets if @prune
+    def resources
+      @resources ||= File.exist?(@ejson_file) ? create_secrets : []
```
dturn: Seems like File.exist?(@ejson_file)` is an implementation detail leaking. ` Maybe `create_secrets` should return `[]` or the `exist?` call should be wrapped in a function

lib/kubernetes-deploy/deploy_task.rb
```diff
@@ -240,20 +243,21 @@ def check_initial_status(resources)
...
         logger: @logger,
-        prune: prune,
       )
-      return unless ejson.secret_changes_required?
 
-      @logger.phase_heading("Deploying kubernetes secrets from #{EjsonSecretProvisioner::EJSON_SECRETS_FILE}")
-      ejson.run
+      unless ejson.resources.empty?
+        @logger.info(
+          "Loaded #{ejson.resources.size} kubernetes secrets from #{EjsonSecretProvisioner::EJSON_SECRETS_FILE}"
```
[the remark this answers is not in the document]
benlangfeld: "Generated" was the term I used.

lib/kubernetes-deploy/ejson_secret_provisioner.rb
benlangfeld: `#with_decrypted_ejson` already does that check, so it's somewhat redundant. I did `with_decrypted_ejson do .... end || []`.

DazWorrall: I wrote this in the distant past of last week 😅iirc I chose this approach because some of the log assertions were failing in cases when e.g. there's no matching key. I'll have another pass and try to tidy this up.

lib/kubernetes-deploy/deploy_task.rb
```diff
@@ -130,11 +133,11 @@ def run!(verify_result: true, allow_protected_ns: false, prune: true)
       @logger.phase_heading("Initializing deploy")
       validate_configuration(allow_protected_ns: allow_protected_ns, prune: prune)
       resources = discover_resources
+      resources += secrets_from_ejson
```
timothysmith0609: I dislike that `discover_resources` is now _not_ discovering all resources and we need to make this additional call. Would the abstraction not be more fitting to generate the EJSON secrets inside `discover_resources` itself? This would require updating `TemplateDiscovery` as well as `KubernetesResource::build`. IMO deploy_task shouldn't be in the business of being aware of secrets

benlangfeld:
I need this and https://github.com/Shopify/kubernetes-deploy/pull/421 quite urgently. Is there anything I can do to help move either forward? I do not have access to buildkite to see what the failures are :/

lib/kubernetes-deploy/deploy_task.rb
DazWorrall: I'll move the secret appending into `discover_resources` to encapsulate all discovery there, but I'll hold off on touching `TemplateDiscovery` for now unless you feel strongly - processing secrets as templates will probably need to be addressed to resolve [this discussion](https://github.com/Shopify/kubernetes-deploy/pull/235#discussion_r161610047), and I don't want this PR to get any bigger if possible. For now this PR does ~one thing: generalise how secrets are deployed, we can change how they're built in a follow up.

DazWorrall:
@benlangfeld some context on timelines: I'm actively working on this, but not exclusively, and as I'm separated from my reviewers by a few timezones this will probably take in the order of ~days to complete - not a few hours, but (hopefully) not ~weeks either.

Also, sorry for not communicating more loudly I was doing this before you started hacking yourself.

DazWorrall:
An annoyance to fix later: I couldn't replicate the policial failures locally, so I guess we have some config drift.

lib/kubernetes-deploy/kubernetes_resource.rb
```diff
@@ -53,6 +50,14 @@ def build(namespace:, context:, definition:, logger:, statsd_tags:, crd: nil)
         end
       end
 
+      def class_for_kind(kind)
+        if KubernetesDeploy.const_defined?(kind)
+          KubernetesDeploy.const_get(kind) # rubocop:disable Sorbet/ConstantsFromStrings
```
DazWorrall: > Sorbet/ConstantsFromStrings: Don't use `const_get`, it make the code harder to understand, less editor-friendly and impossible to analyze. Replace `const_get` with a case/when or a hash.

That suggestion seems out of scope for this PR, so I chose to disable the cop.

DazWorrall:
This is ready for another pass.

rendhalver:
>     * Do we still need/want the 'managed secret' annotation?

I can't find a reference to this annotation?
By the name I would assume it would stop a secret deployed by something else from being removed by `kubernetes-deploy`?
Am I correct?

I would think keeping that annotation is a good idea and would be useful for our setup.

DazWorrall:
> I can't find a reference to this annotation?

Secrets created from ejson are currently [annotated](https://github.com/Shopify/kubernetes-deploy/blob/57d3e9a66de4783ade0fc0c6a3a3ca46ad1c0557/lib/kubernetes-deploy/ejson_secret_provisioner.rb#L98) with `kubernetes-deploy.shopify.io/ejson-secret=true`.

> By the name I would assume it would stop a secret deployed by something else from being removed by `kubernetes-deploy`? Am I correct?

No, it's used by the current implementation to ensure [only 'managed' secrets are pruned](https://github.com/Shopify/kubernetes-deploy/blob/d28b284c50197ad6228d55bcef65e9fa6f55c64a/lib/kubernetes-deploy/ejson_secret_provisioner.rb#L70), but after this the lifecycle of secrets will be managed like any other resource.

So I guess the answer to my question is "no it's not needed", but want to make sure.

lib/kubernetes-deploy/kubernetes_resource.rb
dturn: > That suggestion seems out of scope for this PR

I agree

lib/kubernetes-deploy/ejson_secret_provisioner.rb
```diff
@@ -149,12 +99,10 @@ def generate_secret_yaml(secret_name, secret_type, data)
...
         "data" => encoded_data,
       }
-      secret.to_yaml
-    end
 
-    def secret_exists?(secret_name)
-      _out, _err, st = @kubectl.run("get", "secret", secret_name)
-      st.success?
+      KubernetesDeploy::Secret.build(
+        namespace: @namespace, context: @context, logger: @logger, definition: secret, statsd_tags: [],
```
KnVerey: `statsd_tags` should get the same set as other resources, so either we need to pass that set to this class, or we need to make this class return the definition only and make `DeployTask` build the resource.

lib/kubernetes-deploy/ejson_secret_provisioner.rb
```diff
@@ -17,75 +17,41 @@ class EjsonSecretProvisioner
     EJSON_SECRETS_FILE = "secrets.ejson"
     EJSON_KEYS_SECRET = "ejson-keys"
```
KnVerey: nit: I think it's ok to keep the "management annotation" on the resources, because it might be useful for people to be able to see the secret was created that way. However, I think we should rename the `MANAGEMENT_ANNOTATION` constant because we're no longer using it for that purpose.

lib/kubernetes-deploy/kubernetes_resource/secret.rb
```diff
@@ -0,0 +1,23 @@
+# frozen_string_literal: true
+module KubernetesDeploy
+  class Secret < KubernetesResource
+    TIMEOUT = 10.seconds
```
KnVerey: I think this is reasonable, but it is inconsistent with our timeout for other resources like this (e.g. Configmap). We should probably have a constant they all use. 🤔

test/helpers/fixture_set.rb
```diff
@@ -124,9 +124,8 @@ def assert_pod_templates_present(cm_name)
     end
 
     def assert_secret_present(secret_name, expected_data = nil, type: 'Opaque', managed: false)
```
KnVerey: same nit about the word "managed" being misleading now

test/integration/kubernetes_deploy_test.rb
```diff
@@ -340,6 +341,16 @@ def test_output_of_failed_unmanaged_pod
     ], in_order: true)
   end
 
+  def test_deployment_includes_ejson_secrets
```
KnVerey: What is this proving that `test_create_and_update_secrets_from_ejson` (in the other file) doesn't already cover?

test/integration/kubernetes_deploy_test.rb
```diff
@@ -557,24 +568,28 @@ def test_pruning_of_existing_managed_secrets_when_ejson_file_has_been_deleted
...
+    # Check ejson-keys is not deleted
+    ejson_cloud.assert_secret_present('ejson-keys')
   end
 
   def test_can_deploy_template_dir_with_only_secrets_ejson
     ejson_cloud = FixtureSetAssertions::EjsonCloud.new(@namespace)
     ejson_cloud.create_ejson_keys_secret
     assert_deploy_success(deploy_fixtures("ejson-cloud", subset: ["secrets.ejson"]))
     assert_logs_match_all([
-      "Deploying kubernetes secrets from secrets.ejson",
```
KnVerey: Can you add an assertion on the line you replaced this with either here or in one of the main ejson provisioning tests?

test/unit/kubernetes-deploy/kubectl_test.rb
```diff
@@ -168,6 +168,14 @@ def test_run_with_raise_if_not_found_does_not_raise_on_other_errors
     build_kubectl.run("get", "pod", "foobar", raise_if_not_found: true)
   end
 
+  def test_run_output_is_sensitive_squashes_debug_logs
+    stub_open3(%W(kubectl get pods --namespace=testn --context=testc --request-timeout=#{timeout}),
+      resp: "", err: "oops", success: false)
+    logger.level = 0
```
KnVerey: Interesting! I think we have a few tests that are manipulating the debug env var and rebuilding the logger (which prevents them from being parallelized) where they could be doing this instead. 🤔

CHANGELOG.md
```diff
@@ -1,5 +1,8 @@
 ## next
 
+*Features*
+- Support for deploying Secrets from templates ([#424](https://github.com/Shopify/kubernetes-deploy/pull/424)).
```
KnVerey: I think we also need a flashy "Breaking change" entry here because of the pruning implications and the critical nature of secrets. These are the cases I've thought of so far:
* If you previously used this gem to deploy secrets from EJSON and the first time commit you deploy using this version removes one or more of them, they will not be pruned.
  * We could actually avoid this one by making a transitional release, but I'm not sure it is justified, since the impact is _not_ deleting something, we're not at 1.0, and we have been maintaining this changelog strictly for folks for a long time.
* If you previously manually `kubectl apply`'d secrets that are not passed to kubernetes-deploy, your first deploy using this version is going to delete them
  * We could potentially make a transitional version log warnings about these, but I kinda doubt people would notice them. Users (including us at Shopify) are going to have to audit their cluster(s) for applied secrets before rolling out this version regardless. ⚠️ 
* If you previously passed secrets manifests to kubernetes-deploy (we would have ¯\\\_(ツ)_/¯ applied them) and they are no longer in the set you pass to the first deploy using this version, it will delete them

Can you think of any others? One case I think actually doesn't cause trouble is deploying with the new version and then rolling back to deploying with the old (as long as we keep our ejson secret annotation).

KnVerey:
Did a 🎩 (PRINT_LOGS=1) of some of the tests and have a couple additional thoughts based on the output below.

image

image

image

I think we should:
* Remove the first two lines in the box. Seems on the too-verbose side now that secret creation isn't a separate phase.
* Either suppress the "Discovering templates" when there are none, or even better, make the ejson secrets just show up in that list (maybe like `  - Secret/monitoring-token (from ejson)`)
* Search the provisioner for references to "Creation" that need to be changed to "Generation"

benlangfeld:
@DazWorrall I'm addressing some of those code review comments in https://github.com/Shopify/kubernetes-deploy/pull/425

CHANGELOG.md
DazWorrall: > Can you think of any others? 

I cannot 👍 

> One case I think actually doesn't cause trouble is deploying with the new version and then rolling back to deploying with the old (as long as we keep our ejson secret annotation).

Which is a solid reason to keep the annotation in itself!

lib/kubernetes-deploy/kubernetes_resource/secret.rb
DazWorrall: I ~stole~ took inspiration from `PodDisruptionBudget` here 🤷‍♂️ 

Agreed it would be good to pick a standard.

test/integration/kubernetes_deploy_test.rb
DazWorrall: Nothing really - I started with this test and `resources += secrets_from_ejson` and worked backwards, this can be removed now.

DazWorrall:
I think all your points have been addressed @KnVerey

CHANGELOG.md
```diff
@@ -1,5 +1,11 @@
 ## next
 
+*Features*
```
KnVerey: I think this needs to be rebased or it might incorrectly get merged into the 0.25.0 section

lib/kubernetes-deploy/kubernetes_resource/secret.rb
```diff
@@ -0,0 +1,23 @@
+# frozen_string_literal: true
+module KubernetesDeploy
+  class Secret < KubernetesResource
+    TIMEOUT = 30.seconds
+    KUBECTL_OUTPUT_IS_SENSITIVE = true
+
+    def status
+      exists? ? "Available" : "Unknown"
```
dturn: `Unknown` -> `Not Found`

lib/kubernetes-deploy/deploy_task.rb
```diff
@@ -240,20 +242,22 @@ def check_initial_status(resources)
...
         template_dir: @template_dir,
         logger: @logger,
-        prune: prune,
+        statsd_tags: @namespace_tags
       )
-      return unless ejson.secret_changes_required?
 
-      @logger.phase_heading("Deploying kubernetes secrets from #{EjsonSecretProvisioner::EJSON_SECRETS_FILE}")
-      ejson.run
+      unless ejson.resources.empty?
```
KnVerey: I'd suggest removing this logging entirely, since the inline resource list is showing which ones come from ejson

lib/kubernetes-deploy/kubernetes_resource/secret.rb
KnVerey: A PR we just merged standardized this on "Not Found" instead of "Unknown"

test/integration-serial/serial_deploy_test.rb
```diff
@@ -34,27 +34,14 @@ def test_create_and_update_secrets_from_ejson
...
+      %r{Secret\/catphotoscom\s+Available},
+      %r{Secret\/unused-secret\s+Available},
+      %r{Secret\/monitoring-token\s+Available},
     ])
 
     refute_logs_match(ejson_cloud.test_private_key)
     refute_logs_match(ejson_cloud.test_public_key)
     refute_logs_match(Base64.strict_encode64(ejson_cloud.catphotoscom_key_value))
-
-    # Update secrets
```
KnVerey: nit: the test name should have "and update" removed from it now

test/integration/kubernetes_deploy_test.rb
```diff
@@ -101,8 +101,9 @@ def test_pruning_works
       prune_matcher("statefulset", "apps", "stateful-busybox"),
       prune_matcher("job", "batch", "hello-job"),
       prune_matcher("poddisruptionbudget", "policy", "test"),
+      prune_matcher("secret", "", "hello-secret"),
     ] # not necessarily listed in this order
-    expected_msgs = [/Pruned 10 resources and successfully deployed 6 resources/]
+    expected_msgs = [/Pruned 11 resources and successfully deployed 6 resources/]
     expected_pruned.map do |resource|
       expected_msgs << /The following resources were pruned:.*#{resource}/
     end
```
KnVerey: I can't comment on the specific line, but there's a test we need to change now that secrets are supported:

```ruby
  def test_success_with_unrecognized_resource_type
    # Secrets are intentionally unsupported because they should not be committed to your repo
    secret = {
      "apiVersion" => "v1",
      "kind" => "Secret",
      "metadata" => { "name" => "test" },
      "data" => { "foo" => "YmFy" },
    }

    result = deploy_fixtures("hello-cloud", subset: ["configmap-data.yml"]) do |fixtures|
      fixtures["secret.yml"] = { "Secret" => secret }
    end
    assert_deploy_success(result)

    live_secret = kubeclient.get_secret("test", @namespace)
    assert_equal({ foo: "YmFy" }, live_secret["data"].to_h)
  end
```

Maybe `ReplicationController` would be a good type to use. I can't see us supporting it, since it was effectively replaced by Deployment a long time ago.

Also that test should have failed when this change was made... we should make sure the new version checks for the unsupported resource type logging.

DazWorrall:
> Adjust `DeployTask#apply_all` (and likely a couple things it calls) to be sensitivity-aware. Ideally if you aren't deploying secrets, we can still consider that output not-sensitive.

So squashing the debug output from kubectl in `apply_all` is easy enough, what I'm less sure about is what the right thing might be in `record_apply_failure`. My last commit attempts to remain useful by only logging the errors for which we would print out a useful hint and omitting any others, but I worry that might also leak sensitive content. I've tried to replicate an apply failure locally that might dump the raw secret output, but so far I've not succeeded. The safest option would be to squash all output, but as many deploys will contain secrets this would be a big UX degradation imo.

benlangfeld:
Is there anything left to be done on this besides https://github.com/Shopify/kubernetes-deploy/pull/433 ?

test/fixtures/unrecognized-type/configmap-data.yml
```diff
@@ -0,0 +1,10 @@
+apiVersion: v1
```
KnVerey: what is the point of including this resource?

README.md
```diff
@@ -579,14 +579,16 @@ The list of fully supported types is effectively the list of classes found in `l
...
-4. Add the a basic example of the type to the hello-cloud [fixture set](https://github.com/Shopify/kubernetes-deploy/tree/master/test/fixtures/hello-cloud) and appropriate assertions to `#assert_all_up` in [`hello_cloud.rb`](https://github.com/Shopify/kubernetes-deploy/blob/master/test/helpers/fixture_sets/hello_cloud.rb). This will get you coverage in several existing tests, such as `test_full_hello_cloud_set_deploy_succeeds`.
-5. Add tests for any edge cases you foresee.
+4. Add the new class to list of resources in
+   [`deploy_task.rb`](https://github.com/Shopify/kubernetes-deploy/blob/6a0dd662735bbcc0c0cf110d049a08a044a07dd1/lib/kubernetes-deploy/deploy_task.rb#L8)
+5. Add the a basic example of the type to the hello-cloud [fixture set](https://github.com/Shopify/kubernetes-deploy/tree/master/test/fixtures/hello-cloud) and appropriate assertions to `#assert_all_up` in [`hello_cloud.rb`](https://github.com/Shopify/kubernetes-deploy/blob/master/test/helpers/fixture_sets/hello_cloud.rb). This will get you coverage in several existing tests, such as `test_full_hello_cloud_set_deploy_succeeds`.
+6. Add tests for any edge cases you foresee.
```
KnVerey: Thanks for updating this! We should probably mention the prune whitelist as well, if you don't mind.

KnVerey:
>  I've tried to replicate an apply failure locally that might dump the raw secret output, but so far I've not succeeded. 

Me neither. I can get it to dump the secret's name, but not the content (dry run always catches errors there from what I can tell). 

> My last commit attempts to remain useful by only logging the errors for which we would print out a useful hint and omitting any others, but I worry that might also leak sensitive content. ... The safest option would be to squash all output, but as many deploys will contain secrets this would be a big UX degradation imo.

I've pointed out an additional line we need to adjust in my last review, but overall I think this is very reasonable. Given that most errors (including all errors in Secrets we could think of) will be caught at the validation stage, where we can easily tell whether or not a Secret was involved, the likelihood of us printing secret data despite these precautions strikes me as extremely low.

benlangfeld:
@DazWorrall I'm preparing fixes for the 3 test failures.

DazWorrall:
@KnVerey rebased and addressed your feedback. I had to use some unpleasant stubbing to get the tests done but I'm happy we've covered all the known failure states now.

I had to add a couple of serial tests - they proved flaky during CI when ran in parallel, but [running in isolation](https://buildkite.com/shopify/kubernetes-deploy-gem/builds/1470) they're fine so there's a race somewhere I don't understand.

> @DazWorrall test failures fixed at #435

Sorry I missed this @benlangfeld, I was iterating between meetings and not paying attention to my inbox :( I appreciate the thought!

benlangfeld:
> Sorry I missed this @benlangfeld, I was iterating between meetings and not paying attention to my inbox :( I appreciate the thought!

No problem. Just doing whatever I can to get this thing merged.

test/integration-serial/serial_deploy_test.rb
```diff
@@ -368,6 +368,46 @@ def test_deploying_crs_with_invalid_crd_conditions_fails
...
+    assert_deploy_failure(result)
+    refute_logs_match(/Kubectl err:/)
+    assert_logs_match_all([
+      "Command failed: apply -f",
+      /Invalid template: Deployment-web.*\.yml/,
+    ])
+    refute_logs_match("kind: Deployment") # content of the sensitive template
+  end
+
+  def test_apply_failure_with_sensitive_resources_hides_raw_output
```
KnVerey: It is the stubbing that makes the test above concurrency-unfriendly (`mocha` isn't threadsafe), but this one looks really normal and should be able to run in parallel. What flakes did you see?

test/integration-serial/serial_deploy_test.rb
```diff
@@ -368,6 +368,46 @@ def test_deploying_crs_with_invalid_crd_conditions_fails
...
+    logger.level = 0
+    # An invalid PATCH produces the kind of error we want to catch, so first create a valid secret:
+    assert_deploy_success(deploy_fixtures("hello-cloud", subset: %w(secret.yml)))
+    # Then try to PATCH an immutable field
+    result = deploy_fixtures("hello-cloud", subset: %w(secret.yml)) do |fixtures|
+      secret = fixtures["secret.yml"]["Secret"].first
+      secret["type"] = "something/invalid"
+    end
+    assert_deploy_failure(result)
+    refute_logs_match(/Kubectl err:/)
```
KnVerey: This assertion is too general. There are a handful of other kubectl commands that get run during the test, some of which might fail and be retried. Maybe that was a source of flakiness?

test/integration-serial/serial_deploy_test.rb
```diff
@@ -368,6 +368,46 @@ def test_deploying_crs_with_invalid_crd_conditions_fails
...
+      /Invalid template: Deployment-web.*\.yml/,
+    ])
+    refute_logs_match("kind: Deployment") # content of the sensitive template
+  end
+
+  def test_apply_failure_with_sensitive_resources_hides_raw_output
+    logger.level = 0
+    # An invalid PATCH produces the kind of error we want to catch, so first create a valid secret:
+    assert_deploy_success(deploy_fixtures("hello-cloud", subset: %w(secret.yml)))
+    # Then try to PATCH an immutable field
```
KnVerey: Smart! 👏

lib/kubernetes-deploy/deploy_task.rb
```diff
@@ -302,10 +302,10 @@ def split_templates(filename)
       raise FatalDeploymentError, "Failed to render and parse template"
     end
 
-    def record_invalid_template(err:, filename:, content:)
+    def record_invalid_template(err:, filename:, content: nil)
       debug_msg = ColorizedString.new("Invalid template: #{filename}\n").red
       debug_msg += "> Error message:\n#{FormattedLogger.indent_four(err)}"
```
KnVerey: shouldn't we also be suppressing (or replacing) the error itself? Just because it had a filename in it doesn't really tell us what else it contains

test/integration-serial/serial_deploy_test.rb
```diff
@@ -368,6 +368,46 @@ def test_deploying_crs_with_invalid_crd_conditions_fails
...
+    KubernetesDeploy::Deployment.any_instance.expects(:kubectl_output_is_sensitive?).returns(true).at_least_once
+    result = deploy_fixtures("hello-cloud", subset: ["web.yml.erb"]) do |fixtures|
+      bad_port_name = "http_test_is_really_long_and_invalid_chars"
+      svc = fixtures["web.yml.erb"]["Service"].first
+      svc["spec"]["ports"].first["targetPort"] = bad_port_name
+      deployment = fixtures["web.yml.erb"]["Deployment"].first
+      deployment["spec"]["template"]["spec"]["containers"].first["ports"].first["name"] = bad_port_name
+    end
+    assert_deploy_failure(result)
+    refute_logs_match(/Kubectl err:/)
```
KnVerey: Same comment as below--need to be more specific with this assertion or it will be flakey

benlangfeld:
@KnVerey If I prepare a PR for those last review comments, would this get merged today?

KnVerey:
> @KnVerey If I prepare a PR for those last review comments, would this get merged today?

I'm in a bunch of meetings right now, but I'll do my best

test/integration-serial/serial_deploy_test.rb
benlangfeld: Addressed in https://github.com/Shopify/kubernetes-deploy/pull/438/commits/9e13b30da411ada3cd28134179a390e7674e06c8

test/integration-serial/serial_deploy_test.rb
benlangfeld: Addressed in https://github.com/Shopify/kubernetes-deploy/pull/438/commits/9e13b30da411ada3cd28134179a390e7674e06c8

lib/kubernetes-deploy/deploy_task.rb
benlangfeld: Like this? https://github.com/Shopify/kubernetes-deploy/pull/438/commits/97a4fd30896fc86ab8545b80b0b33079754c126e

test/integration-serial/serial_deploy_test.rb
benlangfeld: These pass for me. https://github.com/Shopify/kubernetes-deploy/pull/438/commits/da575af16d5adb35b17e6279ed10b1b60a4f6a63

benlangfeld:
@KnVerey Could https://github.com/Shopify/kubernetes-deploy/pull/438 possibly get run through CI?

test/integration-serial/serial_deploy_test.rb
benlangfeld: If that commit fails in CI, might I propose that this be merged with the serial tests in place and we come back to this issue at lower priority? I don't think this should be a condition of merging this feature.

README.md
```diff
@@ -579,14 +579,17 @@ The list of fully supported types is effectively the list of classes found in `l
...
     * `sync` -- Gather the data you'll need to determine `deploy_succeeded?` and `deploy_failed?`. The superclass's implementation fetches the corresponding resource, parses it and stores it in `@instance_data`. You can define your own implementation if you need something else.
     * `deploy_succeeded?`
     * `deploy_failed?`
 3. Adjust the `TIMEOUT` constant to an appropriate value for this type.
-4. Add the a basic example of the type to the hello-cloud [fixture set](https://github.com/Shopify/kubernetes-deploy/tree/master/test/fixtures/hello-cloud) and appropriate assertions to `#assert_all_up` in [`hello_cloud.rb`](https://github.com/Shopify/kubernetes-deploy/blob/master/test/helpers/fixture_sets/hello_cloud.rb). This will get you coverage in several existing tests, such as `test_full_hello_cloud_set_deploy_succeeds`.
-5. Add tests for any edge cases you foresee.
+4. Add the new class to list of resources in
+   [`deploy_task.rb`](https://github.com/Shopify/kubernetes-deploy/blob/6a0dd662735bbcc0c0cf110d049a08a044a07dd1/lib/kubernetes-deploy/deploy_task.rb#L8)
```
dturn: any reason these don't point to master?

lib/kubernetes-deploy/deploy_task.rb
```diff
@@ -449,22 +452,35 @@ def log_pruning(kubectl_output)
...
     end
 
-    def record_apply_failure(err)
+    def record_apply_failure(err, resources: [])
       warn_msg = "WARNING: Any resources not mentioned in the error(s) below were likely created/updated. " \
         "You may wish to roll back this deploy."
       @logger.summary.add_paragraph(ColorizedString.new(warn_msg).yellow)
 
       unidentified_errors = []
+      sensitive_filenames = resources.select(&:kubectl_output_is_sensitive?).map { |r| File.basename(r.file_path) }
```
dturn: nit: `sensitive_filenames` -> `filenames_with_sensitive_content`

test/helpers/fixture_sets/hello_cloud.rb
```diff
@@ -21,6 +21,7 @@ def assert_all_up
       assert_stateful_set_up
       assert_job_up
       assert_network_policy_up
+      assert_secret_created
```
dturn: why `_created` and not `_present`

test/integration-serial/serial_deploy_test.rb
KnVerey: It's ok to let the one with stubbing run serially permanently, but I think this one belongs in the regular file. There's not any follow-up work to do--just trying to get them committed in the right place. 😄

README.md
benlangfeld: Addressed in https://github.com/Shopify/kubernetes-deploy/pull/438

lib/kubernetes-deploy/deploy_task.rb
benlangfeld: Addressed in https://github.com/Shopify/kubernetes-deploy/pull/438

test/helpers/fixture_sets/hello_cloud.rb
benlangfeld: Addressed in https://github.com/Shopify/kubernetes-deploy/pull/438

benlangfeld:
Looks like this is ready 💃

benlangfeld:
Thank you to everyone involved. This change is very important for my use case and I'm very grateful for it getting to master ❤️

KnVerey:
@benlangfeld if you need this immediately, can you use a git ref to reference it from master for a few days? I'd really like to include https://github.com/Shopify/kubernetes-deploy/pull/415 in the next release too; it's very nearly ready but Tim was off today.

benlangfeld:
Absolutely. I also need https://github.com/Shopify/kubernetes-deploy/pull/421 , but at least I can now rebase that on less of a moving target.

README.md
DazWorrall: > any reason these don't point to master?

I used a specific commit so the line numbers don't rot, not a huge deal.

PR #425, opened 2019-02-27
Secrets as resources review fixes

benlangfeld:
Addresses some code review comments from https://github.com/Shopify/kubernetes-deploy/pull/424

Issue #271, opened 2018-04-06
Revisit pruning whitelist


KnVerey:
# Bug report

As #259 points out, our pruning whitelist is incomplete, even for the resources that are otherwise fully supported by this gem. In my opinion we should also prune:

* pod templates
* resource quotas
* service accounts
* replica sets

**Expected behavior:** The above resources are pruned if the templates are removed

**Actual behavior:** They are not pruned

PR #426, opened 2019-02-27
Support label based namespace partitioning

### LINKED ISSUE #411: Support label-based sub-namespace resource management

benlangfeld:
Closes #411 

This is just a rebase of https://github.com/Shopify/kubernetes-deploy/pull/421 on https://github.com/Shopify/kubernetes-deploy/pull/425 (for https://github.com/Shopify/kubernetes-deploy/pull/424) to get early feedback and test results.

### DIFF
--- a/lib/kubernetes-deploy/deploy_task.rb
+++ b/lib/kubernetes-deploy/deploy_task.rb
@@ -41,6 +41,7 @@ require 'kubernetes-deploy/ejson_secret_provisioner'
 require 'kubernetes-deploy/renderer'
 require 'kubernetes-deploy/cluster_resource_discovery'
 require 'kubernetes-deploy/template_discovery'
+require 'kubernetes-deploy/utils'
 
 module KubernetesDeploy
   class DeployTask
@@ -102,7 +103,7 @@ module KubernetesDeploy
     end
 
     def initialize(namespace:, context:, current_sha:, template_dir:, logger:, kubectl_instance: nil, bindings: {},
-      max_watch_seconds: nil)
+      max_watch_seconds: nil, selector: nil)
       @namespace = namespace
       @namespace_tags = []
       @context = context
@@ -117,6 +118,7 @@ module KubernetesDeploy
         logger: @logger,
         bindings: bindings,
       )
+      @selector = selector
     end
 
     def run(*args)
@@ -223,7 +225,9 @@ module KubernetesDeploy
     measure_method(:predeploy_priority_resources, 'priority_resources.duration')
 
     def validate_resources(resources)
-      KubernetesDeploy::Concurrency.split_across_threads(resources) { |r| r.validate_definition(kubectl) }
+      KubernetesDeploy::Concurrency.split_across_threads(resources) do |r|
+        r.validate_definition(kubectl, selector: @selector)
+      end
       failed_resources = resources.select(&:validation_failed?)
       return unless failed_resources.present?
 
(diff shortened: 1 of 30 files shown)


lib/kubernetes-deploy/deploy_task.rb
```diff
@@ -428,7 +440,12 @@ def apply_all(resources, prune)
...
 
         if prune
-          command.push("--prune", "--all")
+          command.push("--prune")
+          if @selector
+            command.push("--selector", Utils.selector_to_string(@selector))
+          else
+            command.push("--all")
+          end
           prune_whitelist.each { |type| command.push("--prune-whitelist=#{type}") }
```
benlangfeld: @DazWorrall @KnVerey Is it reasonable to only apply the whitelist when a selector is not provided, in order to satisfy https://github.com/Shopify/kubernetes-deploy/issues/271? This approach gives the user full flexibility to prune any kind of resources they like, such that we don't have to keep updating the whitelist, it being only a safe default.

Alternatively, what if we begin to require a selector and otherwise prune all resources?

KnVerey: > We don't currently prune any `apps/v1/*` resources, for example, to which the likes of `Deployment` have been promoted.

We do actually. Group versions are better thought of as "views" on resources. If a `Deployment` exists, it can be accessed via any version supported by the API server in that cluster. In fact, you _must_ specific a group version whenever you `get` a resource--kubectl hides it from you, but try the following for example: 

```
kubectl get deployment.v1beta1.apps -v=6
GET https://localhost:6443/apis/apps/v1beta1/namespaces/katrina-test/deployments?limit=500 200 OK

kubectl get deployment.v1beta1.extensions -v=6
GET https://localhost:6443/apis/extensions/v1beta1/namespaces/katrina-test/deployments?limit=500 200 OK
```

In any case, so as long as we specify a group version that all of our supported k8s versions recognize (which we have CI coverage to prove), we will prune resources of that kind regardless of the group version used to create/update them.

> or making the prune whitelist configurable

I feel strongly that we should not expose the prune whitelist to end users. Pruning is technically an alpha feature that has some unfortunate gotchas (like the one I explained above where the whitelist requires but ignores GVs), and there's no such thing as pruning with no whitelist: kubectl has one (a much less sensible one) baked in. We should manage (i.e. prune) all kinds we support with few or maybe even no exceptions, and we should do so consistently whether we're managing the whole namespace or a selector-based sub-section of it.

> such that we don't have to keep updating the whitelist, it being only a safe default.

I don't actually think it's a big deal to manage this. We should only need to change it when we add support for new resources, and in the extremely rare cases where the group version we're referencing for a kind is deprecated. https://github.com/Shopify/kubernetes-deploy/issues/271 is really easy to fix and reflects omissions on our part earlier on in the development of this gem.

> Is it reasonable to only apply the whitelist when a selector is not provided

Overall I'm not sure I understand your reasoning with this proposal. Since pruning and selectors are both kubectl features, I (perhaps naively) don't anticipate problems with their interaction. The outstanding problem with secrets from EJSON is that users currently have no way to get them labelled correctly for use with sub-namespace deploys. Perhaps we can do that automatically; otherwise, we need to make it possible in the ejson file itself.

benlangfeld:
This isn't serving any purpose; closing.

PR #428, opened 2019-02-27
Stops modifying KUBECONFIG in integration tests

### LINKED ISSUE #52: Use --kubeconfig instead of env var in kubectl invocations

benlangfeld:
Fixes #52


benlangfeld:
I'll wait for https://github.com/Shopify/kubernetes-deploy/pull/429 to be merged before I do any more on this.

test/helpers/fixture_deploy_helper.rb
```diff
@@ -128,8 +129,8 @@ def write_fixtures_to_dir(fixtures, target_dir)
     end
   end
 
-  def build_kubectl(log_failure_by_default: true, timeout: '5s')
+  def build_kubectl(kubeconfig: ENV["KUBECTL"], log_failure_by_default: true, timeout: '5s')
```
timothysmith0609: `KUBECTL` -> `KUBECONFIG`?

lib/kubernetes-deploy/deploy_task.rb
```diff
@@ -190,6 +191,8 @@ def run!(verify_result: true, allow_protected_ns: false, prune: true)
 
     private
 
+    attr_accessor :kubeconfig
```
benlangfeld: This is a crappy API, fighting against the fact that `KubeclientBuilder` is a mixin when it should probably be a collaborating object. I'll prototype a fix as a second PR against this branch to see what you think.

benlangfeld:
@dturn @timothysmith0609 Review comments addressed.

lib/kubernetes-deploy/kubeclient_builder.rb
```diff
@@ -116,7 +116,26 @@ def _build_kubeclient(api_version:, context:, endpoint_path: nil)
...
     def config_files
       # Split the list by colon for Linux and Mac, and semicolon for Windows.
-      ENV.fetch("KUBECONFIG", "#{Dir.home}/.kube/config").split(/[:;]/).map!(&:strip).reject(&:empty?)
+      (kubeconfig || "#{Dir.home}/.kube/config").split(/[:;]/).map!(&:strip).reject(&:empty?)
+    end
+
+    def kubeconfig
+      # Cannot return a value to ensure that the default is used over env["KUBECONFIG"] when `kubeconfig: nil` is set
+      # in tests explicitly, for the case when ENV["KUBECONFIG"] is nil, since it may be set in the test environment
+    end
```
timothysmith0609: If this is only for tests I would rather mock it out in the tests themselves than add it here. @dturn wdyt?

benlangfeld: What do you think of https://github.com/powerhome/kubernetes-deploy/pull/1 ?

lib/kubernetes-deploy/deploy_task.rb
benlangfeld: Fixed in https://github.com/powerhome/kubernetes-deploy/pull/1

benlangfeld:
I would be very grateful if https://github.com/Shopify/kubernetes-deploy/pull/424 could be merged prior to this. It is infinitely more important.

lib/kubernetes-deploy/deploy_task.rb
timothysmith0609: the KubeclientBuilder-as-object makes sense to me and conceptually makes more sense to me. I'm ok with the change, wdyt @dturn ?

CHANGELOG.md
```diff
@@ -5,6 +5,9 @@
 - Defaults KUBECONFIG to `~/.kube/config` ([#429](https://github.com/Shopify/kubernetes-deploy/pull/429))
 - Uses `TASK_ID` environment variable as the `deployment_id` when rendering resource templates for better [Shipit](https://github.com/Shopify/shipit) integration. ([#430](https://github.com/Shopify/kubernetes-deploy/pull/430))
 
+*Other*
```
dturn: We usually omit non user facing changes from the chagelog.

benlangfeld:
> However, I don't think modifying `DeployTask#initialize` to take an extra param just to make testing easier it the right choice. Instead I think that validating the behavior of `validate_config_files` directly would be enough.

@dturn The tests have always been integration tests. In order to remove that parameter, it is necessary to make them unit tests on `KubeclientBuilder`. That is a regression in test coverage. Are you sure you want me to do that?

dturn:
@timothysmith0609 / @KnVerey thoughts on making these two serial integration tests into unit tests?

benlangfeld:
@dturn Is that the change you wanted?

lib/kubernetes-deploy/kubectl.rb
```diff
@@ -9,6 +9,7 @@ class ResourceNotFoundError < StandardError; end
 
     def initialize(namespace:, context:, logger:, log_failure_by_default:, default_timeout: DEFAULT_TIMEOUT,
       output_is_sensitive_default: false)
+      @kubeconfig = KubeclientBuilder.new.kubeconfig
```
dturn: Should `kubeconfig` be a class method on `KubeclientBuilder`? I can't decide.

benlangfeld: If we're not doing an integration test with `DeployTask`, there's no reason for it not to be. Let me know what you choose.

timothysmith0609: Since it doesn't seem to be impacted by the state of any given `KubeclientBuilder` object, I'm inclined to make it a class method

benlangfeld: Cool. I'm putting the energy I have available into https://github.com/Shopify/kubernetes-deploy/pull/439 because that's the one that is useful to me. Once that is merged, I'll look at these other PRs again.

test/integration-serial/serial_deploy_test.rb
```diff
@@ -44,70 +44,6 @@ def test_create_secrets_from_ejson
     refute_logs_match(Base64.strict_encode64(ejson_cloud.catphotoscom_key_value))
   end
 
-  # This can be run in parallel when we switch to --kubeconfig (https://github.com/Shopify/kubernetes-deploy/issues/52)
-  def test_unreachable_context
```
KnVerey: I really like the change to make `KubeclientBuilder` a class, and the other two tests removed here were about such granular behaviour that I don't feel strongly about their conversion to the unit level. However, I think we should keep this one and just accept that it will need to stay in the serial suite unless we decide to change the tasks' APIs in the future. We don't have any other tests that simulate this "Can't connect" behaviour.

KnVerey:
The following test is failing:

```

Failure:
--
  | KubectlTest#test_run_output_is_sensitive_squashes_debug_logs [/usr/src/app/lib/kubernetes-deploy/kubectl.rb:38]
  | Minitest::Assertion: unexpected invocation: Open3.capture3("kubectl", "get", "pods", "--kubeconfig=/root/.kube/config", "--namespace=testn", "--context=testc", "--request-timeout=15")
  | unsatisfied expectations:
  | - expected exactly once, not yet invoked: Open3.capture3("kubectl", "get", "pods", "--namespace=testn", "--context=testc", "--request-timeout=15")
  | - expected never, invoked once: Open3.capture3(any_parameters)
  | satisfied expectations:
  | - allowed any number of times, not yet invoked: #<Mock:0x56317a4e5370>.success?(any_parameters)


```

test/integration-serial/serial_deploy_test.rb
benlangfeld: So, this PR is not currently delivering any of the value it was supposed to. It was intended to make these integration tests executable in parallel, but instead it's ending up converting two into unit tests and leaving one running in serial.

Is any of this worth doing at all? What would you actually like to get from this?

KnVerey: Thanks for taking a step back and bringing this up. ❤️ You're totally right that we aren't really accomplishing our original goals. I think this is a case where a quick idea we had turned out differently than expected when actually implemented. Looking where this PR ended up, I see the following wins:
- Local testing indicates that converting the two tests to the unit level will save us ~1.7 seconds each (the one that needs to stay happens to be faster, at ~0.5 seconds).
- The refactor to make the `KubeclientBuilder` a class is an improvement in code quality. I can envision further improvements to make the class smarter in the future, now that it can hold state itself.

Although different from the original goals, these improvements seem worth merging to me.

benlangfeld: Done.

benlangfeld:
@KnVerey Should be good now. I'll leave squashing this to whoever merges it.

KnVerey:
The tests that failed are the flakey HPA tests that have been removed on master. The rest are 💚 .

PR #429, opened 2019-02-27
Optional REVISION & KUBECONFIG

### LINKED ISSUE #351: Make setting ENV["REVISION"] optional

benlangfeld:
Closes #351

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,8 +1,8 @@
 ## next
 
 *Enhancements*
-- Setting the REVISION environment variable is now optional
-- Defaults KUBECONFIG to `~/.kube/config`
+- Setting the REVISION environment variable is now optional ([#429](https://github.com/Shopify/kubernetes-deploy/pull/429))
+- Defaults KUBECONFIG to `~/.kube/config` ([#429](https://github.com/Shopify/kubernetes-deploy/pull/429))
 - Uses `TASK_ID` environment variable as the `deployment_id` when rendering resource templates for better [Shipit](https://github.com/Shopify/shipit) integration. ([#430](https://github.com/Shopify/kubernetes-deploy/pull/430))
 
 ## 0.25.0
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@
 
 *Enhancements*
 - Setting the REVISION environment variable is now optional
+- Defaults KUBECONFIG to `~/.kube/config`
 
 *Bug fixes*
 - Attempting to deploy from a directory that only contains `secrets.ejson` will no longer fail deploy ([#416](https://github.com/Shopify/kubernetes-deploy/pull/416))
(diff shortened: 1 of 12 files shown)


test/integration-serial/serial_deploy_test.rb
```diff
@@ -102,16 +102,6 @@ def test_multiple_configuration_files
     ], in_order: true)
     reset_logger
 
-    ENV['KUBECONFIG'] = nil
```
dturn: I think we should modify this test to show that `ENV['KUBECONFIG'] = nil` results is the new default behavior

benlangfeld: Just `assert_deploy_success(result)`, or something else?

dturn: Yeah, I think that would do it. Thought it's probably worth pulling out into its own test. Instead of being in this long one

benlangfeld: I was working on the assumption that all the other tests would fail if it were not set. Is that not true.

dturn: They will, but I'd still like to have an explicit assertion.

benlangfeld: Done. Could I get another review?

CHANGELOG.md
```diff
@@ -6,6 +6,10 @@
 *Enhancements*
 - Uses `TASK_ID` environment variable as the `deployment_id` when rendering resource templates for better [Shipit](https://github.com/Shopify/shipit) integration. ([#430](https://github.com/Shopify/kubernetes-deploy/pull/430))
 
+*Enhancements*
```
dturn: Since we released 0.25.0 this will need to be moved up

benlangfeld: Done.

benlangfeld: Looks like the TASK_ID change didn't actually make it in to the release, but it's in the changelog under that header. Probably needs moving up also.

dturn: Can you move that as well? Once that's done I'll merge this

benlangfeld: Done. Good to go.

benlangfeld:
@dturn Resovled. One more review? I assume you'd be happy to just do a squash merge. If you need any other history cleaning, please let me know ASAP.

test/integration/kubernetes_deploy_test.rb
```diff
@@ -1080,6 +1080,11 @@ def test_not_apply_resource_can_be_pruned
     ])
   end
 
+  def test_no_revision
```
dturn: 👍

PR #434, opened 2019-02-28
Make RenderTask to be loaded when kubernetes-deloy is required

jessie-JNing:
Make `KubernetesDeploy::RenderTask` to be loaded when kubernetes-deploy is required

**How is this accomplished?**
add 'kubernetes-deploy/render_task' to require

### DIFF
--- a/lib/kubernetes-deploy.rb
+++ b/lib/kubernetes-deploy.rb
@@ -21,6 +21,7 @@ require 'kubernetes-deploy/concurrency'
 require 'kubernetes-deploy/bindings_parser'
 require 'kubernetes-deploy/duration_parser'
 require 'kubernetes-deploy/resource_cache'
+require 'kubernetes-deploy/render_task'
 
 module KubernetesDeploy
   MIN_KUBE_VERSION = '1.10.0'


dturn:
> Make `KubernetesDeploy::RenderTask` available to public call

To be clear, this doesn't make it `public`. It just causes the file to be loaded when `require "kubernetes-deploy"` is run.

jessie-JNing:
> To be clear, this doesn't make it `public`. It just causes the file to be loaded when `require "kubernetes-deploy"` is run.

Ah, Got it. thanks for clarifying it.

KnVerey:
What's wrong with making people using this in gem form require the specific task they want? I'd actually be inclined to stop requiring the deploy task rather than to start requiring all of them.

timothysmith0609:
>Can you remote `require 'kubernetes-deploy/render_task'` from `exe/kubernetes-render`

Also from `render_task_test.rb` (`kubernetes-deploy` is included by `test_helper`).

dturn:
> What's wrong with making people using this in gem form require the specific task they want? I'd actually be inclined to stop requiring the deploy task rather than to start requiring all of them.

I don't think there is anything wrong with either approach. I think it's a better ux to not have to include specific parts of our gem. We've done a good job name spacing and the code base is rather small so I don't think there is a big downside to having a bit more code loaded.

KnVerey:
I would prefer to keep the modularity

jessie-JNing:
As we keep the modularity way, there's no need to have this change. I'll close it.
Thanks, everyone~

PR #435, opened 2019-03-01
Secrets as resources test fixes

benlangfeld:


benlangfeld:
@DazWorrall This should now fix all the failing tests from https://github.com/Shopify/kubernetes-deploy/pull/424

Issue #420, opened 2019-02-20
HPA tests are still flakey


dturn:
Our HPA integration tests are flakey. We've attempted to fix in various ways, most recently https://github.com/Shopify/kubernetes-deploy/pull/408.  However, it appears that the tests are rather sensitive to k8s versions see https://buildkite.com/shopify/kubernetes-deploy-gem/builds/1389#40f95e1a-2fee-47f2-bc0e-6dd12d0ae7c3/625-711 where 8/8 tests failed. Let's  move the HPA tests to unit tests. 

Bonus points if we can find a way to keep some amount of integration testing (maybe allowing it to fail in the normal way, but still catching other types of failures).

PR #440, opened 2019-03-02
Retry HPA test only

KnVerey:

Fix https://github.com/Shopify/kubernetes-deploy/issues/420 without deleting the tests after all.

Here's how long the two tests in question took to pass on 50 recent successful runs (values are seconds, tics are occurrences of that length):

```
62: -
63: -
64: -
78: -
79: -
80: --
92: ------
93: -
95: ---
96: -
97: -
107: -
108: -
109: -
110: -
122: ----
123: -
124: ---
125: ----
126: -
127: -
137: --
140: --
152: --
153: -
154: -
155: -
169: -
182: -
184: -
185: -
```

The mode is 92. The mean is 117. 130s covers ~75% of runs.

**How is this accomplished?**

* Add a module you can extend a minitest class with to add retries based on the test name. Kinda sketchy? Yes. If you're aware of a better way, LMK.
* Merge the two HPA tests into one, shorten its timeout to 130s, and let it retry twice. 
* Disable the retry-all logic at the buildkite level. No other tests are flakey (or if they are, we're not finding out because we're hiding it!), and this is making us waste tons of time and compute when someone pushes a legitimate failure.

**What could go wrong?**

This could still not be enough, in which case we can try to tweak this or give up and convert to unit tests like we planned to. I'll retry on Buildkite several times before merging to test the validity of this solution.

### DIFF
--- a/test/helpers/add_retries_test_helper.rb
+++ b/test/helpers/add_retries_test_helper.rb
@@ -1,0 +1,19 @@
+# frozen_string_literal: true
+
+# Usage: Extend a descendant of Minitest::Test with this module
+# and then add "__Nretries" to the end of the names of the tests you want to retry up to N times if they fail
+module AddRetriesTestHelper
+  def run_one_method(klass, method_name, reporter)
+    match_data = method_name.match(/__(?<retries>\d+)retries$/)
+    return super unless match_data && match_data[:retries]
+    retries = match_data[:retries].to_i
+
+    (retries + 1).times do |i|
+      print(" Retry#{i}/#{retries}:") if i > 0 && !ENV["VERBOSE"]
+      result = Minitest.run_one_method(self, method_name)
+      reporter.record(result)
+      print(" (Retry #{i}/#{retries})") if i > 0 && ENV["VERBOSE"]
+      break if result.passed?
+    end
+  end
+end
--- a/test/helpers/add_retries_test_helper.rb
+++ b/test/helpers/add_retries_test_helper.rb
@@ -3,17 +3,31 @@
 # Usage: Extend a descendant of Minitest::Test with this module
 # and then add "__Nretries" to the end of the names of the tests you want to retry up to N times if they fail
 module AddRetriesTestHelper
+  # ENV["VERBOSE"] example:
+  #  KubernetesDeployTest#test_hpa_can_be_successful_and_gets_pruned__2retries 0.58 = F
+  #  KubernetesDeployTest#test_hpa_can_be_successful_and_gets_pruned__2retries (Retry 1/2) 0.53 = F
+  #  KubernetesDeployTest#test_hpa_can_be_successful_and_gets_pruned__2retries (Retry 2/2) 0.52 = F
+
+  # !ENV["VERBOSE"] example:
+  #   F Retry1/2:F Retry2/2:F
   def run_one_method(klass, method_name, reporter)
     match_data = method_name.match(/__(?<retries>\d+)retries$/)
     return super unless match_data && match_data[:retries]
     retries = match_data[:retries].to_i
+    result = nil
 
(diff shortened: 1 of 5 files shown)


test/helpers/add_retries_test_helper.rb
```diff
@@ -0,0 +1,19 @@
...
+# Usage: Extend a descendant of Minitest::Test with this module
+# and then add "__Nretries" to the end of the names of the tests you want to retry up to N times if they fail
+module AddRetriesTestHelper
+  def run_one_method(klass, method_name, reporter)
+    match_data = method_name.match(/__(?<retries>\d+)retries$/)
+    return super unless match_data && match_data[:retries]
+    retries = match_data[:retries].to_i
+
+    (retries + 1).times do |i|
+      print(" Retry#{i}/#{retries}:") if i > 0 && !ENV["VERBOSE"]
```
KnVerey: This is how retries look locally:

test/helpers/add_retries_test_helper.rb
```diff
@@ -0,0 +1,19 @@
...
+      result = Minitest.run_one_method(self, method_name)
+      reporter.record(result)
+      print(" (Retry #{i}/#{retries})") if i > 0 && ENV["VERBOSE"]
```
KnVerey: This is how they look on CI:

image

test/helpers/add_retries_test_helper.rb
```diff
@@ -0,0 +1,47 @@
...
+      break if result.passed?
+    end
+
+    reporter.record(result)
+  end
+
+  # ENV["VERBOSE"] example:
+  #  KubernetesDeployTest#test_hpa_can_be_successful_and_gets_pruned__2retries 0.58 = F
+  #  KubernetesDeployTest#test_hpa_can_be_successful_and_gets_pruned__2retries (Retry 1/2) 0.53 = F
+  #  KubernetesDeployTest#test_hpa_can_be_successful_and_gets_pruned__2retries (Retry 2/2) 0.52 = F
```
KnVerey: This is really annoying to achieve, as you can see below. In case a visual might help, here are the parts we're needing to print to look consistent with the final result printed by the actual verbose reporter:

image

KnVerey:
The verbose printing is still wrong in CI... forgot to account for Parallelizable reporter so you get 💩 like this:

image

test/helpers/add_retries_test_helper.rb
```diff
@@ -0,0 +1,33 @@
...
+# and then add "__Nretries" to the end of the names of the tests you want to retry up to N times if they fail
+module AddRetriesTestHelper
+  # ENV["VERBOSE"] example:
+  #  KubernetesDeployTest#test_hpa_can_be_successful_and_gets_pruned__2retries 0.58 = F
+  #  KubernetesDeployTest#test_hpa_can_be_successful_and_gets_pruned__2retries (Retry 1/2) 0.53 = F
+  #  KubernetesDeployTest#test_hpa_can_be_successful_and_gets_pruned__2retries (Retry 2/2) 0.52 = F
+
+  # !ENV["VERBOSE"] example:
+  #   F Retry1/2:F Retry2/2:F
+  def run_one_method(klass, method_name, reporter)
```
KnVerey: Although a patch like this seems to be the way all the retry-related extensions for minitest are written (it is apparently factored separately from `Minitest.run_one_method` to allow this), doing so is fundamentally incompatible with running tests in parallel. Our `super` here is `Minitest::Parallel`, which has its own special version that just pushes tests to a queue--they aren't meant to actually run right here. That explains why all my CI runs were slowish even though HPA tests only failed once across all of them (the HPA tests were blocking). I would have to mess with `Minitest.run_one_method` instead, which is REALLY sketchy and doesn't have `reporter` access. 👎

PR #441, opened 2019-03-02
Log deployed resources in alphabetical order for ease of comparison

### LINKED ISSUE #261: Apply sorting to overview outputs
I do deployments containing more than 30 resources. With this amount, the output of the overviews is not easily comparable between runs so these don't serve a whole lot.

In my opinion, the following outputs should by alphabetically sorted:
* Discovering templates
* Checking initial resource statuses
* Deploying resources
* Result: Successful resources

benlangfeld:
Closes #261

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
 - Arguments to `--bindings` will now be deep merged. ([#419](https://github.com/Shopify/kubernetes-deploy/pull/419))
 - `kubernetes-deploy` and `kubernetes-render` now support reading templates from STDIN. ([#415](https://github.com/Shopify/kubernetes-deploy/pull/415))
 - Support for specifying a `--selector`, a label with which all deployed resources are expected to have, and by which prunable resources will be filtered. This permits sharing a namespace with resources managed by third-parties, including other kubernetes-deploy deployments. ([#439](https://github.com/Shopify/kubernetes-deploy/pull/439))
+- Lists of resources printed during deployments will now be sorted alphabetically. ([#441](https://github.com/Shopify/kubernetes-deploy/pull/441))
 
 *Features*
 
--- a/lib/kubernetes-deploy/deploy_task.rb
+++ b/lib/kubernetes-deploy/deploy_task.rb
@@ -282,7 +282,7 @@ module KubernetesDeploy
         @logger.warn("Detected non-namespaced #{'resource'.pluralize(global.count)} which will never be pruned:")
         global.each { |r| @logger.warn("  - #{r.id}") }
       end
-      resources
+      resources.sort
     end
     measure_method(:discover_resources)
 
@@ -369,6 +369,9 @@ module KubernetesDeploy
 
       if resources.length > 1
         @logger.info("Deploying resources:")
+        resources.each do |r|
+          @logger.info("- #{r.id} (#{r.pretty_timeout_type})")
+        end
       else
         resource = resources.first
         @logger.info("Deploying #{resource.id} (#{resource.pretty_timeout_type})")
@@ -381,7 +384,6 @@ module KubernetesDeploy
       applyables += individuals.select { |r| pruneable_types.include?(r.type) }
 
       individuals.each do |r|
-        @logger.info("- #{r.id} (#{r.pretty_timeout_type})") if resources.length > 1
         r.deploy_started_at = Time.now.utc
         case r.deploy_method
         when :replace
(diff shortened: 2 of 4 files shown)


benlangfeld:
> I dislike passing an additional arg to initialize and build that really only gets used by secrets created by ejson, It feels like we're already passing too many args to those functions.

Resolved

> Does logging all of the names at once, instead of near the processing of the resource make debugging harder if something raises an exception?

Do you have a suggestion for how to ensure these log messages are sorted while also emitting them immediately while iterating an unsorted set?

KnVerey:
> Do you have a suggestion for how to ensure these log messages are sorted while also emitting them immediately while iterating an unsorted set?

I agree that doesn't sound possible. I'm thinking that there's more value to the discovery step if it continues to print the list as it goes, and that it is very reasonable for a step called "discovery" to list things in the order encountered. I would be inclined to sort right at the end of discovery, which will still result in printing an ordered list very early on (at "checking initial statuses"). We also wouldn't need to make `KubernetesResource` aware of `source` that way.

benlangfeld:
@KnVerey I understood the contention to be around deployment, not discovery: https://github.com/Shopify/kubernetes-deploy/pull/441/files#diff-4c33bbb177fdf882c41d892f3bde9570L379

KnVerey:
That link doesn't point to anything in particular for me, but I'm really not concerned about consolidating the deploy list. In practice, there's only one type that (sadly) still cannot be applied in k8s 1.10+: `PodDisruptionBudget`. So the vast majority are going to appear in a single list no matter what. In fact, I think you've fixed a small UX issue where PDBs would get logged twice, once by the "individuals" section and once by the "apply_all" (which it is also part of, because it is prunable, but is a no-op). If that weren't compelling enough, consider that what could go wrong there is basically a single command (the create/apply). That command should very rarely fail (because we do up-front validations to avoid attempting it at all if it won't succeed), and when it does, we painstakingly parse its errors. I think the risk of user confusion is pretty low.

On the other hand, during template discovery, there are a lot more moving parts and a much higher chance of user error and resulting failure. That's why I think we should keep logging the resources as we go there, even though that means the first listing will stay out of order.

benlangfeld:
@KnVerey Done. With that restriction to the sorting, the implementation turned out really simple.

dturn:
2 test failure:

```

Failure:
--
  | KubernetesDeployTest#test_scale_existing_deployment_down_to_zero [/usr/src/app/test/integration/kubernetes_deploy_test.rb:758]
  | Minitest::Assertion: '(?-mix:Deployment\/web\s+0 replicas)' not found in the expected sequence in the following logs:

...

[INFO][2019-03-14 15:56:40 +0000]
--
  | [INFO][2019-03-14 15:56:40 +0000]	------------------------------------------Result: SUCCESS-------------------------------------------
  | [INFO][2019-03-14 15:56:40 +0000]	Successfully deployed 4 resources
  | [INFO][2019-03-14 15:56:40 +0000]
  | [INFO][2019-03-14 15:56:40 +0000]	Successful resources
  | [INFO][2019-03-14 15:56:40 +0000]	ConfigMap/hello-cloud-configmap-data              Available
  | [INFO][2019-03-14 15:56:40 +0000]	Deployment/web                                    0 replicas
  | [INFO][2019-03-14 15:56:40 +0000]	Ingress/web                                       Created
  | [INFO][2019-03-14 15:56:40 +0000]	Service/web                                       Doesn't require any endpoints

```

```

KubernetesDeployTest#test_can_deploy_template_dir_with_only_secrets_ejson [/usr/src/app/test/integration/kubernetes_deploy_test.rb:628]
--
  | Minitest::Assertion: '(?-mix:Secret\/monitoring-token\s+Available)' not found in the expected sequence in the following logs:

..

[INFO][2019-03-14 15:50:09 +0000]	------------------------------------------Result: SUCCESS-------------------------------------------
--
  | [INFO][2019-03-14 15:50:09 +0000]	Successfully deployed 3 resources
  | [INFO][2019-03-14 15:50:09 +0000]
  | [INFO][2019-03-14 15:50:09 +0000]	Successful resources
  | [INFO][2019-03-14 15:50:09 +0000]	Secret/catphotoscom                               Available
  | [INFO][2019-03-14 15:50:09 +0000]	Secret/monitoring-token                           Available
  | [INFO][2019-03-14 15:50:09 +0000]	Secret/unused-secret                              Available



```

Both look to be just an ordering issue

timothysmith0609:
Seems like a simple improvement that will be much appreciated (I'm definitely happy to see it). I'll hold off ✅ until the tests are passing but this looks good to me.

benlangfeld:
@dturn @timothysmith0609 @KnVerey Failing tests fixed. This should be good to go.

PR #442, opened 2019-03-02
Correct reference in CHANGELOG

benlangfeld:


dturn:
Thanks for fixing my mistake!

PR #445, opened 2019-03-07
Fix changelog

KnVerey:

Correct a caveat in the changelog that turned out to be wrong. The `EjsonSecretProvisioner` used `apply` all along, so removing a secret in the first deploy using the new version will work as expected.

https://github.com/Shopify/kubernetes-deploy/blob/a59c733aa98f81eb616df612c7ec30cf99b0ecb8/lib/kubernetes-deploy/ejson_secret_provisioner.rb#L122

Also be more helpful by suggesting how to check for affected secrets and how to unannotate them.

@Shopify/cloudx

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,7 @@
 
 *Features*
 
-- **[Breaking change]** Support for deploying Secrets from templates ([#424](https://github.com/Shopify/kubernetes-deploy/pull/424)). Because non-ejson secrets are now fully supported and therefore **subject to pruning like any other resource**:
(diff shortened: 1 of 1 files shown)


CHANGELOG.md
```diff
@@ -10,10 +10,10 @@
 
 *Features*
 
-- **[Breaking change]** Support for deploying Secrets from templates ([#424](https://github.com/Shopify/kubernetes-deploy/pull/424)).
-  * If you previously used this gem to deploy secrets from EJSON and the first time commit you deploy using this version removes one or more of them, they will not be pruned.
+- **[Breaking change]** Support for deploying Secrets from templates ([#424](https://github.com/Shopify/kubernetes-deploy/pull/424)). Because non-ejson secrets are now fully supported and therefore **subject to pruning like any other resource**:
```
dturn: ~Because~ Non-ejson secrets ...

PR #446, opened 2019-03-08
Bump ruby

dturn:
Move dev to a more recent ruby version.

Fix meet? for minikube v 0.34.1

**How is this accomplished?**
Update dev.yml

**What could go wrong?**
Not much, just having to  install stuff on the next `dev up`

### DIFF
--- a/dev.yml
+++ b/dev.yml
@@ -1,13 +1,13 @@
 ---
 name: kubernetes-deploy
 up:
-  - ruby: 2.3.3
+  - ruby: 2.6.1
   - bundler
   - homebrew:
     - Caskroom/cask/minikube
   - custom:
       name: Minikube Cluster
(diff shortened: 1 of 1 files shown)


dev.yml
```diff
@@ -1,13 +1,13 @@
 ---
 name: kubernetes-deploy
 up:
-  - ruby: 2.3.3
```
KnVerey: The reason for this version so that we're developing with the minimum supported in our gemspec: `spec.required_ruby_version = '>= 2.3.0'`. Are you saying minikube somehow depends on the ruby version? 😕

dturn: TIL . I'll revert the ruby change, just seemed odd that we were using such an old version of ruby.

eq -> ge change is totally separate.

KnVerey: Reasonable confusion. 😄 Can you add a comment?

PR #439, opened 2019-03-01
Support label-based sub-namespace resource management

benlangfeld:
Closes #411

Rebases https://github.com/Shopify/kubernetes-deploy/pull/421, addresses a review comment and resolves lint warnings.

PR #447, opened 2019-03-08
Never prune ejson-keys

timothysmith0609:
Component of [CloudPlatform#1273](https://github.com/Shopify/cloudplatform/issues/1273)

#424 introduced the ability to treat secrets like any other resource. _This includes pruning secrets with the `last-applied-configuration` annotation_. For historical reasons, a number of important secrets have had this annotation placed on them (our [old app mover script](https://github.com/Shopify/cloudplatform/blob/master/scripts/app-move-1-ns.sh) seems like the likely culprit, though older workflows may have introduced this as well).

A very important secret is `ejson-keys`, which contains the key pair for EJSON encryption/decryption. Since it is unlikely anyone would ever want this to be pruned, this PR introduces a safeguard against this case. During the initial confirmation phase, kubernetes-deploy will check to see if the `ejson-keys` secret exists and contains the `last-applied-configuration` annotation. If so, the logger will warn that the secret has been applied and will remove the annotation before proceeding through the deploy.

**What could go wrong?**
This should protect users of kubernetes-deploy from ever mistakenly pruning their ejson-keys secrets. However, users should not expect this to work with other secrets (which will be pruned indiscriminately).


lib/kubernetes-deploy/deploy_task.rb
```diff
@@ -541,6 +543,29 @@ def confirm_namespace_exists
       @logger.info("Namespace #{@namespace} found")
     end
 
+    # make sure to never prune the ejson-keys secret
+    def check_applied_ejson_keys_secret
```
dturn: The fact that EJSON specific details have leaked into this file seems like an abstraction issue.  Any reason we couldn't move this function into the ejson_secret_provisoner.rb and call it from https://github.com/Shopify/kubernetes-deploy/blob/master/lib/kubernetes-deploy/ejson_secret_provisioner.rb#L135

timothysmith0609: `EjsonSecretProvisioner` only does work if a `secrets.ejson` [file exists in the deploy directory](https://github.com/Shopify/kubernetes-deploy/blob/master/lib/kubernetes-deploy/ejson_secret_provisioner.rb#L42). For our purposes, this isn't necessarily/routinely the case. E.g we use `ejson-keys` to decrypt secrets coming from other files.

I agree we're missing an abstraction layer, but I don't think `EjsonSecretProvisioner` is where this code needs to go, since it's more of a pre-flight check of a well-known resource that, under normal circumstances, _should not be in a position to be pruned_, as opposed to the general workflow of provisioning secrets via EJSON (the provisioner should solely be a consumer of `ejson-keys`, not mutate it in any way).

@KnVerey and I had a conversation regarding the bloat of `deploy_task.rb` in general, and it's certainly due for a refactor. Said that, since this issue deals with the behaviour of a kubernetes resource _at deploy time_, it should be the deploy task's responsibility to ensure correct behaviour. This speaks more towards abstracting away our preflight checks into a more cohesive validation/preadmission framework than anything else, IMO.

lib/kubernetes-deploy/deploy_task.rb
```diff
@@ -541,6 +543,29 @@ def confirm_namespace_exists
...
+      raw_secret, st, err = nil
+      with_retries(2) do
+        raw_secret, err, st = kubectl.run("get", "-n", @namespace, "secret",
```
KnVerey: The `kubectl` instance injects the namespace for you by default, so no need to include in in the args.

lib/kubernetes-deploy/deploy_task.rb
```diff
@@ -541,6 +543,29 @@ def confirm_namespace_exists
...
+          "ejson-keys", "-oyaml", output_is_sensitive: true)
```
KnVerey: Even if this code isn't in the provisioner, we should probably reuse its `EjsonSecretProvisioner::EJSON_KEYS_SECRET` constant

lib/kubernetes-deploy/deploy_task.rb
```diff
@@ -541,6 +543,29 @@ def confirm_namespace_exists
...
+        st.success? || err.include?(KubernetesDeploy::Kubectl::NOT_FOUND_ERROR_TEXT)
+      end
+      if st.success?
+        secret = YAML.safe_load(raw_secret, [Date, Time])
+        return unless secret.dig("metadata", "annotations", LAST_APPLIED_ANNOTATION)
+        @logger.warn("Secret ejson-keys has annotation #{LAST_APPLIED_ANNOTATION}.")
+        @logger.warn("kubernetes-deploy will remove this annotation to ensure this secret does not get pruned.")
+        with_retries(2) do
+          _, err, st = kubectl.run("annotate", "secret", "ejson-keys", "#{LAST_APPLIED_ANNOTATION}-")
```
KnVerey: - same comment about using the constant for the name
- the output of this command likely isn't sensitive in practice, but since we know we are dealing with a secret, maybe we should use that option anyway out of an abundance of precaution

test/fixtures/secrets/ejson-keys.yml
```diff
@@ -0,0 +1,8 @@
+apiVersion: v1
```
KnVerey: Could we re-use the ejson-keys fixture set for this? Both to avoid extra fixture sets, and to show the intended scary scenario explicitly in the test:

```ruby
ejson_fixtures.create_ejson_keys_secret # 1. something creates it normally
# 2. do a kubectl apply that adds keys or something, like a clueless user would
# 3. run a deploy like your test currently does, showing that we fixed the user's mistake
```

lib/kubernetes-deploy/deploy_task.rb
KnVerey: The unique thing about this check is that it needs to run right here, not at the normal validation time, in order to preserve the ideal that our first few stages do not do any mutation. Therefore even when we have a proper shared validation abstraction, I kinda doubt this can be rolled into it. Maybe we should go ahead and create a new generic object now to encapsulate the concept of mutating preflight check or whatever we're calling it. After all this PR is the first time we do something like that.

test/integration/kubernetes_deploy_test.rb
```diff
@@ -907,6 +907,19 @@ def test_ejson_secrets_respects_no_prune_flag
...
+    assert_deploy_success(deploy_fixtures("secrets"))
+    assert_deploy_success(deploy_fixtures("hello-cloud", subset: %w(role.yml)))
+
+    refute_logs_match(%r{The following resources were pruned:.*secret( "|\/)ejson-keys})
+    assert_logs_match_all([
+      "Secret ejson-keys has annotation kubectl.kubernetes.io/last-applied-configuration.",
+      "kubernetes-deploy will remove this annotation to ensure this secret does not get pruned.",
+      "Successfully removed kubectl.kubernetes.io/last-applied-configuration from ejson-keys secret.",
+      %r{The following resources were pruned:.*secret( "|\/)my-secret},
+    ], in_order: true)
```
KnVerey: this test should also actually `get` the secret and confirm the annotation is not there anymore

timothysmith0609:
- Updated Changelog
- Updated README
- Added test to make sure we don't do anything when prune: false
- Better test coverage by actually inspecting the `ejson-keys` secret before and after a deploy (with/without prune)

CHANGELOG.md
```diff
@@ -15,6 +15,9 @@
   * If you previously manually `kubectl apply`'d secrets that are not passed to kubernetes-deploy, your first deploy using this version is going to delete them.
   * If you previously passed secrets manifests to kubernetes-deploy and they are no longer in the set you pass to the first deploy using this version, it will delete them.
 
+*Other*
+
+- The secret `ejson-keys` will never be pruned by kubernetes-deploy. Instead, it will remove the `kubectl.kubernetes.io/last-applied-configuration` annotation if it is present and log a warning. This _only_ affects the `ejson-keys` secret. All other secrets are subject to pruning as normal. ([#447](https://github.com/Shopify/kubernetes-deploy/pull/447))
```
KnVerey: Since pruning is first introduced in this version and is described above, this isn't really a change. Maybe this should be one of the `*` sub-points of the feature above?

benlangfeld:
Is this only necessary in some corner-case? I tested EJSON support on master and the `ejson-keys` secret was not pruned.

timothysmith0609:
>Is this only necessary in some corner-case? I tested EJSON support on master and the ejson-keys secret was not pruned.

This is indeed a corner case. In general, `ejson-keys` should never be `applied`. In practice, we have discovered that this is not always the case (some workflows in our tooling have `applied` it, for example). The conclusion our team came to is that `ejson-keys` is important enough to protect against accidental pruning.

For more context, `ejson-keys` is seeded into our namespaces and never explicitly deployed; that is to say, there's no `yml` for it in our deploy directories. It stands to reason that any `apply` ran against it is most likely erroneous. The risk of accidentally pruning keys outweighs the cost of implementing it here, IMO.

The bad UX component here is that, if `ejson-keys` is routinely deployed from a template directory, the warning will surface on every deploy. Said that, redeploying that secret on every deploy may be considered an anti-pattern, so not sure how concerned we should be with that.

lib/kubernetes-deploy/deploy_task.rb
```diff
@@ -53,14 +53,14 @@ class DeployTask
       kube-system
       kube-public
     )
+    LAST_APPLIED_ANNOTATION = "kubectl.kubernetes.io/last-applied-configuration"
```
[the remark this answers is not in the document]
KnVerey: Even though DeployTask is the only place that uses it, I agree with Danny that semantically it belongs to `KubernetesResource`

README.md
```diff
@@ -76,10 +76,16 @@ This repo also includes related tools for [running tasks](#kubernetes-run) and [
 * Ruby 2.3+
 * Your cluster must be running Kubernetes v1.10.0 or higher<sup>1</sup>
 * Each app must have a deploy directory containing its Kubernetes templates (see [Templates](#using-templates-and-variables))
+* You must remove the` kubectl.kubernetes.io/last-applied-configuration` annotation from any resources in the namespace that are not included in your deploy directory. This annotation is added automatically when you create resources with `kubectl apply`. `kubernetes-deploy` will prune any resources that have this annotation and are not in the deploy directory.<sup>2,3</sup>
+* Each app managed by `kubernetes-deploy` must have its own exclusive Kubernetes namespace.
```
KnVerey: I think you're adding this back accidentally because of the rebase. Please reconcile your changes to this file against  the ones made here: https://github.com/Shopify/kubernetes-deploy/pull/439/files#diff-04c6e90faac2675aa89e2176d2eec7d8

lib/kubernetes-deploy/deploy_task.rb
```diff
@@ -557,6 +558,25 @@ def confirm_namespace_exists
...
+        st.success? || err.include?(KubernetesDeploy::Kubectl::NOT_FOUND_ERROR_TEXT)
+      end
+      if st.success?
+        secret = YAML.safe_load(raw_secret, [Date, Time])
+        return unless secret.dig("metadata", "annotations", LAST_APPLIED_ANNOTATION)
+
+        raise EjsonPrunableError,
+          "Found #{LAST_APPLIED_ANNOTATION} annotation on #{EjsonSecretProvisioner::EJSON_KEYS_SECRET} secret. " \
+          "kubernetes-deploy will not continue since it is extremely unlikely that this secret should be pruned. " \
+          "contact your cluster administrator for further assistance."
```
KnVerey: Nit: capitalize this sentence

timothysmith0609: Passing as an error message downcases everything except the first char. Moving this to a logger entry and then raising an empty error

README.md
timothysmith0609: Fixed

lib/kubernetes-deploy/deploy_task.rb
KnVerey: Ok, but make sure it is still the summary section that is surfacing it

timothysmith0609: I added an error log to phase 1 when the secret is prunable and moved the original error back into the exception object. I simply removed the `contact your cluster administrator for further assistance` since it's pretty redundant.

benlangfeld:
Once this merges, would it be possible to cut a release?

timothysmith0609:
I've removed the annotations on all `ejson-keys` and `service-account-key` secrets. It remains to go through the remaining secrets to determine whether or not they are still required.

**Cluster services**
- [ ] datadog-secrets  
- [x] fluent . _appears to have been replaced by ejson secret `fluentd`_
- [x] mail-client-cert . _appears to have been replaced by ejson secret_`mail-client-tls-cert`_
- [x] splunk-client-cert  _appears to have been replaced by ejson secret `splunk-client-tls-cert`_
- [ ] star-shopifycloud
- [ ] start-shopifycloud-pkcs12

**api-docs-production**
- [ ] star-docs-shopify-io

**argus-server-production**
- [ ] star-shopifycloud

**backup-verification-production**
- [ ] aws-mysql-creds

**captain-hook-pint**
- [ ] captain-hook-ping-shopifycloud-com

**cloud-docs-production**
- [ ] docs-shopifycloud

**cloud-docs-staging**
- [ ] oauth2-proxy

**cloudbosses**
- [ ] webhook-server-secret

**deployer-staging**
- [ ] buildkite

**development-docs-production**
- [ ] star-shopifycloud

**expansion-tool-production**
- [ ] star-shopifycloud

**friendly-ghost-production**
- [ ] star-shopifycloud

**friendly-ghost-staging**
- [ ] star-shopifycloud

**genghis-production**
- [ ] star-shopifycloud
- [ ] star-shopifyio

**goto-cloud-production**
- [ ] database-url
- [ ] identity-config

**hsts-production**
- [ ] star-shopifycloud-com

**kube-system**
- [x] kubernetes-dashboard-certs (we don't prune on this ns)

**orchestrator**
- [ ] orchestrator

**pipa-production**
- [ ] default-token-fnn18

**pipa-test**
- [ ] default-token-5bdb4

**reportify-digest-production**
- [ ] cloudsql-proxy-service-account

**reportify-extractor-production**
- [ ] cloudsql-proxy-service-account

**reportify-query-production**
- [ ] cloudsql-proxy-service-account

**reportify-rollups-production**
- [ ] cloudsql-instance-credentials
- [ ] cloudsql-proxy-service-account

**reportify-streams-production**
- [ ] cloudsql-proxy-service-account

**script-errors-production**
- [ ] cloudsql-proxy-service-account

**services-db-production**
- [ ] cloudsql-proxy-service-account

**shipit-production**
- [ ] shipit-shopifycloud-com

**shopify-core-ingress**
- [ ] star-myshopifycom

**shopify-core-ingress-staging**
- [ ] star-myshopifycom

**tally-production**
- [ ] star-shopifycloud

lib/kubernetes-deploy/deploy_task.rb
```diff
@@ -557,6 +557,27 @@ def confirm_namespace_exists
       @logger.info("Namespace #{@namespace} found")
     end
 
+    # make sure to never prune the ejson-keys secret
+    def confirm_ejson_keys_not_prunable
+      raw_secret, st, err = nil
+      with_retries(2) do
```
dturn: I thought we were going to refactor EjsonSecretProvisioner#run_kubectl_json to do this?

README.md
```diff
@@ -76,9 +76,12 @@ This repo also includes related tools for [running tasks](#kubernetes-run) and [
...
 * Each app must have a deploy directory containing its Kubernetes templates (see [Templates](#using-templates-and-variables))
+* You must remove the` kubectl.kubernetes.io/last-applied-configuration` annotation from any resources in the namespace that you do not wish to be deleted and which are not included in your deploy directory. This annotation is added automatically when you create resources with `kubectl apply`. `kubernetes-deploy` will prune any resources that have this annotation and are not in the deploy directory.<sup>2,3</sup>
 
 <sup>1</sup> We run integration tests against these Kubernetes versions. You can find our
 offical compatibility chart below.
+<sup>2</sup> This requirement can be bypassed with the `--no-prune` option, but it is not recommended.
+<sup>3</sup> Note that kubernetes-deploy will fail if the `ejson-keys` secret is present and contains the `last-applied-configuration` annotation. This is done to prevent accidental pruning of your private keys. If you need to remove `ejson-keys`, contact your cluster administrator or use `kubectl`, if possible. This check is not made when the `--no-prune` flag is set.
```
KnVerey: I'm not sure if this is really relevant to the pre-requisites section. People reading this wouldn't have an ejson-keys secret yet.

README.md
```diff
@@ -76,9 +76,12 @@ This repo also includes related tools for [running tasks](#kubernetes-run) and [
 * Ruby 2.3+
 * Your cluster must be running Kubernetes v1.10.0 or higher<sup>1</sup>
 * Each app must have a deploy directory containing its Kubernetes templates (see [Templates](#using-templates-and-variables))
+* You must remove the` kubectl.kubernetes.io/last-applied-configuration` annotation from any resources in the namespace that you do not wish to be deleted and which are not included in your deploy directory. This annotation is added automatically when you create resources with `kubectl apply`. `kubernetes-deploy` will prune any resources that have this annotation and are not in the deploy directory.<sup>2,3</sup>
```
KnVerey: This was deleted in the PR that introduced the ability to manage a subset of the namespace using labels. I don't feel strongly about bringing it back, but it would need to be adjusted to account for the label-based strategy.

lib/kubernetes-deploy/deploy_task.rb
KnVerey: I also forgot we had discussed something around this. Now that you bring it up though, the idea I remember is to expose a `EjsonSecretProvisioner#ejson_keys` method, which would replace L562-567 and make it so that we only ever fetch that secret once. `run_kubectl_json` is just a private kubectl wrapper.

lib/kubernetes-deploy/deploy_task.rb
```diff
@@ -557,6 +557,27 @@ def confirm_namespace_exists
...
+        raw_secret, err, st = kubectl.run("get", "secret",
+          EjsonSecretProvisioner::EJSON_KEYS_SECRET, "-oyaml", output_is_sensitive: true)
+        st.success? || err.include?(KubernetesDeploy::Kubectl::NOT_FOUND_ERROR_TEXT)
+      end
+      if st.success?
+        secret = YAML.safe_load(raw_secret, [Date, Time])
+        return unless secret.dig("metadata", "annotations", KubernetesResource::LAST_APPLIED_ANNOTATION)
+        @logger.error("Secret #{EjsonSecretProvisioner::EJSON_KEYS_SECRET} will be pruned if deploy proceeds")
+
+        raise EjsonPrunableError.exception(
```
KnVerey: why `.exception` instead of `.new`? I had to look up what `.exception` does and it seems to have the exact same result.

I'm wondering if the pattern you want is actually `@logger.summary.add_paragraph(this_message)` followed by `raise FatalDeploymentError, "Configuration invalid"` like L358

lib/kubernetes-deploy/deploy_task.rb
```diff
@@ -557,6 +557,27 @@ def confirm_namespace_exists
...
+      raw_secret, st, err = nil
+      with_retries(2) do
+        raw_secret, err, st = kubectl.run("get", "secret",
+          EjsonSecretProvisioner::EJSON_KEYS_SECRET, "-oyaml", output_is_sensitive: true)
+        st.success? || err.include?(KubernetesDeploy::Kubectl::NOT_FOUND_ERROR_TEXT)
+      end
+      if st.success?
+        secret = YAML.safe_load(raw_secret, [Date, Time])
+        return unless secret.dig("metadata", "annotations", KubernetesResource::LAST_APPLIED_ANNOTATION)
+        @logger.error("Secret #{EjsonSecretProvisioner::EJSON_KEYS_SECRET} will be pruned if deploy proceeds")
```
KnVerey: > will be pruned if deploy proceeds

It never will though. Might this be unnecessarily alarming?

timothysmith0609: >It never will though. Might this be unnecessarily alarming?

We should have at least one logger entry inside the phase, otherwise it will go directly to the Failure section, which can seem confusing, even with the provided error message below. Maybe this just needs a slight reword, what about `Deploy cannot proceed because protected resource Secret/ejson-keys would be pruned"?

lib/kubernetes-deploy/ejson_secret_provisioner.rb
```diff
@@ -137,21 +137,16 @@ def decrypt_ejson(key_dir)
...
       encoded_private_key = secret["data"][public_key]
       unless encoded_private_key
         raise EjsonSecretError, "Private key for #{public_key} not found in #{EJSON_KEYS_SECRET} secret"
       end
 
       Base64.decode64(encoded_private_key)
     end
-
-    def run_kubectl_json(*args)
-      args += ["--output=json"]
```
timothysmith0609: There's not really a reason for this to be its own method here since the exception handling can be dealt with in the caller and `Kubectl#run` handles the rest.

lib/kubernetes-deploy/kubectl.rb
```diff
@@ -20,14 +20,15 @@ def initialize(namespace:, context:, logger:, log_failure_by_default:, default_t
       raise ArgumentError, "context is required" if context.blank?
     end
 
-    def run(*args, log_failure: nil, use_context: true, use_namespace: true, raise_if_not_found: false, attempts: 1,
-      output_is_sensitive: nil)
+    def run(*args, log_failure: nil, use_context: true, use_namespace: true, output: nil,
+      raise_if_not_found: false, attempts: 1, output_is_sensitive: nil)
```
timothysmith0609: Exposing `output` as an actual parameter seems preferable to expecting callers to manually add the `--output` flag, which feels leaky to me.

README.md
timothysmith0609: I'll just remove these comments. In the end it has more to do with kubernetes concepts than kubernetes-deploy so the information doesn't really need to be here

lib/kubernetes-deploy/deploy_task.rb
```diff
@@ -557,6 +557,23 @@ def confirm_namespace_exists
       @logger.info("Namespace #{@namespace} found")
     end
 
+    # make sure to never prune the ejson-keys secret
+    def confirm_ejson_keys_not_prunable
+      out, err, st = kubectl.run("get", "secret", EjsonSecretProvisioner::EJSON_KEYS_SECRET,
+        output: "json", output_is_sensitive: true)
+      return if err.include?(KubernetesDeploy::Kubectl::NOT_FOUND_ERROR_TEXT)
+
+      if st.success?
```
dturn: Is ignoring a failure that isn't not found actually the right thing to do?

lib/kubernetes-deploy/errors.rb
```diff
@@ -23,6 +23,8 @@ def initialize(name, context)
 
   class DeploymentTimeoutError < FatalDeploymentError; end
 
+  class EjsonPrunableError < FatalDeploymentError; end
```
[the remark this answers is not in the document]
timothysmith0609: The failure mode is related more to deploy task runtime than the provisioner in general, so this seemed appropriate. I'm fine either way so let me know if you would prefer it moved

lib/kubernetes-deploy/ejson_secret_provisioner.rb
```diff
@@ -137,21 +137,16 @@ def decrypt_ejson(key_dir)
     end
 
     def fetch_private_key_from_secret
-      secret = run_kubectl_json("get", "secret", EJSON_KEYS_SECRET)
+      out, err, st = @kubectl.run("get", "secret", EJSON_KEYS_SECRET, output: "json")
+      raise EjsonSecretError, err unless st.success?
+      result = JSON.parse(out)
+      secret = result.fetch('items', result)
```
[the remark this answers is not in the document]
timothysmith0609: This is redundant now. The original `run_kubectl_json` didn't know if it would be getting one or multiple secrets, so had to account for the latter case. Since we only ever get one secret at a time where it used to be called, we can remove it

lib/kubernetes-deploy/deploy_task.rb
```diff
@@ -557,6 +557,27 @@ def confirm_namespace_exists
       @logger.info("Namespace #{@namespace} found")
     end
 
+    # make sure to never prune the ejson-keys secret
+    def confirm_ejson_keys_not_prunable
+      out, err, st = kubectl.run("get", "secret", EjsonSecretProvisioner::EJSON_KEYS_SECRET,
```
KnVerey: Even after this refactor, we're still making this request (and parsing the resulting data) twice in the case where they do have ejson secrets--once here and once when the provisioner needs it. Can we make the provisioner expose it instead? That's what I thought we agreed to last week (though I forgot about it on my last review).

lib/kubernetes-deploy/deploy_task.rb
```diff
@@ -557,6 +557,27 @@ def confirm_namespace_exists
...
+          "Error running validation for Secret/#{EjsonSecretProvisioner::EJSON_KEYS_SECRET}: #{err}"
+      end
+
+      if st.success?
+        secret = JSON.parse(out)
+        return unless secret.dig("metadata", "annotations", KubernetesResource::LAST_APPLIED_ANNOTATION)
+        @logger.error("Deploy cannot proceed because protected resource " \
+          "Secret/#{EjsonSecretProvisioner::EJSON_KEYS_SECRET} would be pruned.")
+
+        raise EjsonPrunableError.exception("Found #{KubernetesResource::LAST_APPLIED_ANNOTATION} annotation on " \
```
KnVerey: I asked this previously but maybe you missed it: why `.exception`? It seems to do the exact same thing as `.new` but I had never seen it before and had to look it up.

lib/kubernetes-deploy/errors.rb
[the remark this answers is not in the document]
KnVerey: I think this is a fine place for it. TBH I'd be ok with reusing `TaskConfigurationError` for this. I'm not sure it's likely our users will want to catch this error specifically.

lib/kubernetes-deploy/ejson_secret_provisioner.rb
```diff
@@ -68,6 +68,14 @@ def private_key
       @private_key ||= fetch_private_key_from_secret
     end
 
+    def ejson_keys_secret
+      @ejson_keys_secret ||= begin
+        out, err, st = @kubectl.run("get", "secret", EJSON_KEYS_SECRET, output: "json")
+        out = JSON.parse(out) if st.success?
+        [out, err, st]
+      end
```
timothysmith0609: Since the error handling is slightly different when this is called, I'm compelled to return the `out, err, st` and let the caller sort it out. Said that, calling this method `ejson_keys_secret` might not be accurate, since it encompasses more information than just the secret itself. Maybe `kubectl_get_ejson_keys`? That's more explicit but exposes implementation details 🤷‍♂️

lib/kubernetes-deploy/ejson_secret_provisioner.rb
```diff
@@ -37,6 +37,14 @@ def resources
       @resources ||= build_secrets
     end
 
+    def ejson_keys_secret
+      @ejson_keys_secret ||= begin
+        out, err, st = @kubectl.run("get", "secret", EJSON_KEYS_SECRET, output: "json")
+        out = JSON.parse(out) if st.success?
+        [out, err, st]
```
timothysmith0609: Since the error handling is slightly different when this is called, I'm compelled to return the `out, err, st` and let the caller sort it out. Said that, calling this method `ejson_keys_secret` might not be accurate, since it encompasses more information than just the secret itself. Maybe `kubectl_get_ejson_keys`? That's more explicit but exposes implementation details 🤷‍♂️

lib/kubernetes-deploy/deploy_task.rb
```diff
@@ -557,6 +560,25 @@ def confirm_namespace_exists
...
 
+    # make sure to never prune the ejson-keys secret
+    def confirm_ejson_keys_not_prunable
+      secret, err, st = ejson_provisioner.ejson_keys_secret
+      unless st.success? || err.include?(KubernetesDeploy::Kubectl::NOT_FOUND_ERROR_TEXT)
+        raise FatalDeploymentError,
+          "Error running validation for Secret/#{EjsonSecretProvisioner::EJSON_KEYS_SECRET}: #{err}"
+      end
+
+      if st.success?
```
dturn: Is this check needed?

lib/kubernetes-deploy/ejson_secret_provisioner.rb
dturn: I think this is ok. A half thought out idea: return the secret if it's found, nil if it's not present, and raise an error for everything else, not sure if that's better.

lib/kubernetes-deploy/deploy_task.rb
timothysmith0609: Yes, since we might have a `NOT_FOUND` error on our hands instead of actual success.

lib/kubernetes-deploy/ejson_secret_provisioner.rb
KnVerey: We can have `Kubectl` raise if not found and ignore that specifically. Here's what I would suggest: https://github.com/Shopify/kubernetes-deploy/compare/dont_prune_ejson_keys...ejson_keys_suggestion. I put it in a commit you can cherry-pick if you like it, for expediency.

joekohlsdorf:
fyi: The behavior described in the PRs description is not what is happening, the discussion also doesn't explain this change:
```
During the initial confirmation phase, kubernetes-deploy will check to see if the ejson-keys secret exists and contains the last-applied-configuration annotation. If so, the logger will warn that the secret has been applied and will remove the annotation before proceeding through the deploy.
```

Instead the deployment just fails:
```
[ERROR][2019-06-05 20:38:33 +0000]	Deploy cannot proceed because protected resource Secret/ejson-keys would be pruned.
[INFO][2019-06-05 20:38:33 +0000]	
[INFO][2019-06-05 20:38:33 +0000]	------------------------------------------Result: FAILURE-------------------------------------------
[FATAL][2019-06-05 20:38:33 +0000]	Found kubectl.kubernetes.io/last-applied-configuration annotation on ejson-keys secret. kubernetes-deploy will not continue since it is extremely unlikely that this secret should be pruned.
```

The changelog should have included more information on how to correct this for ejson-secrets because this annotation gets set by any version of kubernetes-deploy pre 0.26.

PR #449, opened 2019-03-13
Convert HPA to unit tests

KnVerey:
* Remove the flakey tests. Fixes https://github.com/Shopify/kubernetes-deploy/issues/420.
* Remove the full-suite retries. We don't have any other known flakes (yay!) so these retries will now typically be repeating legit failures, wasting a lot of resources and giving us an excuse not to notice/fix any flakes that surface in the future. [EDIT: ironically, I saw a flake I hadn't seen in months on this PR. I still think we should disable the retries]
* Make the new unit tests have coverage of failure cases
* Fix a bug I noticed: if the deployment an HPA is based on was scaled to zero for some reason, the HPA intentionally leaves it alone. This should not make your deploy fail. I think we had misunderstood this condition when we wrote the original implementation.

For reference: [k8s HPA code](https://github.com/kubernetes/kubernetes/blob/a4e3a5cb52d2e36eff85e10b31e1d4d37afec137/pkg/controller/podautoscaler/horizontal.go)

**How is this accomplished?**
🔥 tests and write new ones, nothing unusual

**What could go wrong?**
If the HPA conditions we're looking for change, we won't notice because the unit tests hard code them. I don't think the risk is actually that significant, because HPA is stable and k8s schema change versioning is really strict. 

@Shopify/cloudx


test/unit/kubernetes-deploy/kubernetes_resource/horizontal_pod_autoscaler_test.rb
```diff
@@ -0,0 +1,135 @@
...
+      "status" => "True",
+      "type" => "ScalingActive",
+    }]
+    hpa = build_synced_hpa(build_hpa_template(conditions: conditions))
+    assert(hpa.deploy_succeeded?)
+    refute(hpa.deploy_failed?)
+    assert_equal("Configured", hpa.status)
+  end
+
+  def test_hpa_succeeds_if_scaling_is_explicitly_disabled
```
dturn: With the changes you made to the HPA class. I think we could combine this test and the pruning one as the `test_hpa_can_be_pruned` integration test if we sent the deployment to `0` replicas.

KnVerey: Good point! I gave this a try though, and seemingly the HPA controller is really slow. Even the `ScalingDisabled` condition reliably takes at least 30s to appear when I run the test. I'm not sure covering that condition is worth that much time 😞

dturn: Wonder if that 30 seconds is actually a hard coded value somewhere. But since the integration tests run in parallel, I'd trade the machine time for more fidelity.

KnVerey: It's more fidelity only on the edge case where the've got an HPA that is disabled. If it were on a primary success or failure case I'd agree, but not 30s for a ScalingDisabled test. It's run in parallel yes, but we can't control at which point in the run this particular test will start.

lib/kubernetes-deploy/kubernetes_resource/horizontal_pod_autoscaler.rb
```diff
@@ -2,16 +2,17 @@
 module KubernetesDeploy
   class HorizontalPodAutoscaler < KubernetesResource
     TIMEOUT = 3.minutes
-    RECOVERABLE_CONDITIONS = %w(ScalingDisabled FailedGet)
+    RECOVERABLE_CONDITION_PREFIX = "FailedGet"
 
     def deploy_succeeded?
-      scaling_active_condition["status"] == "True"
+      scaling_active_condition["status"] == "True" || scaling_disabled?
```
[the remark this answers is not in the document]
KnVerey: To me this is analogous to the situation where a service exists and is properly configured but has no back ends because the related deployment is scaled to zero. In both situations, the dependent resource is doing the right thing, but the right thing happens to be "disable". We do also already expose the situation to the user via the status string. I'm hesitant to add a warning about something that is actually expected behavior.

PR #454, opened 2019-03-26
Use raw rendered output

dturn:
Be able to detect invalid yaml. Specifically cases where the document separator `---` is missing in an erb loop. 

**How is this accomplished?**
Currently we load and dump yaml documents before outputing them. Unfortunately, when the parser encounters the same key multiple times its uses the last one. Instead output the raw rendered yaml so a tool like `yaml lint` can be used. We still load and dump for some level of syntax checking. 

**What could go wrong?**
The yaml output is different and that leads to issues. I still need to figure out what's up with test_render_task_with_partials_and_bindings

it seems like the yaml is getting rendered multiple times ...


test/fixtures/hello-cloud/unmanaged-pod-1.yml.erb
```diff
@@ -3,7 +3,7 @@ kind: Pod
 metadata:
   name: unmanaged-pod-1-<%= deployment_id %>
   annotations:
-    kubernetes-deploy.shopify.io/timeout-override: "60s"
+    kubernetes-deploy.shopify.io/timeout-override: 60s
```
dturn: Maybe I should leave this file alone and just fix the expected output in the tests?

test/integration/render_task_test.rb
```diff
@@ -35,7 +35,7 @@ def test_render_task_multiple_templates
     assert_render_success(render.run(mock_output_stream, ['configmap-data.yml', 'unmanaged-pod-1.yml.erb']))
 
     stdout_assertion do |output|
-      assert_equal output, <<~RENDERED
```
dturn: the first arg to [assert_equal](https://ruby-doc.org/stdlib-2.1.1/libdoc/test/unit/rdoc/Test/Unit/Assertions.html#method-i-assert_equal) is the expected value. I can revert this, but leads to harder to read test failures

lib/kubernetes-deploy/render_task.rb
```diff
@@ -69,8 +69,12 @@ def render_filename(filename, stream)
       file_content = File.read(File.join(@template_dir, filename))
       rendered_content = @renderer.render_template(filename, file_content)
       YAML.load_stream(rendered_content, "<rendered> #{filename}") do |doc|
-        stream.puts YAML.dump(doc)
+        YAML.dump(doc)
```
KnVerey: Why do we still need to do this? I think it is the loading that will catch syntax errors, and would expect `parse_stream` below to do so as effectively.

test/integration/render_task_test.rb
KnVerey: This is fine. I always get the expected/actual order backwards.

test/integration/render_task_test.rb
```diff
@@ -98,64 +96,58 @@ def test_render_task_with_partials_and_bindings
...
         apiVersion: v1
         kind: ConfigMap
         metadata:
           name: config-for-pod1
         data:
-          supports_partials: yep
-        ---
+          supports_partials: "yep"
+
+        # This is valid
```
KnVerey: Not stripping the comments is a nice side-effect

test/integration/render_task_test.rb
```diff
@@ -275,6 +267,28 @@ def test_render_valid_fixtures
     end
   end
 
+  def test_render_only_adds_initial_doc_seperator_when_missing
```
KnVerey: This test doesn't currently prove the "only" part

test/integration/render_task_test.rb
```diff
@@ -275,6 +267,28 @@ def test_render_valid_fixtures
...
+    fixture = 'no-doc-seperator.yml'
+    expected = "---\n# This doc has no yaml seperator\nkey1: foo\n"
+
+    assert_render_success(render.run(mock_output_stream, [fixture]))
+    stdout_assertion do |output|
+      assert_equal expected, output
+    end
+  end
+
+  def test_render_preserves_duplicate_keys_adds_doc_initial_seperator_when_needed
```
KnVerey: > _adds_doc_initial_seperator_when_needed

this fixture already has a doc separator, so this name doesn't seem right

test/fixtures/hello-cloud/unmanaged-pod-1.yml.erb
KnVerey: Doesn't really matter to me

test/fixtures/invalid-partials/duplicate-keys.yml
```diff
@@ -0,0 +1,3 @@
+---
```
KnVerey: Let's make these files `.erb`--theoretically the render task could be optimized to treat plain yamls differently (the renderer itself is IIRC)

Suggestion: make this fixture reflect the use case in the PR description: the duplicate keys being generated by an ERB loop without the appropriate separator placement

dturn:
Tests are fixed up. I apparently had a brain-fart last night when I was writing them.

jessie-JNing:
The code LGTM~
If there's a 🎩  that proves `yamllint` capture the duplicate keys, will be great!

dturn:
```
➜  kubernetes-deploy git:(render-raw) ✗ be exe/kubernetes-render  --template-dir test/fixtures/partials duplicate-keys.yml.erb | yamllint -           
[INFO][2019-03-27 11:55:10 -0700]	
[INFO][2019-03-27 11:55:10 -0700]	---------------------------------Phase 1: Initializing render task----------------------------------
[INFO][2019-03-27 11:55:10 -0700]	Validating configuration
[INFO][2019-03-27 11:55:10 -0700]	
[INFO][2019-03-27 11:55:10 -0700]	-----------------------------------Phase 2: Rendering template(s)-----------------------------------
[INFO][2019-03-27 11:55:10 -0700]	Rendering duplicate-keys.yml.erb ...
[INFO][2019-03-27 11:55:10 -0700]	Rendered duplicate-keys.yml.erb
[INFO][2019-03-27 11:55:10 -0700]	
[INFO][2019-03-27 11:55:10 -0700]	------------------------------------------Result: SUCCESS-------------------------------------------
[INFO][2019-03-27 11:55:10 -0700]	Successfully rendered 1 template(s)
[INFO][2019-03-27 11:55:10 -0700]	
stdin
  3:1       error    duplication of key "key1" in mapping  (key-duplicates)
  4:1       error    duplication of key "key1" in mapping  (key-duplicates)
```

PR #455, opened 2019-03-27
Bump version + revise changelog

timothysmith0609:
Bump version to 0.26.0

**How is this accomplished?**
Go through commits, make sure everything is accounted for in Changelog (ignored internal details such as test fixes, etc.)

**What could go wrong?**
Unwanted secret pruning may occur for users who have not audited their clusters for stray resources

cc @benlangfeld as I believe you're interested in some of the changes in this release

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,5 @@
+## next
+
 ## 0.26.0
 
 *Enhancements*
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -1,6 +1,7 @@
-## next
```
benlangfeld: When we do this, can we keep `## next` so that the next PR doesn't have to add it back in?

PR #460, opened 2019-04-02
Fix compatibility with kubectl 1.14

lautis:
Remove use of deprecated -a option. The —show-all option was deprecated in Kubernetes 1.10, and removed in 1.14.

The -a option was deprecated in kubectl 1.10. As 1.9 and older are not supported, this should not cause any issues.

https://github.com/kubernetes/kubernetes/pull/69255

PR #461, opened 2019-04-02
Add 1.14 to CI

dturn:
Support the latest k8s release.

**How is this accomplished?**
Adding it to ci

**What could go wrong?**
Not much, its just CI instructions

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 ## next
 
+*Enhancements*
+- Officially support Kubernetes 1.14 [#461](https://github.com/Shopify/kubernetes-deploy/pull/461)
+
 ## 0.26.4
 
 *Bug fixes*
--- a/.buildkite/pipeline.nightly.yml
+++ b/.buildkite/pipeline.nightly.yml
@@ -1,4 +1,11 @@
 steps:
+  - name: 'Run Test Suite (:kubernetes: 1.14-latest)'
+    command: bin/ci
+    agents:
+      queue: k8s-ci
+    env:
+      LOGGING_LEVEL: 4
+      KUBERNETES_VERSION: v1.14-latest
   - name: 'Run Test Suite (:kubernetes: 1.13-latest)'
     command: bin/ci
     agents:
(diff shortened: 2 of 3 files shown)


dturn:
This is unlikely to pass 1.14 CI w/o https://github.com/Shopify/kubernetes-deploy/pull/460

KnVerey:
True, but it is currently failing during setup, so looks like we have other work to do

dturn:
> True, but it is currently failing during setup, so looks like we have other work to do

Yep, was hoping it would just work, but not at all surprised.

benlangfeld:
Would you mind pasting the other failures here?

dturn:
> Would you mind pasting the other failures here?

#460 was merged. This PR is just to enable 1.14 runs on our internal CI. The failures aren't related to the code in this repo, we just haven't added support for 1.14 to our CI. I'm trying to find some time to look at this, but I don't think it's blocking anything, we just don't have as much coverage as we'd like for now.

RyanBrushett:
- Do we want to deprecate 1.10? I acted as though we did but, checking. I can also do it in a different PR if that's better.
- I'll add a changelog entry tomorrow.
CI will still fail until probably tomorrow. Will update here then.

dturn:
> Do we want to deprecate 1.10? I acted as though we did but, checking. I can also do it in a different PR if that's better.

I think we actually need to keep running 1.10 for a bit longer.

RyanBrushett:
I'll do that in a separate PR. Cleaning this up shortly.

CHANGELOG.md
```diff
@@ -1,5 +1,8 @@
 ## next
 
+*Enhancements*
+- Officially support Kubernetes 1.14 [#461](https://github.com/Shopify/kubernetes-deploy/pull/461)
```
KnVerey: This should also be mentioned in the version compatibility table in the readme

dturn: I think we only track EoL versions in that table `Last officially supported in gem version`. Though we don't have an entry for 1.9 :(

KnVerey: Ah, you're right, my mistake.

Issue #463, opened 2019-04-06
Multiple kubeconfig files broken


dwradcliffe:
# Bug report

Since https://github.com/Shopify/kubernetes-deploy/pull/428, if you have KUBECONFIG with multiple files, kubernetes-deploy does not work at all.

The `--kubeconfig` flag on kubectl only supports a single file, where the env var supports merging multiple files.

**Expected behavior:**
No error.

**Actual behavior:**

```
[INFO][2019-04-06 10:57:05 -0400]	------------------------------------Phase 1: Initializing deploy------------------------------------
[DEBUG][2019-04-06 10:57:05 -0400]	Running command (attempt 1): kubectl config get-contexts -o name --kubeconfig=/Users/davidradcliffe/.kube/config:/Users/davidradcliffe/.kube/config1:/Users/davidradcliffe/.kube/config2 --request-timeout=15
[DEBUG][2019-04-06 10:57:05 -0400]	Kubectl out:
[INFO][2019-04-06 10:57:05 -0400]
[INFO][2019-04-06 10:57:05 -0400]	------------------------------------------Result: FAILURE-------------------------------------------
[FATAL][2019-04-06 10:57:05 -0400]	Context foo is not available. valid contexts: []
```

**Version(s) affected:**
v0.26.2

### Steps to Reproduce

1. `export KUBECONFIG=~/.kube/config1:~/.kube/config2`
2. `kubernetes-deploy foo bar`

benlangfeld:
I propose that we revert #428.

https://kubernetes.io/docs/concepts/configuration/organize-cluster-access-kubeconfig/#merging-kubeconfig-files

@KnVerey thoughts?

dturn:
Yep, that's definitely the problem. I've got an idea for how to fix forward and I'd rather do that than revert #428.  I'll try and push up a branch tomorrow.

dwradcliffe:
👍 Fixed in `0.26.3`

PR #464, opened 2019-04-11
[Bug fix] Support multiple files in kubeconfig env var

dturn:
Right now having have multiple file via the `:` syntax in your kubeconfig env var doesn't work. It should.

fix: https://github.com/Shopify/kubernetes-deploy/issues/463

**How is this accomplished?**
Some refactoring to find the right file to pass to kubectl, apparently using the `--kubeconfig` flag only supports one file. 

**What could go wrong?**
Totally break using kubeconfig env var, but seems unlikely

### DIFF
--- a/lib/kubernetes-deploy/deploy_task.rb
+++ b/lib/kubernetes-deploy/deploy_task.rb
@@ -518,10 +518,7 @@ module KubernetesDeploy
     end
 
     def confirm_context_exists
-      found = kubectl.find_config_for_context(@context, raise_on_missing: false)
-      unless found
-        raise FatalDeploymentError, "Context #{@context} is not available."
-      end
+      kubectl.config_for_context(@context) # Raises if context not found
       confirm_cluster_reachable
       @logger.info("Context #{@context} found")
     end
--- a/lib/kubernetes-deploy/deploy_task.rb
+++ b/lib/kubernetes-deploy/deploy_task.rb
@@ -518,13 +518,9 @@ module KubernetesDeploy
     end
 
     def confirm_context_exists
-      out, err, st = kubectl.run("config", "get-contexts", "-o", "name",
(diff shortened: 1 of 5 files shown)


lib/kubernetes-deploy/deploy_task.rb
```diff
@@ -518,14 +518,7 @@ def find_bad_files_from_kubectl_output(line)
...
     def confirm_context_exists
-      out, err, st = kubectl.run("config", "get-contexts", "-o", "name",
-        use_namespace: false, use_context: false, log_failure: false)
-      available_contexts = out.split("\n")
-      if !st.success?
-        raise FatalDeploymentError, err
-      elsif !available_contexts.include?(@context)
-        raise FatalDeploymentError, "Context #{@context} is not available. Valid contexts: #{available_contexts}"
-      end
+      kubectl.config_for_context(@context) # Raises if context not found
```
dturn: The error message is now a bit less informative since it doesn't list all of the available contexts.

lib/kubernetes-deploy/kubectl.rb
```diff
@@ -21,15 +20,15 @@ def initialize(namespace:, context:, logger:, log_failure_by_default:, default_t
       raise ArgumentError, "context is required" if context.blank?
     end
 
-    def run(*args, log_failure: nil, use_context: true, use_namespace: true, output: nil,
+    def run(*args, log_failure: nil, use_namespace: true, output: nil,
```
dturn: This is technically an api change

dturn:
cc: @benlangfeld (for some reason I can't tag you as a reviewer)

KnVerey:
I merged https://github.com/Shopify/kubernetes-deploy/pull/468 instead

PR #465, opened 2019-04-16
Removes special treatment of GCP authentication

benlangfeld:

Simplify code related to GCP authentication.

**How is this accomplished?**

Upgrading to a version of kubeclient which deals with GCP auth natively: https://github.com/abonas/kubeclient/pull/394#issuecomment-469012306

**What could go wrong?**

GCP auth could break. I'm not in a position to test it, but our unit tests were passing before they were removed.


lib/kubernetes-deploy/kubeclient_builder/kube_config.rb
```diff
@@ -1,21 +1,12 @@
 # frozen_string_literal: true
 
-require 'googleauth'
```
timothysmith0609: We actually still need to import this as it's not included by default in Kubeclient (they didn't want to bloat the gem with a dependency that might not be used by all importers)

lib/kubernetes-deploy/kubeclient_builder/kube_config.rb
```diff
@@ -1,21 +1,12 @@
...
 module KubernetesDeploy
   class KubeclientBuilder
     class KubeConfig < Kubeclient::Config
       def self.read(filename)
         parsed = YAML.safe_load(File.read(filename), [Date, Time])
         new(parsed, File.dirname(filename))
```
timothysmith0609: This is almost a verbatim copy of the Kubeclient version. IIRC we put this here in order to ensure the created object would call our version of `fetch_user_options` instead of `Kubeclient::Config`'s. (`new(parsed, File.dirname(filename)) vs `Config.new(parsed, File.dirname(filename))`.

Now that we don't need that, I think we can get rid of this class entirely.

cc @dturn @KnVerey

test/unit/kubernetes-deploy/kube_config_test.rb
```diff
@@ -4,48 +4,13 @@
...
   def setup
     WebMock.disable_net_connect!
-    set_google_env_vars
   end
 
   def teardown
     WebMock.allow_net_connect!
   end
 
-  def test_auth_use_default_gcp_success
```
KnVerey: Is this test invalid now, or are you removing it because we no longer own the code that makes it work?

lib/kubernetes-deploy/kubeclient_builder/kube_config.rb
benlangfeld: I punted on removing this because https://github.com/Shopify/kubernetes-deploy/pull/464/files would have to put it back again.

test/unit/kubernetes-deploy/kube_config_test.rb
benlangfeld: I removed it because it’s no longer our responsibility, the same way we don’t have test coverage for OIDC for example.

lib/kubernetes-deploy/kubeclient_builder/kube_config.rb
timothysmith0609: Can we do this work upstream in Kubeclient?

benlangfeld: I removed the file entirely since #464 was closed.

KnVerey:
@timothysmith0609 can you please 👀 this?

timothysmith0609:
The diff makes sense to me, in the end we've just moved the same logic out to the kubeclient gem. Since the test run has passed I feel pretty confident.

PR #467, opened 2019-04-17
Do not add doc separator when render result is blank

KnVerey:

Fix a regression from https://github.com/Shopify/kubernetes-deploy/pull/454. It is causing the template validator for my Web PR to fail because a file entire content is rendered conditionally. Before that PR, the render result was an empty file, which I think makes sense. Now it instead contains an empty YAML doc. We should not be adding that header.

**How is this accomplished?**
Add a regression test and a presence check on the render result.

**What could go wrong?**
There's a reason I'm not thinking of why having an extra header could be a good thing?

@Shopify/cloudx

### DIFF
--- a/lib/kubernetes-deploy/render_task.rb
+++ b/lib/kubernetes-deploy/render_task.rb
@@ -70,7 +70,7 @@ module KubernetesDeploy
       rendered_content = @renderer.render_template(filename, file_content)
       implicit = true
       YAML.parse_stream(rendered_content, "<rendered> #{filename}") { |d| implicit = d.implicit }
-      stream.puts "---\n" if implicit
+      stream.puts "---\n" if implicit && rendered_content.present?
       stream.puts rendered_content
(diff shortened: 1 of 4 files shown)


lib/kubernetes-deploy/render_task.rb
```diff
@@ -70,7 +70,7 @@ def render_filename(filename, stream)
       rendered_content = @renderer.render_template(filename, file_content)
       implicit = true
       YAML.parse_stream(rendered_content, "<rendered> #{filename}") { |d| implicit = d.implicit }
-      stream.puts "---\n" if implicit
+      stream.puts "---\n" if implicit && rendered_content.present?
```
timothysmith0609: I'm tempted to log when we come across an empty template. Is that overkill?

KnVerey: I think that's a good idea

KnVerey: Added in the last commit:

image

PR #468, opened 2019-04-18
kubectl --kubeconfig cannot merge multiple files

KnVerey:

Revert only the problematic part of https://github.com/Shopify/kubernetes-deploy/pull/428, i.e. https://github.com/Shopify/kubernetes-deploy/pull/428/files#diff-f5478b2ac59d4babc30470442f3a8656. I still like the Class refactor, but `--kubeconfig` cannot merge multiple files, so if we want to use it we'll need to be more clever about what we pass through. Ironically the test we removed would have caught this regression. I'm sorry I did not catch that. I added it back verbatim (after confirming it does reproduce the bug on master).

Fixes https://github.com/Shopify/kubernetes-deploy/issues/463

**How is this accomplished?**
Making the `kubectl` class go back to depending on the env var. ~~As a bonus, make it proactively validate this dependency.~~ after CI showed this broke default location support, I removed it to keep this PR the smallest viable fix.

I think there is a nicer way to deal with this, but I wasn't able to put together a proposal as quickly as I wanted to. I think we should merge something really simple like this so we have time to discuss the right refactor to do. 

**What could go wrong?**

If somebody consuming this project as a gem switched away from the environment variable, they'll need to switch back (it is still effectively mandatory for CLI usage).

@Shopify/cloudx

### DIFF
--- a/lib/kubernetes-deploy/kubeclient_builder.rb
+++ b/lib/kubernetes-deploy/kubeclient_builder.rb
@@ -102,25 +102,25 @@ module KubernetesDeploy
         errors << "Kube config file name(s) not set in $KUBECONFIG"
       else
         @kubeconfig_files.each do |f|
-          unless File.file?(f)
-            errors << "Kube config not found at #{f}"
-          end
+          # If any files in the list are not valid, we can't be sure the merged context list is what the user intended
+          errors << "Kube config not found at #{f}" unless File.file?(f)
         end
       end
       errors
     end
 
-    private
-
-    def kubeclient_configs
-      @kubeclient_configs ||= @kubeconfig_files.map { |f| KubeConfig.read(f) if File.file?(f) }.compact
+    def validate_config_files!
+      errors = validate_config_files
+      raise TaskConfigurationError, errors.join(', ') if errors.present?
     end
 
-    def build_kubeclient(api_version:, context:, endpoint_path: nil)
-      # Find a context defined in kube conf files that matches the input context by name
-      raise TaskConfigurationError, "No kubeconfig files found in #{@kubeconfig_files}" if kubeclient_configs.empty?
-      config = kubeclient_configs.find { |c| c.contexts.include?(context) }
+    private
 
+    def build_kubeclient(api_version:, context:, endpoint_path: nil)
+      validate_config_files!
+      @kubeclient_configs ||= @kubeconfig_files.map { |f| KubeConfig.read(f) }
+      # Find a context defined in kube conf files that matches the input context by name
+      config = @kubeclient_configs.find { |c| c.contexts.include?(context) }
       raise ContextMissingError.new(context, @kubeconfig_files) unless config
 
       kube_context = config.context(context)
--- a/lib/kubernetes-deploy/kubeclient_builder.rb
+++ b/lib/kubernetes-deploy/kubeclient_builder.rb
@@ -7,18 +7,14 @@ module KubernetesDeploy
     class ContextMissingError < FatalDeploymentError
       def initialize(context_name, kubeconfig)
         super("`#{context_name}` context must be configured in your " \
-          "KUBECONFIG file(s) (#{kubeconfig}).")
+          "KUBECONFIG file(s) (#{kubeconfig.join(', ')}).")
       end
     end
 
-    def self.kubeconfig
-      new.kubeconfig
-    end
-
-    attr_reader :kubeconfig
-
     def initialize(kubeconfig: ENV["KUBECONFIG"])
-      @kubeconfig = kubeconfig || "#{Dir.home}/.kube/config"
+      files = kubeconfig || "#{Dir.home}/.kube/config"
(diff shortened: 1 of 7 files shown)


test/unit/kubernetes-deploy/kubeclient_builder_test.rb
```diff
@@ -28,13 +28,14 @@ def test_build_client_from_multiple_config_files
     Kubeclient::Client.any_instance.stubs(:discover)
     # Set KUBECONFIG to include multiple config files
     dummy_config = File.join(__dir__, '../../fixtures/kube-config/dummy_config.yml')
-    kubeconfig = "#{ENV['KUBECONFIG']}:#{dummy_config}"
```
KnVerey: This wasn't providing great multi-config coverage in our CI, which doesn't set the env var.

lib/kubernetes-deploy/kubectl.rb
```diff
@@ -27,7 +26,6 @@ def run(*args, log_failure: nil, use_context: true, use_namespace: true, output:
       output_is_sensitive = @output_is_sensitive_default if output_is_sensitive.nil?
 
       args = args.unshift("kubectl")
-      args.push("--kubeconfig=#{@kubeconfig}")
```
KnVerey: This is the revert (see https://github.com/Shopify/kubernetes-deploy/pull/428/files#diff-f5478b2ac59d4babc30470442f3a8656)

lib/kubernetes-deploy/kubeclient_builder.rb
```diff
@@ -121,12 +112,16 @@ def validate_config_files
...
 
+    def kubeclient_configs
+      @kubeclient_configs ||= @kubeconfig_files.map { |f| KubeConfig.read(f) if File.file?(f) }.compact
+    end
+
     def build_kubeclient(api_version:, context:, endpoint_path: nil)
       # Find a context defined in kube conf files that matches the input context by name
-      configs = config_files.map { |f| KubeConfig.read(f) }
-      config = configs.find { |c| c.contexts.include?(context) }
+      raise TaskConfigurationError, "No kubeconfig files found in #{@kubeconfig_files}" if kubeclient_configs.empty?
```
KnVerey: This is new. As CI pointed out, without it, if you try to build a kubeclient without validating the config files, and one of them is missing, we would 💥 with an `Errno::ENOENT`. My thinking with the way I addressed this is:
- We don't need to be re-reading the config files every time we build a new client, hence the new memoization above
- If for whatever reason the code did not run the validation before building a client, and we have both a valid config containing the desired context and a missing file, we should go ahead and do what the user wanted. If we can't do what the user wanted, we should throw our own more helpful error, not `Errno::ENOENT`.

lib/kubernetes-deploy/kubeclient_builder.rb
```diff
@@ -100,17 +96,12 @@ def build_networking_v1_kubeclient(context)
...
-
     def validate_config_files
       errors = []
-      if config_files.empty?
+      if @kubeconfig_files.empty?
         errors << "Kube config file name(s) not set in $KUBECONFIG"
       else
-        config_files.each do |f|
+        @kubeconfig_files.each do |f|
           unless File.file?(f)
```
timothysmith0609: This might be nicer as a one-liner
`errors << "Kube config not found at ${f}" unless File.file(f)`

lib/kubernetes-deploy/kubeclient_builder.rb
```diff
@@ -121,12 +112,16 @@ def validate_config_files
 
     private
 
+    def kubeclient_configs
+      @kubeclient_configs ||= @kubeconfig_files.map { |f| KubeConfig.read(f) if File.file?(f) }.compact
```
timothysmith0609: It might be worth rejecting kubeconfig entries that aren't files in `initialize` so that we don't need to make the check multiple times (here and in `validate_config_files`)

KnVerey: If we reject them up front though, any related errors we show will not match the list of files the user thinks they're giving us, which could be confusing. E.g. notice how the error I added below uses `@kubeconfig_files` directly (which is unfiltered). Putting the validation in the initializer would be one way around that problem, but I have found that making initializers raise (other than for malformed input) is unexpected and makes it more complicated for consumers to handle the errors as intended.

KnVerey: Actually, maybe we should just run the whole validation in `build_kubeclient` instead of proceeding as long as the files we can load contain the relevant context. Since contexts are merged across files, it is possible that we have a context with the correct name that is not the one the user expected to use.

timothysmith0609: >but I have found that making initializers raise (other than for malformed input) is unexpected and makes it more complicated for consumers to handle the errors as intended.

Fair point, I had the same idea in the back of my mind when I commented but decided to ask nonetheless.

🚢

PR #469, opened 2019-04-20
Check volumeBindingMode in StorageClass of PVC

redmcg:
If the StorageClass for a PVC has 'volumeBindingMode: WaitForFirstConsumer',
then it won't bind until after a Pod mounts it. But the PVC must be pre-deployed,
as the Pod requires it. So 'Pending' must be treated as a 'Success' state in this
scenario.

Automatic handling of the 'volumeBindingMode: WaitForFirstConsumer' condition in the StorageClass of a PVC resource.

**How is this accomplished?**
By determining the StorageClass of the PVC resource and finding the volumeBindingMode value. If it is WaitForFirstConsumer, then a PVC status of 'Pending' is treated as a success during deployment.

**What could go wrong?**
The code currently assumes the DefaultStorageClass admission plugin is turned on, and looks for a default StorageClass if one is not present in the PVC definition. There doesn't appear to be a way to determine if this admission plugin is turned on programmatically.

The code also checks for two other conditions and provides a warning:
1. if there are multiple Default StorageClasses defined; or
2. if the PVCs defined StorageClass is not found


lib/kubernetes-deploy/kubernetes_resource/persistent_volume_claim.rb
```diff
@@ -3,12 +3,66 @@ module KubernetesDeploy
...
+        else
+          # using default storage class
+          sc_name = default_sc[0]["metadata"]["name"]
+        end
+      else
+        # using storage class from pvc definition
+        sc_name = @definition.dig("spec", "storageClassName")
+
+        # a storageClassName of "" is an explicit way of saying you want a PV
+        # with no defined StorageClass. We won't look up a storage_class
```
dturn: Just to confirm this case is different than using the default storage class?

lib/kubernetes-deploy/kubernetes_resource/persistent_volume_claim.rb
```diff
@@ -3,12 +3,66 @@ module KubernetesDeploy
...
+        end
+
+        if default_sc.length != 1
+          warn_msg = "Multiple default StorageClasses found. If the DefaultStorageClass " \
+            "admission plugin is turned on, all PVC creation will fail."
+          @logger.warn(warn_msg) if default_sc.length > 1
+          return
+        else
+          # using default storage class
+          sc_name = default_sc[0]["metadata"]["name"]
```
dturn: Since we have the instance why not set @storage_class and return early.

lib/kubernetes-deploy/kubernetes_resource/persistent_volume_claim.rb
```diff
@@ -3,12 +3,66 @@ module KubernetesDeploy
...
     TIMEOUT = 5.minutes
 
+    def init_storage_class(cache)
+      @storage_class = {}
+
+      if @definition.dig("spec", "storageClassName").nil?
+        # if no storage class is defined we try to find the default one
+        # FIXME: This assumes the DefaultStorageClass admission plugin is turned on,
+        # need a way to determine this
+        is_default_class = "storageclass.beta.kubernetes.io/is-default-class"
```
dturn: What's the difference between the annotation you're using and `"storageclass.kubernetes.io/is-default-class"`?

lib/kubernetes-deploy/kubernetes_resource/persistent_volume_claim.rb
```diff
@@ -3,12 +3,66 @@ module KubernetesDeploy
...
+        # a storageClassName of "" is an explicit way of saying you want a PV
+        # with no defined StorageClass. We won't look up a storage_class
+        # if that is the case
+        return if sc_name == ""
+      end
+
+      @storage_class = cache.get_instance("StorageClass", sc_name)
+
+      # check the defined StorageClass exists
+      warn_msg = "StorageClass/#{sc_name} not found. This is required for #{id} to deploy."
```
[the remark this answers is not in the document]
redmcg: Re: StorageClass - I wasn't sure what your preference would be here (since StorageClass is not namespaced). But if the StorageClass is present in the deployment, deploying it first would resolve this issue. A change here probably deserves its own PR though.

Re: the warn message - I did want to place it earlier. The reason it's here is because I needed the cache object to examine the StorageClass (and I don't think I can get it earlier?).

In terms of severity - again I wasn't sure what your preference would be. I used `kubectl apply` as my guide here. It just deploys the PVC and leaves it in the PENDING state. But obviously its role is different to `kubernetes-deploy`s.

But I'm happy to change that `warn` to a failure. Would I do that with something similar to `raise_predates_deploy_error` in `pod.rb`?

lib/kubernetes-deploy/kubernetes_resource/persistent_volume_claim.rb
redmcg: Good pick-up - I hadn't even noticed I was using `beta` (I copy and pasted from my [minikube](https://github.com/kubernetes/minikube/blob/9a7b0b91aad20b7a938cd7ba0db203864227ff7f/deploy/addons/storageclass/storageclass.yaml) deployment). I just modified my `minikube` set-up to use `storageclass.kubernetes.io/is-default-class` instead and it worked all the same. So I assume they're the same thing; and this is just a hangover from dynamic provisioning's beta days. So I guess we need support for both?

lib/kubernetes-deploy/kubernetes_resource/persistent_volume_claim.rb
```diff
@@ -3,12 +3,66 @@ module KubernetesDeploy
...
+      @logger.warn(warn_msg) if @storage_class.blank?
+    end
+
+    def sync(cache)
+      super
+
+      # find the storage class (if we haven't already)
+      init_storage_class(cache) if @storage_class.nil?
```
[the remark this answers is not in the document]
redmcg: Apologies - I'm not super familiar with ruby. I think my latest [commit](https://github.com/Shopify/kubernetes-deploy/pull/469/commits/8e8664738ccbbc379f7a90f487b975651b99bb47) addresses this; although I've never used that `||=` operator before.

Although [policial](https://github.com/Shopify/kubernetes-deploy/pull/469/checks?check_run_id=110218543) is now complaining of a redundant return (but I think it's required).

lib/kubernetes-deploy/kubernetes_resource/persistent_volume_claim.rb
redmcg: Yeah - I believe this is used to explicitly opt out of using the default storage class (or any storage class):
https://kubernetes.io/docs/concepts/storage/persistent-volumes/#class-1

lib/kubernetes-deploy/kubernetes_resource/persistent_volume_claim.rb
redmcg: Again - another good pick-up. Not sure what I was thinking here. Fixed with:
https://github.com/Shopify/kubernetes-deploy/pull/469/commits/05f28f9adfc734855fa8f33c233d1967597171a8

redmcg:
I've finally got a test for the successful path - but I don't think it's usable.

The default provisioner in minikube doesn't seem to honour the `volumeBindingMode: WaitForFirstConsumer` declaration, so instead I used the [local](https://kubernetes.io/docs/concepts/storage/volumes/#local) volume type. This however requires `nodeAffinity`, which I set as `minikube`. But this means this test will fail on any other cluster.

At this point - I'm not really sure how to proceed. I'll keep thinking. But if you have any ideas - please let me know.

lib/kubernetes-deploy/kubernetes_resource/persistent_volume_claim.rb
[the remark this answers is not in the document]
dturn: I think the right thing to do here is set `storage_class` back to nil since it could be created during the deploy. Then add override the default `timeout_message` with this error.

dturn:
Can you commit the test, I usually find it easier to provide guidance when I can see the code.

redmcg:
> Can you commit the test, I usually find it easier to provide guidance when I can see the code.

I've done that now. Apologies - as (I feel) it's really ugly.

The `storage_class.yml` test runs first and creates:
- the `k8s-deploy-test` `StorageClass`
- a privileged pod to create the `/mnt/k8s-deploy-test` on the host; and
- finally the `PersistentVolume`

Then `pvc.yml` runs and tests two different PVCs (one with `storageClassName` defined and one without). The default `storageClassName` on `minikube` uses `volumeBindingMode: Immediately`.

So the problems are:
1. `nodeAffinity` on the PV (assumes a hostname of `minikube`);
2. the pod creating `/mnt/k8s-deploy-test` on the host (assumes a single host cluster); and
3. the default `storageClassName` test (assumes cluster has a default storage class)

redmcg:
I was thinking this would be easier to test if I didn't use the `local` PV plugin. But I need a plugin that supports `volumeBindingMode: WaitForFirstConsumer` - and as you can see - it's not a very long list:
https://kubernetes.io/docs/concepts/storage/storage-classes/#volume-binding-mode

dturn:
So we actually have code to help create PersistentVolumes https://github.com/Shopify/kubernetes-deploy/blob/master/test/helpers/test_provisioner.rb#L42 take a look at how its used. I suspect this will solve several of your problems. Though you'll probably have to modify it to accept a storge class. (Also please be sure to clean up global resources at the end of each test).

lib/kubernetes-deploy/kubernetes_resource/persistent_volume_claim.rb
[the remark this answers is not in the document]
redmcg: I finally understood what policial meant (I initially thought it meant the whole line was redundant). This is fixed in https://github.com/Shopify/kubernetes-deploy/pull/469/commits/cf90b451b704b8f6b3402d058e756bbd8664a831

redmcg:
> So we actually have code to help create PersistentVolumes https://github.com/Shopify/kubernetes-deploy/blob/master/test/helpers/test_provisioner.rb#L42 take a look at how its used.

In https://github.com/Shopify/kubernetes-deploy/pull/469/commits/4208c0b79cce69c86f22f52e975d4a9820a9fc18, I've started using the `prepare_pv` you linked too (although I had to make it public so I could use it within the test).

> Also please be sure to clean up global resources at the end of each test

I've changed the test to clean-up the PVC afterwards too.

I'm not sure how to use the kubeclient to create a StorageClass though - so this is still being created via yaml and is not being cleaned up afterwards.

I've also made it a single test, so it can be ran in parallel, thus I've moved the test out of `integration-serial`.

I was also able to address a couple of the assumptions I listed:
1. I changed nodeAffinity on the PV so it'll land on any node; and
2. I'm using an initContainer to create `/data/local0001` - so this will always take place on the same node as the pod claiming the PVC

The following assumptions still stand however:
3. the default storageClassName test (assumes cluster has a default storage class); and
4. the default class will have `volumeBindingMode: Immediately` (to test that scenario)

lib/kubernetes-deploy/kubernetes_resource/persistent_volume_claim.rb
redmcg: Added support for both annotations in https://github.com/Shopify/kubernetes-deploy/pull/469/commits/4d88029021d91fc02dbb483ebd80ffbafce00c3c

lib/kubernetes-deploy/kubernetes_resource/persistent_volume_claim.rb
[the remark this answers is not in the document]
redmcg: I think my initial understanding of how this works was incorrect. In doing more testing, I've found that not deploying the StorageClass at all can still allow success. It appears a default StorageClass definition is being used that has no provisioner and a 'volumeBindingMode: Immediate' (I should check the k8s code to confirm).

As a result, I have found that as long as the PV is present, the PVC will bind to it immediately and the deployment will succeed.

So I think the `timeout_message` should suggest that Dynamic Volume Provisioning may have failed because the StorageClass is not defined. But even if it succeeds - it may have used a volumeBindingMode of Immediate when then intention was to use WaitForFirstConsumer. Should I add a WARN for that somewhere? For example: "StorageClass not found; volumeBindingMode: Immediate will be used"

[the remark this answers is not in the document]
redmcg: In ebb89e78026b64a19b3cccf3a4e78e3d2d4d997d I replaced the warn with a timeout_msg

dturn:
@redmcg Sorry I haven't responded in a few days. If its ok with you, I've got an idea for a small refactor. I'll try and push it up today / tomorrow

redmcg:
Sounds good. No need to apologise.

redmcg:
Replaced by #479

PR #473, opened 2019-04-24
Update rubocop and re-lint files

KnVerey:
Update our rubocop compliance. I noticed that it was correcting untouched code when I ran it on my last PR.

**How is this accomplished?**
`gem install rubocop`, delete the styleguide cache, `rubocop -a` and then fix what couldn't be auto-corrected.

**What could go wrong?**
Committing unintentional changes, because I actually did this on my other branch and pulled it over from there.

@Shopify/cloudx

### DIFF
--- a/lib/kubernetes-deploy/kubernetes_resource.rb
+++ b/lib/kubernetes-deploy/kubernetes_resource.rb
@@ -53,7 +53,7 @@ module KubernetesDeploy
 
       def class_for_kind(kind)
         if KubernetesDeploy.const_defined?(kind)
-          KubernetesDeploy.const_get(kind) # rubocop:disable Sorbet/ConstantsFromStrings
+          KubernetesDeploy.const_get(kind)
         end
       rescue NameError
         nil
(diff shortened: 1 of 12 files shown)


lib/kubernetes-deploy/kubernetes_resource.rb
```diff
@@ -53,7 +53,7 @@ def build(namespace:, context:, definition:, logger:, statsd_tags:, crd: nil)
 
       def class_for_kind(kind)
         if KubernetesDeploy.const_defined?(kind)
-          KubernetesDeploy.const_get(kind) # rubocop:disable Sorbet/ConstantsFromStrings
```
KnVerey: Rubocop locally complains that disabling this cop has no effect, but then Policial fails without it. I added a general exclusion to the main rubocop config instead. We're not using Sorbet.

test/unit/kubernetes-deploy/kubernetes_resource_test.rb
```diff
@@ -399,7 +399,7 @@ def dummy_events(start_time)
         count: 3,
         last_seen: start_time + 3.seconds,
         reason: "FailedSync",
-        message: <<~STRING
+        message: <<~STRING,
```
dturn: Is this really where the `,` goes?

KnVerey: Yep--it's like when you chain a method on a heredoc, this is where it goes (`<<~STRING.strip` for example). I agree this comma looks stupid though. 😞

PR #476, opened 2019-04-25
Improve kubectl retries

KnVerey:

Make our kubectl use more resilient.

**How is this accomplished?**
* ~~Automatically retrying client timeouts specifically if the caller didn't ask us to retry errors in general.~~ EDIT: [Made `Kubectl#run` take a whitelist of error matchers that constrains retries, and used it to make validation calls retry client timeouts specifically.] We particularly saw the lack of client timeout retries affecting `apply --dry-run` validation in production, and we can't add generic retries on those (because validation failures should not be retried!). 
* Making our backoff delay more sophisticated. We believe our api servers are getting overloaded during our biggest deploys (because of the amount of churn they cause, not this gem's calls specifically). This is a small thing that I'm guessing could help, though I don't have evidence.

**What could go wrong?**
* The new retry backoff could slow down doomed deploys without actually helping the API server.
* An error in the retry logic could result in retrying things that are not safe to be retried.

@Shopify/cloudx

### DIFF
--- a/lib/kubernetes-deploy/deploy_task.rb
+++ b/lib/kubernetes-deploy/deploy_task.rb
@@ -552,13 +552,18 @@ module KubernetesDeploy
     end
 
     def confirm_namespace_exists
-      st, err = nil
-      with_retries(2) do
-        _, err, st = kubectl.run("get", "namespace", @namespace, use_namespace: false, log_failure: true)
-        st.success? || err.include?(KubernetesDeploy::Kubectl::NOT_FOUND_ERROR_TEXT)
-      end
-      raise FatalDeploymentError, "Failed to find namespace. #{err}" unless st.success?
+      raise FatalDeploymentError, "Failed to find namespace. #{err}" unless namespace_definition.present?
       @logger.info("Namespace #{@namespace} found")
+    end
+
+    def namespace_definition
+      @namespace_definition ||= begin
+        definition, _err, st = kubectl.run("get", "namespace", @namespace, use_namespace: false,
+          log_failure: true, raise_if_not_found: true, attempts: 3, output: 'json')
+        st.success? ? JSON.parse(definition, symbolize_names: true) : nil
+      end
+    rescue Kubectl::ResourceNotFoundError
+      nil
     end
 
     # make sure to never prune the ejson-keys secret
@@ -574,14 +579,8 @@ module KubernetesDeploy
     end
 
     def tags_from_namespace_labels
-      namespace_info = nil
-      with_retries(2) do
-        namespace_info, _, st = kubectl.run("get", "namespace", @namespace, "-o", "json", use_namespace: false,
-          log_failure: true)
-        st.success?
-      end
-      return [] if namespace_info.blank?
-      namespace_labels = JSON.parse(namespace_info, symbolize_names: true).fetch(:metadata, {}).fetch(:labels, {})
+      return [] if namespace_definition.blank?
+      namespace_labels = namespace_definition.fetch(:metadata, {}).fetch(:labels, {})
       namespace_labels.map { |key, value| "#{key}:#{value}" }
     end
 
--- a/lib/kubernetes-deploy/deploy_task.rb
+++ b/lib/kubernetes-deploy/deploy_task.rb
@@ -552,7 +552,7 @@ module KubernetesDeploy
     end
 
     def confirm_namespace_exists
-      raise FatalDeploymentError, "Failed to find namespace. #{err}" unless namespace_definition.present?
+      raise FatalDeploymentError, "Namespace #{@namespace} not found" unless namespace_definition.present?
       @logger.info("Namespace #{@namespace} found")
     end
 
--- a/lib/kubernetes-deploy/kubectl.rb
+++ b/lib/kubernetes-deploy/kubectl.rb
@@ -95,8 +95,11 @@ module KubernetesDeploy
     end
 
     def retriable_err?(err, retry_whitelist)
-      return retry_whitelist.any? { |retriable| err.match(retriable) } if retry_whitelist
-      !err.match(ERROR_MATCHERS[:not_found])
+      return !err.match(ERROR_MATCHERS[:not_found]) if retry_whitelist.nil?
+      retry_whitelist.any? do |retriable|
+        raise NotImplementedError, "No matcher defined for #{retriable.inspect}" unless ERROR_MATCHERS.key?(retriable)
+        err.match(ERROR_MATCHERS[retriable])
+      end
     end
 
     def extract_version_info_from_kubectl_response(response)
--- a/lib/kubernetes-deploy/kubectl.rb
+++ b/lib/kubernetes-deploy/kubectl.rb
@@ -2,9 +2,11 @@
 
 module KubernetesDeploy
   class Kubectl
+    ERROR_MATCHERS = {
+      not_found: /NotFound/,
+      client_timeout: /Client\.Timeout exceeded while awaiting headers/,
+    }
     DEFAULT_TIMEOUT = 15
-    NOT_FOUND_ERROR_TEXT = 'NotFound'
-    CLIENT_TIMEOUT_ERROR_MATCHER = /Client\.Timeout exceeded while awaiting headers/
     MAX_RETRY_DELAY = 16
 
     class ResourceNotFoundError < StandardError; end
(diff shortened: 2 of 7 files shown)


lib/kubernetes-deploy/kubectl.rb
```diff
@@ -24,40 +27,37 @@ def run(*args, log_failure: nil, use_context: true, use_namespace: true, output:
...
+        current_attempt += 1
       end
 
       [out.chomp, err.chomp, st]
     end
 
     def retry_delay(attempt)
-      attempt
+      # exponential backoff starting at 1s with cap at 16s, offset by up to 0.5s
+      [2**(attempt - 1), MAX_RETRY_DELAY].min - JITTER_RANGE.sample
```
KnVerey: What do you think of this? Too fancy? Not sophisticated enough?

lib/kubernetes-deploy/kubectl.rb
```diff
@@ -24,40 +27,37 @@ def run(*args, log_failure: nil, use_context: true, use_namespace: true, output:
       raise_if_not_found: false, attempts: 1, output_is_sensitive: nil)
```
KnVerey: I refactored this method a little to make it easier to follow (and accommodate adding a retry on timeout). There are two small functional changes that I think are good things:
- Previously, if log_failure && not_404 && debug_logging_enabled, the error would be printed twice, once at the warn level and once at the debug level. Now it will only be printed at the warn level.
- Previously, if log_failure && 404 && raise_on_404, the failed command would be logged. Now it won't be.

lib/kubernetes-deploy/kubectl.rb
dturn: `rand(0.5)` seems like a cleaner way to add jitter. 

16s seems like a lot of backoff, where did that value come from?

lib/kubernetes-deploy/kubectl.rb
```diff
@@ -24,40 +27,37 @@ def run(*args, log_failure: nil, use_context: true, use_namespace: true, output:
...
-          @logger.warn("The following command failed (attempt #{attempt}/#{attempts}): #{Shellwords.join(args)}")
+          escaped_cmd = Shellwords.join(cmd)
+          @logger.warn("The following command failed (attempt #{current_attempt}/#{attempts}): #{escaped_cmd}")
           @logger.warn(err) unless output_is_sensitive
-        end
-
-        if err.match(NOT_FOUND_ERROR_TEXT)
-          raise(ResourceNotFoundError, err) if raise_if_not_found
         else
           @logger.debug("Kubectl err: #{err}") unless output_is_sensitive
```
dturn: should we log something in debug mode?

lib/kubernetes-deploy/kubectl.rb
```diff
@@ -24,40 +27,37 @@ def run(*args, log_failure: nil, use_context: true, use_namespace: true, output:
...
-          StatsD.increment('kubectl.error', 1, tags: { context: @context, namespace: @namespace, cmd: args[1] })
         end
-        sleep(retry_delay(attempt)) unless attempt == attempts
+        StatsD.increment('kubectl.error', 1, tags: { context: @context, namespace: @namespace, cmd: cmd[1] })
+
+        attempts += 1 if err.match(CLIENT_TIMEOUT_ERROR_MATCHER) && retry_timeout?(current_attempt, attempts)
```
dturn: Modifying attempts doesn't seem like the right way to do this. If nothing else its going to generate confusing warning messages:
`The following command failed (attempt 1/1) ` to `"The following command failed (attempt 2/2)` .

lib/kubernetes-deploy/kubectl.rb
```diff
@@ -24,40 +27,37 @@ def run(*args, log_failure: nil, use_context: true, use_namespace: true, output:
...
+        sleep(retry_delay(current_attempt)) unless current_attempt == attempts
+        current_attempt += 1
       end
 
       [out.chomp, err.chomp, st]
     end
 
     def retry_delay(attempt)
```
dturn: Why isn't this private?

lib/kubernetes-deploy/kubectl.rb
KnVerey: What do you mean? That's what this line does

lib/kubernetes-deploy/kubectl.rb
KnVerey: True, that is confusing. I can move the conditional increment up to address that though. Fundamentally we are adding an additional attempt--how else would you do that?

lib/kubernetes-deploy/kubectl.rb
dturn: I think I wanted to say: Should we log something in debug mode _if the output_is_sensitive_ ? 
`@logger.debug("Kubectl err: #{output_is_sensitive  ? '<sensitive>': err}")`

lib/kubernetes-deploy/kubectl.rb
KnVerey: I'm guessing it was to make retries testable (some tests stub this to be zero). I even added unit tests on it in this PR. It is odd to have as part of the interface of this class, but I think those tests are valuable. Maybe this or even the whole retry logic should be some sort of utility object with generic tests of its own? I hesitate to block this PR on changing this though.

lib/kubernetes-deploy/kubectl.rb
KnVerey: Yeah, I could do that. The actual sleep offsets will be really excessively specific (e.g. `0.46836499688841826`) but I guess that doesn't matter. Or I could round it.

> 16s seems like a lot of backoff, where did that value come from?

It's just what seemed reasonable to me given that our default timeout for the calls themselves is 15s.

lib/kubernetes-deploy/kubectl.rb
KnVerey: Ah! sure, we can do that

lib/kubernetes-deploy/kubectl.rb
dturn: If it's not easy to change, I don't it needs to be changed it right now.

lib/kubernetes-deploy/kubectl.rb
dturn: It might be enough to change the log message to be clear this is an auto retry. But this interface isn't great since I can say i want `1` attempt but if there is a timeout I'll get a second.

KnVerey: Yeah, the more I think about this the more I agree with you. I'm also thinking that automatically retrying _client_ timeouts like this isn't necessarily a good idea. Since k8s is declarative most of the time it should be a no-op even to resubmit actions, but it wouldn't be ok for an unmanaged pod for example. I'll take another stab at this.

KnVerey:
@dturn pushed a different approach: Made `Kubectl#run` take a whitelist of error matchers that constrains retries, and used it to make validation calls retry client timeouts specifically.

lib/kubernetes-deploy/deploy_task.rb
```diff
@@ -552,15 +552,20 @@ def confirm_cluster_reachable
     end
 
     def confirm_namespace_exists
```
KnVerey: As part of this PR I looked over where we use `Kubectl#run`, and I noticed that this file requests namespaces twice and implements retries itself instead of using the built-in mechanism.

lib/kubernetes-deploy/kubectl.rb
```diff
@@ -21,43 +25,45 @@ def initialize(namespace:, context:, logger:, log_failure_by_default:, default_t
     end
 
     def run(*args, log_failure: nil, use_context: true, use_namespace: true, output: nil,
-      raise_if_not_found: false, attempts: 1, output_is_sensitive: nil)
+      raise_if_not_found: false, attempts: 1, output_is_sensitive: nil, retry_whitelist: nil)
```
KnVerey: maybe `attempts` should be `max_attempts` to be even clearer

lib/kubernetes-deploy/kubernetes_resource.rb
```diff
@@ -433,6 +423,19 @@ def validate_selector(selector)
       end
     end
 
+    def validate_spec_with_kubectl(kubectl)
+      command = ["create", "-f", file_path, "--dry-run", "--output=name"]
+      _, err, st = kubectl.run(*command, log_failure: false, output_is_sensitive: sensitive_template_content?,
+        retry_whitelist: [Kubectl::ERROR_MATCHERS[:client_timeout]], attempts: 3)
```
dturn: It feels strange to have to pass a value of a hash. I think `retry_whitelist: [:client_timeout]` or `[Kubectl::ERROR_MATCHERS::CLIENT_TIMEOUT]` seem more natural. But its a minor thing.

KnVerey: Good idea, I'll make the caller pass a symbol and the class will internally look it up in `ERROR_MATCHERS`

PR #479, opened 2019-05-01
Check volume binding mode

dturn:
If the StorageClass for a PVC has 'volumeBindingMode: WaitForFirstConsumer',
then it won't bind until after a Pod mounts it. But the PVC must be pre-deployed,
as the Pod requires it. So 'Pending' must be treated as a 'Success' state in this
scenario.

What are you trying to accomplish with this PR?
Automatic handling of the 'volumeBindingMode: WaitForFirstConsumer' condition in the StorageClass of a PVC resource.

How is this accomplished?
By determining the StorageClass of the PVC resource and finding the volumeBindingMode value. If it is WaitForFirstConsumer, then a PVC status of 'Pending' is treated as a success during deployment.

What could go wrong?
The code currently assumes the DefaultStorageClass admission plugin is turned on, and looks for a default StorageClass if one is not present in the PVC definition. There doesn't appear to be a way to determine if this admission plugin is turned on programmatically.

The code also checks for two other conditions and provides a warning:

if there are multiple Default StorageClasses defined; or
if the PVCs defined StorageClass is not found

@redmcg . When I pushed it went to origin instead of your repo. This still isn't done, needs better error messages, a test or two to catch some of the failure cases, and the test needs to clean up the storage class since its not namespaced. But thought I'd get your eyes early


redmcg:
This looks much much better. My only concern was with using `hostPath` instead of `local` in the PV - as when reading this:
https://kubernetes.io/docs/concepts/storage/storage-classes/#volume-binding-mode

it sounded like `volumeBindingMode: WaitForFirstConsumer` wasn't supported by `hostPath`. But I did a couple of tests and it seems to be supported. So we can get rid of that initContainer too.

How do you want to proceed with the PR? Am I able to contribute to this branch? And should I delete my own PR?

dturn:
> How do you want to proceed with the PR? Am I able to contribute to this branch? And should I delete my own PR?

If its ok with you, using this branch is easier for us (makes running CI simpler). You should be able to push to this PR, just pull my changes into your branch and we'll make it work. It would be great if you could make the test changes you suggested!

redmcg:
I couldn't push directly to this branch - I get:
`Permission to Shopify/kubernetes-deploy.git denied to redmcg.`

But I can create a new pull request against it. That does mean you'll have a PR against a PR though. I'll try that - but will of course close it if you prefer a different approach.

lib/kubernetes-deploy/kubernetes_resource/persistent_volume_claim.rb
```diff
@@ -3,16 +3,76 @@ module KubernetesDeploy
...
-      status == "Bound"
+      return false if deploy_failed?
+      return true if status == "Bound"
+
+      # if the StorageClass has volumeBindingMode: WaitForFirstConsumer,
+      # it won't bind until after a Pod mounts it. But it must be pre-deployed,
+      # as the Pod requires it. So 'Pending' must be treated as a 'Success' state
+      if storage_class&.volume_binding_mode == "WaitForFirstConsumer"
+        status == "Pending" || status == "Bound"
+      end
```
RyanBrushett: Should this explicitly return either true or false if it gets to this point?

lib/kubernetes-deploy/kubernetes_resource/persistent_volume_claim.rb
```diff
@@ -3,16 +3,76 @@ module KubernetesDeploy
...
 
     def deploy_failed?
-      status == "Lost"
+      status == "Lost" || failure_message.present?
+    end
+
+    def failure_message
+      if storage_class_name.nil? && @storage_classes.count(&:default?) > 1
+        "PVC has no StorageClass specified and there are multiple StorageClasses" \
+        "with annoated as default. This is an invalid cluster configuration."
```
RyanBrushett: Don't need the word `with`
```suggestion
        "annoated as default. This is an invalid cluster configuration."
```

lib/kubernetes-deploy/kubernetes_resource/persistent_volume_claim.rb
```diff
@@ -3,16 +3,76 @@ module KubernetesDeploy
...
     end
 
     def deploy_failed?
-      status == "Lost"
+      status == "Lost" || failure_message.present?
+    end
+
+    def failure_message
+      if storage_class_name.nil? && @storage_classes.count(&:default?) > 1
+        "PVC has no StorageClass specified and there are multiple StorageClasses" \
```
RyanBrushett: Needs a space.

```suggestion
        "PVC has no StorageClass specified and there are multiple StorageClasses " \
```

lib/kubernetes-deploy/kubernetes_resource/persistent_volume_claim.rb
```diff
@@ -3,16 +3,76 @@ module KubernetesDeploy
...
+      @storage_classes = cache.get_all("StorageClass").map { |sc| StorageClass.new(sc) }
+    end
+
     def status
       exists? ? @instance_data["status"]["phase"] : "Not Found"
     end
 
     def deploy_succeeded?
-      status == "Bound"
+      return false if deploy_failed?
```
timothysmith0609: Do we need to explicitly invoke `deploy_failed?` inside `deploy_succeeded?`? I don't see this pattern anywhere else and it seems that this check will occur inside `resource_watcher.rb`, anyway ([ref](https://github.com/Shopify/kubernetes-deploy/blob/master/lib/kubernetes-deploy/resource_watcher.rb#L34)).

lib/kubernetes-deploy/kubernetes_resource/persistent_volume_claim.rb
```diff
@@ -3,16 +3,76 @@ module KubernetesDeploy
...
+      return true if status == "Bound"
+
+      # if the StorageClass has volumeBindingMode: WaitForFirstConsumer,
+      # it won't bind until after a Pod mounts it. But it must be pre-deployed,
+      # as the Pod requires it. So 'Pending' must be treated as a 'Success' state
+      if storage_class&.volume_binding_mode == "WaitForFirstConsumer"
+        status == "Pending" || status == "Bound"
```
timothysmith0609: As far as kubernetes-deploy is concerned, this case is treated as `Success`. However, the critical point of actually binding the volume has still not occurred so it might be misleading to indicate success carte blanche. To that end, should we log a warning to the output?

Conversely, if the binding fails, the pod that's using the claim will fail anyway so 🤷‍♂

lib/kubernetes-deploy/kubernetes_resource/persistent_volume_claim.rb
```diff
@@ -3,16 +3,76 @@ module KubernetesDeploy
...
+      end
     end
 
     def deploy_failed?
-      status == "Lost"
+      status == "Lost" || failure_message.present?
+    end
+
+    def failure_message
```
timothysmith0609: nit: more descriptive name (e.g. multiple_defaults_error_message).

lib/kubernetes-deploy/kubernetes_resource/persistent_volume_claim.rb
timothysmith0609: nit: also `annoated` -> `annotated`

lib/kubernetes-deploy/kubernetes_resource/persistent_volume_claim.rb
```diff
@@ -3,16 +3,76 @@ module KubernetesDeploy
...
+    private
+
+    def storage_class_name
+      @definition.dig("spec", "storageClassName")
+    end
+
+    def storage_class
+      if storage_class_name.present?
+        @storage_classes.detect { |sc| sc.name == storage_class_name }
+      elsif storage_class_name != ""
```
timothysmith0609: why not just a bare `else`?

test/integration/kubernetes_deploy_test.rb
```diff
@@ -1287,6 +1287,129 @@ def test_missing_name_on_unknown_resource_prints_metadata_but_not_body
     refute_logs_match("YWRtaW4=")
   end
 
+  # Note: These tests assume a default storage class with a dynamic provisioner and 'Immediate' bind
+  def test_pvc
+    pvname = "local0001"
+    scname = "k8s-deploy-test"
```
timothysmith0609: nits:
`pv_name`?
`sc_name` or `storage_class_name`?

test/fixtures/pvc/storage_class.yml
```diff
@@ -0,0 +1,6 @@
+apiVersion: storage.k8s.io/v1
+kind: StorageClass
+metadata:
+  name: k8s-deploy-test
+provisioner: kubernetes.io/no-provisioner
+volumeBindingMode: WaitForFirstConsumer
```
timothysmith0609: Since this field is important to the logic of the program, it makes sense to reflect its value in the filename.
Alternatively, leave this field blank and add it dynamically when you load the fixture.

lib/kubernetes-deploy/kubernetes_resource/persistent_volume_claim.rb
dturn: This is overriding a method in `KubernetesResource`, or do you think I should move the string to a constant?

lib/kubernetes-deploy/kubernetes_resource/persistent_volume_claim.rb
dturn: I agree there is room to 🔫 👣 here. But I don't think its our job to warn people of the dangers of using `"WaitForFirstConsumer"`

lib/kubernetes-deploy/kubernetes_resource/persistent_volume_claim.rb
dturn: `storage_class_name == ""` is actually a different case than `storage_class_name == nil`. The first says I don't want any storage class, the second is I want the default one.

lib/kubernetes-deploy/kubernetes_resource/persistent_volume_claim.rb
timothysmith0609: I think it's fine. I forgot it was an overridden method

lib/kubernetes-deploy/kubernetes_resource/persistent_volume_claim.rb
timothysmith0609: I think that's worth a comment in the code

lib/kubernetes-deploy/kubernetes_resource/persistent_volume_claim.rb
```diff
@@ -3,16 +3,76 @@ module KubernetesDeploy
...
     def deploy_failed?
-      status == "Lost"
+      status == "Lost" || failure_message.present?
+    end
+
+    def failure_message
+      if storage_class_name.nil? && @storage_classes.count(&:default?) > 1
+        "PVC has no StorageClass specified and there are multiple StorageClasses " \
+        "annotated as default. This is an invalid cluster configuration."
+      end
```
timothysmith0609: Is there an `else` case to consider? If it just returns nil will that hide worthwhile information from the user?

lib/kubernetes-deploy/kubernetes_resource/persistent_volume_claim.rb
RyanBrushett: A comment and/or explicitly handling that case.

lib/kubernetes-deploy/kubernetes_resource/persistent_volume_claim.rb
dturn: The else case is `status == "Lost"` which didn't have any message before.  I'm also not sure what we would helpfully say in the lost case.

lib/kubernetes-deploy/kubernetes_resource/persistent_volume_claim.rb
dturn: done.

redmcg:
Thanks all for your work on this

lib/kubernetes-deploy/kubernetes_resource/persistent_volume_claim.rb
```diff
@@ -3,16 +3,78 @@ module KubernetesDeploy
...
+    def storage_class_name
+      @definition.dig("spec", "storageClassName")
+    end
+
+    def storage_class
+      if storage_class_name.present?
+        @storage_classes.detect { |sc| sc.name == storage_class_name }
+      # storage_class_name = "" is an explicit request for no storage class
+      # storage_class_name = nil is an impplicit request for default storage class
+      elsif storage_class_name != ""
```
KnVerey: Isn't this more specifically `storage_class_name.nil?`?

test/integration/kubernetes_deploy_test.rb
```diff
@@ -1287,6 +1287,129 @@ def test_missing_name_on_unknown_resource_prints_metadata_but_not_body
...
+
+    result = deploy_fixtures("pvc", subset: ["pvc.yml", "pod.yml"]) do |fixtures|
+      pvc = fixtures["pvc.yml"]["PersistentVolumeClaim"].first
+      pvc["spec"]["storageClassName"] = storage_class_name
+    end
+    assert_deploy_failure(result)
+
+    assert_logs_match_all([
+      "Failed to deploy 1 priority resource",
+      "PersistentVolumeClaim/with-storage-class: TIMED OUT (timeout: 20s)",
```
KnVerey: Is this test taking a full 20s? Can we use a custom timeout to cut that down?

dturn: Good eye, it does :(. Should be able to decrease it, just haven't tried yet.

redmcg: This is already using a custom timeout via a `kubernetes-deploy.shopify.io/timeout-override`:
https://github.com/Shopify/kubernetes-deploy/blob/ef17097b42b5875e15c138b74f7a9eb80485af9c/test/fixtures/pvc/pvc.yml#L5

The default was 300s - so I shortened it to 20s. I chose 20s to be sure it wasn't so short as to create a false positive - but 20s is still pretty generous.

PR #480, opened 2019-05-01
Check volume binding mode (additional tests)

redmcg:
Contributing to PR#479

**How is this accomplished?**
With this PR

**What could go wrong?**
It's a little bit ugly - but it might work

### DIFF
--- a/lib/kubernetes-deploy/kubeclient_builder.rb
+++ b/lib/kubernetes-deploy/kubeclient_builder.rb
@@ -96,6 +96,14 @@ module KubernetesDeploy
       )
     end
 
+    def build_storage_v1_kubeclient(context)
+      build_kubeclient(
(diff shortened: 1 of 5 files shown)


redmcg:
In https://github.com/Shopify/kubernetes-deploy/pull/480/commits/39a4677b9509ea8d82513e2c8c8f7b404ff01c3b I added a number of tests.

I had to use unique PV and storage class names, as even when running serially there seemed to be issues when using the same name (I think the resources weren't getting cleaned up prior to the next test starting). This at least allows it to also run in parallel.

The tests still assume a default storage class exists in the test cluster. And because I don't want to modify existing cluster settings, I've not included a test for the absence of a default storage class.

Is it better to not make assumptions about the test cluster and to remove all tests around default storage class?

PR #487, opened 2019-05-24
Easier ruby use

KnVerey:

Make it easier to use our Ruby interface, i.e. `DeployTask`, `RunnerTask`, `RenderTask` and `RestartTask` by:
- Moving `require`s around so that requiring a specific task (e.g. `require 'kubernetes-deploy/deploy_task'`) actually gets you the dependencies you need to run it (and not dependencies you don't need).
- Making all the tasks give you a logger out of the box if you don't pass one in. People rarely want to customize the logger (outside our tests), so making them construct one of our fancy loggers is unnecessary friction.
- Making `require kubernetes-deploy` require ALL the tasks so that it still has a purpose in the world where the correct way to require any specific task--including DeployTask--is `require kubernetes-deploy/TASK`.

cc @stefanmb 

**What could go wrong?**

The requires could be incomplete, causing the tasks to blow up in production when executing a line my manual tests didn't cover. I tried making the test files only require the code for their respective tasks, but in addition to being a bit of a mess, it was pointless because all but the RenderTask suite need to require DeployTask (i.e. most of the code) for setup purposes.

**Todo**

- [x] Look carefully about which files each task is losing access to
- [ ] ~~Investigate the OJ conflict with Rails apps and consider requiring that file from common~~ While a valid issue, this is not actually related to this PR.

After review but before shipping, test the tasks:
- [x] invoke each of the tasks from the command line with various options to make sure none are broken
- [x] invoke each of the tasks from a Ruby console session to make sure this PR works as stated
- [ ] use cumulus-cat's Shipit task to test this branch

### DIFF
--- a/exe/kubernetes-deploy
+++ b/exe/kubernetes-deploy
@@ -1,7 +1,11 @@
 #!/usr/bin/env ruby
 # frozen_string_literal: true
 
-require 'kubernetes-deploy'
+require 'kubernetes-deploy/deploy_task'
+require 'kubernetes-deploy/options_helper'
+require 'kubernetes-deploy/bindings_parser'
+require 'kubernetes-deploy/label_selector'
+
 require 'optparse'
 
 skip_wait = false
--- a/exe/kubernetes-render
+++ b/exe/kubernetes-render
@@ -22,12 +22,10 @@ ARGV.options do |opts|
 end
 
 templates = ARGV
-logger = KubernetesDeploy::FormattedLogger.build(verbose_prefix: false)
 
 begin
   KubernetesDeploy::OptionsHelper.with_validated_template_dir(template_dir) do |dir|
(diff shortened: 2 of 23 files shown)


KnVerey:
@Shopify/cloudx This is ready for review. @dturn kindly offered to take over getting it merged. I already manually tested invoking all the tasks with all possible args from the command line, plus requiring only the stated files in the console and successfully invoking the task from ruby. If changes are required after review, some of those tests may need to be repeated, but otherwise this just needs a test deploy using cumulus-cat.

We need to document our Ruby interface, both with inline docs and in the readme. That could be added to this PR, but at this point I'd be inclined to save it for Krane.

dturn:
> I wasn't able to test kubernetes-run in my local, will try to figure it
out


I used:

be exe/kubernetes-run cloud-portal-staging tierstaging-us-east1-2
flush-cache

PR #488, opened 2019-05-25
Fix: Undefined method "merge" on LabelSelector

ramontayag:
When deploying with a `--selector`, fix exceptions:
- `undefined method "merge"` on `LabelSelector` and
- `undefined method `reverse_merge!` on `Hash`

**How is this accomplished?**
- convert LabelSelector to hash before merging
- require ActiveSupport's `reverse_merge` lib

**What could go wrong?**
Not sure - it's already busted now


dturn:
Thanks for the bug fix!

PR #496, opened 2019-06-05
Define Memcached and Redis as CustomResource

jerr:
Be able to validate the custom resource status for Redis and Memcached resources.

_Related PR:_ [cloudbuddies#1783](https://github.com/Shopify/cloudbuddies/pull/1783)

**How is this accomplished?**
Define `Memcached` and `Redis` as `CustomResource` and call the `deploy_succeeded` function to validate the status.

**What could go wrong?**
Nothing special, this is compatible with the existing CRDs
I tested with and without the `kubernetes-deploy.shopify.io/instance-rollout-conditions` annotation for both CRDs.

### DIFF
--- a/lib/kubernetes-deploy/kubernetes_resource.rb
+++ b/lib/kubernetes-deploy/kubernetes_resource.rb
@@ -38,7 +38,11 @@ module KubernetesDeploy
         opts = { namespace: namespace, context: context, definition: definition, logger: logger,
                  statsd_tags: statsd_tags }
         if (klass = class_for_kind(definition["kind"]))
-          return klass.new(**opts)
+          if (klass <= CustomResource)
+            return klass.new(crd: crd, **opts)
+          else
+            return klass.new(**opts)
+          end
         end
         if crd
           CustomResource.new(crd: crd, **opts)
--- a/lib/kubernetes-deploy/kubernetes_resource/memcached.rb
+++ b/lib/kubernetes-deploy/kubernetes_resource/memcached.rb
@@ -1,6 +1,6 @@
 # frozen_string_literal: true
 module KubernetesDeploy
-  class Memcached < KubernetesResource
+  class Memcached < CustomResource
     TIMEOUT = 5.minutes
     CONFIGMAP_NAME = "memcached-url"
 
@@ -16,6 +16,7 @@ module KubernetesDeploy
     end
 
     def deploy_succeeded?
+      return false unless super
       deployment_ready? && service_ready? && configmap_ready?
     end
 
--- a/lib/kubernetes-deploy/kubernetes_resource/redis.rb
+++ b/lib/kubernetes-deploy/kubernetes_resource/redis.rb
@@ -1,6 +1,6 @@
 # frozen_string_literal: true
 module KubernetesDeploy
-  class Redis < KubernetesResource
+  class Redis < CustomResource
     TIMEOUT = 5.minutes
     UUID_ANNOTATION = "redis.stable.shopify.io/owner_uid"
 
@@ -19,6 +19,7 @@ module KubernetesDeploy
     end
 
     def deploy_succeeded?
+      return false unless super
       deployment_ready? && service_ready?
     end
 


lib/kubernetes-deploy/kubernetes_resource.rb
```diff
@@ -38,7 +38,11 @@ def build(namespace:, context:, definition:, logger:, statsd_tags:, crd: nil)
         opts = { namespace: namespace, context: context, definition: definition, logger: logger,
                  statsd_tags: statsd_tags }
         if (klass = class_for_kind(definition["kind"]))
-          return klass.new(**opts)
+          if klass <= CustomResource
```
maximebedard: I think you could remove this, and burn both Memcached and Redis instead no? I guess this is for compatibility reason only?

jerr: Memcached and Redis check also the deployment (nb replicas == available pods).  We need to keep this because `CustomResource` does not handle that.

```ruby
   def deploy_succeeded?
      return false unless super
      deployment_ready? && service_ready?
    end
```

dturn: Would it be possible to make the buddy check these conditions. We really want to 🔥 these files since they are in an open source repo, but are only useable to shopify.

jerr: > Would it be possible to make the buddy check these conditions?
Let's go for that.

KnVerey:
@jerr are you planning on revisiting this PR soon? I see the "buddies v2" project this was part of was moved to Done last week.

jerr:
> @jerr are you planning on revisiting this PR soon? I see the "buddies v2" project this was part of was moved to Done last week.

yes, everything is done on the buddies side and shipped.  I have to find a good way to test the `kubernetes-deploy` changes.

jerr:
## Test a simple case

1. create a namespace
```
$ kubectl create namespace kubedeploy-test
namespace/kubedeploy-test created
```
2. create a redis yaml file
```
$ cat << EOF >> redis1.yaml 
{
  "apiVersion": "stable.shopify.io/v1",
  "kind": "Redis",
  "metadata": {
    "name": "test1"
  }
}
EOF
``` 
3. deploy it with `kubernetes-deploy`
```
$ REVISION=12345 ./exe/kubernetes-deploy kubedeploy-test minikube  --template-dir=~/tmp/kubernetes-deploy_496
```
image
🎉 **Works as expected!**

jerr:
## Test wrong configuration case

1. create the wrong configuration
```
cat << EOF > redis2.yaml 
{
  "apiVersion": "stable.shopify.io/v1",
  "kind": "Redis",
  "metadata": {
    "name": "test2"
  }, 
  "spec": {
    "max_memory": "1Mi"
  }
}
EOF
``` 
2. deploy it with `kubernetes-deploy`
```
$ REVISION=12345 ./exe/kubernetes-deploy kubedeploy-test minikube  --template-dir=~/tmp/kubernetes-deploy_496
```
image
🎉 **Works as expected!**

3. fix the configuration
```
$ cat << EOF > redis2.yaml  
{
  "apiVersion": "stable.shopify.io/v1",
  "kind": "Redis",
  "metadata": {
    "name": "test2"
  },
  "spec": {
    "max_memory": "1Gi"
  }
}
EOF
```

4. redeploy
```
$ REVISION=12345 ./exe/kubernetes-deploy kubedeploy-test minikube  --template-dir=~/tmp/kubernetes-deploy_496
```
image
🎉 **Works as expected!**

jerr:
## Test the pruning case

1. delete one Redis
```
$ rm redis1.yaml
```

2. deploy it with kubernetes-deploy
```
$ REVISION=12345 ./exe/kubernetes-deploy kubedeploy-test minikube  --template-dir=~/tmp/kubernetes-deploy_496
```
image
🎉 Works as expected!

test/unit/kubernetes-deploy/kubernetes_resource_test.rb
```diff
@@ -321,7 +321,7 @@ def test_build_handles_hardcoded_and_core_and_dynamic_objects
     redis_cr = KubernetesDeploy::KubernetesResource.build(namespace: "test", context: "test",
       logger: @logger, statsd_tags: [], crd: redis_crd,
       definition: { "kind" => "Redis", "metadata" => { "name" => "test" } })
-    assert_equal(redis_cr.class, KubernetesDeploy::Redis)
+    assert_equal(redis_cr.class, KubernetesDeploy::CustomResource)
```
KnVerey: This assertion shouldn't be changed to the generic class--this part of the test is proving that if a hardcoded class exists, it takes precedence. Therefore, Redis can't be used for it anymore. Simply change the example to use Cloudsql or Elasticsearch, which are our last to hardcoded Shopify CRs.

PR #498, opened 2019-06-05
Add more targets to prune whitelist

RyanBrushett:

Quick win: resolve https://github.com/Shopify/kubernetes-deploy/issues/271
Basically, expand the prune whitelist to include ServiceAccount, PodTemplate, and ReplicaSet.

**How is this accomplished?**

I just added them to the whitelist. Seems consistent.

**What could go wrong?**

Could prune stuff it shouldn't or miss things it should prune, I suppose. Does this cover enough?

### DIFF
--- a/lib/kubernetes-deploy/deploy_task.rb
+++ b/lib/kubernetes-deploy/deploy_task.rb
@@ -58,7 +58,6 @@ module KubernetesDeploy
     # core/v1/Endpoints -- managed by services
     # core/v1/PersistentVolumeClaim -- would delete data
     # core/v1/ReplicationController -- superseded by deployments/replicasets
-    # extensions/v1beta1/ReplicaSet -- managed by deployments
 
     def predeploy_sequence
       before_crs = %w(
@@ -85,7 +84,10 @@ module KubernetesDeploy
         core/v1/Service
         core/v1/ResourceQuota
         core/v1/Secret
+        core/v1/ServiceAccount
+        core/v1/PodTemplate
         batch/v1/Job
+        extensions/v1beta1/ReplicaSet
         extensions/v1beta1/DaemonSet
         extensions/v1beta1/Deployment
         extensions/v1beta1/Ingress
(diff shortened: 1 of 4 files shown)


KnVerey:
> Does this cover enough?

Can you explain more how you came up with the list? Which namespaced resources, if any, are you excluding from it?

RyanBrushett:
@KnVerey Your issue here: https://github.com/Shopify/kubernetes-deploy/issues/271
Didn't think much harder about it than that, honestly. Should that list be expanded?

KnVerey:
Ah, I see. Yeah, that issue was opened more than a year ago. I'm thinking for the purposes of this PR, what we need is a quick audit of the resources we support vs. what is on the whitelist. I think all namespaced resources we support should be on it, but maybe I'm forgetting something.

RyanBrushett:
@KnVerey Good shout, I'll do some digging and see what I can come up with 👍

RyanBrushett:
@KnVerey I went through our `kubernetes_resource` directory to see what k8s resources we support (aside from custom stuff) and what wasn't in our prune whitelist.
I added `role` and `rolebinding` which are in our `kubernetes_resource` directory, are namespaced, but aren't currently in our prune whitelist. Does that make sense?

lib/kubernetes-deploy/deploy_task.rb
```diff
@@ -58,7 +58,6 @@ class DeployTask
     # core/v1/Endpoints -- managed by services
     # core/v1/PersistentVolumeClaim -- would delete data
```
KnVerey: I wonder if we should re-assess this choice as well. It is true that deleting a PVC can result in data loss (depending on the persistentVolumeReclaimPolicy on the underlying PV), but failing to prune them can result in continuing costs for undesired, unused storage.

RyanBrushett: Oooo good call! I'll do some more digging.

RyanBrushett:
I'm leaning towards pruning persistent volume claims.
See https://kubernetes.io/docs/concepts/storage/persistent-volumes/#reclaiming

TL;DR One can set a reclaim policy on the volume to either retain or delete. If people want to be really careful, a retain reclaim policy will save them from losing data upon pruning a PVC.
Dynamically-provisioned volumes inherit the reclaim policy from their [storage class](https://kubernetes.io/docs/concepts/storage/storage-classes/) so, again, there is an opportunity to protect data.

Thoughts?

RyanBrushett:
After some digging, it seems like it’d be a good idea to split this apart. Pruning PVCs could easily lead to data loss and I think that’s going to warrant some additional legwork.
See https://github.com/Shopify/kubernetes-deploy/issues/519 
Meanwhile, I think that’ll unblock this PR :+1: 

Thoughts anyone?

PR #502, opened 2019-06-27
Update kubernetes-render instructions on README.md

viniciusgama:
Adding extra information about the behavior of the kubernetes-render command

Make it the behavior of the command a little more explicit. It's just a small thing but it could have saved me some time if I found that information more easily. Just thought about contributing back and making others lives easier as well.

**How is this accomplished?**
By updating the documentation.

**What could go wrong?**
Nothing

### DIFF
--- a/README.md
+++ b/README.md
@@ -501,6 +501,12 @@ To render some templates in a template dir, run kubernetes-render with the names
 kubernetes-render --template-dir=./path/to/template/dir this-template.yaml.erb that-template.yaml.erb
 ```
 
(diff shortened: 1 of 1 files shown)


viniciusgama:
thanks for the quick reply @dturn. I made the suggested change, it really seems like a better idea. I'm quite bad at writing documentation and also english is not my main language so any other feedback will be greatly appreciated.

dturn:
Thanks again for the PR!

Note: I merged this w/o CI since the PR only changes `README.md`

PR #505, opened 2019-07-02
Allow customising which CRs are pre-deployed

benlangfeld:

Allow users to determine that certain CRDs should not be considered for pre-deploy. Our specific use-case is a variation on `CronJob` which we want to deploy only once our bare Pod deployment tasks (Rails migrations) have run.

**How is this accomplished?**

CRDs support an annotation which when set to any value other than `true` causes custom resources of that type not to be selected for pre-deploy.

**What could go wrong?**

Not sure.


benlangfeld:
A corporate CLA was just signed to cover this.

lib/kubernetes-deploy/kubernetes_resource/custom_resource_definition.rb
```diff
@@ -56,6 +56,11 @@ def prunable?
       prunable == "true"
     end
 
+    def predeployed?
+      predeployed = @definition.dig("metadata", "annotations", "kubernetes-deploy.shopify.io/predeployed")
+      predeployed.nil? || predeployed == "true"
```
dturn: `predeployed.downcase`

Should we raise if the value isn't in `nil, 'false', 'true'` ?

test/integration-serial/serial_deploy_test.rb
```diff
@@ -161,6 +161,24 @@ def test_crd_pruning
     wait_for_all_crd_deletion
   end
 
+  def test_custom_resources_predeployed
+    assert_deploy_success(deploy_fixtures("crd-predeploy", subset: %w(mail.yml things.yml widgets.yml)))
```
dturn: I'd prefer if you used the fixtures in `test/fixtures/crd` and added the annotations in the test, rather than create a new directory.

lib/kubernetes-deploy/kubernetes_resource/custom_resource_definition.rb
benlangfeld: We don't do any of that for `prunable`. Do you need those things implemented there too?

test/integration-serial/serial_deploy_test.rb
benlangfeld: Done.

lib/kubernetes-deploy/kubernetes_resource/custom_resource_definition.rb
dturn: Good point. Its probably a good idea to do it in both places, but I won't make it a blocker.

benlangfeld: `prunable` is explicitly documented as "any other value means false"; restricting to only `nil, 'false', 'true'` would be BC-breaking. I'm also not sure why we would want to be case insensitive.

dturn:
The CI failure is because the lint step can't check out code from remote repos, I'll need to dig into that. For now you can run `bundle exec rubocop`.

benlangfeld:
@dturn Resolved.

CHANGELOG.md
```diff
@@ -2,6 +2,7 @@
 
 *Enhancements*
 - Officially support Kubernetes 1.14 [#461](https://github.com/Shopify/kubernetes-deploy/pull/461)
+- Allow customising which custom resources are deployed in the pre-deploy phase.
```
timothysmith0609: Can you add a link to the PR here, please? I often find it helpful during releases when going over things in detail.

benlangfeld: Done

benlangfeld:
Could someone maybe trigger another CI run on this?

timothysmith0609:
CI on updated commit is all green (except for lint, as above)

benlangfeld:
@timothysmith0609 What lint issue?

```
{12:49}~/code/kubernetes-deploy:crd-select-predeploy ✓ [2.5.0]
➭ be rubocop
.rubocop-http---shopify-github-io-ruby-style-guide-rubocop-yml: Style/RedundantSortBy has the wrong namespace - should be Performance
.rubocop-http---shopify-github-io-ruby-style-guide-rubocop-yml: Style/Sample has the wrong namespace - should be Performance
Warning: unrecognized cop Layout/IndentFirstArgument found in .rubocop-http---shopify-github-io-ruby-style-guide-rubocop-yml
Warning: unrecognized cop Layout/IndentFirstArrayElement found in .rubocop-http---shopify-github-io-ruby-style-guide-rubocop-yml
Warning: unrecognized cop Layout/IndentFirstHashElement found in .rubocop-http---shopify-github-io-ruby-style-guide-rubocop-yml
Warning: unrecognized cop Style/Strip found in .rubocop-http---shopify-github-io-ruby-style-guide-rubocop-yml
Warning: unrecognized cop Sorbet/ConstantsFromStrings found in .rubocop.yml
Warning: unrecognized cop Layout/IndentFirstArgument found in .rubocop.yml
Warning: unrecognized cop Layout/IndentFirstArrayElement found in .rubocop.yml
Warning: unrecognized cop Layout/IndentFirstHashElement found in .rubocop.yml
Warning: unrecognized cop Style/Strip found in .rubocop.yml
Inspecting 112 files
................................................................................................................

112 files inspected, no offenses detected
```

benlangfeld:
Thank you @dturn ! Do you have an estimate for when this might make it into a release?

timothysmith0609:
>@timothysmith0609 What lint issue? 

Just the fact that lint for external contrib is broken in our CI, not that there was an actual lint error in the code

dturn:
> Do you have an estimate for when this might make it into a release?

We're going to try and release 0.26.5 today.

Issue #483, opened 2019-05-15
Service assumes pod selector is in sync with deployment's labels.


dturn:
# Bug report

The service class checks that a Service's selector matches pod since failing to do so is likely a bad spec. However, there is a cases where 0 pods is ok, e.g. when the deployment that the service is fronting is scaled down to zero. 

Currently, if the pods selector matches multiple deployments, we assume that the 0 case is wrong. However, the multiple deployments might not use the same selector to select their pods.  We likely need to better filter deployments when multiple match.

https://github.com/Shopify/kubernetes-deploy/blob/master/lib/kubernetes-deploy/kubernetes_resource/service.rb#L8-L10

**Expected behavior:** 

I _think_ that we actually need to fetch the selector from each deployment to find linked pods.

```
➜  k8s-template-validator git:(resource-format) ✗ k --context XXX -n XXX describe service 
web-perf  

Name:                     web-perf
...
Selector:                 app=web-perf,name=web-perf
```

```
➜  kubernetes-deploy git:(master) ✗ k --context XXX -n XXX get deployment -lname=web-perf,app=web-perf

NAME            DESIRED   CURRENT   UP-TO-DATE   AVAILABLE   AGE
web-perf        0         0         0            0           1y
web-perf-ht     1         1         1            1           2h
web-perf-noht   1         1         1            1           2h
```

```
 k --context XXX -n XXX get pod  -lapp=web-perf 

NAME                            READY   STATUS    RESTARTS   AGE
web-perf-ht-6695bfb76c-btc88    2/2     Running   0          2h
web-perf-noht-754597c64-4bdbv   2/2     Running   0          2h


k --context XXX -nXXX get pod  -lapp=web-perf,name=web-perf

No resources found.
```

```
apiVersion: extensions/v1beta1
kind: Deployment
metadata:
  labels:
    app: web-perf
    name: web-perf
    run: web-perf
  name: web-perf-noht
...
  template:
    metadata:
      labels:
        app: web-perf
        name: web-perf-noht
        run: web-perf
```

Steps to reproduce
1. Make a service and deployment that has replicas 0 and the labels are different form its pods labels
2. Run a deploy
3. Deploy will fail because no deployments will be found and therefore no healthy pods will be found

PR #509, opened 2019-07-03
change logic for services backed by multiple deployments

dwradcliffe:
Attempt to fix https://github.com/Shopify/kubernetes-deploy/issues/483

**How is this accomplished?**
Change the logic for related deployments to sum the replicas instead of bailing if there are more than one.

**What could go wrong?**
...

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -11,3 +11,7 @@ Naming/FileName:
 
 Sorbet/ConstantsFromStrings:
   Enabled: false
+
+Layout/Tab:
+  Exclude:
+    - test/integration/kubernetes_deploy_test.rb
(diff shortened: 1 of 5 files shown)


dturn:
> Attempt to fix #483

This PR as is part of the fix, but still doesn't address, fix the issue the assumption that the service's pod's selectors are being used to match deployments too.

KnVerey:
CI is complaining with `NoMethodError: undefined method `sum' for []:Array`. That's because we still officially support [ruby 2.3.3](https://github.com/Shopify/kubernetes-deploy/blob/master/kubernetes-deploy.gemspec#L25). I think we can change that, given that that version is [past EOL](https://www.ruby-lang.org/en/news/2019/03/31/support-of-ruby-2-3-has-ended/).

PR #510, opened 2019-07-09
Allow for yaml aliases to be loaded using YAML.safe_load

jmreid:
As of https://github.com/Shopify/kubernetes-deploy/pull/504 and [`0.26.5` ](https://github.com/Shopify/kubernetes-deploy/blob/master/CHANGELOG.md#0265), YAML aliases no longer load and instead throw a `Psych::BadAlias`.

This PR reenabled YAML aliases based on the Psych docs:
https://ruby-doc.org/stdlib-2.1.0/libdoc/psych/rdoc/Psych.html#method-c-safe_load

**How is this accomplished?**
Changes `YAML.safe_load`

**What could go wrong?**
Matches the YAML alias behaviour pre-`0.26.5`, so I'm hoping this is pretty safe. It looks like the dropping of YAML alias support wasn't intentional.

### DIFF
--- a/lib/kubernetes-deploy/bindings_parser.rb
+++ b/lib/kubernetes-deploy/bindings_parser.rb
@@ -41,7 +41,7 @@ module KubernetesDeploy
         when '.json'
           bindings = parse_json(File.read(file_path))
         when '.yaml', '.yml'
-          bindings = YAML.safe_load(File.read(file_path))
+          bindings = YAML.safe_load(File.read(file_path), [], [], true, file_path)
         else
           raise ArgumentError, "Supplied file does not appear to be JSON or YAML"
         end
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 ## next
 
+*Bug Fixes*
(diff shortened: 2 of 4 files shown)


DazWorrall:
I think we want this so added a test to prevent a future regression, I'll leave the approval to people with more context.

lib/kubernetes-deploy/bindings_parser.rb
```diff
@@ -41,7 +41,7 @@ def parse_file(string)
         when '.json'
           bindings = parse_json(File.read(file_path))
         when '.yaml', '.yml'
-          bindings = YAML.safe_load(File.read(file_path))
+          bindings = YAML.safe_load(File.read(file_path), [], [], true)
```
dturn: let's pass 'file_path' as the final arg https://ruby-doc.org/stdlib-2.1.0/libdoc/psych/rdoc/Psych.html#method-c-safe_load it should lead to better error messages (the filename gets passed all the way to parse_stream if you don't want to trace what happens).

jmreid: Good idea!

PR #514, opened 2019-07-23
`kubectl -f`-style support for deploy task

timothysmith0609:
- Implement a new flag, `-f`, with similar semantics to `kubectl -f`
  - `-f` accepts a comma-separated list of filenames and/or directories
  - Must support proper partials scoping for both file and directory arguments
  - Must work with templates passed from stdin (tbh this is mostly for free)
- Currently this is only for `deploy_task`, but the `-f` parsing is done in such a way as to be reusable (via `TemplateDiscovery#templates`)
- Required some changes to method signatures in the tests. I tried to make these as minor as possible.
- `--template-dir` is still supported (both at the CLI and API level), but should be considered deprecated once this PR merges

**Update (Aug 15, 2019)**
The scope of this PR expanded from just supporting multiple template directory to full fledged support for passing in both files and directories. This approach necessitated a few changes to some internal abstractions, namely `TemplateDiscovery` and the new `LocalResourceDiscovery`.

`TemplateDiscover::templates(paths)` returns a mapping of the form `dir -> [filenames]`. As an example, let's say the `paths` argument is `[my/dir, other/foo.yml]` and the contents of `my/dir` and `other` are `[bar.yml, baz.yml`] and `[foo.yml, more.yml]`. The return value of `templates` will thus be:
```
{
  "my/dir": ["bar.yml", "baz.yml"],
  "other": ["foo.yml"]
}
```
`TemplateDiscovery` performs proper aggregation of resources and eliminates the possibility of duplicate values existing. Furthermore, proper partial scoping is preserved since we maintain a mapping of directories for all kinds of resources (both dirs and files).

**How is this accomplished?**
Initially, I tried passing multiple template directories around `deploy_task`, but this became cumbersome: I had to create a mapping of template directories to their respective partials dirs, wrap them together, and pass them to and from `TemplateContext` objects and back to the (somewhat-global) renderer.

A cleaner approach was to simply instantiate a separate renderer for each template directory passed in, then, when appropriate, concatenate the result of each template directory rendering operation and proceed through the deploy as normal. This way, the changes to the codebase are much smaller and the logic for each individual element stays simpler.

_Edit:_ What I ended up doing in the end is a bit of a mix of the first 2 ideas. In the deploy task constructor, we declare a mapping of renderers to template directories. The renderers are then lazily instantiated directly at the point of rendering (in `#split_templates`), leveraging the closure provided by the default block of the `@renderers` Hash (thank you Blake for the great suggestion).

Please see the unresolved comments from my review for more fine-grained issues regarding this PR.

**What could go wrong?**
How defensive should we be? I think that giving people the ability to specify multiple template directories places the responsibility of ensuring they end up with a sane and collision-free experience squarely on them.

**TODO**
- [x] Add tests


lib/kubernetes-deploy/options_helper.rb
```diff
@@ -6,22 +6,35 @@ class OptionsError < StandardError; end
...
+            dirs << template_dir
+          end
+        end
+
+        if template_dirs.include?("-")
           Dir.mktmpdir("kubernetes-deploy") do |dir|
             template_dir_from_stdin(temp_dir: dir)
-            yield dir
+            dirs << dir
+            yield dirs
```
timothysmith0609: This control flow feels a bit clumsy, but it accomplishes the goal of wrapping `mktmpdir` inside the execution of the deploy task, ensuring it will be cleaned up on program exit

lib/kubernetes-deploy/renderer.rb
```diff
@@ -16,11 +16,12 @@ def initialize(msg, parents: [], content: nil, filename:)
...
 
-    def initialize(current_sha:, template_dir:, logger:, bindings: {})
+    def initialize(current_sha:, template_dirs:, logger:, bindings: {})
       @current_sha = current_sha
-      @template_dir = template_dir
-      @partials_dirs =
-        %w(partials ../partials).map { |d| File.expand_path(File.join(@template_dir, d)) }
+      @template_dirs = template_dirs
+      @partials_dirs = @template_dirs.each_with_object({}) do |template_dir, partials|
+        partials[template_dir] = %w(partials ../partials).map { |d| File.expand_path(File.join(template_dir, d)) }
```
timothysmith0609: This establishes the mapping of `@partials_dir[template_dir] -> [partials_dirs...]`

lib/kubernetes-deploy/renderer.rb
```diff
@@ -46,13 +47,14 @@ def render_template(filename, raw_template)
       raise InvalidTemplateError.new(err.message, filename: filename, content: raw_template)
     end
 
-    def render_partial(partial, locals)
+    def render_partial(template_dir, partial, locals)
```
timothysmith0609: This PR adds `template_dir` to the method signature of `render_partial, find_partial, TemplateContext#new, and TemplateContext#partial`. It's annoying to have to pass this context around to so many places, but it seems unavoidable. Interested to hear any alternatives anyone may have.

lib/kubernetes-deploy/deploy_task.rb
```diff
@@ -204,15 +206,17 @@ def cluster_resource_discoverer
...
-        namespace: @namespace,
-        context: @context,
-        template_dir: @template_dir,
-        logger: @logger,
-        statsd_tags: @namespace_tags,
-        selector: @selector,
-      )
+    def ejson_provisioners
+      @ejson_provisioners ||= @template_dirs.map do |template_dir|
+        EjsonSecretProvisioner.new(
```
timothysmith0609: It seemed more straightforward to leave the provisioner itself alone and simply create as many instances as there are template directories (which will conceivably be a small number). We can then aggregate all the resources together on [line 274](https://github.com/Shopify/kubernetes-deploy/pull/514/files#diff-4c33bbb177fdf882c41d892f3bde9570R274) and proceed as normal.

An alternative would be to separate certain common methods out into the `EjsonSecretProvisioner` metaclass to avoid having to make excessive calls to e.g. `ejson-keys`. In practice, I don't think this is worth it, but I can be swayed easily if others feel strongly.

lib/kubernetes-deploy/deploy_task.rb
```diff
@@ -568,7 +578,7 @@ def namespace_definition
 
     # make sure to never prune the ejson-keys secret
     def confirm_ejson_keys_not_prunable
-      secret = ejson_provisioner.ejson_keys_secret
+      secret = ejson_provisioners.first.ejson_keys_secret
```
timothysmith0609: This is based on the assumption that `ejson_keys_secret` is always going to get `ejson-keys`. As above, this is currently going to be called once per template dir, but we can change that if so desired

lib/kubernetes-deploy/renderer.rb
```diff
@@ -15,6 +15,7 @@ def initialize(msg, parents: [], content: nil, filename:)
       end
     end
     class PartialNotFound < InvalidTemplateError; end
+    attr_reader :template_dir
```
timothysmith0609: This is currently needed inside `split_templates`, since we need to find the renderer associated with a given template directory. This feels like it could be refactored, though. Instead of passing `@template_dirs` around, we could simply pass around the renderer... we would still need this `attr_reader`, but it might read a bit easier.

lib/kubernetes-deploy/deploy_task.rb
```diff
@@ -102,22 +102,24 @@ def server_version
...
       @logger = logger
       @kubectl = kubectl_instance
       @max_watch_seconds = max_watch_seconds
-      @renderer = KubernetesDeploy::Renderer.new(
-        current_sha: @current_sha,
-        template_dir: @template_dir,
-        logger: @logger,
-        bindings: bindings,
-      )
+      @renderers = @template_dirs.map do |template_dir|
```
BlakeMesdag: It might be worth switching to a hash with a default block: 
``` ruby
@renderers = Hash.new do |hash, template_dir|
  hash[template_dir] = KubernetesDeploy::Renderer.new(
    current_sha: @current_sha,
    logger: @logger,
    template_dir: template_dir,
    bindings: bindings,
  )
end
```

Accessing it with `@renderers[template_dir]` further down would instantiate it the first time and return the first one from then on.

test/integration/kubernetes_deploy_test.rb
```diff
@@ -1095,7 +1103,7 @@ def test_jobs_can_fail
       "Result: FAILURE",
       "Job/hello-job: FAILED",
       "Final status: Failed",
-      %r{\[Job/hello-job\]	DeadlineExceeded: Job was active longer than specified deadline \(\d+ events\)},
+      %r{\[Job/hello-job\]\tDeadlineExceeded: Job was active longer than specified deadline \(\d+ events\)},
```
timothysmith0609: This (and the other change like this) were to satisfy the linter, which was complaining about the tab literal

lib/kubernetes-deploy/deploy_task.rb
timothysmith0609: This turned out to be a really good idea, since the closure exposes `current_sha, logger, and bindings` so they don't have to be passed around elsewhere

lib/kubernetes-deploy/render_task.rb
```diff
@@ -29,7 +29,7 @@ def run!(stream, only_filenames = [])
       @logger.phase_heading("Initializing render task")
 
       filenames = if only_filenames.empty?
-        TemplateDiscovery.new(@template_dir).templates
+        TemplateDiscovery.new([@template_dir]).templates[@template_dir]
```
timothysmith0609: This is the result of the change in contract of `TemplateDiscovery#new` requiring an array of template dirs. 
- Do we want the render task to accept multiple template dirs. Note that the render task can also accept an arbitrary list of filenames as a filter. For multiple template dirs, we'll just apply the filter to each directory.
- If we do want the render task to accept multiple template directories as well, do we want to tackle that in this PR or in a follow-up?

test/helpers/fixture_deploy_helper.rb
```diff
@@ -82,19 +82,19 @@ def deploy_dir_without_profiling(dir, wait: true, allow_protected_ns: false, pru
     )
   end
 
-  # Deploys all fixtures in the given directory via KubernetesDeploy::DeployTask
+  # Deploys all fixtures in the given directories via KubernetesDeploy::DeployTask
   # Exposed for direct use only when deploy_fixtures cannot be used because the template cannot be loaded pre-deploy,
   # for example because it contains an intentional syntax error
-  def deploy_dir(dir, **args)
+  def deploy_dir(*dirs, **args)
```
timothysmith0609: Multiple template directories are exceptional, so it seemed more straightforward to change the signature to accept an arbitrary number of `dirs` as opposed to enforcing an actual array. I can do the latter if that's what we ultimately want.

test/integration/kubernetes_deploy_test.rb
```diff
@@ -802,6 +802,18 @@ def test_deploy_successful_with_partial_availability
...
   end
 
+  def test_deploy_successful_with_multiple_template_dirs
+    result = deploy_dir(fixture_path("test-partials"), fixture_path("cronjobs"),
+      bindings: { 'supports_partials' => 'yep' })
+    assert_deploy_success(result)
+  end
+
+  def test_deploy_successful_with_multiple_template_dirs_multiple_partials
+    result = deploy_dir(fixture_path("test-partials"), fixture_path("test-partials2"),
```
timothysmith0609: The goal of this test is to ensure that partials with identical names are actually rendered according to the relative path of the given template directory. This was done by using an identically-named partial that renders different parts of the spec; if the 'wrong' partial is used, the deploy would fail. This convinces me enough that the renderer for a given template directory is behaving as expected (e.g. properly scoped to the right dir)

dturn:
My original concern with this PR was how will it interact with our goal of supporting a `-f` flag in the Krane CLI https://docs.google.com/document/d/1oInUsKplYGNWTymPY48xtDCx3x4Cwc-phcEKUkt2v4U/edit#. I think this actually makes it easier @timothysmith0609 thoughts?

timothysmith0609:
>My original concern with this PR was how will it interact with our goal of supporting a -f flag in the Krane CLI https://docs.google.com/document/d/1oInUsKplYGNWTymPY48xtDCx3x4Cwc-phcEKUkt2v4U/edit#. I think this actually makes it easier @timothysmith0609 thoughts?

At the very least, I don't think this PR makes things _any more difficult_ in achieving that end. The sole complicating factor between `kubectl -f` and `krane deploy -f` is our support of partials. I think there's a few ways we can tackle that:
- When given a directory/directories, proceed as we currently do (this PR actually enables this behaviour)
- When given concrete filenames (_not_ directories), collect them together by directory, and treat those collections as more `template_dir`s (this will require more thought than just this, of course).

test/fixtures/test-partials2/partials/pod.yaml.erb
```diff
@@ -0,0 +1,15 @@
+metadata:
+  name: web-from-partial
+spec:
+  replicas: 1
+  template:
```
KnVerey: Nitpick: only this part is a podspec, despite the name of the file.

lib/kubernetes-deploy/deploy_task.rb
```diff
@@ -102,22 +102,24 @@ def server_version
       kubectl.server_version
     end
 
-    def initialize(namespace:, context:, current_sha:, template_dir:, logger:, kubectl_instance: nil, bindings: {},
+    def initialize(namespace:, context:, current_sha:, template_dirs:, logger:, kubectl_instance: nil, bindings: {},
```
KnVerey: I was trying to think of ways to avoid this breaking API change, but the options that occur to me aren't pretty:
* Keep `template_dir:,` but let it accept an array and do `@template_dirs = Array(template_dir)` in the initializer
* have both `template_dir:,` and `template_dirs:`, either the initializer merges them or it raises an error if they're both specified

It kinda irks me to make a breaking API change like this soon before we make another one (with Krane, which is a single clean cutover in a way). I'd be kinda ok with something like the second option above as a result. If we keep it breaking though, let's make it very prominent in the changelog.

test/unit/kubernetes-deploy/options_helper_test.rb
```diff
@@ -5,32 +5,32 @@
 class OptionsHelperTest < KubernetesDeploy::TestCase
```
KnVerey: To be exhaustive, I think we're missing test cases for:
- More than one dir
- More than one dir, and one of them is stdin
- More than one stdin specified

lib/kubernetes-deploy/deploy_task.rb
timothysmith0609: The avoidance of class-level state is what prompted me to choose not to do that in the first place. Even if we're able to create a stateless solution here, it opens the door/makes it easier for changes down the line that might violate that principle

lib/kubernetes-deploy/options_helper.rb
```diff
@@ -6,22 +6,35 @@ class OptionsError < StandardError; end
 
     STDIN_TEMP_FILE = "from_stdin.yml.erb"
     class << self
-      def with_validated_template_dir(template_dir)
-        if template_dir == '-'
+      def with_validated_template_dirs(template_dirs)
+        if template_dirs.select { |dir| dir == "-" }.length > 2
```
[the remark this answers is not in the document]
timothysmith0609: Just doing `.uniq` is probably the right thing to do here

test/fixtures/test-partials2/partials/pod.yaml.erb
timothysmith0609: The goal of the test is to ensure that it's not loading the identically named partial in the other directory, so I kept the partial name the same but ensured the render would fail to validate if it grabbed the wrong partial

dturn:
> A cleaner approach was to simply instantiate a separate renderer for each template directory passed in, then, when appropriate, concatenate the result of each template directory rendering operation and proceed through the deploy as normal. This way, the changes to the codebase are much smaller and the logic for each individual element stays simpler.
>
>Edit: What I ended up doing in the end is a bit of a mix of the first 2 ideas.

Can you comment on why the cleaner approach didn't work out?

timothysmith0609:
>Can you comment on why the cleaner approach didn't work out?

The renderer takes a few arguments that are easily grabbed in the task initializer. In the end, it seemed easier to wrap those in the closure provided by the default hash block than try to pass them into `TemplateDiscovery` further down the stack. That's really the only difference

lib/kubernetes-deploy/deploy_task.rb
```diff
@@ -267,23 +271,26 @@ def check_initial_status(resources)
...
       @logger.info("Discovering templates:")
 
-      TemplateDiscovery.new(@template_dir).templates.each do |filename|
-        split_templates(filename) do |r_def|
-          crd = crds[r_def["kind"]]&.first
-          r = KubernetesResource.build(namespace: @namespace, context: @context, logger: @logger, definition: r_def,
-            statsd_tags: @namespace_tags, crd: crd)
-          resources << r
-          @logger.info("  - #{r.id}")
+      TemplateDiscovery.new(@template_dirs).templates.each do |template_dir, filenames|
```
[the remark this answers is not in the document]
timothysmith0609: I'm :+1: on pushing the template splitting to the renderer. I don't think `TemplateDiscovery` should own resource instantiation, however.

IMO, `TemplateDiscovery` shouldn't own anything other than returning a list of valid template files, organized by directory. For now, `deploy_task` is the only thing that actually needs to instantiate resources, so I'm ok with having such instantiation logic live there.

TBH, the more I think about it the more I think we should move rendering out of `deploy_task`. Coupling rendering with deployment is a bad anti-pattern (as we've discussed ad nauseam).

As it stands, I'm pretty comfortable with the current solution of creating a renderer-constructor in `DeployTask#new` and opportunistically creating renderers from provided template dirs at their actual point of use. Moving template splitting into the renderer itself jives well with this approach and let's us kick the tougher question of whether we should be rendering here at all down the road.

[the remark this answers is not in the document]
timothysmith0609: I have a few more ideas after thinking about this a bit more.

Perhaps `TemplateDiscovery` is poorly named, and should instead be something like `ResourceDiscovery`. `ResourceDiscovery` is then responsible for finding all relevant resource files in a template directories, splitting the templates in each file, calling the renderer on them, and returning a list of fully realized `KubernetesResource`s.

This allows us to completely remove any explicit references to the renderer in `DeployTask`. This is beneficial since rendering logic shouldn't need to be directly in `DeployTask`, anyway (it should only care about fully realized resources). Also, if we ever decide to disallow rendering of ERB inside `DeployTask`, we can simple pass in `render_erb: false` and ignore any `.erb` files. WDYT?

dturn:
>Perhaps TemplateDiscovery is poorly named, and should instead be something like ResourceDiscovery. ResourceDiscovery is then responsible for finding all relevant resource files in a template directories, splitting the templates in each file, calling the renderer on them, and returning a list of fully realized KubernetesResources.

I can't comment in the thread but, I like the idea. As an fyi we have a class called `ClusterResourceDiscovery` so I'd prefer something more specific  `LocalResouceDiscovery`?

lib/kubernetes-deploy/options_helper.rb
```diff
@@ -6,24 +6,37 @@ class OptionsError < StandardError; end
...
+        if template_dirs.include?("-")
           Dir.mktmpdir("kubernetes-deploy") do |dir|
             template_dir_from_stdin(temp_dir: dir)
-            yield dir
+            dirs << dir
+            yield dirs
           end
-        elsif template_dir
-          yield template_dir
         else
```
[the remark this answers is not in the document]
timothysmith0609: The last part of the `if` is the closing of the wrapped `mktmpdir` block, so not quite the same unfortunately :(

test/fixtures/ejson-cloud2/secrets.ejson
```diff
@@ -0,0 +1,12 @@
+{
+  "_public_key": "65f79806388144edf800bf9fa683c98d3bc9484768448a275a35d398729c892a",
+	"kubernetes_secrets": {
```
dturn: ws

lib/kubernetes-deploy/render_task.rb
timothysmith0609: I think you're absolutely right. The current semantics of `kubernetes-render` are incompatible with a `kubectl -f`-like model. For instance, there would be no way to know which bare file names provided to `kubernetes-render` point to which provided template dirs :/

To me, this means we need to make a fundamental change to the CLI contract of `kubernetes-render`. In particular, dropping support for providing bare filenames as arguments and instead force the (conventional) semantics of the `-f` flag. From your above example, this would be:
`krane render -f dirA/one.yml,dirB/two.yml` (not that bad).

In light of this, I'd like to keep `kubernetes-render` _as-is_ (only supporting a single template dir), and make a follow-up PR to make `krane render` behave with `-f` semantics.

test/unit/kubernetes-deploy/options_helper_test.rb
timothysmith0609: Done

lib/kubernetes-deploy/deploy_task.rb
```diff
@@ -102,22 +102,17 @@ def server_version
...
-    def initialize(namespace:, context:, current_sha:, template_dir:, logger:, kubectl_instance: nil, bindings: {},
-      max_watch_seconds: nil, selector: nil)
+    def initialize(namespace:, context:, current_sha:, logger:, kubectl_instance: nil, bindings: {},
+      max_watch_seconds: nil, selector: nil, template_dirs: [], template_dir: nil)
       @namespace = namespace
       @namespace_tags = []
       @context = context
       @current_sha = current_sha
-      @template_dir = File.expand_path(template_dir)
+      @template_dirs = (template_dirs.map { |dir| File.expand_path(dir) } << template_dir).compact
```
timothysmith0609: We don't need to worry about the case of `template_dir` and `template_dirs` both existing when invoking via CLI, but I've done it this way to make it accommodate the case of someone using the API itself.

lib/kubernetes-deploy/deploy_task.rb
```diff
@@ -333,10 +307,12 @@ def validate_configuration(allow_protected_ns:, prune:)
...
-      if !File.directory?(@template_dir)
-        errors << "Template directory `#{@template_dir}` doesn't exist"
-      elsif Dir.entries(@template_dir).none? { |file| file =~ /(\.ya?ml(\.erb)?)$|(secrets\.ejson)$/ }
-        errors << "`#{@template_dir}` doesn't contain valid templates (secrets.ejson or postfix .yml, .yml.erb)"
+      @template_dirs.each do |template_dir|
+        if !File.directory?(template_dir)
+          errors << "Template directory `#{template_dir}` doesn't exist"
+        elsif Dir.entries(template_dir).none? { |file| file =~ /(\.ya?ml(\.erb)?)$|(secrets\.ejson)$/ }
+          errors << "`#{template_dir}` doesn't contain valid templates (secrets.ejson or postfix .yml, .yml.erb)"
+        end
```
timothysmith0609: This whole block is another thing that should be extracted out of `DeployTask` if possible

lib/kubernetes-deploy/local_resource_discovery.rb
```diff
@@ -0,0 +1,66 @@
...
+        )
+      end
+      @crds = crds
+    end
+
+    def resources
+      resources = []
+      templates.each do |template_dir, filenames|
+        filenames.each do |filename|
+          split_templates(template_dir, filename) do |r_def|
```
timothysmith0609: Thinking about implementing `-f` (e.g. include _filenames_, not just _template_dirs_):

I think we'll be able to do it here, at least in the context of `DeployTask`. We can process directory arguments as we currently do, but for the remaining filenames, we can gather them up by directory and create a similiar `hash[template_dir]: [ONLY_SPECIFIED_FILES]` and proceed as normal.

I'm wondering if I'm painting myself into a corner where I'll have to reimplement `-f` differently for `krane render`, which wouldn't be the greatest (since this class is intrinsically linked with _resource_ discovery, which is not a concern of the render task).

In fact, `krane render` couldn't even instantiate a `LocalResourceDiscovery` object without passing in a ton of `nil` fields (to say nothing of it being semantic nonsense) :( 

Maybe what we need to actually do is move the creation of `template_dir -> filenames` mappings further up the stack instead of in `LocalResourceDiscovery#templates`. I don't want to write the `-f` filter in multiple places

lib/kubernetes-deploy/template_discovery.rb
```diff
@@ -1,14 +1,39 @@
...
+          end
+        end
+        # Filename arg
+        file_args.each_with_object(template_paths) do |filename, hash|
+          dir_name = File.dirname(filename)
+          hash[dir_name] ||= []
+          hash[dir_name] << File.basename(filename) unless hash[dir_name].include?(filename)
+        end
+
+        template_paths
```
timothysmith0609: Now any task should be able to call `TemplateDiscovery#templates` and receive back a mapping of `dirname: [files]`. The input of `templates` is an array whose elements are either file paths or directories (abs and/or relative). The nice thing here is that there's zero other dependencies for generating the dir->files mappings

lib/kubernetes-deploy/deploy_task.rb
```diff
@@ -204,15 +200,17 @@ def cluster_resource_discoverer
...
-      @ejson_provisioner ||= EjsonSecretProvisioner.new(
-        namespace: @namespace,
-        context: @context,
-        template_dir: @template_dir,
-        logger: @logger,
-        statsd_tags: @namespace_tags,
-        selector: @selector,
-      )
+    def ejson_provisioners
+      @ejson_provisioners ||= TemplateDiscovery.templates(@template_paths).keys.map do |template_dir|
```
timothysmith0609: This feels a bit messy, but perhaps this is something to look at in the future. We need to call `TemplateDiscovery` in order to get a nice aggregation of explicit file arguments into a single template dir. If we don't do that, we'd need to add additional logic in the secret provisioner as to whether it's dealing with a directory or a file.... that's what `TemplateDiscovery` is supposed to take care of!

dturn: What happens if I pass `-f config/deploy/staging/` and `-f config/deploy/production/i_dont_like_sym_links.yml` and there are ejson files in each directory?

timothysmith0609: Ah, you're right, that will deploy the undeclared secret :( I've got a (failing) test for this written now, just trying to decide how best to solve it.

It seems like we need to pass the full template mappings around... I'm not sure I want to call `@templates = TemplateDiscovery.new(@template_paths).templates` right in the constructor (is it doing too much to be there?). What are your thoughts. I recall Katrina mentioning that we should do as little as possible in the constructors. Said that, `TemplateDiscovery#templates` can't really throw any errors so it might be safe to do

dturn: If we're trying not to break backwards compatibility adding a method like `TemplateDiscovery.ejson_templates(@template_paths).keys.map do ...` might be the simplest solution. If we we're ok breaking backwards compatibility I could see `EjsonProvisioner` taking a file instead of a directory in which its looking for a specific file.

lib/kubernetes-deploy/deploy_task.rb
```diff
@@ -204,15 +200,18 @@ def cluster_resource_discoverer
...
-        statsd_tags: @namespace_tags,
-        selector: @selector,
-      )
+    def ejson_provisioners
+      @ejson_provisoners ||= TemplateDiscovery.ejson_secret_templates(@template_paths).map do |secret_template|
+        EjsonSecretProvisioner.new(
+          namespace: @namespace,
+          context: @context,
+          ejson_keys_secret: ejson_keys_secret,
+          template_dir: File.dirname(secret_template),
```
tsontario: I think passing the ejson_keys_secret in from DeployTask is the right thing to do, but I'm actually going to revert this change and push it forward to a follow-up PR. The reason for that being it involves a number of changes to the test suite and makes also makes a change to the EjsonSecretProvisioner#new signature which requires a discussion all its own. We can accomplish this PRs goals without having this scope-creep in

lib/kubernetes-deploy/deploy_task.rb
timothysmith0609: Let's not break backwards-compat just yet. I like your idea of `TemplateDiscovery.ejson_templates`

README.md
```diff
@@ -109,6 +109,7 @@ official compatibility chart below.
 Refer to `kubernetes-deploy --help` for the authoritative set of options.
 
 - `--template-dir=DIR`: Used to set the deploy directory. Set `$ENVIRONMENT` instead to use `config/deploy/$ENVIRONMENT`. This flag also supports reading from STDIN. You can do this by using `--template-dir=-`. Example: `cat templates_from_stdin/*.yml | kubernetes-deploy ns ctx --template-dir=-`.
+- (alpha feature) `-f [PATHS]`: Accepts a comma-separated list of directories and/or filenames to specify the set of directories/files that will be deployed (use `-` to read from STDIN). Cannot be combined with `--template-dir`. Example: `cat templates_from_stdin/*.yml | kubernetes-deploy ns ctx -f -,path/to/dir,path/to/file.yml`
```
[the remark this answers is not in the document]
tsontario: I believe both ways should work. AFAIK kubectl -f accepts both multiple invocations and a list of filenames/dirs. The same behaviour should work here as well, since invoking `-f` is a purely additive affair

exe/kubernetes-deploy
```diff
@@ -47,17 +52,24 @@ ARGV.options do |opts|
   bindings = parser.parse
 end
 
+# Deprecation path: this can be removed when --template-dir is fully replaced by -f
```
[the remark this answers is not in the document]
timothysmith0609: Yes, exactly

exe/kubernetes-render
```diff
@@ -13,21 +13,25 @@ ARGV.options do |opts|
...
-    template_dir = d
+    template_dir = [d]
   end
   opts.parse!
   bindings = parser.parse
 end
 
 templates = ARGV
 logger = KubernetesDeploy::FormattedLogger.build(verbose_prefix: false)
+if template_dir && !File.directory?(template_dir.first)
```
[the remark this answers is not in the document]
timothysmith0609: You're right, this doesn't handle the default case of using `config/deploy/$ENVIRONMENT`.
Unintuitively, the template directory is passed to `RenderTask#new` whereas the list of filenames is passed as an (optional) argument to `RenderTask#run`. This doesn't square with how we want `-f` to be used so IMO precludes implementing that feature until we make the move over to Krane, where we have the opportunity to make such a breaking change without having to accommodate this older method.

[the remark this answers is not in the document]
timothysmith0609: Yes, that's what it was trying to do, but failed since it wouldn't transparently use `config/deploy/$ENVIRONMENT` if no template dir argument was given (it would erroneously fail). There's a cleaner way to do this in the end, which is setting `template_dir = []` instead of nil and proceeding as normal. `with_validated_template_paths` should do the right thing in that case and maintain compatibility with the CLI/API interface of the render task. Note that render task still only accepts a _single_ template directory, since we can't change that without breaking the current API (specifically, the trailing optional list of filenames)

lib/kubernetes-deploy/options_helper.rb
```diff
@@ -6,27 +6,37 @@ class OptionsError < StandardError; end
 
     STDIN_TEMP_FILE = "from_stdin.yml.erb"
     class << self
-      def with_validated_template_dir(template_dir)
-        if template_dir == '-'
+      def with_validated_template_paths(template_paths)
```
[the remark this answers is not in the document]
timothysmith0609: Nothing in particular, but we do (possibly) perform some transformations on the input. e.g. we try to infer `default_template_dir` and deal with `-` (STDIN) case.

`with_processed_template_paths`, maybe? or just `with_template_paths`?

lib/kubernetes-deploy/template_set.rb
```diff
@@ -0,0 +1,88 @@
...
+module KubernetesDeploy
+  class TemplateSet
+    VALID_TEMPLATES = %w(.yml.erb .yml .yaml .yaml.erb)
+    class << self
+      def with_dirs_and_files(dirs_and_files)
+        resource_templates = {}
+        dir_paths, file_paths = dirs_and_files.partition { |path| File.directory?(path) }
+
+        # Directory paths
+        dir_paths.each_with_object(resource_templates) do |template_dir, hash|
```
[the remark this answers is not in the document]
timothysmith0609: I originally thought it would make my intentions clearer but it's actually a little more compact to do this the 'simpler' way.

README.md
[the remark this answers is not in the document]
timothysmith0609: 🤦‍♂ moment for me. I wasn't using `+=` when parsing the options, instead using raw assignment, so of course it would never work. Changing to increment-operator makes multiple invocations of `-f` now work

lib/kubernetes-deploy/ejson_secret_provisioner.rb
```diff
@@ -17,10 +17,11 @@ class EjsonSecretProvisioner
     EJSON_SECRETS_FILE = "secrets.ejson"
     EJSON_KEYS_SECRET = "ejson-keys"
 
-    def initialize(namespace:, context:, template_dir:, logger:, statsd_tags:, selector: nil)
+    def initialize(namespace:, context:, ejson_keys_secret:, ejson_file:, logger:, statsd_tags:, selector: nil)
       @namespace = namespace
       @context = context
-      @ejson_file = "#{template_dir}/#{EJSON_SECRETS_FILE}"
+      @ejson_keys_secret = ejson_keys_secret
```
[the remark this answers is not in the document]
timothysmith0609: We need both the public and private key. I don't have an issue passing it around but I'd prefer to leave that to another PR

lib/kubernetes-deploy/template_sets.rb
```diff
@@ -0,0 +1,139 @@
...
+        end
+        template_sets
+      end
+    end
+
+    def <<(template_set)
+      if template_set.is_a?(TemplateSet)
+        @template_sets << template_set
+      else
+        raise InvalidTemplateError, "Expected TemplateSet but got #{template_set.class}"
```
timothysmith0609: Wondering if I should create a specific error class for this case :/

lib/kubernetes-deploy/template_sets.rb
```diff
@@ -0,0 +1,139 @@
...
+    end
+
+    private
+
+    def initialize
+      @template_sets = []
+    end
+  end
+
+  class TemplateSet
```
timothysmith0609: Ideally this would be a private inner class but that's not really something you can do in Ruby. Scoping it as `TemplateSets::TemplateSet` also felt wrong. I don't think this should live in its own file, though, since it's intrinsically linked to `TemplateSets` and it's the latter that we should expose and make visible to the rest of the library

lib/kubernetes-deploy/template_sets.rb
```diff
@@ -0,0 +1,139 @@
...
+                bindings: bindings,
+                template_dir: template_dir
+              ))
+        end
+        template_sets
+      end
+    end
+
+    def <<(template_set)
+      if template_set.is_a?(TemplateSet)
```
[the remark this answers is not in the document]
timothysmith0609: Now that I can make this a private class `TemplateSets` should be the only thing using it so I can get rid of the check

lib/kubernetes-deploy/template_sets.rb
```diff
@@ -0,0 +1,139 @@
+# frozen_string_literal: true
+
+module KubernetesDeploy
+  class TemplateSets
+    VALID_TEMPLATES = %w(.yml.erb .yml .yaml .yaml.erb)
+    class << self
+      def new_from_dirs_and_files(paths:, logger:, current_sha:, bindings:)
```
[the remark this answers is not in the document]
timothysmith0609: It is nice to get everything instantiated all at once, but I agree that if you don't need a renderer then there's no point in instantiating it and we won't know if we want it _until_ we call `with_resource_definitions`. I've moved renderer instantiation there.

lib/kubernetes-deploy/template_sets.rb
```diff
@@ -0,0 +1,138 @@
...
+
+        template_sets = TemplateSets.new
+        resource_templates.each do |template_dir, files|
+          template_sets << TemplateSet.new(template_dir: template_dir, file_whitelist: files, logger: logger)
+        end
+        template_sets
+      end
+    end
+
+    def <<(template_set)
```
dturn: This is a bit odd to be exposed since the `TemplateSet`  is a private class. Should `TemplateSets.new` take an array of `TemplateSet`

timothysmith0609: I could go either way, but as you point out, since its a private class and we only intend on constructing these entities in `#from_dirs_and_files` it does make more sense to simple pass an array of `TemplateSet` objects into the (private) constructor

PR #516, opened 2019-07-24
fix prunable true

shashankkoppar:
...
prunable is hardcoded as true 

**How is this accomplished?**
...check if nil first 

**What could go wrong?**
...
Deletes all resources other than the one deploying


dturn:
We've decided to make Custom Resource pruning opt-in only. The change you're proposing would  make CR pruning opt-out not opt-in. I'm gong to close this PR but if there is more you'd like to discuss feel free to re-open/comment.

PR #520, opened 2019-08-05
Add ClusterRoleBinding to Phase 3

krillw:
What are you trying to accomplish with this PR?
Add ClusterRoleBinding to Phase 3: Predeploying. ClusterRoleBinding should be used with RoleBinding simultaneously.


dturn:
Thanks for contributing this and signing the CLA. Seems like we'd want to pre-deploy ClusterRole as well? 

I'd also appreciate if you could add a test like `test_role_and_role_binding_predeployed_before_unmanaged_pod` as well.

dturn:
As of version 1.0 deploy no longer handles non-namespaced resources.

Issue #256, opened 2018-03-21
Krane CLI


KnVerey:
We've decided to try Thor for this (see discussion on this issue). The new CLI must:
- be easy to test (and well-tested)
- enforce required args (addressing https://github.com/Shopify/kubernetes-deploy/issues/482)
- implement the changes to the command and flags detailed in our [Design doc](https://docs.google.com/document/d/1oInUsKplYGNWTymPY48xtDCx3x4Cwc-phcEKUkt2v4U/edit#)

ramontayag:
Add this to the list too https://github.com/commander-rb/commander

KnVerey:
Interesting find, Tim. The maintainers (which apparently include Rafael) argue the CVE isn't actually relevant to a CLI tool: [Thor issue](https://github.com/erikhuda/thor/issues/514). Apparently it only affects Thor's `get` command, which we wouldn't have a use for.

benlangfeld:
@KnVerey Has a library on which to build this been chosen? I would be happy to prep a PR for this if so.

KnVerey:
We haven't chosen yet, but I'll bring it up again with my team and try to get back to you soon.

KnVerey:
I think Thor makes the most sense for this project. However, please note that we have a few other changes that we'd to take the opportunity to make in this transition, beyond just splitting things like `kubernetes-deploy` -> `krane deploy`. Unfortunately these aren't documented in a single place yet. I've added a work item to our internal tracker to produce a proper design document for this.

benlangfeld:
@KnVerey I have a little time available to deal with this. Is it worth me making a start? Do you have any references to those other changes you'd like to make?

KnVerey:
Hey @benlangfeld thanks for letting me know you'd like to start this soon. I started a draft of the design doc and will post it here for comment by the end of the week.

KnVerey:
Here is the proposed design as promised @benlangfeld : https://docs.google.com/document/d/1oInUsKplYGNWTymPY48xtDCx3x4Cwc-phcEKUkt2v4U/edit?usp=sharing. Let me know if you have any trouble accessing it. Everyone should be able to add comments and suggestions.

cc @Shopify/cloudplatform

benlangfeld:
Excellent, thanks @KnVerey . Here's what I propose as an approach for attacking this, each step being a PR with the intent of being merged before moving on to the following step (to reduce rebase overhead) and the possibility of being included in intermediate releases:

[...]

PR #524, opened 2019-08-13
Add a Krane CLI with one command

dturn:

Start down the road of adding a new CLI `Krane`. Our first step is adding `krane version`

See https://github.com/Shopify/kubernetes-deploy/issues/256 for lots of details

**How is this accomplished?**
We've picked `Thor` to help make the cli as easy as possible. We've picked `version` to be the first command because it's just about trivial.

**What could go wrong?**
Its new code, so the only real issue is if the dependency causes people headaches. 

**CLI**

```
➜  kubernetes-deploy git:(cli) ✗ be krane version         
[INFO][2019-08-14 11:06:56 -0700]	krane 0.26.7

➜  kubernetes-deploy git:(cli) ✗ be exe/krane  
Krane commands:
  krane help [COMMAND]  # Describe available commands or one specific command
  krane version         # Prints the version

➜  kubernetes-deploy git:(cli) ✗ be exe/krane version
[INFO][2019-08-13 16:43:39 -0700]	krane 0.26.7

➜  kubernetes-deploy git:(cli) ✗ be exe/krane version -v
ERROR: "krane version" was called with arguments ["-v"]
Usage: "krane version"
```

A fun question here is what exactly should the output be. I looked at `kubectl`, `ruby`, `node`, `bash`, and `grep` and each of these prints something different. Most are of the form `program_name version_string`. So that's what I went for.


Rakefile
```diff
@@ -20,7 +20,7 @@ desc("Run unit tests")
 Rake::TestTask.new(:unit_test) do |t|
   t.libs << "test"
   t.libs << "lib"
-  t.test_files = FileList['test/unit/**/*_test.rb']
+  t.test_files = FileList['test/unit/**/*_test.rb', 'test/exe/**/*_test.rb']
```
dturn: Is this a unit test or should it be a new category?

exe/krane
```diff
@@ -0,0 +1,22 @@
...
+  end
+
+  private
+
+  def logger(verbose: false)
+    @logger ||= KubernetesDeploy::FormattedLogger.build(@namespace, @context, verbose_prefix: verbose)
+  end
+end
+
+Krane.start(ARGV) if __FILE__ == $PROGRAM_NAME
```
timothysmith0609: Does this work? I manually bumped the gem and installed it locally and got this (added debug lines to check `__FILE__` and `$PROGRAM_NAME` when I noticed the version wouldn't print):
```
$ krane version
__FILE__:	/Users/timothysmith/.gem/ruby/2.3.3/gems/kubernetes-deploy-0.27.0/exe/krane
$PROGRAM_NAME: /Users/timothysmith/.gem/ruby/2.3.3/bin/krane
```

`bundle exec krane` yields the same result, whereas `bundle exec exe/krane version` behaves as expected.

CHANGELOG.md
```diff
@@ -6,6 +6,12 @@
   * To see what resources may be affected, run `kubectl get $RESOURCE -o jsonpath='{ range .items[*] }{.metadata.namespace}{ "\t" }{.metadata.name}{ "\t" }{.metadata.annotations}{ "\n" }{ end }' --all-namespaces | grep "last-applied"`
   * To exclude a resource from kubernetes-deploy (and kubectl apply) management, remove the last-applied annotation `kubectl annotate $RESOURCE $SECRET_NAME kubectl.kubernetes.io/last-applied-configuration-`.
 
+*Other*
+
+- We've added a new Krane cli. This code is in alpha. We are providing
+no warranty at this time and reserve the right to make major breaking changes including
+removing it entirely at any time.
```
timothysmith0609: Do we have any links to provide for the Krane redesign or is all that stuff still internal? If not, it's no problem

Rakefile
timothysmith0609: `test/cli`, perhaps?

exe/krane
dturn: Good catch! I've re-worked the file layout,  let me know what you think.

lib/krane/cli/base.rb
```diff
@@ -0,0 +1,19 @@
...
+    package_name "Krane"
+
+    def self.exit_on_failure?
+      true
+    end
+
+    private
+
+    def logger(verbose: false)
+      @logger ||= KubernetesDeploy::FormattedLogger.build(@namespace, @context, verbose_prefix: verbose)
```
timothysmith0609: Not sure if it's better to keep the base `krane` command without `@namespace, @context` and override as needed in each command or just do it all here.

timothysmith0609:
Oops, didn't notice the failed CI before I :+1: but it's just a simple rubocop fix

lib/krane/cli/base.rb
dturn: 🤞 this method goes away with a TaskContext object.

lib/krane/cli/version_command.rb
```diff
@@ -0,0 +1,15 @@
+# frozen_string_literal: true
+require 'active_support/concern'
+
+module Krane
+  module VersionCommand
```
dturn: Given the folder structure I'm wondering if this should be `module krane; module CLI; module VersionCommand` and then have the `Krane::CLI::Krane` instead of `Krane::CLI`

timothysmith0609: I think that makes sense, since `Krane` (top-level) is the entire repo, including API accessible elements, whereas `Krane::CLI::Krane` is the actual scoping of the command-line entrypoint. It's a bit wordy but makes the separation of responsibilities clearer IMO

test/exe/krane_test.rb
```diff
@@ -0,0 +1,20 @@
...
+    assert_logs_match_all([
+      "Krane Version: #{KubernetesDeploy::VERSION}",
+    ], in_order: true)
+  end
+
+  private
+
+  def krane
+    @krane = Krane::CLI.new.tap do |krane|
+      krane.instance_variable_set('@logger', @logger)
```
timothysmith0609: Hmmm, why do we need to do it like this? Won't it be instantiated transparently on the call to `#logger`?

lib/krane/cli/version_command.rb
dturn: Fixed

lib/krane/cli/version_command.rb
```diff
@@ -0,0 +1,21 @@
...
+        #check: { type: :string, banner: 'context',
+        #  desc: 'Verify if Krane is compatible with your local and remote k8s versions.' }
+      }
+
+      def self.from_options(options)
+        # This code will use the Validtor object being added in a different PR.
+        #if options[:check]
+        #  $stderr.puts("Checking if context: #{options[:check]} is valid")
+        #end
+        $stderr.puts("Krane Version: #{KubernetesDeploy::VERSION}")
```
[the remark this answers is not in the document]
dturn: Good question: https://www.jstorimer.com/blogs/workingwithcode/7766119-when-to-use-stderr-instead-of-stdout has some thoughts on each. But I think you're right, the version is the desired output of this command so it should go to stdout. (The output for the `--check` flag is probably destined for stderr)

lib/krane.rb
```diff
@@ -0,0 +1,5 @@
+# frozen_string_literal: true
+require 'kubernetes-deploy/common'
```
dturn: This is a partial optimization to make loading time faster. If we require `kubernetes-deploy` it takes about 0.8 seconds to print the `krane help`. As is about 0.4 and if we require nothing 0.3. I feel like this strikes a good balance between not feeling to slugish to print help but not have to cherry pick exactly what we want when want it.

dturn:
This is ready for 👀 though I don't want to merge it till after the `-f` flag PR. I played with what it might look like to add the `--check` flag in https://github.com/Shopify/kubernetes-deploy/commit/e6306af47deee1776bce9d5414c36b2843b3e5f5 and feel good that there is a reasonable way forward for more than just trivial commands. @KnVerey took a stab at deploy in https://github.com/Shopify/kubernetes-deploy/compare/cli...knverey_cli and this pr took some but not all of the ideas.

Issue #344, opened 2018-10-09
Consistent preflight validations for all tasks


KnVerey:
Follow-up to https://github.com/Shopify/kubernetes-deploy/pull/337

The deploy task already checks whether it can reach the API server before proceeding, so that it can fail with a relevant message instead of failing whichever other request happens to be made first. We should do the same in the runner task. In fact, we should make sure that all globally applicable preflight validations are done consistently by all four of our tasks. One possible approach would be to introduce a config validator object for this purpose, as prototyped here: https://github.com/Shopify/kubernetes-deploy/compare/config_validator.

- Determine all of the appropriate pre-flight validations should be in place across all tasks
- Make sure they are being executed consistently (including order)

Considerations
- Maybe we make a generic validation class that can be reused or something specific for each task

dturn:
https://github.com/Shopify/kubernetes-deploy/pull/533 Added a `TaskConfig` and `TaskConfigValidator` to each task. But it only added the full set of pre-flight validations to the RestartTask. We need to add it to the remaining tasks to close out this issue. 

When working on adding the full set of pre-flight validations to the remaining tasks, consider if we need sub classes like `DeployTaskConfig` and `DeployTaskConfigValidator` (or punt this to a new issue).

dturn:
Done.

PR #533, opened 2019-08-16
First pass at shared validation

dturn:
[Consistent preflight validations for all tasks](https://github.com/Shopify/kubernetes-deploy/issues/344). Inspiration for this PR taken from https://github.com/Shopify/kubernetes-deploy/compare/config_validator

I've started by just adding this to restart task. If we like where this is going I can add it to the remaining tasks in this PR or subsequent ones. 

**How is this accomplished?**

Our run/restart/deploy tasks have overlapping state they need,  I've attempted to extract it into TaskConfig object. For now this is context, namespace, and logger. ~I've also added methods for generating `kubectl`, and `kubeclient_builder`, its possible that this is backwards and those methods should take a task_config.~  (We decided against generating kubectl, kubeclient_builder in task_config, see the comments for the discussion about why). When we go to `Krane` I think the TaskConfig should be passed to the task instead of it building it. (Though maybe not since that would require users using this as a library to construct a 'random' object, perhaps two constructors for each task).

I've also created a Validator object that can validate, kubeconfig, context, namespace and kubectl version. This looks to be the shared set of validations.  Each task has additional validations, this PR leaves open several approaches to solving that problem.

**What could go wrong?**

Because we need to unify validations that are done slightly differently we are changing the error class that gets raised as well as the wording. If people were rescuing these errors or matching text this will be a breaking change.

### DIFF
--- a/lib/kubernetes-deploy/restart_task.rb
+++ b/lib/kubernetes-deploy/restart_task.rb
@@ -41,10 +41,10 @@ module KubernetesDeploy
       @logger.reset
 
       @logger.phase_heading("Initializing restart")
-      errors = Validator.new(@task_config).errors
-      unless errors.empty?
+      task_config_validator = TaskConfigValidator.new(@task_config)
+      unless task_config_validator.valid?
         @logger.summary.add_action("Configuration invalid")
-        @logger.summary.add_paragraph(errors.map { |err| "- #{err}" }.join("\n"))
+        @logger.summary.add_paragraph(task_config_validator.errors.map { |err| "- #{err}" }.join("\n"))
         raise KubernetesDeploy::TaskConfigurationError
       end
 
@@ -178,7 +178,7 @@ module KubernetesDeploy
     end
 
     def kubectl
-      @task_config.kubectl
+      @kubectl ||= Kubectl.new(namespace: @namespace, context: @context, logger: @logger, log_failure_by_default: true)
     end
 
     def v1beta1_kubeclient
@@ -186,7 +186,7 @@ module KubernetesDeploy
     end
 
     def kubeclient_builder
-      @task_config.kubeclient_builder
+      @kubeclient_builder ||= KubeclientBuilder.new
     end
   end
 end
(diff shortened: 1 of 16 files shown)


lib/kubernetes-deploy/task_config.rb
```diff
@@ -0,0 +1,29 @@
...
+      @context = context
+      @namespace = namespace
+      @logger = logger
+    end
+
+    def logger
+      @logger ||= KubernetesDeploy::FormattedLogger.build(@namespace, @context)
+    end
+
+    def kubectl(log_failure_by_default: true)
```
KnVerey: What is the reason for adding kubectl/kubeclient in here? They themselves aren't configuration--they're really powerful cluster access objects. `TaskConfig` strikes me as something we could pass around to all our business objects instead of passing context, namespace and logger individually, but many of those are not supposed to talk to the cluster.

lib/kubernetes-deploy/restart_task.rb
```diff
@@ -37,11 +38,14 @@ def perform!(deployments_names = nil, selector: nil)
       @logger.reset
 
       @logger.phase_heading("Initializing restart")
-      verify_namespace
-      deployments = identify_target_deployments(deployments_names, selector: selector)
-      if kubectl.server_version < Gem::Version.new(MIN_KUBE_VERSION)
-        @logger.warn(KubernetesDeploy::Errors.server_version_warning(kubectl.server_version))
+      errors = Validator.new(@task_config).errors
```
KnVerey: Nit: can we make this name more specific? `TaskConfigValidator` by the looks of it? If this only validates the TaskConfig object though, why not make the interface `task_config.valid?` directly?

lib/kubernetes-deploy/validator.rb
```diff
@@ -0,0 +1,86 @@
...
+          "Cloud not find Namespace: #{namespace} in Context: #{context}"
+        else
+          "Could not connect to kubernetes cluster. #{err}"
+        end
+      end
+    end
+
+    def validate_server_version
+      if kubectl.server_version < Gem::Version.new(MIN_KUBE_VERSION)
+        logger.warn(KubernetesDeploy::Errors.server_version_warning(kubectl.server_version))
```
KnVerey: Maybe this `server_version_warning` thing should belong to this class now?

Issue #534, opened 2019-08-17
Ruby >=2.4.4 required


basert:
# Bug report

Installing the gem on ruby < 2.4.x errors due to zeitwerk requiring ruby 2.4.4. zeitwerk is a transitive dependency of https://rubygems.org/gems/activesupport which recently got updated to version 6.x. The version of the dependency should be locked.

**Expected behavior:** Installing on ruby < 2.4.x should be possible

**Actual behavior:** Installation fails

### Steps to Reproduce

1. Use ruby < 2.4.x
2. run `gem install kubernetes-deploy`

dturn:
As part of the 1.0 release we are officially not supporting ruby 2.3

PR #535, opened 2019-08-17
Adjust activesupport dependency to limit upper version to <6.0.0

basert:
This pull request tries to solve issue https://github.com/Shopify/kubernetes-deploy/issues/534

### DIFF
--- a/kubernetes-deploy.gemspec
+++ b/kubernetes-deploy.gemspec
@@ -23,7 +23,7 @@ Gem::Specification.new do |spec|
   spec.require_paths = %w(lib)
 
   spec.required_ruby_version = '>= 2.3.0'
-  spec.add_dependency("activesupport", ">= 5.0")
+  spec.add_dependency("activesupport", "~> 5.0")
   spec.add_dependency("kubeclient", "~> 4.3")
   spec.add_dependency("googleauth", "~> 0.8.0")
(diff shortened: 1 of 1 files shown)


dturn:
@basert Do you actually need ruby 2.3? (All support for Ruby 2.3 ended in March https://www.ruby-lang.org/en/news/2019/03/31/support-of-ruby-2-3-has-ended/). If not I'm thinking bumping our min ruby version 2.4 might be a bette approach here.

basert:
Unfortunately company requires the use of Debian stretch and no backports. https://packages.debian.org/stretch/ruby seems to only include 2.3.3

dturn:
I'm actually against capping the active support version to support an EoL version of Ruby. I'll leave it up to the other maintainers to voice an opinion on when we're going to drop support for ruby 2.3. For now can we use something like: 

```
if RUBY_VERSION >= '2.4'
  spec.add_dependency("activesupport", "~> 5.0")
else
  spec.add_dependency("activesupport", ">= 5.0")
end
```

KnVerey:
> I'll leave it up to the other maintainers to voice an opinion on when we're going to drop support for ruby 2.3. 

How about we do as you suggest (conditional) right now and drop 2.3 with the move to `krane`?

dturn:
As part of the 1.0 release we are officially not supporting ruby 2.3

PR #536, opened 2019-08-19
Ruby 2.4

dturn:
End our support of Ruby 2.3 since Ruby 2.3 is no longer supported.

**How is this accomplished?**
Bump some things.

**What could go wrong?**
There isn't anything that requires 2.4, though there appears to be some [gem version fun](https://github.com/Shopify/kubernetes-deploy/pull/535). We could since the burden to support 2.3 is trivial today we could keep supporting. However, I'm of the belief that we shouldn't tie ourselves to unsupported versions.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -2,7 +2,7 @@ inherit_from:
   - http://shopify.github.io/ruby-style-guide/rubocop.yml
 
 AllCops:
-  TargetRubyVersion: 2.3
+  TargetRubyVersion: 2.4
 
 Naming/FileName:
   Enabled: true
--- a/.shopify-build/kubernetes-deploy.yml
+++ b/.shopify-build/kubernetes-deploy.yml
@@ -1,6 +1,6 @@
 containers:
   default:
-    docker: circleci/ruby:2.5.1-node-browsers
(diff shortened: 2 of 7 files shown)


lib/kubernetes-deploy/options_helper.rb
```diff
@@ -26,7 +26,7 @@ def default_template_dir(template_dir)
           template_dir = File.join("config", "deploy", ENV['ENVIRONMENT'])
         end
 
-        if !template_dir || template_dir.empty?
```
KnVerey: This isn't related, is it? This change just makes the options helper depend on `ActiveSupport`. It's probably always available, but you'd need to tophat all the exes to confirm.

dturn: New version of rubocop flagged this.

KnVerey: Really? That's so weird... `blank` still isn't part of ruby core afaict. Also, it looks like the `lint` step passed on your first two commits on CI.

dturn: huh, so turns out rubocop is now fine with this file reverted.

Issue #525, opened 2019-08-14
Add support `krane.shopify.io` annotations


dturn:
## Problem
We're transitioning from `kubernetes-deploy` to `krane`. This means that we need to support `krane.shopify.io` annotations. 

## Possible solution
Create new constants and use both.

## Acceptance criteria
- A user can use current annotations
- A user can use `krane.shopify.io` for annotations 

## Additional context
https://github.com/Shopify/kubernetes-deploy/issues/256#issuecomment-475580770

Issue #527, opened 2019-08-14
Rename code references of `kubernetes-deploy` to `krane`


dturn:
## Problem

We're transitioning from kubernetes-deploy to krane. This means we need to remove all references to `kubernetes-deploy` in the code base. 

This issue excludes annotations which are handled in #526 .
This issue excludes the gemspec which is handled in #528 .

## Possible solution
- Update the docs
- Update classes & modules

## Alternative solutions

Split the documentation updates into its own PR. 

## Acceptance criteria
- Results from _greping_  'kubernetes-deploy' returns only annotations, historical change log entries, and the gemspec, shipit.yml. 

## Additional context
https://github.com/Shopify/kubernetes-deploy/issues/256#issuecomment-475580770

KnVerey:
Note: whoever pointed out the other day that our entrypoint methods still don't all have the same name was correct. `RestartTask`'s methods need to be renamed to `run`/`run!` as part of the Krane rename. Before we do the actual release, we should also audit the arguments to make sure they're sensible and consistent (same names for same thing, required vs. optional, on initialize vs on run).

dturn:
>Note: whoever pointed out the other day that our entrypoint methods still don't all have the same name was correct. RestartTask's methods need to be renamed to run/run! as part of the Krane rename. 

Tracking the interface rename in https://github.com/Shopify/kubernetes-deploy/issues/530

>Before we do the actual release, we should also audit the arguments to make sure they're sensible and consistent (same names for same thing, required vs. optional, on initialize vs on run).

Tracking the audit in https://github.com/Shopify/kubernetes-deploy/issues/528. But ideally any work that needs to be done to bring our public interfaces into compliance will be done with the issue that adds the new task.

PR #539, opened 2019-08-28
Add support for Krane annotations

adrianna-chang-shopify:
Closes: https://github.com/Shopify/kubernetes-deploy/issues/525

As described in the issue, we want to support new Krane annotations while allowing the current `kubernetes-deploy` annotations to remain in use.

I think we should break this down into three steps:
1) Add support for new Krane annotations
- [x] kubernetes-deploy.shopify.io/timeout-override
- [x] kubernetes-deploy.shopify.io/prunable
- [x] kubernetes-deploy.shopify.io/predeployed
- [x] kubernetes-deploy.shopify.io/instance-rollout-conditions 
- [x] kubernetes-deploy.shopify.io/instance-timeout
- [x] kubernetes-deploy.shopify.io/required-rollout
2) Add documentation to indicate that the new annotations are preferred, and that the old ones will be deprecated
3) (Maybe?) Raise a warning in the CLI if a `kubernetes-deploy` annotation is used, indicating that `krane.shopify.io` should be used instead

**How is this accomplished?**
- Need to add support for both annotations at the same time, determining which one is being used in the template when performing validations
- Need to duplicate tests for annotation-specific behaviour
(Chatted with @dturn and we decided that duplicating _every_ test that makes use of an annotated resource is unnecessary)
- Add some docs / warnings

**What could go wrong?**
- I may have missed an annotation so double check?
- Quick note: There is currently an EJSON annotation (`kubernetes-deploy.shopify.io/ejson-secret`). Discussion about it's purpose can be found here: https://github.com/Shopify/kubernetes-deploy/pull/424. I've excluded it from this PR - thinking we might want to tackle that for this issue instead https://github.com/Shopify/kubernetes-deploy/issues/527 since it's really just a tracking reference.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@
 - We've added a new Krane cli. This code is in alpha. We are providing
 no warranty at this time and reserve the right to make major breaking changes including
 removing it entirely at any time. ([#256](https://github.com/Shopify/kubernetes-deploy/issues/256))
+- Deprecate `kubernetes-deploy.shopify.io` annotations in favour of `krane.shopify.io`.
 
 ## 0.27.0
 *Enhancements*
--- a/lib/kubernetes-deploy/deploy_task.rb
+++ b/lib/kubernetes-deploy/deploy_task.rb
@@ -255,6 +255,11 @@ module KubernetesDeploy
       KubernetesDeploy::Concurrency.split_across_threads(resources) do |r|
         r.validate_definition(kubectl, selector: @selector)
       end
+
+      resources.select(&:has_warnings?).each do |resource|
+        record_warnings(warning: resource.validation_warning_msg, filename: File.basename(resource.file_path))
+      end
+
       failed_resources = resources.select(&:validation_failed?)
       return unless failed_resources.present?
 
@@ -317,6 +322,12 @@ module KubernetesDeploy
         end
       end
       @logger.summary.add_paragraph(debug_msg)
+    end
+
+    def record_warnings(warning:, filename:)
+      warn_msg = "Template warning: #{filename}\n"
+      warn_msg += "> Warning message:\n#{FormattedLogger.indent_four(warning)}"
+      @logger.summary.add_paragraph(ColorizedString.new(warn_msg).yellow)
     end
 
     def validate_configuration(allow_protected_ns:, prune:)
--- a/lib/kubernetes-deploy/kubernetes_resource.rb
+++ b/lib/kubernetes-deploy/kubernetes_resource.rb
@@ -30,7 +30,9 @@ module KubernetesDeploy
       If you have reason to believe it will succeed, retry the deploy to continue to monitor the rollout.
       MSG
 
-    TIMEOUT_OVERRIDE_ANNOTATION = "kubernetes-deploy.shopify.io/timeout-override"
+    TIMEOUT_OVERRIDE_ANNOTATION_SUFFIX = "timeout-override"
+    TIMEOUT_OVERRIDE_ANNOTATION_DEPRECATED = "kubernetes-deploy.shopify.io/#{TIMEOUT_OVERRIDE_ANNOTATION_SUFFIX}"
+    TIMEOUT_OVERRIDE_ANNOTATION = "krane.shopify.io/#{TIMEOUT_OVERRIDE_ANNOTATION_SUFFIX}"
     LAST_APPLIED_ANNOTATION = "kubectl.kubernetes.io/last-applied-configuration"
     SENSITIVE_TEMPLATE_CONTENT = false
 
@@ -93,7 +95,8 @@ module KubernetesDeploy
 
     def timeout_override
       return @timeout_override if defined?(@timeout_override)
-      @timeout_override = DurationParser.new(timeout_annotation).parse!.to_i
+
+      @timeout_override = DurationParser.new(krane_annotation_value(TIMEOUT_OVERRIDE_ANNOTATION_SUFFIX)).parse!.to_i
     rescue DurationParser::ParsingError
       @timeout_override = nil
     end
@@ -113,6 +116,7 @@ module KubernetesDeploy
       @statsd_report_done = false
       @disappeared = false
       @validation_errors = []
+      @validation_warnings = []
       @instance_data = {}
     end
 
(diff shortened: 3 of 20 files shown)


RyanBrushett:
## Warning

Screenshot 2019-09-06 at 10 34 17 AM

lib/kubernetes-deploy/kubernetes_resource.rb
```diff
@@ -400,16 +413,34 @@ def validate_timeout_annotation
...
+      end
     end
 
     def timeout_annotation
-      @definition.dig("metadata", "annotations", TIMEOUT_OVERRIDE_ANNOTATION)
+      @definition.dig("metadata", "annotations", TIMEOUT_OVERRIDE_ANNOTATION_DEPRECATED) ||
+        @definition.dig("metadata", "annotations", TIMEOUT_OVERRIDE_ANNOTATION)
+    end
+
+    def annotation_prefix
```
dturn: I think we should DRY this method and `timeout_annotation` to just take the part after the `/`

lib/kubernetes-deploy/kubernetes_resource.rb
```diff
@@ -122,12 +125,22 @@ def to_kubeclient_resource
 
     def validate_definition(kubectl, selector: nil)
```
dturn: Nor for this PR: I think we'll be able to significantly shrink the size of this method via the `TaskConfigValidator` which is probably the right point to decide if this really belongs in this class or some new object.

test/integration-serial/serial_deploy_test.rb
```diff
@@ -135,6 +135,32 @@ def test_crd_can_fail
...
 
+  def test_crd_pruning_deprecated
+    assert_deploy_success(deploy_fixtures("crd", subset: %w(mail.yml widgets_deprecated.yml)))
+    assert_logs_match_all([
+      "Phase 1: Initializing deploy",
+      "Detected non-namespaced resources which will never be pruned:",
+      " - CustomResourceDefinition/mail.stable.example.io",
+      "Phase 3: Deploying all resources",
+      "CustomResourceDefinition/mail.stable.example.io (timeout: 120s)",
+      %r{CustomResourceDefinition/mail.stable.example.io\s+Names accepted},
```
dturn: should we assert for the deprecation warning?

test/unit/kubernetes-deploy/kubernetes_resource/deployment_test.rb
```diff
@@ -365,6 +422,34 @@ def test_deploy_timed_out_based_on_progress_deadline_ignores_statuses_for_older_
 
   private
 
+  def build_deployment_template_deprecated(status: { 'replicas' => 3 }, rollout: nil,
```
dturn: Why a new method over a new param to the exiting method?

test/unit/kubernetes-deploy/kubernetes_resource_test.rb
```diff
@@ -97,6 +97,35 @@ def test_fetch_events_returns_empty_hash_when_kubectl_results_empty
     assert_operator(events, :empty?)
   end
 
+  def test_deprecated_annotation_generates_warning
+    customized_resource = DummyResource.new(definition_extras: build_timeout_metadata_deprecated("60S"))
+    customized_resource.validate_definition(kubectl)
+    assert(customized_resource.has_warnings?, "Deprecated annotation did not generate a warning")
+    assert_equal("kubernetes-deploy.shopify.io/timeout-override is deprecated: "\
+      "Use 'krane.shopify.io' annotations instead",
```
dturn: I'm worried the  current message could be mistaken for depreciating the entire annotation instead of just the prefix what about something like: `kubernetes-deploy.shopify.io` prefix for annotations deprecated please switch to the `krane.shopify.io` prefix?

CHANGELOG.md
```diff
@@ -8,6 +8,9 @@
 - **[Breaking change]** Added ServiceAccount, PodTemplate, ReplicaSet, Role, and RoleBinding to the prune whitelist.
   * To see what resources may be affected, run `kubectl get $RESOURCE -o jsonpath='{ range .items[*] }{.metadata.namespace}{ "\t" }{.metadata.name}{ "\t" }{.metadata.annotations}{ "\n" }{ end }' --all-namespaces | grep "last-applied"`
   * To exclude a resource from kubernetes-deploy (and kubectl apply) management, remove the last-applied annotation `kubectl annotate $RESOURCE $SECRET_NAME kubectl.kubernetes.io/last-applied-configuration-`.
+  
+*Other*
```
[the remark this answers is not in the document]
RyanBrushett: Shit I rebased and missed that, sorry.

test/unit/kubernetes-deploy/kubernetes_resource_test.rb
RyanBrushett: Great shout, will adjust 👍

RyanBrushett:
Not at all, I went back over it and got it sorted out. Sorry for overlooking that.

Ready for another 👀

lib/kubernetes-deploy/deploy_task.rb
```diff
@@ -255,6 +255,11 @@ def validate_resources(resources)
       KubernetesDeploy::Concurrency.split_across_threads(resources) do |r|
         r.validate_definition(kubectl, selector: @selector)
       end
+
+      resources.select(&:has_warnings?).each do |r|
```
dturn: style nit: `r` -> `resource` (or at least something that isn't a single letter var for a multi-line block).

RyanBrushett: I'll fix it in this one place. Fixing it all over the file will make it harder to review for others. In a smaller PR or a separate one I can pick at cleaning it up everywhere.

lib/kubernetes-deploy/deploy_task.rb
```diff
@@ -319,6 +324,12 @@ def record_invalid_template(err:, filename:, content: nil)
       @logger.summary.add_paragraph(debug_msg)
     end
 
+    def record_warnings(warning:, filename:)
```
dirceu: Do we need to support this for any other task? For example, we have access to warnings here as well:

https://github.com/Shopify/kubernetes-deploy/blob/3d5cf5964932658a91e3e5621719ffaaa7ee0aac/lib/kubernetes-deploy/runner_task.rb#L40

lib/kubernetes-deploy/kubernetes_resource.rb
```diff
@@ -396,20 +410,42 @@ def global?
...
-      @validation_errors << "#{TIMEOUT_OVERRIDE_ANNOTATION} annotation is invalid: #{e}"
+      @validation_errors << "#{timeout_annotation_key} annotation is invalid: #{e}"
+    end
+
+    def validate_annotation_version
+      annotation_keys = @definition.dig("metadata", "annotations")&.keys
+      annotation_keys&.each do |annotation|
+        if annotation.include?("kubernetes-deploy.shopify.io")
+          annotation_prefix = annotation.split('/').first
+          @validation_warnings << "#{annotation_prefix} as a prefix for annotations is deprecated: "\
```
jessie-JNing: from my understanding of the function, if there're two annotation keys both with deprecated prefix, this warning message will show twice, do we want to do that?

Issue #265, opened 2018-04-04
Services for zero-replica statefulsets time out


KnVerey:
# Bug report

If you deploy a StatefulSet + Service pair where the SS's replicas is set to zero, the service will eventually time out with a message like this:

```
[FATAL][2018-04-04 18:52:53 +0000]	Service/thing: TIMED OUT (timeout: 420s)
[FATAL][2018-04-04 18:52:53 +0000]	This service does not seem to select any pods. This means its spec.selector is probably incorrect.
[FATAL][2018-04-04 18:52:53 +0000]	  - Final status: Selects 0 pods
[FATAL][2018-04-04 18:52:53 +0000]	  - Events: None found. Please check your usual logging service (e.g. Splunk).
```
We've explicitly handled this for zero-replica deployments, but services can be placed in front of any pod-generating controller.

**Expected behavior:** The deploy recognizes that these services aren't meant to have endpoints and succeeds.

**Actual behavior:** The deploy times out.

**Version(s) affected:** 0.18.0

### Steps to Reproduce

Deploy a StatefulSet + Service pair where the SS's replicas is set to zero

KnVerey:
In fact, the logic we're using for services is flawed and doesn't acknowledge the full range of service types possible. We should revisit it in general.

dturn:
> We've explicitly handled this for zero-replica deployments, but services can be placed in front of any pod-generating controller.

This is true, but sticking a service in front of a CronJob/Job seems unusual enough that I don't think we need to handle this case.

PR #540, opened 2019-08-28
Support StatefulSets with 0 replicas

dturn:

Currently when one deploys a StatefulSet + Service pair where the SS's replicas is set to zero, the service will eventually time out. This behavior is wrong, its valid for a service to point at a SS that's intentionally been scaled down. We should treat this case the same as a deployment. 

fixes: https://github.com/Shopify/kubernetes-deploy/issues/265

**How is this accomplished?**
Have the service check to see if it's actually pointing at a deployment or statefulset that has been scaled down. 

**What could go wrong?**

- We're intentionally not handling the case of a service in front of other pod generating controllers: CronJob/Job.  The assumption here is its unusual to point a service at  collection of pods that normally isn't running. Further since they normally aren't running, I'd have expected a bug report about it by now.

- This keeps the same bad assumption that the pod selector will also select the deployment / SS that was raised in https://github.com/Shopify/kubernetes-deploy/issues/483. I've got a second PR to solve it https://github.com/Shopify/kubernetes-deploy/pull/541 so feel ok doing this in two steps.

### DIFF
--- a/lib/kubernetes-deploy/kubernetes_resource/service.rb
+++ b/lib/kubernetes-deploy/kubernetes_resource/service.rb
@@ -7,9 +7,11 @@ module KubernetesDeploy
       super
       if exists? && selector.present?
         @related_deployments = cache.get_all(Deployment.kind, selector)
+        @related_statefulsets = cache.get_all(StatefulSet.kind, selector)
         @related_pods = cache.get_all(Pod.kind, selector)
       else
         @related_deployments = []
+        @related_statefulsets = []
         @related_pods = []
       end
     end
@@ -30,7 +32,7 @@ module KubernetesDeploy
       return false unless exists?
       return exists? unless requires_endpoints?
       # We can't use endpoints if we want the service to be able to fail fast when the pods are down
-      exposes_zero_replica_deployment? || selects_some_pods?
+      exposes_zero_replica_workload? || selects_some_pods?
     end
 
     def deploy_failed?
@@ -43,7 +45,7 @@ module KubernetesDeploy
 
     private
 
-    def exposes_zero_replica_deployment?
+    def exposes_zero_replica_workload?
       return false unless related_replica_count
       related_replica_count == 0
     end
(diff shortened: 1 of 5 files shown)


lib/kubernetes-deploy/kubernetes_resource/service.rb
```diff
@@ -7,9 +7,11 @@ def sync(cache)
       super
       if exists? && selector.present?
         @related_deployments = cache.get_all(Deployment.kind, selector)
+        @related_statefulsets = cache.get_all(StatefulSet.kind, selector)
```
timothysmith0609: I think this is fine for now, but if we find ourselves having to accommodate some other kind of workload we should consider aggregating things into a `@related_workloads = {}` to simplify processing

dturn: Yep, tackling that in https://github.com/Shopify/kubernetes-deploy/pull/541/files#diff-bfab9900d04f0a61190aa04f896aed1fR10

timothysmith0609: :+1: Yep just saw that and was on my way to delete this comment

dturn:
>Should we warn when a service is targeting a workload with (explicitly-set) 0 replicas, or do we trust the user to know what's going on?

We've got:
```
 Service/stateful-busybox Doesn't require any endpoint
 StatefulSet/stateful-busybox 0 replicas
```
in the output already, that seems like enough to me.

>The timeout message currently states: This service does not seem to select any pods. This means its spec.selector is probably incorrect.. This is not necessarily the timeout condition, though. What we want to say is that no pods are selected but there should be.


How about:

"This service does not seem to select any pods. This is likely invalid please confirm the spec.selector is correct and the targeted workload is healthy."

lib/kubernetes-deploy/kubernetes_resource/service.rb
```diff
@@ -30,20 +32,22 @@ def deploy_succeeded?
...
     end
 
     def deploy_failed?
       false
     end
 
     def timeout_message
-      "This service does not seem to select any pods. This means its spec.selector is probably incorrect."
+      "This service does not seem to select any pods. "\
+      "This is likely invalid please confirm the spec.selector is correct and "\
```
RyanBrushett:
```suggestion
      "This is likely invalid. Please confirm the spec.selector is correct and "\
```

⛏

PR #541, opened 2019-08-28
Search all workloads if Pod selector doesn't match any workloads

dturn:

Currently a service assumes that the pod selector will also match the workload. This isn't always true. Because it's possible for there to be no pods that match the service's pod-selector (when the workload is scaled to 0). We need to check every workload instead of assuming no such workload exists. 

Closes: https://github.com/Shopify/kubernetes-deploy/issues/483

**How is this accomplished?**
Search all workloads to find the correct set.

**What could go wrong?**

- If the namespace has many StatefulSets and/or Deployments  this could be a lot of data to fetch and check.

- Should we always search everything? We could only search everything when there are no pods. And if there are pods walk the owner references back. However, walking the owner refs would be an extra API call though with less data...

### DIFF
--- a/lib/kubernetes-deploy/kubernetes_resource/service.rb
+++ b/lib/kubernetes-deploy/kubernetes_resource/service.rb
@@ -9,9 +9,7 @@ module KubernetesDeploy
       super
       if exists? && selector.present?
         @related_pods = cache.get_all(Pod.kind, selector)
-        # Optimization: assume that the pod selector matches the workload selector
-        @related_workloads = workloads_fetcher(cache, selector)
-        @related_workloads = workloads_fetcher(cache, nil) if @related_workloads.blank?
+        @related_workloads = fetch_related_workloads(cache)
       else
         @related_pods = []
         @related_workloads = []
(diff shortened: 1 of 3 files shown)


lib/kubernetes-deploy/kubernetes_resource/service.rb
```diff
@@ -6,13 +6,13 @@ class Service < KubernetesResource
     def sync(cache)
       super
       if exists? && selector.present?
-        @related_deployments = cache.get_all(Deployment.kind, selector)
-        @related_statefulsets = cache.get_all(StatefulSet.kind, selector)
         @related_pods = cache.get_all(Pod.kind, selector)
+        # Optimization: assume that the pod selector matches the workload selector
+        @related_workloads = workloads_fetcher(cache, selector)
```
dturn: This 'optimization' wont always do the right thing. I suspect it needs to be removed

timothysmith0609:
>If the namespace has many StatefulSets and/or Deployments this could be a lot of data to fetch and check.

I would take the hit and we can reconsider this if it becomes problematic

>Should we always search everything? We could only search everything when there are no pods. And if there are pods walk the owner references back. However, walking the owner refs would be an extra API call though with less data...

I think we should grab everything we (could) need in `sync`; I don't really see a clean way around that. Trying to selectively grab resources only when there's no pods feels like it will open us up to strange edge cases.

PR #545, opened 2019-09-04
v0.27.0

timothysmith0609:
Cut a release that includes:

- `-f` support (#514)
- Add ServiceAccount, PodTemplate, ReplicaSet, Role, and RoleBinding to prune whitelist (#498)
- Initial implementation of shared validations for tasks (#533)
- Better usage of `require` statements for tasks. We no longer require code we don't need for a given task. (#487)

**How is this accomplished?**
Bump version followed by Shipit deploy once this merges

**What could go wrong?**
Let's do a bit of tophatting before releasing this and make sure everything we want it in the changelog

cc @Shopify/production-excellence

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,5 @@
+## next
+
 ## 0.27.0
 *Enhancements*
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -1,6 +1,8 @@
-## next
```
dturn: We usually keep this at the top.

Issue #543, opened 2019-08-30
Ensure Service class successfully deploys for all valid scenarios  


dturn:
The `Service` class needs some ❤️ . Currently it works for the most common uses, but there are real but less frequent cases that we do not handle correctly. #265 & #483 target two documented issues, but this issue is to audit the class to ensure we haven't missed any other service configurations that are valid but would lead to a failed deploy.

dturn:
I read through https://kubernetes.io/docs/concepts/services-networking/service/ and the only case we aren't covering that jumped out at me was:
- [ ] `type: LoadBalancer` we could verify that `status.LoadBalancer` is populated.

dturn:
Not required but exploring https://github.com/Shopify/kubernetes-deploy/issues/135 as part of this task might be worthwhile

dirceu:
The only other thing I found specifically related to Service is that NodePort may require port validation. I 🎩 and the tool already works as expected, so I think there's nothing to do for this case.

With invalid port:

```
[FATAL][2019-09-10 09:05:08 -0400]      WARNING: Any resources not mentioned in the error(s) below were likely created/updated. You may wish to roll back this deploy.
[FATAL][2019-09-10 09:05:08 -0400]
[FATAL][2019-09-10 09:05:08 -0400]      Unidentified error(s):
[FATAL][2019-09-10 09:05:08 -0400]          The Service "test-lb" is invalid: spec.ports[0].nodePort: Invalid value: 123123: provided port is not in the valid range. The range of valid ports is 30000-32767
```

With a port already in use:

```
[FATAL][2019-09-10 09:07:28 -0400]      WARNING: Any resources not mentioned in the error(s) below were likely created/updated. You may wish to roll back this deploy.
[FATAL][2019-09-10 09:07:28 -0400]
[FATAL][2019-09-10 09:07:28 -0400]      Unidentified error(s):
[FATAL][2019-09-10 09:07:28 -0400]          The Service "test-lb2" is invalid: spec.ports[0].nodePort: Invalid value: 30001: provided port is already allocated
```

dturn:
I'm surprised 'port is not in the valid range' isn't detected in `--dry-run`. Do you know if  `--server-dry-run` detects it. Might add strength to the argument we should be using that everywhere.

dirceu:
> Do you know if --server-dry-run detects it?

It does, but only if the initial Service that required that port was already scheduled:

If neither exists:
```
[...]

PR #547, opened 2019-09-06
Verify status.loadBalancer when deploying a service of type LoadBalancer

dirceu:

Fixes part of https://github.com/Shopify/kubernetes-deploy/issues/543.

Make sure that we only declare a service of type `LoadBalancer` as deployed after its IP address is published (which is an asynchronous process in the case of most cloud providers).

According to the documentation:

> The actual creation of the load balancer happens asynchronously, and information about the provisioned balancer is published in the Service’s .status.loadBalancer field

After being fully provisioned, this will show as something like this:

```yaml
status:
  loadBalancer:
    ingress:
      - ip: 146.148.47.155
```

**How is this accomplished?**

I added an extra condition that will only run if the type of the Service is `LoadBalancer`; if that's the case, it will check for the presence of `status.loadBalancer.ingress.ip` to determine if the Service has been fully provisioned.

**What could go wrong?**

It seems that testing this particular condition is [not possible with `minikube`](https://kubernetes.io/docs/tasks/access-application-cluster/create-external-load-balancer/#finding-your-ip-address), as (specifically in that environment) the IP address is never published to the resource data and you need a special command to retrieve it.

**Live test**

```
$ kubernetes-deploy dirceu-test tierstaging-us-east1-2 --template-dir ~/Desktop
[INFO][2019-09-09 08:28:54 -0400]
[INFO][2019-09-09 08:28:54 -0400]	------------------------------------Phase 1: Initializing deploy------------------------------------
[INFO][2019-09-09 08:28:55 -0400]	Context tierstaging-us-east1-2 found
[INFO][2019-09-09 08:28:55 -0400]	Namespace dirceu-test found
[INFO][2019-09-09 08:28:55 -0400]	All required parameters and files are present
[INFO][2019-09-09 08:28:55 -0400]	Discovering resources:
[INFO][2019-09-09 08:28:56 -0400]	  - Service/test-lb
[INFO][2019-09-09 08:28:56 -0400]
[INFO][2019-09-09 08:28:56 -0400]	----------------------------Phase 2: Checking initial resource statuses-----------------------------
[INFO][2019-09-09 08:28:56 -0400]	Service/test-lb                                   Not found
[INFO][2019-09-09 08:28:56 -0400]
[INFO][2019-09-09 08:28:56 -0400]	----------------------------------Phase 3: Deploying all resources----------------------------------
[INFO][2019-09-09 08:28:56 -0400]	Deploying Service/test-lb (timeout: 420s)
[INFO][2019-09-09 08:29:30 -0400]	Still waiting for: Service/test-lb
[INFO][2019-09-09 08:30:00 -0400]	Still waiting for: Service/test-lb
[INFO][2019-09-09 08:30:30 -0400]	Still waiting for: Service/test-lb
[INFO][2019-09-09 08:30:54 -0400]	Successfully deployed in 118.0s: Service/test-lb
[INFO][2019-09-09 08:30:54 -0400]
[INFO][2019-09-09 08:30:54 -0400]	------------------------------------------Result: SUCCESS-------------------------------------------
[INFO][2019-09-09 08:30:54 -0400]	Pruned 1 resource and successfully deployed 1 resource
[INFO][2019-09-09 08:30:54 -0400]
[INFO][2019-09-09 08:30:54 -0400]	Successful resources
[INFO][2019-09-09 08:30:54 -0400]	Service/test-lb                                   Doesn't require any endpoints

$ kubectl --context tierstaging-us-east1-2 -n dirceu-test get service test-lb -o yaml
apiVersion: v1
kind: Service
metadata:
  annotations:
    kubectl.kubernetes.io/last-applied-configuration: |
      {"apiVersion":"v1","kind":"Service","metadata":{"annotations":{},"labels":{"app":"test-lb","name":"test-lb"},"name":"test-lb","namespace":"dirceu-test"},"spec":{"ports":[{"name":"http","port":80,"targetPort":"http"}],"type":"LoadBalancer"}}
  creationTimestamp: "2019-09-09T12:28:57Z"
  labels:
    app: test-lb
    name: test-lb
  name: test-lb
  namespace: dirceu-test
  resourceVersion: "39073112"
  selfLink: /api/v1/namespaces/dirceu-test/services/test-lb
  uid: 64fe58ff-d2fd-11e9-8940-58010a8e017b
spec:
  clusterIP: <REDACTED>
  externalTrafficPolicy: Cluster
  ports:
  - name: http
    nodePort: 32731
    port: 80
    protocol: TCP
    targetPort: http
  sessionAffinity: None
  type: LoadBalancer
status:
  loadBalancer:
    ingress:
    - ip: <REDACTED>
```

### DIFF
--- a/lib/kubernetes-deploy/kubernetes_resource/service.rb
+++ b/lib/kubernetes-deploy/kubernetes_resource/service.rb
@@ -19,6 +19,8 @@ module KubernetesDeploy
     def status
       if !exists?
         "Not found"
+      elsif requires_publishing? && !published?
+        "LoadBalancer IP address is not provisioned yet"
       elsif !requires_endpoints?
         "Doesn't require any endpoints"
       elsif selects_some_pods?
@@ -30,6 +32,7 @@ module KubernetesDeploy
 
     def deploy_succeeded?
       return false unless exists?
+      return published? if requires_publishing?
       return exists? unless requires_endpoints?
       # We can't use endpoints if we want the service to be able to fail fast when the pods are down
       exposes_zero_replica_workload? || selects_some_pods?
@@ -88,6 +91,14 @@ module KubernetesDeploy
 
     def external_name_svc?
       @definition["spec"]["type"] == "ExternalName"
+    end
+
+    def requires_publishing?
+      @definition["spec"]["type"] == "LoadBalancer"
+    end
+
+    def published?
+      @instance_data.dig('status', 'loadBalancer', 'ingress').present?
     end
   end
 end
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 ## next
 
+*Enhancements*
+- Make sure that we only declare a Service of type LoadBalancer as deployed after its IP address is published. [#547](https://github.com/Shopify/kubernetes-deploy/pull/547)
+
 *Other*
 
 - We've added a new Krane cli. This code is in alpha. We are providing
--- a/test/fixtures/for_unit_tests/service_test.yml
+++ b/test/fixtures/for_unit_tests/service_test.yml
@@ -200,3 +200,12 @@ spec:
       containers:
       - name: app
         image: busybox
+---
+apiVersion: v1
+kind: Service
+metadata:
+  name: standard-lb
+spec:
+  type: LoadBalancer
+  selector:
+    type: standard
--- a/test/unit/kubernetes-deploy/kubernetes_resource/service_test.rb
+++ b/test/unit/kubernetes-deploy/kubernetes_resource/service_test.rb
@@ -151,6 +151,39 @@ class ServiceTest < KubernetesDeploy::TestCase
     assert_equal("Doesn't require any endpoints", svc.status)
   end
 
+  def test_ensures_populated_status_for_load_balancers
+    svc_def = service_fixture('standard-lb')
+    svc = build_service(svc_def)
+
+    stub_kind_get("Service", items: [svc_def])
+    stub_kind_get("Deployment", items: deployment_fixtures)
+    stub_kind_get("Pod", items: pod_fixtures)
+    stub_kind_get("StatefulSet", items: [])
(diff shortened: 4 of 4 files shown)


lib/kubernetes-deploy/kubernetes_resource/service.rb
```diff
@@ -89,5 +92,13 @@ def related_replica_count
     def external_name_svc?
       @definition["spec"]["type"] == "ExternalName"
     end
+
+    def requires_publishing?
+      @definition["spec"]["type"] == "LoadBalancer"
+    end
+
+    def published?
+      !@instance_data.dig('status', 'loadBalancer', 'ingress', 'ip').nil?
```
dturn: I find `present?`  easier to read than  `!X.nil?`

dirceu: Actually had to change this completely; the data structure had a different shape than what I expected. 🎩 FTW!

lib/kubernetes-deploy/kubernetes_resource/service.rb
```diff
@@ -89,5 +92,14 @@ def related_replica_count
...
       @definition["spec"]["type"] == "ExternalName"
     end
+
+    def requires_publishing?
+      @definition["spec"]["type"] == "LoadBalancer"
+    end
+
+    def published?
+      ingresses = @instance_data.dig('status', 'loadBalancer', 'ingress')
+      ingresses && !ingresses.empty?
```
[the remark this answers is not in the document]
dirceu: As discussed offline - this was changed to `ingresses` because `ingress` is actually a (seemingly improperly named) array, not a single value. `@instance_data.dig('status', 'loadBalancer', 'ingress').present?` should work as well though, I'll update that! 👍

dturn:
You can merge this PR as is, but ❤️ if you'd add an update to `CHANGELOG.md` with this enhancement. (We're trying to do this in each PR instead since doing them in the PR that bumps the gem version is a bit painful).

Issue #549, opened 2019-09-09
Delete Shopify Specific Resources


dturn:
We should not have any Shopify specific code in this repo. It appears that two of the Shopify specific classes may now implement the default behavior. Investigate and remove or work with teams to determine what needs to be done to remove the code.

https://github.com/Shopify/kubernetes-deploy/blob/master/lib/kubernetes-deploy/kubernetes_resource/topic.rb 

https://github.com/Shopify/kubernetes-deploy/blob/master/lib/kubernetes-deploy/kubernetes_resource/elasticsearch.rb

PR #550, opened 2019-09-09
Delete resources that mirror default behavior

dturn:

We should not have any Shopify specific code in this repo. 

Closes: https://github.com/Shopify/kubernetes-deploy/issues/549

**How is this accomplished?**
Delete three Shopify specific resources. These resources appear to behave the same as the default CustomResource code. 

**What could go wrong?**
The default behavior is slightly different.


dturn:
I 🎩 this with `cumulus-cat` and didn't see any difference. Though `cumulus-cat`  only uses ES.

Issue #484, opened 2019-05-21
Error logs incomplete when there are sensitive info


peiranliushop:
# Bug report
Shipit stack deployment failure error message is not sufficient.

**Description:**
Deploy failure is reported [here](https://discourse.shopify.io/t/deploy-is-failing-with-a-running-kubernetes-job/3322)

kubernetes-deploy outputs error message 

`[FATAL][2019-05-17 17:52:25 +0000]	WARNING: There was an error applying some or all resources. The raw output may be sensitive and so cannot be displayed.
`

is not specific about which yaml file caused the failure and why. The error message is vague because there are k8s secrets in the template set `kubectl` applies to that are considered sensitive data. 
Currently kubernetes-deploy just blocks the kubectl error as a whole. It is hard to identify the cause of failure.

**Expected behavior:** 
Output error message should include at least which yaml files were failed during `kubectl -apply`

**Actual behavior:** 
Output a standard error message without details.

**Version(s) affected:**
0.26.4 and earlier versions

# Feature request

- [ ] If the maintainers agree with the feature as described here, I intend to submit a Pull Request myself.<sup>1</sup>

**Proposal:** [provide details on the behaviour you'd like to see and why it would be useful]

<sup><small>1</small></sup> <sub>This is the quickest way to get a new feature! We reserve the right to close feature requests, even ones we like, if the proposer does not intend to contribute to the feature and it doesn't fit in our current roadmap.</sub>

KnVerey:
> Output error message should include at least which yaml files were failed during kubectl -apply

It's necessarily all of them--we pass all the templates to apply, and it returns us a single error. That error is the only source of further information, and it is very difficult if not impossible to be certain that it does not reference secret content. Do you have more information about what the underlying error was in this case?

yfxie:
@peiranliushop same issue.
I found it because the keys of data in ejson can't include slash character.
My original data key is like `some/path/file.json`. After I changed it to be only filename `file.json`, it works now.

@KnVerey 
before I get using into ejson, I can create secret with key including slash character via:
```
kubectl create secret generic some-secret --from-file=some/path/file.json
```
This will create a secret with key equals `some/path/file.json` and its value is content of the file.
Any workaround can I keep slash character in the key of ejson?

peiranliushop:
> > Output error message should include at least which yaml files were failed during kubectl -apply
> 
[...]

PR #553, opened 2019-09-11
Add --server-dry-run feature 

jessie-JNing:
Work on: https://github.com/Shopify/kubernetes-deploy/issues/484
Make Krane use `--server-dry-run` during template validation phrase. 

The `--dry-run` we use before, the validation is only processed on client side, some failure during actual `apply` step cannot be captured, since `apply` happens in kubernetes server side.
The `--server-dry-run` will hit the server and tell us about the result of apply,  while not actually persisting it. 

Details about [--server-dry-run](https://kubernetes.io/docs/reference/using-api/api-concepts/#dry-run).

`--server-dry-run` was released in version [v1.12](https://github.com/kubernetes/kubernetes/blob/33541bdd34c32ad1dbef20a1282d7917b300d952/CHANGELOG-1.12.md#new-features).
while enabled as beta feature in version [v1.13](https://kubernetes.io/docs/reference/using-api/api-concepts/#dry-run).

In current stage, we attempt to apply server dry on to resources with sensitive content. This is because when some failure happen at deploy stage, logs info will be omitted if any of the template content is sensitive. Those validation failure should be caught during template validation, so try to run `--server-dry-run` if possible.  
 
**How is this accomplished?**
Add an overridable property `SERVER_DRY_RUNNABLE` to kubernetes-resource, default to be false. Make it true, if we would like to try `--server-dry-run` on that type of resource.

Breaks a number of tests.
- For unit tests, make it test against version above v1.13
- For integration tests, they are compatible with all supported versions

**What could go wrong?**
try to 🎩 as much as possible, may still miss some.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 - Officially support Kubernetes 1.15 ([#546](https://github.com/Shopify/kubernetes-deploy/pull/546))
 - Make sure that we only declare a Service of type LoadBalancer as deployed after its IP address is published. [#547](https://github.com/Shopify/kubernetes-deploy/pull/547)
 - Add more validations to `RunnerTask`. [#554](https://github.com/Shopify/kubernetes-deploy/pull/554)
+- Validate secrets with `--server-dry-run` on supported clusters. [#553](https://github.com/Shopify/kubernetes-deploy/pull/553) 
 
 
 *Bug Fixes*
--- a/lib/kubernetes-deploy/kubernetes_resource.rb
+++ b/lib/kubernetes-deploy/kubernetes_resource.rb
@@ -466,17 +466,16 @@ module KubernetesDeploy
     end
 
     def validate_spec_with_kubectl(kubectl)
-      _, err, st = validate_with_dry_run_option(kubectl, "--server-dry-run")
-      if !st.success? && err.match(SERVER_DRY_RUN_DISABLED_ERROR)
-        _, err, st = validate_with_dry_run_option(kubectl, "--dry-run")
+      _, err, st = validate_with_dry_run_option(kubectl, "--dry-run")
+      if st.success? && sensitive_template_content?
+        _, err, st = validate_with_dry_run_option(kubectl, "--server-dry-run")
+        unless st.success? || err.match(SERVER_DRY_RUN_DISABLED_ERROR)
+          @validation_errors << "Validation for #{id} failed. Detailed information is unavailable as the raw error may contain sensitive data."
+        end
       end
 
       return true if st.success?
-      @validation_errors << if sensitive_template_content?
-        "Validation for #{id} failed. Detailed information is unavailable as the raw error may contain sensitive data."
-      else
-        err
-      end
+      @validation_errors << err
     end
 
     def validate_with_dry_run_option(kubectl, dry_run_option)
--- a/lib/kubernetes-deploy/kubernetes_resource.rb
+++ b/lib/kubernetes-deploy/kubernetes_resource.rb
@@ -15,6 +15,8 @@ module KubernetesDeploy
     GLOBAL = false
     TIMEOUT = 5.minutes
     LOG_LINE_COUNT = 250
+    SERVER_DRY_RUN_DISABLED_ERROR =
+      /(unknown flag: --server-dry-run)|(doesn't support dry-run)|(dryRun alpha feature is disabled)/
 
     DISABLE_FETCHING_LOG_INFO = 'DISABLE_FETCHING_LOG_INFO'
     DISABLE_FETCHING_EVENT_INFO = 'DISABLE_FETCHING_EVENT_INFO'
@@ -464,9 +466,10 @@ module KubernetesDeploy
     end
 
     def validate_spec_with_kubectl(kubectl)
-      command = ["create", "-f", file_path, "--dry-run", "--output=name"]
-      _, err, st = kubectl.run(*command, log_failure: false, output_is_sensitive: sensitive_template_content?,
-        retry_whitelist: [:client_timeout], attempts: 3)
+      _, err, st = validate_with_dry_run_option(kubectl, "--server-dry-run")
+      if !st.success? && err.match(SERVER_DRY_RUN_DISABLED_ERROR)
+        _, err, st = validate_with_dry_run_option(kubectl, "--dry-run")
+      end
 
       return true if st.success?
       @validation_errors << if sensitive_template_content?
@@ -474,6 +477,12 @@ module KubernetesDeploy
       else
         err
       end
+    end
+
+    def validate_with_dry_run_option(kubectl, dry_run_option)
+      command = ["apply", "-f", file_path, dry_run_option, "--output=name"]
+      kubectl.run(*command, log_failure: false, output_is_sensitive: sensitive_template_content?,
+                               retry_whitelist: [:client_timeout], attempts: 3)
     end
 
     def labels
--- a/lib/kubernetes-deploy/kubernetes_resource.rb
+++ b/lib/kubernetes-deploy/kubernetes_resource.rb
@@ -469,13 +469,17 @@ module KubernetesDeploy
       _, err, st = validate_with_dry_run_option(kubectl, "--dry-run")
       if st.success? && sensitive_template_content?
         _, err, st = validate_with_dry_run_option(kubectl, "--server-dry-run")
-        unless st.success? || err.match(SERVER_DRY_RUN_DISABLED_ERROR)
-          @validation_errors << "Validation for #{id} failed. Detailed information is unavailable as the raw error may contain sensitive data."
+        if st.success? || err.match(SERVER_DRY_RUN_DISABLED_ERROR)
+          return true
         end
       end
 
       return true if st.success?
-      @validation_errors << err
+      @validation_errors << if sensitive_template_content?
+        "Validation for #{id} failed. Detailed information is unavailable as the raw error may contain sensitive data."
+      else
+        err
+      end
     end
 
     def validate_with_dry_run_option(kubectl, dry_run_option)
--- a/lib/kubernetes-deploy/kubernetes_resource.rb
+++ b/lib/kubernetes-deploy/kubernetes_resource.rb
@@ -37,6 +37,7 @@ module KubernetesDeploy
     TIMEOUT_OVERRIDE_ANNOTATION = "krane.shopify.io/#{TIMEOUT_OVERRIDE_ANNOTATION_SUFFIX}"
     LAST_APPLIED_ANNOTATION = "kubectl.kubernetes.io/last-applied-configuration"
     SENSITIVE_TEMPLATE_CONTENT = false
+    SERVER_DRY_RUNNABLE = false
 
     class << self
       def build(namespace:, context:, definition:, logger:, statsd_tags:, crd: nil)
@@ -345,6 +346,10 @@ module KubernetesDeploy
       self.class::SENSITIVE_TEMPLATE_CONTENT
     end
 
+    def server_dry_runnable?
+      self.class::SERVER_DRY_RUNNABLE
+    end
+
     class Event
       EVENT_SEPARATOR = "ENDEVENT--BEGINEVENT"
       FIELD_SEPARATOR = "ENDFIELD--BEGINFIELD"
@@ -467,7 +472,7 @@ module KubernetesDeploy
 
     def validate_spec_with_kubectl(kubectl)
       _, err, st = validate_with_dry_run_option(kubectl, "--dry-run")
-      if st.success? && sensitive_template_content?
+      if st.success? && server_dry_runnable?
         _, err, st = validate_with_dry_run_option(kubectl, "--server-dry-run")
         if st.success? || err.match(SERVER_DRY_RUN_DISABLED_ERROR)
           return true
--- a/test/integration/kubernetes_deploy_test.rb
+++ b/test/integration/kubernetes_deploy_test.rb
@@ -36,7 +36,7 @@ class KubernetesDeployTest < KubernetesDeploy::IntegrationTest
     assert_logs_match(%r{ConfigMap/hello-cloud-configmap-data\s+Available}, 1)
   end
 
-  def test_service_account_predeployed_before_unmanaged_pod
+  def test_unautomounted_service_account_predeployed_without_server_dry_run
     # Add a valid service account in unmanaged pod
     service_account_name = "build-robot"
     result = deploy_fixtures("hello-cloud",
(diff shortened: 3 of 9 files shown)


lib/kubernetes-deploy/kubernetes_resource.rb
```diff
@@ -471,5 +471,12 @@ def statsd_tags
       tags = %W(context:#{context} namespace:#{namespace} type:#{type} status:#{status})
       tags | @optional_statsd_tags
     end
+
+    def dry_run_option(kubectl)
```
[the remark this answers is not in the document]
jessie-JNing: > Can we detect the situation where it fails do to a custom plugin and fallback to the old behavior?

Do you mean we can do something like: https://kubernetes.io/docs/tasks/extend-kubectl/kubectl-plugins/  ?

I kind of feel it's overkilled. Shall we just give a "server-dry-run" available check during [validate_configuration](https://github.com/Shopify/kubernetes-deploy/blob/master/lib/kubernetes-deploy/deploy_task.rb#L333)?
The check can be done via a noop apply, and see if it succeeds.

lib/kubernetes-deploy/kubernetes_resource.rb
```diff
@@ -15,6 +15,8 @@ class KubernetesResource
     GLOBAL = false
     TIMEOUT = 5.minutes
     LOG_LINE_COUNT = 250
+    SERVER_DRY_RUN_DISABLED_ERROR =
+      /(unknown flag: --server-dry-run)|(doesn't support dry-run)|(dryRun alpha feature is disabled)/
```
jessie-JNing: Does it look OK to you?

jessie-JNing:
🎩  test logs show useful info(compare between minikube `v1.11.10` and  docker-for-desktop `v1.14.6`)

Deploy secret below to namespace
<details><summary>secret.yaml</summary>
<p>

```
---
apiVersion: v1
kind: Secret
metadata:
  name: hello-secret
type: Opaque
data:
  username: my_username
  password: cGFzc3dvcmQ=
```
</p>
</details>

Change type to be `invalid/type`, and deploy it again.
It should fail as type is an immutable field. 

- when running under minikube `v1.11.10`, logs didn't show which file got failed.
```
REVISION="4f7d6697fb5a20d4c960cd1cc6530774069489b6" bundle exec kubernetes-deploy cloud-portal-standby minikube --template-dir=/Users/jessie.ning/Documents/k8s/quick-resource/sensitive_info                                                                                                                              
[INFO][2019-09-16 14:31:20 -0400]       
[INFO][2019-09-16 14:31:20 -0400]       ------------------------------------Phase 1: Initializing deploy------------------------------------
[INFO][2019-09-16 14:31:21 -0400]       Context minikube found
[INFO][2019-09-16 14:31:21 -0400]       Namespace cloud-portal-standby found
[INFO][2019-09-16 14:31:21 -0400]       All required parameters and files are present
[INFO][2019-09-16 14:31:21 -0400]       Discovering resources:
[INFO][2019-09-16 14:31:21 -0400]         - Secret/hello-secret
[INFO][2019-09-16 14:31:21 -0400]         - Secret/hello-bad-secret
[INFO][2019-09-16 14:31:22 -0400]       
[INFO][2019-09-16 14:31:22 -0400]       ----------------------------Phase 2: Checking initial resource statuses-----------------------------
[INFO][2019-09-16 14:31:22 -0400]       Secret/hello-bad-secret                           Available
[INFO][2019-09-16 14:31:22 -0400]       Secret/hello-secret                               Not Found
[INFO][2019-09-16 14:31:22 -0400]       
[INFO][2019-09-16 14:31:22 -0400]       ------------------------------Phase 3: Predeploying priority resources------------------------------
[INFO][2019-09-16 14:31:22 -0400]       Deploying resources:
[INFO][2019-09-16 14:31:22 -0400]       - Secret/hello-bad-secret (timeout: 30s)
[INFO][2019-09-16 14:31:22 -0400]       - Secret/hello-secret (timeout: 30s)
[INFO][2019-09-16 14:31:22 -0400]       
[INFO][2019-09-16 14:31:22 -0400]       ------------------------------------------Result: FAILURE-------------------------------------------
[FATAL][2019-09-16 14:31:22 -0400]      Command failed: apply -f /var/folders/l2/ym6whn6d4bs697_w9w_9f73w0000gn/t/d20190916-51430-qxnmxh
[FATAL][2019-09-16 14:31:22 -0400]      
[FATAL][2019-09-16 14:31:22 -0400]      WARNING: Any resources not mentioned in the error(s) below were likely created/updated. You may wish to roll back this deploy.
[FATAL][2019-09-16 14:31:22 -0400]      
[FATAL][2019-09-16 14:31:22 -0400]      WARNING: There was an error applying some or all resources. The raw output may be sensitive and so cannot be displayed.
```

- when running under docker-for-desktop `v1.14.6`, error was caught at template validation stage, since we use `--server-dry-run`, the failing file can be displayed in log.
```
REVISION="4f7d6697fb5a20d4c960cd1cc6530774069489b6" bundle exec kubernetes-deploy docker docker-for-desktop --template-dir=/Users/jessie.ning/Documents/k8s/quick-resource/sensitive_info
[INFO][2019-09-16 14:25:20 -0400]       
[INFO][2019-09-16 14:25:20 -0400]       ------------------------------------Phase 1: Initializing deploy------------------------------------
[INFO][2019-09-16 14:25:20 -0400]       Context docker-for-desktop found
[INFO][2019-09-16 14:25:20 -0400]       Namespace docker found
[INFO][2019-09-16 14:25:20 -0400]       All required parameters and files are present
[INFO][2019-09-16 14:25:20 -0400]       Discovering resources:
[INFO][2019-09-16 14:25:20 -0400]         - Secret/hello-secret
[INFO][2019-09-16 14:25:20 -0400]         - Secret/hello-bad-secret
[INFO][2019-09-16 14:25:21 -0400]       
[INFO][2019-09-16 14:25:21 -0400]       ------------------------------------------Result: FAILURE-------------------------------------------
[FATAL][2019-09-16 14:25:21 -0400]      Template validation failed
[FATAL][2019-09-16 14:25:21 -0400]      
[FATAL][2019-09-16 14:25:21 -0400]      Invalid template: Secret-hello-bad-secret20190916-50753-1pbnnsz.yml
[FATAL][2019-09-16 14:25:21 -0400]      > Error message:
[FATAL][2019-09-16 14:25:21 -0400]          Validation for Secret/hello-bad-secret failed. Detailed information is unavailable as the raw error may contain sensitive data.


```

test/test_helper.rb
```diff
@@ -199,6 +199,11 @@ def stub_kubectl_response(*args, kwargs: {}, resp:, err: "", success: true, json
     def build_runless_kubectl
       obj = KubernetesDeploy::Kubectl.new(namespace: 'test', context: KubeclientHelper::TEST_CONTEXT,
         logger: logger, log_failure_by_default: false)
+
+      def obj.server_version
+        Gem::Version.new(KubernetesResource::SERVER_DRY_RUN_MIN_VERSION)
+      end
```
jessie-JNing: dead code, I'll remove it

test/integration/kubernetes_deploy_test.rb
```diff
@@ -36,6 +36,7 @@ def test_full_hello_cloud_set_deploy_succeeds
     assert_logs_match(%r{ConfigMap/hello-cloud-configmap-data\s+Available}, 1)
   end
 
+  # test for running validation when --server-dry-run is not available
```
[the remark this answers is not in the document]
jessie-JNing: agree that "test names to describe what they are testing".
Let me rename this test and one below.

test/unit/kubernetes-deploy/ejson_secret_provisioner_test.rb
```diff
@@ -97,8 +97,18 @@ def test_run_with_selector_does_not_raise_exception
 
   private
 
+  def stub_kubectl_version_request
```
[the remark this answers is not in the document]
jessie-JNing: dead code, I'll remove it

test/integration_test_helper.rb
```diff
@@ -36,6 +36,14 @@ def kube_server_version
       _kubectl.server_version
     end
 
+    def server_dry_run_available?
+      return @server_dry_run unless @server_dry_run.nil?
```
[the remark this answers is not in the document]
jessie-JNing: Ah, yes, that's ruby pattern. Thanks Danny

jessie-JNing:
@dturn  I update the tests, could you help to 👀 it again?  Thank you :)

test/integration/kubernetes_deploy_test.rb
```diff
@@ -45,16 +45,39 @@ def test_service_account_predeployed_before_unmanaged_pod
...
-    hello_cloud = FixtureSetAssertions::HelloCloud.new(@namespace)
-    hello_cloud.assert_configmap_data_present
-    hello_cloud.assert_all_service_accounts_up
-    hello_cloud.assert_unmanaged_pod_statuses("Succeeded", 1)
-    assert_logs_match_all([
-      %r{Successfully deployed in \d.\ds: ServiceAccount/build-robot},
-      %r{Successfully deployed in \d+.\ds: Pod/unmanaged-pod-.*},
-    ], in_order: true)
+
+    unless server_dry_run_available?
```
dturn: Did you have to chnage this test because the unmanaged pod is actually invalid, but we never cared about the success/failure status/. If so let's just use a valid unmanged pod.

test/integration/kubernetes_deploy_test.rb
```diff
@@ -667,24 +714,26 @@ def test_deploy_result_logging_for_mixed_result_deploy
...
-    assert_logs_match_all([
+    expect_log = [
       %r{Deployment/missing-volumes: TIMED OUT \(progress deadline: \d+s\)},
       "Timeout reason: ProgressDeadlineExceeded",
       /Latest ReplicaSet: missing-volumes-\w+/,
       "Final status: 1 replica, 1 updatedReplica, 1 unavailableReplica",
-      /FailedMount.*secrets? "catphotoscom" not found/, # event
-    ], in_order: true)
+    ]
+    expect_log << /FailedMount.*secrets? "catphotoscom" not found/ unless server_dry_run_available?
```
dturn: why does server_dry_run being available change the output of this tests?

test/integration/kubernetes_deploy_test.rb
```diff
@@ -1048,16 +1107,15 @@ def test_streams_unmanaged_pod_logs_when_only_one
   def test_bad_container_on_daemon_sets_fails
     assert_deploy_failure(deploy_fixtures("invalid", subset: ["crash_loop_daemon_set.yml"]))
     num_ds = expected_daemonset_pod_count
-    assert_logs_match_all([
+    expect_log = [
       "Failed to deploy 1 resource",
       "DaemonSet/crash-loop: FAILED",
       "crash-loop-back-off: Crashing repeatedly (exit 1). See logs for more information.",
       "Final status: #{num_ds} updatedNumberScheduled, #{num_ds} desiredNumberScheduled, 0 numberReady",
-      "Events (common success events excluded):",
```
dturn: If events aren't being printed something is wrong, I don't see why your code change would impact this

test/integration_test_helper.rb
```diff
@@ -36,6 +36,15 @@ def kube_server_version
       _kubectl.server_version
     end
 
+    def server_dry_run_available?
```
dturn: Can we just assume its available if we're on version 1.13 or greater?

test/integration/kubernetes_deploy_test.rb
jessie-JNing: sounds makes sense, invalid unmanaged pod could also make it fail, which is not something we would like to test here.

test/integration/kubernetes_deploy_test.rb
jessie-JNing: may flacky, let me verify

test/integration_test_helper.rb
jessie-JNing: Yes, after 1.13, this feature is enable by default.

test/integration/kubernetes_deploy_test.rb
jessie-JNing: this output is from event, fetched by resource watcher. Events happen after template validation, so if server-dry-run catches the error during template validation. We won't have the event log

test/integration/kubernetes_deploy_test.rb
jessie-JNing: actually not flacky test, they are event log, same reason as https://github.com/Shopify/kubernetes-deploy/pull/553/files#r325686855

test/integration/kubernetes_deploy_test.rb
dturn: Setting `PRINT_LOGS=1` will show you the logs (but don't rely on the pass/fail result because it causes the assert_log_match_all to be skipped). It appears what's happening is the server-dry-run is detecting that the service account does not yet exist and correctly failing: `error looking up service account k8sdeploy-test-unautomounted-service-account-p-e8e71a2e74b65512/build-robot: serviceaccount "build-robot" not found`. 

But this stinks because we know this can happen and are going to pre-deploy the service account so it will exist by the time the deployment goes out. I suspect this is going to be a problem for anything that can rely on something we pre-deploy.

To keep us from getting stuck 1 idea, but feel free to come up with something better, would be to just server-dry-run the items that are in the pre-deployed since they aren't supposed to be ordered / have dependencies.

jessie-JNing: Think about this problem, server-dry-run only pre-deploy resource could solve the problem. 
But one idea come to my mind is separating pre-deploy and deploy resource at earlier stage. 
Before: initial deploy -> validate resource -> pre-deploy -> deploy all
Can we change to: initial deploy -> validate pre-deploy resource -> deploy pre-deploy -> validate other resource -> deploy other resource

lib/kubernetes-deploy/kubernetes_resource.rb
```diff
@@ -464,9 +466,13 @@ def validate_selector(selector)
     end
 
     def validate_spec_with_kubectl(kubectl)
-      command = ["create", "-f", file_path, "--dry-run", "--output=name"]
-      _, err, st = kubectl.run(*command, log_failure: false, output_is_sensitive: sensitive_template_content?,
-        retry_whitelist: [:client_timeout], attempts: 3)
+      _, err, st = validate_with_dry_run_option(kubectl, "--dry-run")
+      if st.success? && sensitive_template_content?
```
dturn: I don't think we should re-use `sensitive_template_content` I think we should add a new method / constant like `SERVER_DRY_RUNABLE`

test/integration_test_helper.rb
dturn: I mean can this code just check that the server is on version >= 1.13 ?

jessie-JNing: We could do that. Since the possibility of people disable server-dry-run feature when >=1.13 is pretty low. 
Sure, let me change to the version check.

lib/kubernetes-deploy/kubernetes_resource.rb
jessie-JNing: I'll make a 🎩against different version, better to cover from v1.11 to v1.15. Will list the result

jessie-JNing: The updated 🎩 is https://github.com/Shopify/kubernetes-deploy/pull/553#issuecomment-531903409

CHANGELOG.md
```diff
@@ -4,6 +4,7 @@
 - Officially support Kubernetes 1.15 ([#546](https://github.com/Shopify/kubernetes-deploy/pull/546))
 - Make sure that we only declare a Service of type LoadBalancer as deployed after its IP address is published. [#547](https://github.com/Shopify/kubernetes-deploy/pull/547)
 - Add more validations to `RunnerTask`. [#554](https://github.com/Shopify/kubernetes-deploy/pull/554)
+- Validate resource template with sensitive content with `--server-dry-run` feature. [#553](https://github.com/Shopify/kubernetes-deploy/pull/553)
```
dturn: "Validate Secrets with `--server-dry-run` on supported clusters. "

PR #554, opened 2019-09-12
Use TaskConfigValidator for RunnerTask

dirceu:

Fix part of https://github.com/Shopify/kubernetes-deploy/issues/344 by using all supported `TaskConfigValidator` validations in `kubernetes-deploy run` (plus additional ones to validate `args` and `template`).

**How is this accomplished?**

In the first commit I enabled all validations from `TaskConfigValidator` and updated error messages in tests accordingly. I also chose to delete the unit tests we had for `RunnerTask` because:
1. There were only two and they were related to validation.
2. It would be too much effort to mock all kubectl and kubeclient calls for a low ROI.

In the second commit I created `RunnerTaskConfig` and `RunnerTaskConfigValidator` and moved the validation of `args` and `template` there to follow the same pattern for all validations.

**What could go wrong?**

The same as #533: because we need to unify validations that are done slightly differently we are changing the error class that gets raised as well as the wording. If people were rescuing these errors or matching text this will be a breaking change.


lib/kubernetes-deploy/common.rb
```diff
@@ -17,6 +17,8 @@
 require 'kubernetes-deploy/statsd'
 require 'kubernetes-deploy/task_config'
 require 'kubernetes-deploy/task_config_validator'
+require 'kubernetes-deploy/runner_task_config'
```
dturn: Are these actually common or should they be required by the runner_task?

lib/kubernetes-deploy/runner_task.rb
```diff
@@ -17,7 +17,7 @@ class TaskTemplateMissingError < TaskConfigurationError; end
 
     def initialize(namespace:, context:, logger: nil, max_watch_seconds: nil)
       @logger = logger || KubernetesDeploy::FormattedLogger.build(namespace, context)
-      @task_config = KubernetesDeploy::TaskConfig.new(context, namespace, @logger)
+      @task_config = KubernetesDeploy::RunnerTaskConfig.new(context, namespace, @logger)
```
dturn: The idea behind the TaskConfig was an object we could pass to other classes instead of a bunch of individual params. The fact that the only class that uses the task_template & args is the validator is making me wonder if we should just pass those two values to the Validator instead of create a new RunnerTaskConfig. Not being able to fully populate the RunnerTaskConfig until `Run!` seems odd, but might not really be an issue.

lib/kubernetes-deploy/runner_task_config_validator.rb
```diff
@@ -0,0 +1,23 @@
+# frozen_string_literal: true
+module KubernetesDeploy
+  class RunnerTaskConfigValidator < TaskConfigValidator
+    DEFAULT_VALIDATIONS += %i(
```
dturn: I can't believe this is valid Ruby. Should we have a class method for adding new validations instead of doing so upsetting things to constants.

lib/kubernetes-deploy/runner_task.rb
dirceu: I removed `RunnerTaskConfig` and added those two values to the validator. Thanks for the context!

lib/kubernetes-deploy/runner_task_config_validator.rb
dirceu: I just sent a commit with a change to make the addition of those two new validations happen during `initialize`; not sure yet if that's the best way. I'll check the class method idea!

dirceu:
> also that we haven't lost coverage by the removed tests.

👍 

I now added integration tests that make essentially the same assertions we had before, just in case.

lib/kubernetes-deploy/runner_task_config_validator.rb
```diff
@@ -0,0 +1,25 @@
+# frozen_string_literal: true
+module KubernetesDeploy
+  class RunnerTaskConfigValidator < TaskConfigValidator
+    attr_accessor :template, :args
```
dturn: I'm curious about why you chose to make these  `attr_accessor` vs. args to `initialize`?

dirceu: I don't really have a good answer for that 🙂 and it _does_ look better, so I just sent a commit changing it. Thanks!

lib/kubernetes-deploy/runner_task_config_validator.rb
```diff
@@ -0,0 +1,25 @@
+# frozen_string_literal: true
+module KubernetesDeploy
+  class RunnerTaskConfigValidator < TaskConfigValidator
+    def initialize(template, args, *arguments)
+      super(*arguments)
+      @template = template
+      @args = args
+      @validations += %i(validate_template validate_args)
```
dturn: This is way better than modify the constant!

dirceu:
Fixed rubocop warning and added changelog entry. Will merge once CI is green.

Issue #529, opened 2019-08-14
krane render


dturn:
## Problem

We need to create `krane render`. See the design docs for flags https://docs.google.com/document/d/1oInUsKplYGNWTymPY48xtDCx3x4Cwc-phcEKUkt2v4U/view

This will need to wait for #524 to merge as it will determine the file structure for new commands and their tests. 

This should also wait till #514 merges as it will provide code that will be helpful in supporting the `-f` flag.

Ensure that `new` & `run` use the same conventions as the other tasks, if there is a mismatch that needs to be fixed it can be done in a second PR.

## Possible solution
Build a thor task that calls `RenderTask.new`. 

## Acceptance criteria
- The command must work with all flags
- The command should have tests.

## Additional context
#256

dturn:
Krane render is done. Once we delete exe/kubernetes-render we can refactor the code to make the task simpler.

PR #555, opened 2019-09-13
Krane Render & make render task work with template_paths

dturn:
As part of the Krane 1.0 work render task will render from for multiple dirs. This requires the render task to make use of TemplateSets.

**How is this accomplished?**
I've attempted to make this backwards compatible so the first step was hooking up the render_task to TemplateSets.

This also required some slight modification of TemplateSet (and passing of args down to it). The biggest one is that when a rendering error occurs its not raised until all other templates have been processed. 

**What could go wrong?**
I need to add tests to show the new behavior.

### DIFF
--- a/lib/enumerable/with_delayed_exceptions.rb
+++ b/lib/enumerable/with_delayed_exceptions.rb
@@ -1,0 +1,14 @@
+# frozen_string_literal: true
+
+module Enumerable
+  def with_delayed_exceptions(*catch, &block)
+    exceptions = []
+    each do |i|
+      begin
+        block.call(i)
+      rescue *catch => e
+        exceptions << e
+      end
+    end.tap { raise exceptions.first if exceptions.first }
+  end
+end
--- a/lib/krane/cli/render_command.rb
+++ b/lib/krane/cli/render_command.rb
@@ -1,0 +1,31 @@
+# frozen_string_literal: true
+
+module Krane
+  module CLI
+    class RenderCommand
+      OPTIONS = {
+        bindings: { type: :array, desc: 'Bindings for erb' },
+        f: { type: :array, required: true, desc: 'Directories and files to render' },
+      }
+
+      def self.from_options(options)
+        require 'kubernetes-deploy/render_task'
+        require 'kubernetes-deploy/template_sets'
+        require 'kubernetes-deploy/bindings_parser'
+        require 'kubernetes-deploy/options_helper'
+
+        bindings_parser = KubernetesDeploy::BindingsParser.new
+        options[:bindings]&.each { |b| bindings_parser.add(b) }
+
+        KubernetesDeploy::OptionsHelper.with_processed_template_paths(options[:f]) do |paths|
+          runner = KubernetesDeploy::RenderTask.new(
+            current_sha: ENV["REVISION"],
+            template_paths: paths,
+            bindings: bindings_parser.parse,
+          )
+          runner.run(STDOUT)
+        end
+      end
+    end
+  end
+end
--- a/lib/krane/cli/render_command.rb
+++ b/lib/krane/cli/render_command.rb
@@ -23,7 +23,7 @@ module Krane
             template_paths: paths,
             bindings: bindings_parser.parse,
           )
-          runner.run(STDOUT)
+          runner.run!(STDOUT)
         end
       end
     end
--- a/lib/krane/cli/render_command.rb
+++ b/lib/krane/cli/render_command.rb
@@ -5,19 +5,18 @@ module Krane
     class RenderCommand
       OPTIONS = {
         bindings: { type: :array, desc: 'Bindings for erb' },
-        f: { type: :array, required: true, desc: 'Directories and files to render' },
+        filenames: { type: :array, required: true, aliases: 'f', desc: 'Directories and files to render' },
       }
 
       def self.from_options(options)
         require 'kubernetes-deploy/render_task'
-        require 'kubernetes-deploy/template_sets'
         require 'kubernetes-deploy/bindings_parser'
         require 'kubernetes-deploy/options_helper'
 
         bindings_parser = KubernetesDeploy::BindingsParser.new
         options[:bindings]&.each { |b| bindings_parser.add(b) }
 
-        KubernetesDeploy::OptionsHelper.with_processed_template_paths(options[:f]) do |paths|
+        KubernetesDeploy::OptionsHelper.with_processed_template_paths(options[:filenames]) do |paths|
           runner = KubernetesDeploy::RenderTask.new(
             current_sha: ENV["REVISION"],
             template_paths: paths,
--- a/lib/kubernetes-deploy/render_task.rb
+++ b/lib/kubernetes-deploy/render_task.rb
@@ -3,18 +3,16 @@ require 'tempfile'
 
 require 'kubernetes-deploy/common'
 require 'kubernetes-deploy/renderer'
+require 'kubernetes-deploy/template_sets'
 
 module KubernetesDeploy
   class RenderTask
-    def initialize(logger: nil, current_sha:, template_dir:, bindings:)
+    def initialize(logger: nil, current_sha:, template_dir: nil, template_paths: [], bindings:)
       @logger = logger || KubernetesDeploy::FormattedLogger.build
       @template_dir = template_dir
-      @renderer = KubernetesDeploy::Renderer.new(
-        current_sha: current_sha,
-        bindings: bindings,
-        template_dir: @template_dir,
-        logger: @logger,
-      )
+      @template_paths = template_paths.map { |path| File.expand_path(path) }
+      @bindings = bindings
+      @current_sha = current_sha
     end
 
     def run(*args)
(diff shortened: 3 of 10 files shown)


dturn:
```
➜  kubernetes-deploy git:(multi-render) ✗ find test/fixtures/for_unit_tests/ | grep yml | grep -v broken  | grep -v 'factorial' | xargs bundle exec krane render --bindings foo=1 n=2 -f > /dev/null
[INFO][2019-09-13 16:37:07 -0700]	
[INFO][2019-09-13 16:37:07 -0700]	---------------------------------Phase 1: Initializing render task----------------------------------
[INFO][2019-09-13 16:37:07 -0700]	Validating configuration
[INFO][2019-09-13 16:37:07 -0700]	
[INFO][2019-09-13 16:37:07 -0700]	-----------------------------------Phase 2: Rendering template(s)-----------------------------------
[INFO][2019-09-13 16:37:07 -0700]	Rendering pod_disruption_budget_test.yml...
[INFO][2019-09-13 16:37:07 -0700]	Rendered pod_disruption_budget_test.yml
[INFO][2019-09-13 16:37:07 -0700]	Rendering replica_set_test.yml...
[INFO][2019-09-13 16:37:07 -0700]	Rendered replica_set_test.yml
[INFO][2019-09-13 16:37:07 -0700]	Rendering pod_test.yml...
[INFO][2019-09-13 16:37:07 -0700]	Rendered pod_test.yml
[INFO][2019-09-13 16:37:07 -0700]	Rendering deployment_test.yml...
[INFO][2019-09-13 16:37:07 -0700]	Rendered deployment_test.yml
[INFO][2019-09-13 16:37:07 -0700]	Rendering crd_test.yml...
[INFO][2019-09-13 16:37:07 -0700]	Rendered crd_test.yml
[INFO][2019-09-13 16:37:07 -0700]	Rendering horizontal_pod_autoscaler_test.yml...
[INFO][2019-09-13 16:37:07 -0700]	Rendered horizontal_pod_autoscaler_test.yml
[INFO][2019-09-13 16:37:07 -0700]	Rendering daemon_set.yml...
[INFO][2019-09-13 16:37:07 -0700]	Rendered daemon_set.yml
[INFO][2019-09-13 16:37:07 -0700]	Rendering service_test.yml...
[INFO][2019-09-13 16:37:07 -0700]	Rendered service_test.yml
[INFO][2019-09-13 16:37:07 -0700]	Rendering stateful_set_test.yml...
[INFO][2019-09-13 16:37:07 -0700]	Rendered stateful_set_test.yml
[INFO][2019-09-13 16:37:07 -0700]	Rendering bindings.yml...
[INFO][2019-09-13 16:37:07 -0700]	Rendered bindings.yml
[INFO][2019-09-13 16:37:07 -0700]	
[INFO][2019-09-13 16:37:07 -0700]	------------------------------------------Result: SUCCESS-------------------------------------------
[INFO][2019-09-13 16:37:07 -0700]	Successfully rendered 10 template(s)
[INFO][2019-09-13 16:37:07 -0700]
```

lib/enumerable/with_delayed_exceptions.rb
```diff
@@ -0,0 +1,14 @@
+# frozen_string_literal: true
+
+module Enumerable
+  def with_delayed_exceptions(*catch, &block)
```
timothysmith0609: Not sure what the common convention is, but I'm strongly tempted to scope this into our own module and include it in `TemplateSets` instead of monkey patching `Enumerable`. (Or is this already scoped to the k8s-deploy package?)

lib/kubernetes-deploy/template_sets.rb
```diff
@@ -59,7 +59,7 @@ def validate
 
       private
 
-      def templates(filename:)
+      def templates(filename:, raw:)
```
timothysmith0609: Could you explain the semantics of `raw` for me, please? It's not entirely clear to me what the intent of it is.

lib/enumerable/with_delayed_exceptions.rb
dturn: This pollutes the global `Enumerable` module in ruby. So anyone requiring it or something that requires it will get it.  

I think you're right and I'll look into scoping this down, though I suspect the signature will change (e.g. you pass in the list to iterate over)  since the current receiver is anything that includes `enumerable` not a TemplateSet(s).

lib/kubernetes-deploy/template_sets.rb
dturn: Currently this function returns a ruby object that results from parsing the yaml. `raw` means I wan't the string that comes directly from the erb rendering. 

This is useful for 2 reasons over just calling `Yaml.dump` on the ruby object:
 1. it preserves the whitespace & comments.  
 2. You can put duplicate keys into a hash in yaml. Because ruby doesn't support this (nor does any language I know of), it silently throws away the first once which means when you go to lint it, it will pass even though its not valid.

lib/kubernetes-deploy/template_sets.rb
```diff
@@ -106,18 +109,26 @@ def from_dirs_and_files(paths:, logger:)
...
-          yield r_def
+          bindings: bindings,
+          raw: raw
+        ) do |r_def, filename|
+          yield r_def, filename
         end
       end
     end
 
+    def with_resource_definitions(render_erb: false, current_sha: nil, bindings: nil, raw: false)
```
dturn: Open question should I remove this function and just change everything using it to actually use `with_resource_definitions_and_filename `

timothysmith0609: I'm either/or on this. I think the clarity it brings to the code (e.g. not needing to care about `filename` in certain cases) makes it worthwhile. Since it's really just a wrapper around `with_resource_definitions_and_filename` it's not really duplicating _too much_.

test/exe/render_test.rb
```diff
@@ -0,0 +1,55 @@
...
+    test_sha = "TEST"
+    with_env("REVISION", test_sha) do
+      install_krane_render_expecations
+      krane_render!
+    end
+  end
+
+  private
+
+  def install_krane_render_expecations(new_args = {})
```
dirceu: minor typo

test/integration/render_task_test.rb
```diff
@@ -317,6 +318,65 @@ def test_render_does_not_generate_extra_blank_documents_when_file_is_empty
     assert_logs_match("Rendered effectively_empty.yml.erb successfully, but the result was blank")
   end
 
+  def test_render_task_template_paths_xor_templae_dir
```
dirceu: minor typo

lib/krane/cli/render_command.rb
```diff
@@ -0,0 +1,31 @@
+# frozen_string_literal: true
+
+module Krane
+  module CLI
+    class RenderCommand
+      OPTIONS = {
+        bindings: { type: :array, desc: 'Bindings for erb' },
+        f: { type: :array, required: true, desc: 'Directories and files to render' },
```
dirceu: should we support it with an alias of `:filenames` like we intend to do with `krane deploy`?

lib/kubernetes-deploy/render_task.rb
```diff
@@ -3,18 +3,16 @@
 
 require 'kubernetes-deploy/common'
 require 'kubernetes-deploy/renderer'
+require 'kubernetes-deploy/template_sets'
 
 module KubernetesDeploy
   class RenderTask
-    def initialize(logger: nil, current_sha:, template_dir:, bindings:)
+    def initialize(logger: nil, current_sha:, template_dir: nil, template_paths: [], bindings:)
```
timothysmith0609: Now that we can make breaking changes should we start retiring the deprecated `template_dir` argument in favour of exclusively using `template_paths`

RyanBrushett:
See https://github.com/Shopify/kubernetes-deploy/issues/529

lib/kubernetes-deploy/render_task.rb
dturn: I'd like to do handle this in a follow-up pr. I think this one is big enough as is.

lib/kubernetes-deploy/render_task.rb
```diff
@@ -46,29 +40,38 @@ def run!(stream, only_filenames = [])
...
       end
+
+      count
+    rescue KubernetesDeploy::InvalidTemplateError => exception
+      log_invalid_template(exception)
+      raise
     end
 
-    def render_filename(filename, stream)
+    def render_filename(rendered_content, filename, stream)
```
timothysmith0609: Minor point but this function is no longer doing what it says: the renderer has already run by the time we get here. Said that, we are _rendering_ to the stream, but that's muddying the waters IMO. Not sure what we should call it instead, though (`write_to_stream`?)

test/integration/render_task_test.rb
```diff
@@ -317,6 +318,65 @@ def test_render_does_not_generate_extra_blank_documents_when_file_is_empty
...
+    render = KubernetesDeploy::RenderTask.new(
+      logger: logger,
+      current_sha: "k#{SecureRandom.hex(6)}",
+      bindings: {},
+    )
+
+    assert_render_failure(render.run(mock_output_stream))
+    assert_logs_match_all([
+      "template_dir or template_paths must be set",
+    ], in_order: true)
```
timothysmith0609: redundant `in_order`

Issue #531, opened 2019-08-14
Krane run


dturn:
## Problem

Add the `krane run` command, https://docs.google.com/document/d/1oInUsKplYGNWTymPY48xtDCx3x4Cwc-phcEKUkt2v4U/view# has details on flags that need to be supported.

Ensure that `new` & `run` use the same conventions as the other tasks, if there is a mismatch that needs to be fixed it can be done in a second PR.

This is blocked by #524.

## Possible solution
Build a thor task that calls `RunnerTask.new`. See #530 for thoughts about `--verify-result=false`. 

## Acceptance criteria
- The command must work with all flags
- The command should have tests.

## Additional context

#256 is the tracking issue for the new CLI.

RyanBrushett:
I'll cut an issue for `-f` and try to get time to hack on it myself.

PR #559, opened 2019-09-18
Add krane run command

RyanBrushett:

Add krane run, which is part of https://github.com/Shopify/kubernetes-deploy/issues/531

I think I'd like to refactor a bit more so I won't call that issue complete yet.

**How is this accomplished?**

In an effort to make these PRs small and digestible, I ported over the runner task to `krane` in this PR and added a bunch of tests. I'd like to do refactors in another PR.

**What could go wrong?**

It could not work, I guess.

**Some things to note:**

- I wanted to make entrypoint and args optional. Is what I did here sufficient for that?
- I wanted to use `type: :array` for arguments but that was failing for some cases. Thor has a lot of trouble parsing arrays whose first element is a string that starts with `-`, e.g. `--command=echo --arguments=-n hello` does not work. I fought with this for a while, I'm convinced it's something to do with how Thor parses arrays. I don't think it's worth blocking this on a Thor bug but it's something I might look into shortly.
- env-vars should be `type: :hash` but that's a refactor I'll do in my next PR
- I removed `validate_args` because I want args to be optional cc @dirceu 
- All the edits in `restart test` are a typo fix. You can focus on the second commit.

### DIFF
--- a/lib/krane/cli/krane.rb
+++ b/lib/krane/cli/krane.rb
@@ -4,6 +4,7 @@ require 'krane'
 require 'thor'
 require 'krane/cli/version_command'
 require 'krane/cli/restart_command'
+require 'krane/cli/run_command'
 
 module Krane
   module CLI
@@ -28,6 +29,14 @@ module Krane
       def restart(namespace, context)
         rescue_and_exit do
           RestartCommand.from_options(namespace, context, options)
+        end
+      end
+
+      desc("run NAMESPACE CONTEXT", "Run a pod that exits upon completing a task")
+      expand_options(RunCommand::OPTIONS)
+      def run_command(namespace, context)
+        rescue_and_exit do
+          RunCommand.from_options(namespace, context, options)
         end
       end
 
--- a/lib/krane/cli/run_command.rb
+++ b/lib/krane/cli/run_command.rb
@@ -1,0 +1,53 @@
+# frozen_string_literal: true
+
+module Krane
+  module CLI
+    class RunCommand
+      DEFAULT_RUN_TIMEOUT = '300s'
+
+      OPTIONS = {
+        "global-timeout" => {
+          type: :string,
+          banner: "duration",
+          desc: "Timeout error is raised if the pod runs for longer than the specified number of seconds",
+          default: DEFAULT_RUN_TIMEOUT,
+        },
+        "arguments" => {
+          type: :string,
+          banner: '"ARG1 ARG2 ARG3"',
+          desc: "Override the default arguments for the command with a space-separated list of arguments",
+        },
+        "verify-result" => { type: :boolean, desc: "Wait for completion and verify pod success", default: true },
+        "command" => { type: :array, desc: "Override the default command in the container image" },
+        "template" => {
+          type: :string,
+          desc: "The template file you'll be rendering",
+          default: 'task-runner-template',
+        },
+        "env-vars" => {
+          type: :string,
+          banner: "VAR=val,FOO=bar",
+          desc: "A Comma-separated list of env vars",
+          default: '',
+        },
+      }
+
+      def self.from_options(namespace, context, options)
+        require "kubernetes-deploy/runner_task"
+        runner = KubernetesDeploy::RunnerTask.new(
+          namespace: namespace,
+          context: context,
+          max_watch_seconds: KubernetesDeploy::DurationParser.new(options["global-timeout"]).parse!.to_i,
+        )
+
+        runner.run!(
+          verify_result: options['verify-result'],
+          task_template: options['template'],
+          entrypoint: options['command'],
+          args: options['arguments']&.split(" "),
+          env_vars: options['env-vars'].split(','),
+        )
+      end
+    end
+  end
+end
--- a/lib/kubernetes-deploy/runner_task.rb
+++ b/lib/kubernetes-deploy/runner_task.rb
@@ -154,8 +154,8 @@ module KubernetesDeploy
         raise TaskConfigurationError, message
       end
 
-      container.command = entrypoint
-      container.args = args
+      container.command = entrypoint if entrypoint
+      container.args = args if args
 
       env_args = env_vars.map do |env|
         key, value = env.split('=', 2)
--- a/test/exe/restart_test.rb
+++ b/test/exe/restart_test.rb
@@ -4,21 +4,21 @@ require 'krane/cli/krane'
 
 class RestartTest < KubernetesDeploy::TestCase
   def test_restart_with_default_options
-    set_krane_restart_expecations
+    set_krane_restart_expectations
     krane_restart!
   end
 
   def test_restart_parses_global_timeout
-    set_krane_restart_expecations(new_args: { max_watch_seconds: 10 })
+    set_krane_restart_expectations(new_args: { max_watch_seconds: 10 })
     krane_restart!(flags: '--global-timeout 10s')
-    set_krane_restart_expecations(new_args: { max_watch_seconds: 60**2 })
+    set_krane_restart_expectations(new_args: { max_watch_seconds: 60**2 })
     krane_restart!(flags: '--global-timeout 1h')
   end
 
   def test_restart_passes_deployments_transparently
-    set_krane_restart_expecations(deployments: ['web'])
+    set_krane_restart_expectations(deployments: ['web'])
     krane_restart!(flags: '--deployments web')
-    set_krane_restart_expecations(deployments: ['web', 'jobs'])
+    set_krane_restart_expectations(deployments: ['web', 'jobs'])
     krane_restart!(flags: '--deployments web jobs')
   end
 
@@ -33,9 +33,9 @@ class RestartTest < KubernetesDeploy::TestCase
   end
 
   def test_restart_passes_verify_result
-    set_krane_restart_expecations(run_args: { verify_result: true })
+    set_krane_restart_expectations(run_args: { verify_result: true })
     krane_restart!(flags: '--verify-result true')
-    set_krane_restart_expecations(run_args: { verify_result: false })
+    set_krane_restart_expectations(run_args: { verify_result: false })
     krane_restart!(flags: '--verify-result false')
   end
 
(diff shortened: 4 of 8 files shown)


lib/krane/cli/run_command.rb
```diff
@@ -0,0 +1,53 @@
...
+
+      OPTIONS = {
+        "global-timeout" => {
+          type: :string,
+          banner: "duration",
+          desc: "Timeout error is raised if the pod runs for longer than the specified number of seconds",
+          default: DEFAULT_RUN_TIMEOUT,
+        },
+        "arguments" => {
+          type: :string,
```
RyanBrushett: See PR description for why this is `:string` and not `:array`

lib/krane/cli/run_command.rb
```diff
@@ -0,0 +1,53 @@
...
+        },
+        "verify-result" => { type: :boolean, desc: "Wait for completion and verify pod success", default: true },
+        "command" => { type: :array, desc: "Override the default command in the container image" },
+        "template" => {
+          type: :string,
+          desc: "The template file you'll be rendering",
+          default: 'task-runner-template',
+        },
+        "env-vars" => {
+          type: :string,
```
RyanBrushett: Will refactor this to hash in a future PR.

lib/kubernetes-deploy/runner_task.rb
```diff
@@ -154,8 +154,8 @@ def set_container_overrides!(pod_definition, entrypoint, args, env_vars)
         raise TaskConfigurationError, message
       end
 
-      container.command = entrypoint
-      container.args = args
+      container.command = entrypoint if entrypoint
+      container.args = args if args
```
RyanBrushett: Is this sufficient to make these optional?

test/integration/krane_test.rb
```diff
@@ -16,8 +16,24 @@ def test_restart_black_box
     refute(fetch_restarted_at("redis"), "RESTARTED_AT env is present")
   end
 
+  def test_run_black_box
+    assert_empty(task_runner_pods)
+    assert_deploy_success(deploy_fixtures("hello-cloud", subset: ["template-runner.yml", "configmap-data.yml"]))
+    template = "hello-cloud-template-runner"
+    out, err, status = krane_black_box("run",
+      "#{@namespace} #{KubeclientHelper::TEST_CONTEXT} --template #{template} --command ls --arguments \"-a /\"")
+    assert_match("Successfully ran", err)
```
RyanBrushett: Why does this match on `err` and not `out`? See the test above this one too.

test/exe/run_test.rb
```diff
@@ -0,0 +1,89 @@
...
+    set_krane_run_expectations(run_args: { task_template: 'some-name' })
+    krane_run!(flags: '--template some-name')
+  end
+
+  def test_run_parses_env_vars
+    set_krane_run_expectations(run_args: { env_vars: %w(SOMETHING=8000 FOO=bar) })
+    krane_run!(flags: '--env-vars SOMETHING=8000,FOO=bar')
+  end
+
+  def test_run_failure_as_black_box
```
dturn: What is actually being tested here? (e.g. that the exit status is 1, `-q` isn't a valid arg, ...)

test/integration/krane_test.rb
dturn: Curious why you tested for `Successfully ran`  over `Success`?

re stdout vs. stderr: It has to do with the original purpose of both. I think the render task is the easiest one to talk about. The rendered templates to go stdout because its actually what the program produces while the status / phase messages go to stderr since they're more metadata. https://www.jstorimer.com/blogs/workingwithcode/7766119-when-to-use-stderr-instead-of-stdout is a blog that does a better job than me at explaining it.

RyanBrushett: > Curious why you tested for Successfully ran over Success?

No particular reason, really, other than 'Successfully ran' appears in the output. I can change it for consistency if you like!

> re stdout vs. stderr...

Ooo word, that makes good sense. Duly noted.

test/exe/run_test.rb
RyanBrushett: That's it. Test that it fails and returns status 1 when it's used incorrectly (see [here](https://github.com/Shopify/kubernetes-deploy/blob/master/test/exe/restart_test.rb#L42-L47) for where I knocked this off from heh)

I could make the test less arbitrary though, for sure. I'll make it more realistic and add tests failing on my own flags 😛

test/integration/runner_task_test.rb
```diff
@@ -238,7 +221,7 @@ def test_run_fails_if_task_template_is_blank
     ], in_order: true)
   end
 
-  def test_run_bang_fails_if_task_template_or_args_are_invalid
+  def test_run_bang_fails_if_task_template_is_invalid
```
dirceu: I think this test can be removed now, as this is already tested in the method above.

dirceu: Nevermind, one is `run` and the other `run!`.

lib/krane/cli/run_command.rb
```diff
@@ -0,0 +1,53 @@
...
+          default: '',
+        },
+      }
+
+      def self.from_options(namespace, context, options)
+        require "kubernetes-deploy/runner_task"
+        runner = KubernetesDeploy::RunnerTask.new(
+          namespace: namespace,
+          context: context,
+          max_watch_seconds: KubernetesDeploy::DurationParser.new(options["global-timeout"]).parse!.to_i,
```
dturn: I did this too but we probably shouldn't have a bad arg causing a stack dump. We can defer fixing this to the refactor you're planning. 

```
➜  kubernetes-deploy git:(add-krane-run-command) ✗ be krane run blah blah --global-timeout 3
bundler: failed to load command: krane (/Users/danielturner/.gem/ruby/2.3.3/bin/krane)
KubernetesDeploy::DurationParser::ParsingError: Invalid ISO 8601 duration: "3"
  /Users/danielturner/src/github.com/Shopify/kubernetes-deploy/lib/kubernetes-deploy/duration_parser.rb:25:in `rescue in rescue in parse!'
  /Users/danielturner/src/github.com/Shopify/kubernetes-deploy/lib/kubernetes-deploy/duration_parser.rb:22:in `rescue in parse!'
  /Users/danielturner/src/github.com/Shopify/kubernetes-deploy/lib/kubernetes-deploy/duration_parser.rb:20:in `parse!'
```

test/exe/restart_test.rb
```diff
@@ -4,21 +4,21 @@
 
 class RestartTest < KubernetesDeploy::TestCase
   def test_restart_with_default_options
-    set_krane_restart_expecations
+    set_krane_restart_expectations
```
dturn: Thanks for fixing my typos!

lib/krane/cli/run_command.rb
dturn: Sure seems like it.

lib/kubernetes-deploy/runner_task.rb
dturn: Seems like it.

test/integration/krane_test.rb
dturn: I think consistency would be good.

test/exe/run_test.rb
dturn: If you're just testing the right exit code I think this is good. I just wanted to make sure we are testing our code and not thor.

lib/krane/cli/krane.rb
```diff
@@ -31,6 +32,14 @@ def restart(namespace, context)
         end
       end
 
+      desc("run NAMESPACE CONTEXT", "Run a pod that exits upon completing a task")
+      expand_options(RunCommand::OPTIONS)
+      def run_command(namespace, context)
```
jessie-JNing: Looks great to me! Just a nit, I feel it should be call `run` instead of `run_command`.

RyanBrushett: We can't, `run` is a protected method name within Thor so this needs to be called something like `run_task` or `run_command` 😞 @jessie-JNing

test/exe/run_test.rb
RyanBrushett: Fair!

lib/krane/cli/run_command.rb
RyanBrushett: 😬 I'll make that a part of my next PR for sure.

RyanBrushett:
@dirceu I think in the interest of brevity, the krane CLI changelog entry that's there + the rename notification on our readme will be what we stick with for now.
I'd say on release we'll have to add something a bit clearer though for sure! 👍

philss:
> I wanted to use type: :array for arguments but that was failing for some cases. Thor has a lot of trouble parsing arrays whose first element is a string that starts with -, e.g. --command=echo --arguments=-n hello does not work. I fought with this for a while, I'm convinced it's something to do with how Thor parses arrays. I don't think it's worth blocking this on a Thor bug but it's something I might look into shortly.

I got a bug related to this parsing of arguments in the version 1.0.0: the arguments with just one "word", without spaces, was causing `krane run` to not work correctly. I had to add a space in the list of arguments at the end to make it work.

So this was *not working*:

```
$ krane run my-app my-context --template="template" --arguments="-l" --command="ls"
```

But this *works good* (notice the space in arguments): 

```
$ krane run my-app my-context --template="template" --arguments="-l " --command="ls"
```

The error with the broken command was something like this:

```
ERROR: "krane run_command" was called with arguments ["my-app", "my-context", "-l"]
Usage: "krane run NAMESPACE CONTEXT f, --template=TEMPLATE"
```

I'm going to investigate more where this is failing and try to open a fix for that.

KnVerey:
Thanks for the bug report @philss. If you don't end up opening a PR, can you please let us know by opening an issue so we can keep track of this problem?

dturn:
@timothysmith0609  can confirm, but I believe https://github.com/erikhuda/thor/pull/691 fixes the issue in thor, we are just waiting for the next thor release.

PR #560, opened 2019-09-20
[Feature] Add `kubectl diff` implementation as `kubernetes-diff` and `krane diff`

midN:
This PR implements `kubectl diff` functionality to both older `kubernetes-deploy` version as `kubernetes-diff` as well as to the new and fancy `krane`.
We are currently running slightly modified version of `kubernetes-deploy` with very minor custom changes, but this is a completely new feature, therefore it would be great to get it properly contributed into upstream.

Examples:
`kubernetes-diff` with default `diff` command

`kubernetes-diff` with custom `diff` command

`krane diff` with default `diff` command

-----------

**How is this accomplished?**
By applying enormous amount of focus and discipline to click-clacking on the keyboard while your 3y old is running around, singing very loud and trying to force you into writing bugs.

**What could go wrong?**
Mr T. could win 2020 election.

### DIFF
--- a/exe/kubernetes-diff
+++ b/exe/kubernetes-diff
@@ -67,7 +67,6 @@ begin
 
     runner.run!(stream: STDOUT)
   end
-  # TODO: Rename errors
 rescue KubernetesDeploy::DeploymentTimeoutError
   exit(70)
 rescue KubernetesDeploy::FatalDeploymentError
--- a/exe/kubernetes-diff
+++ b/exe/kubernetes-diff
@@ -1,0 +1,78 @@
+#!/usr/bin/env ruby
+# frozen_string_literal: true
+
+require 'kubernetes-deploy/diff_task'
+require 'kubernetes-deploy/options_helper'
+require 'kubernetes-deploy/bindings_parser'
+require 'kubernetes-deploy/label_selector'
+
+require 'optparse'
+
+template_dir = nil
+template_paths = []
+bindings = {}
+verbose_log_prefix = false
+selector = nil
+
+ARGV.options do |opts|
+  parser = KubernetesDeploy::BindingsParser.new
+  opts.on("--bindings=BINDINGS", "Expose additional variables to ERB templates " \
+    "(format: k1=v1,k2=v2, JSON string or file (JSON or YAML) path prefixed by '@')") { |b| parser.add(b) }
+
+  opts.on("--verbose-log-prefix", "Add [context][namespace] to the log prefix") { verbose_log_prefix = true }
+  opts.on("--template-dir=DIR", "Set the template dir (default: config/deploy/$ENVIRONMENT)") { |v| template_dir = v }
+  opts.on("-f [PATHS]", Array, "comma separated list of template directories and/or filenames") do |paths|
+    template_paths += paths
+  end
+
+  opts.on("--selector=SELECTOR", "Ensure that all resources in your template dir match the given selector, " \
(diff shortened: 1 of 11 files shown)


lib/kubernetes-deploy/diff_task.rb
```diff
@@ -0,0 +1,315 @@
...
+    def server_version
+      kubectl.server_version
+    end
+
+    def initialize(
+      namespace:, context:, current_sha:, logger: nil, kubectl_instance: nil, bindings: {},
+      selector: nil, template_paths: [], template_dir: nil
+    )
+      template_dir = File.expand_path(template_dir) if template_dir
+      template_paths = (template_paths.map { |path| File.expand_path(path) } << template_dir).compact
```
timothysmith0609: You should just be able to use template_paths and drop template_dir. The reason we use both in deploy_task.rb is to ensure the change was backwards compatible at the API-level. Since this is a new task you can just use the newer template_paths (with equivalent -f flag in the CLI)

midN: @timothysmith0609 👍 Will do that, what do you think of using "."/current_dir as default in-case of missing though? I did that in Krane version, but i'm not sure if you guys have account for any defaults in newer version, the default for TEMPLATE_DIR was quite a pain tbh :)

midN: @timothysmith0609 updated

exe/kubernetes-diff
```diff
@@ -0,0 +1,68 @@
...
+verbose_log_prefix = false
+selector = nil
+
+ARGV.options do |opts|
+  parser = KubernetesDeploy::BindingsParser.new
+  opts.on("--bindings=BINDINGS", "Expose additional variables to ERB templates " \
+    "(format: k1=v1,k2=v2, JSON string or file (JSON or YAML) path prefixed by '@')") { |b| parser.add(b) }
+
+  opts.on("--verbose-log-prefix", "Add [context][namespace] to the log prefix") { verbose_log_prefix = true }
+  opts.on("-f [PATHS]", Array, "comma separated list of template directories and/or filenames ( Default: current directory )") do |paths|
```
timothysmith0609: I'm apprehensive about muddling the semantics of our default dir across different tools. To me, it would be surprising to have the same flag yield different defaults for different commands. I'm tempted to keep the default the same as deploy/render/etc. You should be able to get this for free by calling `with_validated_paths` from `options_helper.rb` (see `exe/kubernetes-deploy`, `exe/kubernetes-render`).

midN:
@timothysmith0609 That's true for Deploy as well though, it is rendering the files without `kubernetes-render`, which technically could be used to also pipe it to apply.
I've decided to keep it simple and separate, same as other commands for now.

I agree though, a substantial rewrite for all commands to use shared rendering of files before passing over to any kubectl command would be nice, but that's much bigger task to tackle

exe/kubernetes-diff
midN: This can be removed yeah, we can keep current dir default in fork, i'm glad you removed it for template_paths, was not fan of `config/deploy/env` default :)

KnVerey:
Hey @midN , I'm sorry I missed Tim's original ping back in September. We did a substantial rewrite to release Krane 1.0 in the fall, so this PR would need to be redone if you're still interested in getting it merged. Before you do that though, I echo Tim's question: what is the value-add of `krane diff` over `krane render | kubectl diff`? I'm going to close this PR since it is so stale, but I'll continue to watch it--feel free to reply here or to open an issue for discussion.

PR #557, opened 2019-09-17
Krane Deploy CLI

dirceu:
**What are you trying to accomplish with this PR?**

Add `krane deploy` command using the design [proposed here](https://docs.google.com/document/d/1oInUsKplYGNWTymPY48xtDCx3x4Cwc-phcEKUkt2v4U/view#).

Related to https://github.com/Shopify/kubernetes-deploy/issues/532.

**How is this accomplished?**

- Add new command to Krane.
- Add the ability to specify a list of protected namespaces.

**What could go wrong?**

- Not enough test coverage for all combinations of command line options?
- Not sure if the help messages are clear enough.

**:tophat:ing examples**

Help message:

```bash
$ bundle exec krane help deploy
Usage:
  krane deploy NAMESPACE CONTEXT f, --filenames=/tmp/my-resource.yml

Options:
      [--bindings=foo=bar abc=def]                               # Expose additional variables to ERB templates (format: k1=v1 k2=v2, JSON string or file (JSON or YAML) path prefixed by '@')
  f, --filenames=/tmp/my-resource.yml                            # Path to file or directory that contains the configuration to apply
      [--global-timeout=duration]                                # Max duration to monitor workloads correctly deployed
                                                                 # Default: 300s
      [--protected-namespaces=namespace1 namespace2 namespaceN]  # Enable deploys to a list of selected namespaces; set to an empty string to disable
                                                                 # Default: ["default", "kube-system", "kube-public"]
      [--prune], [--no-prune]                                    # Enable deletion of resources that do not appear in the template dir
                                                                 # Default: true
      [--render-erb], [--no-render-erb]                          # Enable ERB rendering
      [--selector='label=value']                                 # Select workloads by selector(s)
      [--verbose-log-prefix], [--no-verbose-log-prefix]          # Add [context][namespace] to the log prefix
                                                                 # Default: true
      [--verify-result], [--no-verify-result]                    # Verify workloads correctly deployed
                                                                 # Default: true

Ship resources to a namespace
```

Specifying a directory with `-f`:

```bash
$ bundle exec krane deploy dirceu-test-krane-deploy remote-cluster -f ~/Desktop
[INFO][2019-09-18 12:57:03 -0400][remote-cluster][dirceu-test-krane-deploy]
[INFO][2019-09-18 12:57:03 -0400][remote-cluster][dirceu-test-krane-deploy]     ------------------------------------Phase 1: Initializing deploy------------------------------------
[...]

PR #562, opened 2019-09-23
Update `DeployTask` test setup to use `render_erb: false` by default

dirceu:

As a follow-up to https://github.com/Shopify/kubernetes-deploy/pull/557, I want to be explicit whenever we use ERB rendering in the tests as this is becoming `false` by default in 1.0

**How is this accomplished?**

I changed the default value for `render_erb` to `false` in the test setup, and passed the value as `true` only in the places where we actually needed it.

**What could go wrong?**

🤷‍♂

### DIFF
--- a/test/integration/kubernetes_deploy_test.rb
+++ b/test/integration/kubernetes_deploy_test.rb
@@ -3,7 +3,7 @@ require 'integration_test_helper'
 
 class KubernetesDeployTest < KubernetesDeploy::IntegrationTest
   def test_full_hello_cloud_set_deploy_succeeds
-    assert_deploy_success(deploy_fixtures("hello-cloud"))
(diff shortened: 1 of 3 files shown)


test/integration/kubernetes_deploy_test.rb
```diff
@@ -1,9 +1,9 @@
 # frozen_string_literal: true
 require 'integration_test_helper'
 
-class KubernetesDeployTest < KubernetesDeploy::IntegrationTest
+class DeployTaskTest < KubernetesDeploy::IntegrationTest
```
dturn: By convention class names should match file names.

jessie-JNing:
I rebase the PR, and it LGTM~

Issue #522, opened 2019-08-08
Refuse to deploy global resources


KnVerey:
## Proposal

This project's mission is to make it easy to ship changes to a Kubernetes **namespace** and understand the result. Our contributors' guide explicitly explains that we do not support global resources:

> Global resources
> 
> This project is intended to manage a namespace (or a label-delimited subsection of one). It does not officially support non-namespaced resources. In practice, we do model custom resource definitions specifically, because it helps us better handle custom resources (which typically are namespaced). However, our intent is to keep this the only exception to the rule.

However, in practice, we blindly `kubectl apply` _any_ resource we don't recognize, including those that are global. That means that invoking us can have (intended or otherwise) side-effects in the global namespace. Because of this (and likely the fact that our user-facing documentation makes no mention of globals one way or the other), we've received more than one request to predeploy global resources.

I strongly believe that our deploy command should keep its mission of managing a namespace, and I would like to introduce safeguards against global side-effects as part of the transition to Krane.

## Resources affected

`kubectl api-resources --namespaced=false`

```
componentstatuses
namespaces
nodes
persistentvolumes
mutatingwebhookconfigurations.admissionregistration.k8s.io
validatingwebhookconfigurations.admissionregistration.k8s.io
customresourcedefinitions.apiextensions.k8s.io
apiservices.apiregistration.k8s.io
tokenreviews.authentication.k8s.io
selfsubjectaccessreviews.authorization.k8s.io
selfsubjectrulesreviews.authorization.k8s.io
subjectaccessreviews.authorization.k8s.io
certificatesigningrequests.certificates.k8s.io
podsecuritypolicies.extensions
runtimeclasses.node.k8s.io
podsecuritypolicies.policy
clusterrolebindings.rbac.authorization.k8s.io
clusterroles.rbac.authorization.k8s.io
priorityclasses.scheduling.k8s.io
csidrivers.storage.k8s.io
csinodes.storage.k8s.io
storageclasses.storage.k8s.io
volumeattachments.storage.k8s.io
```

## Options

A) Somehow (TBD) identify global resources during validation, and fail fast if the template set includes any. In the error, explain global resource management best practices and suggest using `kubectl apply` directly for them.
[...]

PR #567, opened 2019-09-25
Disable deploying global resources by default

dturn:
As part of the 1.0 push we are going to require a new command to deploy global resources https://github.com/Shopify/kubernetes-deploy/issues/522

This is the first step, preventing `krane deploy` from deploying global resources. This does maintain backwards compatibility with kubernetes-deploy. 

**How is this accomplished?**
Fetch global resources from the cluster and warn/raise if any are part of the deploy set.

**What could go wrong?**
We miss global resources or this change upsets people.

### DIFF
--- a/lib/krane/deploy_task.rb
+++ b/lib/krane/deploy_task.rb
@@ -1,0 +1,11 @@
+# frozen_string_literal: true
+
+require 'kubernetes-deploy/deploy_task'
+
+module Krane
+  class DeployTask < KubernetesDeploy::DeployTask
+    def initialize(**args)
+      super(args.merge(allow_globals: false))
+    end
+  end
+end
--- a/lib/krane/deploy_task.rb
+++ b/lib/krane/deploy_task.rb
@@ -5,6 +5,7 @@ require 'kubernetes-deploy/deploy_task'
 module Krane
   class DeployTask < KubernetesDeploy::DeployTask
     def initialize(**args)
+      raise "Use Krane::DeployGlobalTask to deploy global resources" if args[:allow_globals]
       super(args.merge(allow_globals: false))
     end
   end
--- a/lib/kubernetes-deploy/cluster_resource_discovery.rb
+++ b/lib/kubernetes-deploy/cluster_resource_discovery.rb
@@ -16,7 +16,23 @@ module KubernetesDeploy
       end
     end
 
+    def globals
+      @globals ||= fetch_globals.map do |gv|
+        kind, group = gv.split(".", 2)
+        { group: group, kind: kind.singularize }
+      end
+    end
+
     private
+
+    def fetch_globals
+      raw_names, _, st = kubectl.run("api-resources", "--namespaced=false", output: "name", attempts: 5)
+      if st.success?
+        raw_names.split("\n")
+      else
+        []
+      end
+    end
 
     def fetch_crds
       raw_json, _, st = kubectl.run("get", "CustomResourceDefinition", output: "json", attempts: 5)
--- a/lib/kubernetes-deploy/cluster_resource_discovery.rb
+++ b/lib/kubernetes-deploy/cluster_resource_discovery.rb
@@ -17,18 +17,20 @@ module KubernetesDeploy
     end
 
     def global_resource_names
-      @globals ||= fetch_globals.map do |gv|
-        kind, _group = gv.split(".", 2)
-        kind.singularize
-      end
+      @globals ||= fetch_globals.map { |g| g["kind"] }
     end
 
     private
 
     def fetch_globals
-      raw_names, _, st = kubectl.run("api-resources", "--namespaced=false", output: "name", attempts: 5)
+      raw, _, st = kubectl.run("api-resources", "--namespaced=false", output: "wide", attempts: 5)
       if st.success?
-        raw_names.split("\n")
+        rows = raw.split("\n")
+        header = rows.shift.downcase.scan(/[a-z]+[\W]*/).each_with_object({}) do |match, hash|
+          start = hash.values.map(&:last).max.to_i
+          hash[match.strip] = [start, start + match.length]
+        end
+        rows.map { |r| header.map { |k, (s, e)| [k.strip, r[s...e].strip] }.to_h }
       else
         []
       end
--- a/lib/kubernetes-deploy/cluster_resource_discovery.rb
+++ b/lib/kubernetes-deploy/cluster_resource_discovery.rb
@@ -16,7 +16,7 @@ module KubernetesDeploy
       end
     end
 
-    def global_resource_names
+    def global_resource_kinds
       @globals ||= fetch_globals.map { |g| g["kind"] }
     end
 
@@ -26,11 +26,16 @@ module KubernetesDeploy
       raw, _, st = kubectl.run("api-resources", "--namespaced=false", output: "wide", attempts: 5)
       if st.success?
         rows = raw.split("\n")
-        header = rows.shift.downcase.scan(/[a-z]+[\W]*/).each_with_object({}) do |match, hash|
-          start = hash.values.map(&:last).max.to_i
-          hash[match.strip] = [start, start + match.length]
+        header = rows[0]
+        resources = rows[1..-1]
+        full_width_field_names = header.downcase.scan(/[a-z]+[\W]*/)
+        cursor = 0
+        fields = full_width_field_names.each_with_object({}) do |name, hash|
+          start = cursor
+          cursor = start + name.length
+          hash[name.strip] = [start, cursor - 1]
         end
-        rows.map { |r| header.map { |k, (s, e)| [k.strip, r[s...e].strip] }.to_h }
+        resources.map { |r| fields.map { |k, (s, e)| [k.strip, r[s..e].strip] }.to_h }
       else
         []
       end
--- a/lib/kubernetes-deploy/cluster_resource_discovery.rb
+++ b/lib/kubernetes-deploy/cluster_resource_discovery.rb
@@ -16,10 +16,10 @@ module KubernetesDeploy
       end
     end
 
-    def globals
+    def global_resource_names
       @globals ||= fetch_globals.map do |gv|
-        kind, group = gv.split(".", 2)
-        { group: group, kind: kind.singularize }
+        kind, _group = gv.split(".", 2)
+        kind.singularize
       end
     end
 
--- a/lib/kubernetes-deploy/deploy_task.rb
+++ b/lib/kubernetes-deploy/deploy_task.rb
@@ -102,10 +102,6 @@ module KubernetesDeploy
       wl + cluster_resource_discoverer.crds.select(&:prunable?).map(&:group_version_kind)
     end
 
-    def global_resource_names
-      cluster_resource_discoverer.global_resource_names
-    end
-
     def server_version
       kubectl.server_version
     end
@@ -226,6 +222,10 @@ module KubernetesDeploy
     end
 
     private
+
+    def global_resource_names
+      cluster_resource_discoverer.global_resource_kinds
+    end
 
     def kubeclient_builder
       @kubeclient_builder ||= KubeclientBuilder.new
--- a/lib/kubernetes-deploy/deploy_task.rb
+++ b/lib/kubernetes-deploy/deploy_task.rb
@@ -30,7 +30,6 @@ require 'kubernetes-deploy/kubernetes_resource'
   custom_resource_definition
   horizontal_pod_autoscaler
   secret
-  global_kubernetes_resource
 ).each do |subresource|
   require "kubernetes-deploy/kubernetes_resource/#{subresource}"
 end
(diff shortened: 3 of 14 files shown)


lib/kubernetes-deploy/deploy_task.rb
```diff
@@ -100,13 +101,17 @@ def prune_whitelist
...
+    end
+
     def server_version
       kubectl.server_version
     end
 
     def initialize(namespace:, context:, current_sha:, logger: nil, kubectl_instance: nil, bindings: {},
       max_watch_seconds: nil, selector: nil, template_paths: [], template_dir: nil, protected_namespaces: nil,
-      render_erb: true)
+      render_erb: true, allow_globals: false)
```
dturn: Allows us to maintain backward compatibility of kubernetes-deploy. This will be removed when the exe files go away

KnVerey: 🤔Makes sense, but we should probably make sure that `Krane::DeployTask` never has this option. Maybe the transitional version we release will also need to have two classes in addition to two executables, where the `Krane` one is just delegating to `KubernetesDeploy`. Maybe we release the transitional version as `krane`, but as `0.1.0` and then `1.0.0` is when kubernetes-deploy is ripped out?

dturn: I know rails used to release `X.99.0` right before `X+1.0.0` so you could bump to `X.99.0` and check for deprecation warnings before bumping to `X+1.0.0`. 

We're in a bit of branch from brach crunch. But, that could be a good pattern for making breaking changes.

lib/kubernetes-deploy/cluster_resource_discovery.rb
```diff
@@ -16,8 +16,24 @@ def crds
       end
     end
 
+    def globals
+      @globals ||= fetch_globals.map do |gv|
+        kind, group = gv.split(".", 2)
+        { group: group, kind: kind.singularize }
```
KnVerey: `singularize` isn't necessarily going to give us the real kind. You can get it from kubectl if you use `-o wide` instead of `-o name`. Seems like the kind is the only part you actually use anyway, so maybe this should just be `global_kinds` and return an array?

Incidentally, I noticed the docs have a worrying disclaimer: "(This output can be retrieved from kubectl api-resources, and is accurate as of Kubernetes 1.13.3.)" but it [turns out](https://github.com/kubernetes/website/pull/13223) that was supposed to mean that the copy of the output mentioned in the docs was snapshotted from 1.13.3. So this should be fine to use, as you'd hope.

lib/krane/deploy_task.rb
```diff
@@ -0,0 +1,11 @@
+# frozen_string_literal: true
+
+require 'kubernetes-deploy/deploy_task'
+
+module Krane
+  class DeployTask < KubernetesDeploy::DeployTask
+    def initialize(**args)
```
KnVerey: Not sure how much this matters, but technically doesn't this mean `Krane::DeployTask` _does_ have the `allow_globals` argument (you won't get an argument error), but we overwrite it? We could manually raise the argument error I guess.

lib/kubernetes-deploy/deploy_task.rb
```diff
@@ -261,16 +267,35 @@ def validate_resources(resources)
...
 
+    def validate_globals(resources)
+      return unless (global = resources.select(&:global?).presence)
+      global_names = global.map do |resource|
+        "#{resource.name} (#{resource.type}) in #{File.basename(resource.file_path)}"
+      end
+      global_names = FormattedLogger.indent_four(global_names.join("\n"))
+
+      if @allow_globals
+        msg = "The ability for this task to deploy global resources will be removed in the next version:"
```
KnVerey:
```suggestion
        msg = "The ability for this task to deploy global resources will be removed in the next version, which will affect the following resources:"
```

lib/kubernetes-deploy/deploy_task.rb
```diff
@@ -261,16 +267,35 @@ def validate_resources(resources)
...
+        msg += "\n#{global_names}"
+        @logger.summary.add_paragraph(ColorizedString.new(msg).yellow)
+      else
+        @logger.summary.add_paragraph(ColorizedString.new("Global resources:\n#{global_names}").yellow)
+        raise FatalDeploymentError, "Deploying global resource is not allowed from this command."
```
KnVerey:
```suggestion
        raise FatalDeploymentError, "This command is namespaced and cannot be used to deploy global resources."
```

lib/kubernetes-deploy/kubernetes_resource.rb
```diff
@@ -50,8 +50,13 @@ def build(namespace:, context:, definition:, logger:, statsd_tags:, crd: nil)
         if crd
           CustomResource.new(crd: crd, **opts)
         else
-          inst = new(**opts)
-          inst.type = definition["kind"]
+          type = definition["kind"]
+          inst = if globals.include?(type.downcase)
+            KubernetesDeploy::GlobalKubernetesResource.new(**opts)
```
KnVerey: What if instead of having a new class, we make `global` settable, and handle it like we do `type` below? I think might be a bit simpler overall.

test/integration/kubernetes_deploy_test.rb
```diff
@@ -1593,6 +1592,25 @@ def test_fail_erb_templates_if_rendering_is_disabled
     ], in_order: true)
   end
 
+  def test_deploy_allow_globals_warns
+    result = deploy_fixtures("globals")
```
KnVerey: Don't we already have a fixture set with a global resource in it (CRD or storage class probably?)?

test/helpers/fixture_deploy_helper.rb
```diff
@@ -62,7 +62,7 @@ def deploy_raw_fixtures(set, wait: true, bindings: {}, subset: nil)
 
   def deploy_dirs_without_profiling(dirs, wait: true, allow_protected_ns: false, prune: true, bindings: {},
     sha: "k#{SecureRandom.hex(6)}", kubectl_instance: nil, max_watch_seconds: nil, selector: nil,
-    protected_namespaces: nil, render_erb: true)
+    protected_namespaces: nil, render_erb: true, allow_globals: true)
```
KnVerey: Is the plan to change the default when we remove `KubernetesDeploy`?

lib/krane/deploy_task.rb
```diff
@@ -0,0 +1,11 @@
+# frozen_string_literal: true
+
+require 'kubernetes-deploy/deploy_task'
+
+module Krane
+  class DeployTask < KubernetesDeploy::DeployTask
```
KnVerey: Does this implicitly get coverage from your CLI tests, or should we be using the krane version of the code a new integration test?

lib/kubernetes-deploy/deploy_task.rb
KnVerey: For backwards compatibility, shouldn't the default be true?

lib/krane/deploy_task.rb
dturn: It gets coverage form the black_box_tests

test/helpers/fixture_deploy_helper.rb
dturn: yep.

test/integration/kubernetes_deploy_test.rb
dturn: We do, my thought was this would be the start of the `hello-cloud` but for global resources. I'm fine to save that for the first person to actually add real code to a global resource.

lib/kubernetes-deploy/kubernetes_resource.rb
dturn: I went back and forth on this couldn't decide which was better. This felt more future proof, though I can't think of any special behavior we'd want from a GlobalResource., but the approach of not using a new class was definitely simpler. I'll try the other approach and see  how it goes.

lib/kubernetes-deploy/cluster_resource_discovery.rb
dturn: I'd prefer to use `singularize` over parsing the output to wide, but 👍 to the array approach.

KnVerey: It really isn't safe to assume that `singularize` (some logic somebody wrote into ruby) will generate the real `kind` name (which is explicitly chosen by someone writing a new resource). In theory the plural, the singular and the kind could all be completely different words. I'm not sure there are examples of unexpected plural forms in core resources, but CRs can be global too.

lib/kubernetes-deploy/kubernetes_resource.rb
KnVerey: 🤔 If we considered server resource lists part of the task's global operating context and put them into our new task config object, then we'd have a third option that might read even more straightforwardly: inside the normal superclass, we could do essentially `def global?; @task_config.global_resource_kinds.include?(kind); end`.

test/integration/kubernetes_deploy_test.rb
KnVerey: True, we're going to have to separate that out entirely, so a purpose-built fixture set makes sense.

lib/kubernetes-deploy/kubernetes_resource.rb
dturn: I think that makes sense to do in https://github.com/Shopify/kubernetes-deploy/issues/572.

lib/kubernetes-deploy/cluster_resource_discovery.rb
dturn: Good point, code now parses wide. This was a bit tricker than expected, some columns don't have values so just splitting on whitespace doesn't work, instead I took advantage of the fixed width of each column.

lib/kubernetes-deploy/deploy_task.rb
```diff
@@ -100,13 +100,17 @@ def prune_whitelist
       wl + cluster_resource_discoverer.crds.select(&:prunable?).map(&:group_version_kind)
     end
 
+    def global_resource_names
```
KnVerey: why is this method public?

lib/kubernetes-deploy/kubernetes_resource.rb
```diff
@@ -50,8 +50,10 @@ def build(namespace:, context:, definition:, logger:, statsd_tags:, crd: nil)
         if crd
           CustomResource.new(crd: crd, **opts)
         else
+          type = definition["kind"]
           inst = new(**opts)
-          inst.type = definition["kind"]
+          inst.type = type
+          inst.global = true if global_names.map(&:downcase).include?(type.downcase)
```
KnVerey:
```suggestion
          inst.global = global_names.map(&:downcase).include?(type.downcase)
```

lib/kubernetes-deploy/cluster_resource_discovery.rb
```diff
@@ -16,8 +16,26 @@ def crds
       end
     end
 
+    def global_resource_names
```
KnVerey: nit: this is a list of kinds, not names (which strictly speaking are defined separately)

lib/kubernetes-deploy/cluster_resource_discovery.rb
```diff
@@ -16,8 +16,26 @@ def crds
...
+      @globals ||= fetch_globals.map { |g| g["kind"] }
+    end
+
     private
 
+    def fetch_globals
+      raw, _, st = kubectl.run("api-resources", "--namespaced=false", output: "wide", attempts: 5)
+      if st.success?
```
KnVerey: I _think_ I understand what you're doing here, but this code is so dense that it is hard to follow. Can you please expand it to do fewer things per line. stick to descriptive variables, and maybe add a comment or two?

I'm also thinking this warrants a few unit tests showing it working with different column widths.

lib/kubernetes-deploy/cluster_resource_discovery.rb
```diff
@@ -16,8 +16,26 @@ def crds
...
+        rows = raw.split("\n")
+        header = rows.shift.downcase.scan(/[a-z]+[\W]*/).each_with_object({}) do |match, hash|
+          start = hash.values.map(&:last).max.to_i
```
KnVerey: This seems to be reflecting on the results of the block up to this point in the iteration. Could we be storing this value as we iterate instead of computing it like this? Isn't it the length of the header string we've processed so far, aka the end of the match we last processed?

test/unit/cluster_resource_discovery_test.rb
```diff
@@ -0,0 +1,51 @@
...
+  def test_global_resource_kinds_failure
+    crd = mocked_cluster_resource_discovery(nil, success: false)
+    kinds = crd.global_resource_kinds
+    assert_equal(kinds, [])
+  end
+
+  def test_global_resource_kinds_success
+    crd = mocked_cluster_resource_discovery(full_response)
+    kinds = crd.global_resource_kinds
+    %w(MutatingWebhookConfiguration ComponentStatus ComponentStatus).each do |kind|
```
KnVerey: The second and third are the same, presumably not intentional? I'd also suggest adding an assertion on the total number of kinds discovered.

PR #569, opened 2019-09-25
Add DeployTaskConfigValidator to deploy task

lei-lo:

Part of #344 to be consistent about preflight validations.

**How is this accomplished?**

Added `DeployTaskConfigValidator` which takes care of namespace and context checks via `TaskConfigValidator` plus the deploy-specific check on protected namespaces. This allowed us to remove the duplicate validations from `deploy_task.rb`.

We've decided to leave `template_sets` and `ejson` validations in `deploy_task.rb` because doing so in this PR would make it really big. It's open to refactoring in a later PR if deemed necessary.

**What could go wrong?**

We didn't refactor correctly and validations aren't carried out properly?

### DIFF
--- a/lib/kubernetes-deploy/deploy_task.rb
+++ b/lib/kubernetes-deploy/deploy_task.rb
@@ -332,8 +332,8 @@ module KubernetesDeploy
     end
 
     def validate_configuration(allow_protected_ns:, prune:)
-      task_config_validator =
-        DeployTaskConfigValidator.new(@namespace, allow_protected_ns, prune, @task_config, kubectl, kubeclient_builder)
+      task_config_validator = DeployTaskConfigValidator.new(@protected_namespaces, allow_protected_ns, prune,
+        @task_config, kubectl, kubeclient_builder)
       errors = []
       errors += task_config_validator.errors
       errors += @template_sets.validate
--- a/lib/kubernetes-deploy/deploy_task.rb
+++ b/lib/kubernetes-deploy/deploy_task.rb
@@ -40,6 +40,7 @@ require 'kubernetes-deploy/ejson_secret_provisioner'
 require 'kubernetes-deploy/renderer'
 require 'kubernetes-deploy/cluster_resource_discovery'
 require 'kubernetes-deploy/template_sets'
+require 'kubernetes-deploy/deploy_task_config_validator'
 
 module KubernetesDeploy
   class DeployTask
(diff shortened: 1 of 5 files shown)


dirceu:
There are still some tests to update; I'll follow up on that.

lib/kubernetes-deploy/deploy_task.rb
```diff
@@ -328,36 +329,17 @@ def record_warnings(warning:, filename:)
     end
 
     def validate_configuration(allow_protected_ns:, prune:)
+      task_config_validator =
+        DeployTaskConfigValidator.new(@namespace, allow_protected_ns, prune, @task_config, kubectl, kubeclient_builder)
```
dturn: You shouldn't need to pass the items already passed to `@task_config`

lib/kubernetes-deploy/deploy_task_config_validator.rb
```diff
@@ -0,0 +1,35 @@
+# frozen_string_literal: true
+module KubernetesDeploy
+  class DeployTaskConfigValidator < TaskConfigValidator
+    PROTECTED_NAMESPACES = %w(
```
dturn: Does this constant make sense with the changes to `krane deploy` to pass in the list of protected namespaces?

dirceu: No, we'll be able to remove that after the `krane deploy` changes are merged.

dturn:
I'm going to hold off on reviewing this again till after it gets rebased on top of the `krane deploy` pr

exe/kubernetes-deploy
```diff
@@ -82,7 +82,6 @@ begin
 
     runner.run!(
       verify_result: !skip_wait,
-      allow_protected_ns: allow_protected_ns,
```
dturn: I think we need to maintain backwards compatibility of the exe commands until we delete them.

dirceu: Makes sense. Updating...

lib/kubernetes-deploy/deploy_task.rb
```diff
@@ -528,37 +510,6 @@ def find_bad_files_from_kubectl_output(line)
...
-      end
-      raise FatalDeploymentError, "Failed to reach server for #{@context}" unless success
-      TaskConfigValidator.new(@task_config, kubectl, kubeclient_builder, only: [:validate_server_version]).valid?
-    end
-
-    def confirm_namespace_exists
-      raise FatalDeploymentError, "Namespace #{@namespace} not found" unless namespace_definition.present?
-      @logger.info("Namespace #{@namespace} found")
-    end
-
```
jessie-JNing: The cluster and namespace exist validation is done by https://github.com/Shopify/kubernetes-deploy/blob/cb254bb07450ce99ffaaa32e3c218c46a5494a46/lib/kubernetes-deploy/task_config_validator.rb#L19
That's cool~

PR #570, opened 2019-09-25
Display full deploy error log after server dry run

jessie-JNing:

Follow up work: https://github.com/Shopify/kubernetes-deploy/issues/484

As we bring in server-dry-run in https://github.com/Shopify/kubernetes-deploy/pull/553 for resources type with sensitive information. Spec validation error is able to be caught during template validation phrase, we can display a full error message if deploy failure. 

The pre-condition is we trust kubernetes will not leak sensitive info. 

**How is this accomplished?**
- Add instance variable `server_dry_run_validated` to kubernetes_resource, default to be false, if server-dry-run succeeds during template validation, set it to true.
- During deploy, if detect `server_dry_run_validated` is true, display full error message. 

**What could go wrong?**
Pre-condition is not true, Kubernetes may leak sensitive info .

### DIFF
--- a/lib/kubernetes-deploy/deploy_task.rb
+++ b/lib/kubernetes-deploy/deploy_task.rb
@@ -483,10 +483,15 @@ module KubernetesDeploy
         bad_files = find_bad_files_from_kubectl_output(line)
         if bad_files.present?
           bad_files.each do |f|
-            if filenames_with_sensitive_content.include?(f[:filename]) && !server_dry_run_validated_resource.include?(f[:filename])
+            if filenames_with_sensitive_content.include?(f[:filename])
               # Hide the error and template contents in case it has sensitive information
-              # we display full error messages from kubernetes as we assume there's no sensitive info leak after server-dry-run
-              record_invalid_template(err: "SUPPRESSED FOR SECURITY", filename: f[:filename], content: nil)
+              # we display full error messages as we assume there's no sensitive info leak after server-dry-run
+              err_msg = if server_dry_run_validated_resource.include?(f[:filename])
+                f[:err]
+              else
+                "SUPPRESSED FOR SECURITY"
+              end
+              record_invalid_template(err: err_msg, filename: f[:filename], content: nil)
             else
               record_invalid_template(err: f[:err], filename: f[:filename], content: f[:content])
             end
--- a/lib/kubernetes-deploy/deploy_task.rb
+++ b/lib/kubernetes-deploy/deploy_task.rb
@@ -475,12 +475,17 @@ module KubernetesDeploy
         .select(&:sensitive_template_content?)
         .map { |r| File.basename(r.file_path) }
 
+      server_dry_run_validated_resource = resources
+        .select(&:server_dry_run_validated?)
+        .map { |r| File.basename(r.file_path) }
+
       err.each_line do |line|
         bad_files = find_bad_files_from_kubectl_output(line)
         if bad_files.present?
           bad_files.each do |f|
-            if filenames_with_sensitive_content.include?(f[:filename])
-              # Hide the error and template contents in case it has senitive information
+            if filenames_with_sensitive_content.include?(f[:filename]) && !server_dry_run_validated_resource.include?(f[:filename])
+              # Hide the error and template contents in case it has sensitive information
+              # we display full error messages from kubernetes as we assume there's no sensitive info leak after server-dry-run
               record_invalid_template(err: "SUPPRESSED FOR SECURITY", filename: f[:filename], content: nil)
             else
               record_invalid_template(err: f[:err], filename: f[:filename], content: f[:content])
@@ -490,12 +495,13 @@ module KubernetesDeploy
           unidentified_errors << line
         end
       end
+      return unless unidentified_errors.any?
 
-      if unidentified_errors.present? && filenames_with_sensitive_content.any?
+      if filenames_with_sensitive_content.any? { |f| !server_dry_run_validated_resource.include?(f) }
         warn_msg = "WARNING: There was an error applying some or all resources. The raw output may be sensitive and " \
           "so cannot be displayed."
         @logger.summary.add_paragraph(ColorizedString.new(warn_msg).yellow)
-      elsif unidentified_errors.present?
+      else
         heading = ColorizedString.new('Unidentified error(s):').red
         msg = FormattedLogger.indent_four(unidentified_errors.join)
         @logger.summary.add_paragraph("#{heading}\n#{msg}")
--- a/lib/kubernetes-deploy/deploy_task.rb
+++ b/lib/kubernetes-deploy/deploy_task.rb
@@ -481,23 +481,21 @@ module KubernetesDeploy
 
       err.each_line do |line|
         bad_files = find_bad_files_from_kubectl_output(line)
-        if bad_files.present?
-          bad_files.each do |f|
-            if filenames_with_sensitive_content.include?(f[:filename])
-              # Hide the error and template contents in case it has sensitive information
-              # we display full error messages as we assume there's no sensitive info leak after server-dry-run
-              err_msg = if server_dry_run_validated_resource.include?(f[:filename])
-                f[:err]
-              else
-                "SUPPRESSED FOR SECURITY"
-              end
-              record_invalid_template(err: err_msg, filename: f[:filename], content: nil)
-            else
-              record_invalid_template(err: f[:err], filename: f[:filename], content: f[:content])
-            end
-          end
-        else
+        unless bad_files.present?
           unidentified_errors << line
+          next
+        end
+
+        bad_files.each do |f|
+          err_msg = f[:err]
+          if filenames_with_sensitive_content.include?(f[:filename])
+            # Hide the error and template contents in case it has sensitive information
+            # we display full error messages as we assume there's no sensitive info leak after server-dry-run
+            err_msg = "SUPPRESSED FOR SECURITY" unless server_dry_run_validated_resource.include?(f[:filename])
+            record_invalid_template(err: err_msg, filename: f[:filename], content: nil)
+          else
+            record_invalid_template(err: err_msg, filename: f[:filename], content: f[:content])
+          end
         end
       end
       return unless unidentified_errors.any?
--- a/lib/kubernetes-deploy/deploy_task.rb
+++ b/lib/kubernetes-deploy/deploy_task.rb
@@ -502,7 +502,7 @@ module KubernetesDeploy
       end
       return unless unidentified_errors.any?
 
-      if filenames_with_sensitive_content.any? { |f| !server_dry_run_validated_resource.include?(f) }
+      if (filenames_with_sensitive_content - server_dry_run_validated_resource).present?
         warn_msg = "WARNING: There was an error applying some or all resources. The raw output may be sensitive and " \
           "so cannot be displayed."
         @logger.summary.add_paragraph(ColorizedString.new(warn_msg).yellow)
--- a/lib/kubernetes-deploy/kubernetes_resource.rb
+++ b/lib/kubernetes-deploy/kubernetes_resource.rb
@@ -15,6 +15,7 @@ module KubernetesDeploy
     GLOBAL = false
     TIMEOUT = 5.minutes
     LOG_LINE_COUNT = 250
+    SERVER_DRY_RUN_MIN_VERSION = "1.13"
     SERVER_DRY_RUN_DISABLED_ERROR =
       /(unknown flag: --server-dry-run)|(doesn't support dry-run)|(dryRun alpha feature is disabled)/
 
@@ -347,7 +348,7 @@ module KubernetesDeploy
       self.class::SENSITIVE_TEMPLATE_CONTENT
     end
 
-    def server_dry_runnable?
+    def server_dry_runnable_resource?
       self.class::SERVER_DRY_RUNNABLE
     end
 
@@ -476,11 +477,15 @@ module KubernetesDeploy
     end
 
     def validate_spec_with_kubectl(kubectl)
-      _, err, st = validate_with_dry_run_option(kubectl, "--dry-run")
-      if st.success? && server_dry_runnable?
+      err = ""
+      if server_dry_run_enabled?(kubectl) && server_dry_runnable_resource?
         _, err, st = validate_with_dry_run_option(kubectl, "--server-dry-run")
         @server_dry_run_validated = st.success?
-        return true if st.success? || err.match(SERVER_DRY_RUN_DISABLED_ERROR)
+        return true if st.success?
+      end
+
+      if err.empty? || err.match(SERVER_DRY_RUN_DISABLED_ERROR)
+        _, err, st = validate_with_dry_run_option(kubectl, "--dry-run")
       end
 
       return true if st.success?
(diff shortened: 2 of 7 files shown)


lib/kubernetes-deploy/deploy_task.rb
```diff
@@ -475,13 +475,23 @@ def record_apply_failure(err, resources: [])
...
+        .map { |r| File.basename(r.file_path) }
+
       err.each_line do |line|
         bad_files = find_bad_files_from_kubectl_output(line)
         if bad_files.present?
           bad_files.each do |f|
             if filenames_with_sensitive_content.include?(f[:filename])
-              # Hide the error and template contents in case it has senitive information
-              record_invalid_template(err: "SUPPRESSED FOR SECURITY", filename: f[:filename], content: nil)
+              # Hide the error and template contents in case it has sensitive information
```
dturn: I don't think we should pass the file content. I don't think I've ever seen it triggered so I think being overly cautions here is ok. My main concern is that in the future there is a new sensitive type that we correctly know is sensitive but doesn't match the regexp check `/kind:\s*Secret/`. So `record_invalid_template` blindly dumps the file to logs. 

I think passing the err is fine since the entire PR is based on the idea that with `server-dry-run` the errors at the apply stage wont contain sensitive values.

lib/kubernetes-deploy/deploy_task.rb
```diff
@@ -490,12 +500,13 @@ def record_apply_failure(err, resources: [])
           unidentified_errors << line
         end
       end
+      return unless unidentified_errors.any?
 
-      if unidentified_errors.present? && filenames_with_sensitive_content.any?
+      if filenames_with_sensitive_content.any? { |f| !server_dry_run_validated_resource.include?(f) }
```
dturn: I think this is the same and easier to read`(filenames_with_sensitive_content - server_dry_run_validated_resource).present?`

jessie-JNing: Ah, that's is an easier way 👍

lib/kubernetes-deploy/deploy_task.rb
jessie-JNing: Yes, I think we are good to pass the error if resource is able to go through `server-dry-run`.

About the concern in `record_invalid_template`, I feel the reason we have `/kind:\s*Secret/` is we worry there's resource who has `/kind:\s*Secret/` but we didn't recognize it as `secret`. I kind of think it's over caution.

KnVerey: > I feel the reason we have /kind:\s*Secret/ is we worry there's resource who has /kind:\s*Secret/ but we didn't recognize it as secret. 

IIRC the regex is there because `record_invalid_template` is used for errors occurring during rendering too, and at that point it might be possible for the resource to have been completely invalid and therefore not recognized as anything in particular. The `kind: Secret` scan is a last-ditch effort to prevent leaks when our more robust upstream checks failed. Is it necessary? I hope not. But since there is in reality only this one sensitive resource, I believe the consensus when it went it was that it doesn't hurt.

KnVerey: More importantly, didn't we agree to trust kubectl output for the error message itself (and just suppress the template content)? https://github.com/Shopify/kubernetes-deploy/issues/484#issuecomment-496594881

lib/kubernetes-deploy/kubernetes_resource.rb
```diff
@@ -474,9 +479,8 @@ def validate_spec_with_kubectl(kubectl)
       _, err, st = validate_with_dry_run_option(kubectl, "--dry-run")
```
KnVerey: Two questions related to the previous PR:
- If server dry run is available, is there actually any value in running the client-side dry run as well?
- For a server-side dry run to be possible, both the resource and the server have to support it, but we seem to only look at the resource, and blindly make an API request per secret even when we could have known in advance that all of those requests will be doomed. I'm wondering if the context object we were discussing passing around (namespace, context, logger IIRC) should also contain (or be able to get) the server version for use in places like this. Was any work done on that object?

jessie-JNing: Thanks @KnVerey for reviewing it. 
- > If server dry run is available, is there actually any value in running the client-side dry run as well?

I believe `--server-dry-run` does work including something client dry-run does. I didn't find direct answer, but from help doc:
```
$ k apply --help | grep dry-run
      --dry-run=false: If true, only print the object that would be sent, without sending it. Warning: --dry-run cannot accurately output the result of merging the local manifest and the server-side data. Use --server-dry-run to get the merged result instead.
      --server-dry-run=false: If true, request will be sent to server with dry-run flag, which means the modifications won't be persisted. This is an alpha feature and flag.
```
My local test also shows error caught during client dry-run can also be caught by server dry-run

- > For a server-side dry run to be possible ...

This is a good question. `--server-dry-run` is an alpha feature in `v1.12`, which is disable by default. It becomes to be a beta feature in `v1.13`, which enable by default then. 
We consider about checking the availability of server-dry-run by comparing server version, but we concern that if people use version after `v1.13` but cluster admins disable this feature, that'll be sad. 

But you remind me, we can do a version check before. If version is after `v1.13` and we find cluster admin disable it, we should catch feature disable error and go back to client dry-run. @KnVerey @dturn  what do you think?

lib/kubernetes-deploy/deploy_task.rb
jessie-JNing: `record_invalid_template` is also used during rendering, yeah, agree that `/kind:\sSecret/ ` is a last-ditch effort we need.  

> didn't we agree to trust kubectl output for the error message itself (and just suppress the template content)? #484 (comment)

We did an experiment and raise issue https://github.com/kubernetes/kubernetes/issues/82421, it looks we can error message after server dry-run. Did I understand your comment correctly?

lib/kubernetes-deploy/kubernetes_resource.rb
KnVerey: I'm not sure whether we put this into our contributor guidelines, but we had a case in the past where someone was arguing for support of an alpha resource, and we took the position that we would not add code to support alphas. For consistency with that, I'd be inclined to say that we don't support using the server side dry run against a 1.12 cluster, even if it is enabled. 

> But you remind me, we can do a version check before. If version is after v1.13 and we find cluster admin disable it, we should catch feature disable error and go back to client dry-run.

I agree with that. So:
- If server version >= 1.13, attempt to use server-side dry run _instead of_ client-side
- If server version < 1.13, or server-side dry run returns the "not supported" error, run client-side dry run

lib/kubernetes-deploy/deploy_task.rb
KnVerey: In issue 484, I identified three related problems:
1- was kubernetes-deploy right to fail this deploy?
2-the error was censored, and didn't need to be
3-secret keys are subject to a validation that dry run isn't catching

Server-side dry run addresses problem 3. What I'm talking about is our conclusion on problem 2, which I think would considerably simplify your code here. tl;dr Danny and I concluded on the issue that upstream is being careful about what they put in the error message they give us, and suppressing the error message was overzealous on our part.

lib/kubernetes-deploy/kubernetes_resource.rb
dturn: I didn't know about our alpha restrictions, and its not in the docs so we should add it to contributing.md.

lib/kubernetes-deploy/deploy_task.rb
KnVerey: Never mind -- I missed the part of the issue you opened that _proved_ that apply can dump secret content at least in one specific case. That's enough to warrant what you're doing here. 👍

lib/kubernetes-deploy/deploy_task.rb
```diff
@@ -475,13 +475,23 @@ def record_apply_failure(err, resources: [])
         .select(&:sensitive_template_content?)
         .map { |r| File.basename(r.file_path) }
 
+      server_dry_run_validated_resource = resources
+        .select(&:server_dry_run_validated?)
+        .map { |r| File.basename(r.file_path) }
+
       err.each_line do |line|
         bad_files = find_bad_files_from_kubectl_output(line)
         if bad_files.present?
```
KnVerey: This conditional is getting very nested at this point. Maybe it would be clearer if we turn this piece into an `unless` (and used `next` in it), to at least get rid of one layer?

lib/kubernetes-deploy/kubernetes_resource.rb
jessie-JNing: Sounds good, thanks for the reminding and making the decision. I added the version check, we'll run server-dry-run when version >=1.13

lib/kubernetes-deploy/deploy_task.rb
jessie-JNing: Yeah, the apply error message does leak secret content unfortunately. We can only trust output after server-dry-run for now.

lib/kubernetes-deploy/deploy_task.rb
jessie-JNing: I update the nested if logic here, easy to check out the change 🙂 : https://github.com/Shopify/kubernetes-deploy/pull/570/commits/ff7bbb1e32e7970d62eeae14aa3c7418417aacf0

jessie-JNing:
Deploy secret below to namespace
<details><summary>secret.yaml</summary>
<p>

```
---
apiVersion: v1
kind: Secret
metadata:
  name: hello-secret
type: Opaque
data:
  username: my_username
  password: cGFzc3dvcmQ=
```
</p>
</details>

Change type to be `invalid/type`, and deploy it again.
It should fail as type is an immutable field. 

🎩 on `v1.13.6` minikube:
```
REVISION="4f7d6697fb5a20d4c960cd1cc6530774069489b6" bundle exec kubernetes-deploy cloud-portal-standby minikube --template-dir=/Users/jessie.ning/Documents/k8s/quick-resource/sensitive_info
[INFO][2019-09-27 14:02:01 -0400]       
[INFO][2019-09-27 14:02:01 -0400]       ------------------------------------Phase 1: Initializing deploy------------------------------------
[INFO][2019-09-27 14:02:01 -0400]       Context minikube found
[INFO][2019-09-27 14:02:01 -0400]       Namespace cloud-portal-standby found
[INFO][2019-09-27 14:02:01 -0400]       All required parameters and files are present
[INFO][2019-09-27 14:02:01 -0400]       Discovering resources:
[INFO][2019-09-27 14:02:01 -0400]         - Secret/hello-secret
[INFO][2019-09-27 14:02:01 -0400]         - Secret/hello-bad-secret
[INFO][2019-09-27 14:02:02 -0400]       
[INFO][2019-09-27 14:02:02 -0400]       ------------------------------------------Result: FAILURE-------------------------------------------
[FATAL][2019-09-27 14:02:02 -0400]      Template validation failed
[FATAL][2019-09-27 14:02:02 -0400]      
[FATAL][2019-09-27 14:02:02 -0400]      Invalid template: Secret-hello-bad-secret20190927-11636-1avzmq4.yml
[FATAL][2019-09-27 14:02:02 -0400]      > Error message:
[FATAL][2019-09-27 14:02:02 -0400]          Validation for Secret/hello-bad-secret failed. Detailed information is unavailable as the raw error may contain sensitive data.

```

🎩  on `v1.12.6`
```
REVISION="4f7d6697fb5a20d4c960cd1cc6530774069489b6" bundle exec kubernetes-deploy cloud-portal-standby minikube --template-dir=/Users/jessie.ning/Documents/k8s/quick-resource/sensitive_info
[INFO][2019-09-27 14:13:00 -0400]       
[INFO][2019-09-27 14:13:00 -0400]       ------------------------------------Phase 1: Initializing deploy------------------------------------
[INFO][2019-09-27 14:13:01 -0400]       Context minikube found
[INFO][2019-09-27 14:13:01 -0400]       Namespace cloud-portal-standby found
[INFO][2019-09-27 14:13:01 -0400]       All required parameters and files are present
[INFO][2019-09-27 14:13:01 -0400]       Discovering resources:
[INFO][2019-09-27 14:13:01 -0400]         - Secret/hello-secret
[INFO][2019-09-27 14:13:01 -0400]         - Secret/hello-bad-secret
[INFO][2019-09-27 14:13:01 -0400]       
[INFO][2019-09-27 14:13:01 -0400]       ----------------------------Phase 2: Checking initial resource statuses-----------------------------
[INFO][2019-09-27 14:13:01 -0400]       Secret/hello-bad-secret                           Available
[INFO][2019-09-27 14:13:01 -0400]       Secret/hello-secret                               Available
[INFO][2019-09-27 14:13:01 -0400]       
[INFO][2019-09-27 14:13:01 -0400]       ------------------------------Phase 3: Predeploying priority resources------------------------------
[INFO][2019-09-27 14:13:01 -0400]       Deploying resources:
[INFO][2019-09-27 14:13:01 -0400]       - Secret/hello-bad-secret (timeout: 30s)
[INFO][2019-09-27 14:13:01 -0400]       - Secret/hello-secret (timeout: 30s)
[INFO][2019-09-27 14:13:01 -0400]       
[INFO][2019-09-27 14:13:01 -0400]       ------------------------------------------Result: FAILURE-------------------------------------------
[FATAL][2019-09-27 14:13:01 -0400]      Command failed: apply -f /var/folders/l2/ym6whn6d4bs697_w9w_9f73w0000gn/t/d20190927-17904-1jym9ng
[FATAL][2019-09-27 14:13:01 -0400]      
[FATAL][2019-09-27 14:13:01 -0400]      WARNING: Any resources not mentioned in the error(s) below were likely created/updated. You may wish to roll back this deploy.
[FATAL][2019-09-27 14:13:01 -0400]      
[FATAL][2019-09-27 14:13:01 -0400]      WARNING: There was an error applying some or all resources. The raw output may be sensitive and so cannot be displayed.

```

test/integration/kubernetes_deploy_test.rb
```diff
@@ -320,7 +320,7 @@ def test_dynamic_erb_collection_works
   # https://github.com/kubernetes/kubernetes/issues/42057 shows how this manifested for a particular field,
   # and although that particular case has been fixed, other invalid specs still aren't caught until apply.
   def test_multiple_invalid_k8s_specs_fails_on_apply_and_prints_template
-    result = deploy_fixtures("hello-cloud", subset: ["web.yml.erb"]) do |fixtures|
+    result = deploy_fixtures("hello-cloud", subset: ["web.yml.erb", "secret.yml"]) do |fixtures|
```
dturn: I gave you bad advice about which test to use. Turns outs this test only checks the case where all of the errors match in `find_bad_files_from_kubectl_output` so there is functionally no change here. 

It looks like the test that exercises `(filenames_with_sensitive_content - server_dry_run_validated_resource).present?` is `test_apply_failure_with_sensitive_resources_hides_raw_output` except the only resource it has in a secret. So I think we need a new test. Something like: 


```
  def test_apply_failure_with_sensitive_resources_does_not_hide_apply_error
    result = deploy_fixtures("hello-cloud", subset: %w(redis.yml secret.yml))
    assert_deploy_success(result)

    result = deploy_fixtures("hello-cloud", subset: %w(redis.yml secret.yml)) do |fixtures|
      service = fixtures["redis.yml"]["Service"].detect { |s| s.dig("metadata", "name") == "redis-endpoints" }
      service['spec']['clusterIP'] = ''
    end

    assert_deploy_failure(result)
    assert_logs_match('The Service "redis-endpoints" is invalid: spec.clusterIP: Invalid value: "": field is immutable')
    refute_logs_match(/WARNING: .* The raw output may be sensitive and so cannot be displayed./)
    refute_logs_match("password")
  end
```

the redis.yml might not be the best fixture, I was just quickly looking for something with an immutable field. (e.g I'm testing the apply is failing not the template validation).

PR #571, opened 2019-09-26
Force dependency on activesupport 5.x

douglas:
Fix build errors. ActiveSupport 6.0.0 no longer supports Ruby 2.4, which is still supported by this gem:

```
Gem::RuntimeRequirementNotMetError: activesupport requires Ruby version >=
  | 2.5.0. The current ruby version is 2.4.6.354.
```

**How is this accomplished?**
Changing `activesupport` dependency in `kubernetes-deploy.gemspec` to `< 6.0`.

**What could go wrong?**
Nothing.

### DIFF
--- a/kubernetes-deploy.gemspec
+++ b/kubernetes-deploy.gemspec
@@ -23,7 +23,7 @@ Gem::Specification.new do |spec|
   spec.require_paths = %w(lib)
 
   spec.required_ruby_version = '>= 2.4.0'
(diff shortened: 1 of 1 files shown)


douglas:
I talked with Rafael about this and seems it could be a problem on bundler - but somehow we could not reproduce the problem anymore.

Either way I will just close this PR as it was not the right fix.

PR #574, opened 2019-09-27
Krane global deploy

dturn:

Part of the krane 1.0 work is refusing to allow `krane deploy` to deploy global resources. See https://github.com/Shopify/kubernetes-deploy/issues/522 for more of an explanation.

To allow our users to have a better experience than just `kubectl apply` we're adding a new command `krane global-deploy`. This command will not support erb rendering, for that you can pipe the results of `krane render`.  This version does not support pruning or `--label-selector` thought we would accept the feature in future versions.

**How is this accomplished?**
Add a krane global-deploy command and a krane/global_deploy_task. The global_deploy_task is a thin wrapper around `kubernetes_deploy/deploy_task.

**What could go wrong?**

This assumes the approach we want is to share deploy_task with the namespaced version and only override the parts we want. 

We could chop up the deploy task and then make the deploy_task and global_deploy_task be more fully featured.

### DIFF
--- a/lib/krane/cli/global_deploy_command.rb
+++ b/lib/krane/cli/global_deploy_command.rb
@@ -34,6 +34,7 @@ module Krane
 
           deploy.run!(
             verify_result: options["verify-result"],
+            prune: false,
           )
         end
       end
--- a/lib/krane/cli/global_deploy_command.rb
+++ b/lib/krane/cli/global_deploy_command.rb
@@ -1,0 +1,42 @@
+# frozen_string_literal: true
+
+module Krane
+  module CLI
+    class GlobalDeployCommand
+      DEFAULT_DEPLOY_TIMEOUT = '300s'
+      OPTIONS = {
+        "filenames" => { type: :string, banner: '/tmp/my-resource.yml', aliases: :f, required: true,
+                         desc: "Path to file or directory that contains the configuration to apply" },
+        "global-timeout" => { type: :string, banner: "duration", default: DEFAULT_DEPLOY_TIMEOUT,
+                              desc: "Max duration to monitor workloads correctly deployed" },
+        "verify-result" => { type: :boolean, default: true,
+                             desc: "Verify workloads correctly deployed" },
+        "selector" => { type: :string, banner: "'label=value'", required: true,
+                        desc: "Select workloads by selector(s)" },
+      }
+
+      def self.from_options(context, options)
+        require 'krane/global_deploy_task'
+        require 'kubernetes-deploy/options_helper'
+        require 'kubernetes-deploy/label_selector'
+
+        selector = KubernetesDeploy::LabelSelector.parse(options[:selector])
+
+        KubernetesDeploy::OptionsHelper.with_processed_template_paths([options[:filenames]],
+          require_explicit_path: true) do |paths|
+          deploy = ::Krane::GlobalDeployTask.new(
+            context: context,
+            current_sha: ENV["REVISION"],
+            template_paths: paths,
+            max_watch_seconds: KubernetesDeploy::DurationParser.new(options["global-timeout"]).parse!.to_i,
+            selector: selector,
+          )
+
+          deploy.run!(
+            verify_result: options["verify-result"],
+          )
+        end
+      end
+    end
+  end
+end
--- a/lib/krane/cli/krane.rb
+++ b/lib/krane/cli/krane.rb
@@ -7,6 +7,7 @@ require 'krane/cli/restart_command'
 require 'krane/cli/run_command'
 require 'krane/cli/render_command'
 require 'krane/cli/deploy_command'
+require 'krane/cli/global_deploy_command'
 
 module Krane
   module CLI
@@ -55,6 +56,14 @@ module Krane
       def deploy(namespace, context)
         rescue_and_exit do
           DeployCommand.from_options(namespace, context, options)
+        end
+      end
+
+      desc("global-deploy CONTEXT", "Ship global resources to a cluster")
+      expand_options(GlobalDeployCommand::OPTIONS)
+      def global_deploy(context)
+        rescue_and_exit do
+          GlobalDeployCommand.from_options(context, options)
         end
       end
 
--- a/lib/krane/global_deploy_task.rb
+++ b/lib/krane/global_deploy_task.rb
@@ -1,5 +1,6 @@
 # frozen_string_literal: true
 require 'kubernetes-deploy/deploy_task'
+require 'kubernetes-deploy/global_deploy_task_config_validator'
 
 module Krane
   class GlobalDeployTask < KubernetesDeploy::DeployTask
@@ -7,11 +8,11 @@ module Krane
       super(args.merge(allow_globals: true))
     end
 
-    private
-
-    def namespace_definition
-      nil
+    def run!(**args)
+      super(args.merge(task_config_validator: GlobalDeployTaskConfigValidator))
     end
+
+    private
 
     def validate_globals(resources)
       return unless (namespaced = resources.reject(&:global?).presence)
--- a/lib/krane/global_deploy_task.rb
+++ b/lib/krane/global_deploy_task.rb
@@ -9,20 +9,7 @@ module Krane
     end
 
     def run!(**args)
-      super(args.merge(task_config_validator: GlobalDeployTaskConfigValidator))
-    end
-
-    private
-
-    def validate_globals(resources)
-      return unless (namespaced = resources.reject(&:global?).presence)
-      namespaced_names = namespaced.map do |resource|
-        "#{resource.name} (#{resource.type}) in #{File.basename(resource.file_path)}"
-      end
-      namespaced_names = KubernetesDeploy::FormattedLogger.indent_four(namespaced_names.join("\n"))
-
-      @logger.summary.add_paragraph(ColorizedString.new("Namespaced resources:\n#{namespaced_names}").yellow)
-      raise KubernetesDeploy::FatalDeploymentError, "Deploying namespaced resource is not allowed from this command."
+      super(args.merge(task_config_validator: KubernetesDeploy::GlobalDeployTaskConfigValidator))
     end
   end
 end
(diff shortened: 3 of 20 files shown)


lib/krane/cli/krane.rb
```diff
@@ -70,12 +79,6 @@ def rescue_and_exit
         exit(TIMEOUT_EXIT_CODE)
       rescue KubernetesDeploy::FatalDeploymentError
         exit(FAILURE_EXIT_CODE)
-      rescue KubernetesDeploy::DurationParser::ParsingError => e
```
dturn: rebase issue, I'll fix if we like this approach.

lib/krane/global_deploy_task.rb
```diff
@@ -0,0 +1,23 @@
...
+
+module Krane
+  class GlobalDeployTask < KubernetesDeploy::DeployTask
+    def initialize(**args)
+      super(args.merge(allow_globals: true, namespace: 'default', protected_namespaces: ''))
+    end
+
+    private
+
+    def validate_globals(resources)
```
dturn: This method name should be more generic.

lib/kubernetes-deploy/cluster_resource_discovery.rb
```diff
@@ -2,16 +2,16 @@
 
 module KubernetesDeploy
   class ClusterResourceDiscovery
-    def initialize(namespace:, context:, logger:, namespace_tags:)
-      @namespace = namespace
-      @context = context
-      @logger = logger
+    delegate :namespace, :context, :logger, to: :@task_config
+
+    def initialize(task_config:, namespace_tags: [])
```
dturn: Took a pass at using `task_config` this can be reverted if too much is going on in the pr

KnVerey:
> This version does not support pruning or --label-selector thought we would accept the feature in future versions.

Neither needs to be in the initial PR, but we should chat more about whether they're required for the release. I was actually thinking that label selector might be _mandatory_ for a sane global command, given that you're pretty much guaranteed to not be the only actor modifying the global namespace. As for pruning, I'm not sure whether it should be the default, but our value prop definitely goes down without it, having it off by default would be inconsistent, and making it the default would be a horrifying breaking change if it isn't that way to start. One thing I'm sure about is that if we do have pruning, using a label selector with it should be mandatory. 

> This assumes the approach we want is to share deploy_task with the namespaced version and only override the parts we want.

Won't that be impossible once we 🔥 the kubernetes-deploy version, since the krane version should not be capable at all of deploying globals?

dturn:
> Neither needs to be in the initial PR, but we should chat more about whether they're required for the release. I was actually thinking that label selector might be mandatory for a sane global command, given that you're pretty much guaranteed to not be the only actor modifying the global namespace. As for pruning, I'm not sure whether it should be the default, but our value prop definitely goes down without it, having it off by default would be inconsistent, and making it the default would be a horrifying breaking change if it isn't that way to start. One thing I'm sure about is that if we do have pruning, using a label selector with it should be mandatory.

selector I believe is trivial to add back, so I'll do that here. Pruning is not, so that's going to need to be a second PR either way. 

> Won't that be impossible once we 🔥 the kubernetes-deploy version, since the krane version should not be capable at all of deploying globals?

I did not explain what I was thinking very well at all: 

`kubernetes_deploy/deploy_task` -> `krane/private_deploy_task` or some better name that's clear its not to be used directly and possibly always raises an error if used directly. And then have `deploy_task` and `global_deploy_task` inherit from the private version and among other things override the bit that raises. I just can't decide if this could work, or is going to be a mess as each grows.

lib/kubernetes-deploy/cluster_resource_discovery.rb
KnVerey: What do you think about doing this the other way around: moving all of the stuff in this class into TaskConfig? Everything in here is effectively collected global context about the task's execution environment.

lib/krane/global_deploy_task.rb
```diff
@@ -0,0 +1,23 @@
+# frozen_string_literal: true
+require 'kubernetes-deploy/deploy_task'
+
+module Krane
+  class GlobalDeployTask < KubernetesDeploy::DeployTask
+    def initialize(**args)
+      super(args.merge(allow_globals: true, namespace: 'default', protected_namespaces: ''))
```
KnVerey: Hm, it's pretty weird that we have to set a namespace name, especially a real one, on a non-namespaced command. Come to think of it, that may also be an issue with reusing the Resource superclass (though `TaskConfig` may already be the right tool for solving it in that case).

KnVerey:
> kubernetes_deploy/deploy_task -> krane/private_deploy_task or some better name that's clear its not to be used directly and possibly always raises an error if used directly. And then have deploy_task and global_deploy_task inherit from the private version and among other things override the bit that raises. I just can't decide if this could work, or is going to be a mess as each grows.

I think it could work if you make `initialize` and or `run` raise (so that it is an abstract superclass), but I'm not sure that it will be reasonably easy to reason about, as you say. Have you tried the other option yet? My design inkling is that we'll end up with better code overall if we separate them and encapsulate the bits they need to share.

lib/krane/global_deploy_task.rb
dturn: Yeah, once https://github.com/Shopify/kubernetes-deploy/pull/569 merges I can take a second look at removing the requirement.

lib/krane/cli/global_deploy_command.rb
```diff
@@ -0,0 +1,43 @@
...
+            context: context,
+            current_sha: ENV["REVISION"],
+            template_paths: paths,
+            max_watch_seconds: KubernetesDeploy::DurationParser.new(options["global-timeout"]).parse!.to_i,
+            selector: selector,
+          )
+
+          deploy.run!(
+            verify_result: options["verify-result"],
+            prune: false,
```
dturn: This will get flipped to true in a follow-up PR

lib/kubernetes-deploy/deploy_task.rb
```diff
@@ -122,15 +122,15 @@ def server_version
...
+    def initialize(namespace: nil, context:, current_sha:, logger: nil, kubectl_instance: nil, bindings: {},
       max_watch_seconds: nil, selector: nil, template_paths: [], template_dir: nil, protected_namespaces: nil,
       render_erb: true, allow_globals: false)
       template_dir = File.expand_path(template_dir) if template_dir
       template_paths = (template_paths.map { |path| File.expand_path(path) } << template_dir).compact
 
       @logger = logger || KubernetesDeploy::FormattedLogger.build(namespace, context)
       @template_sets = TemplateSets.from_dirs_and_files(paths: template_paths, logger: @logger)
-      @task_config = KubernetesDeploy::TaskConfig.new(context, namespace, @logger)
+      @task_config = KubernetesDeploy::TaskConfig.new(context, namespace, @logger, allow_globals && namespace.blank?)
```
dturn: `allow_globals && namespace.blank?` is the result of having to support the legacy behavior of `kubernetes-deploy`.

lib/kubernetes-deploy/cluster_resource_discovery.rb
dturn: I think this is a reasonable change for a follow-up PR. My original concern was that TaskConfig should actually be the holder of static information, but I don't actually think there would be issues with it having dynamic abilities.

lib/kubernetes-deploy/deploy_task.rb
jessie-JNing: should we also make https://github.com/Shopify/kubernetes-deploy/pull/574/files#diff-4c33bbb177fdf882c41d892f3bde9570R144 be `allow_globals && namespace.blank?`

I'm not very sure about the `@allow_globals` and `@global_mode` in TaskConfig

jessie-JNing: 👍  Thanks for the explanation.

lib/kubernetes-deploy/task_config.rb
```diff
@@ -1,12 +1,14 @@
 # frozen_string_literal: true
 module KubernetesDeploy
   class TaskConfig
-    attr_reader :context, :namespace
+    attr_reader :context, :namespace, :global_mode
+    attr_accessor :namespace_definition
 
-    def initialize(context, namespace, logger = nil)
+    def initialize(context, namespace, logger = nil, global_mode = false)
```
KnVerey: Is "global mode" anything beyond the absence of the namespace?

lib/kubernetes-deploy/task_config_validator.rb
```diff
@@ -11,12 +11,12 @@ class TaskConfigValidator
...
     delegate :context, :namespace, :logger, to: :@task_config
 
-    def initialize(task_config, kubectl, kubeclient_builder, only: nil)
+    def initialize(task_config, kubectl, kubeclient_builder, only: nil, skip: [])
       @task_config = task_config
       @kubectl = kubectl
       @kubeclient_builder = kubeclient_builder
       @errors = nil
-      @validations = only || DEFAULT_VALIDATIONS
+      @validations = (only || DEFAULT_VALIDATIONS) - skip
```
KnVerey: I'm kinda surprised to see that in addition to subclassing, we also need the ability to modify the defaults with both `only` and `skip` options. Am I reading correctly that the `only` option is just used by tests? Could this work:

```ruby
class TaskConfigValidator
  def validations
    DEFAULT_VALIDATIONS
  end
end

class SomeValidator < TaskConfigValidator
  def validations
    DEFAULT_VALIDATIONS - [:validation_to_skip] # or + something, or just a specific list
  end
end
```

Then the superclass isn't messing around with constructing the list, when it should in practice be a static property of each subclass.

lib/kubernetes-deploy/task_config_validator.rb
```diff
@@ -70,10 +70,12 @@ def validate_namespace_exists
...
       end
 
-      _, err, st = @kubectl.run("get", "namespace", "-o", "name", namespace,
-        use_namespace: false, log_failure: false)
+      definition, err, st = @kubectl.run("get", "namespace", namespace,
+        use_namespace: false, log_failure: false, attempts: 3, output: 'json')
 
-      unless st.success?
+      if st.success?
+        @task_config.namespace_definition = JSON.parse(definition, symbolize_names: true)
```
KnVerey: It isn't intuitive that `validate_namespace_exists` would have the side-effect of setting a property on an entirely different object. Could we make `task_config` capable of caching/returning the namespace object, and raising errors we can catch here?

dturn:
Closing for #602

PR #578, opened 2019-09-30
Version 0.29.0

dturn:
Change log for v0.29.0. 

Note: Due to the need to release a hot fix this change-log is the commit immediately after the 0.29.0 tag. Normally it would be the commit the tag is applied to.

**How is this accomplished?**
...

**What could go wrong?**
...

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,11 +3,20 @@
 *Important!*
 - The next release will be 1.0.0, which means that master will contain breaking changes.
 
+## 0.29.0
+
+*Enchantments*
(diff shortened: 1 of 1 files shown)


dalehamel:
@dturn i think this also eliminates support for below Ruby 2.4, should we mention that?

dturn:
Dropping 2.3 was already in the changelog https://github.com/Shopify/kubernetes-deploy/pull/578/files#diff-4ac32a78649ca5bdd8e0ba38b7006a1eR18 . We've been trying to do a better job of PRs updating the changelogs so we don't have much to write on a release. Or were you suggesting making it more obvious?

KnVerey:
Nope, that's fine. Somehow my eyes missed it since it wasn't in a green line, sorry.

Issue #568, opened 2019-09-25
Krane run -f


RyanBrushett:
# Problem 

Command is there but `-f` flag as explained in https://docs.google.com/document/d/1oInUsKplYGNWTymPY48xtDCx3x4Cwc-phcEKUkt2v4U/view# was not added yet.

# Possible solution

- Add the flag. 
    - I imagine it co-existing with the `--template` option and having one or the other be required, though that is not what the document says. Happy to chat about this but I don't think we should remove `--template`.
- Should be repeatable and be able to point to multiple files.

# Alternative solutions

Don't add the flag but I think we're pretty determined that it should be an option for `run`.

KnVerey:
You're right, the current `--template` flag lets you specify a template that is already live in the target namespace _by name_, which is super handy and not what `-f` will cover. I agree that we should keep it, but maybe we can improve the name? I probably had a hand in it, and _I_ find it unclear what the value of the flag should be. `--fetch-pod-template / -t`? Also its default value is kinda random (`task-runner-template`); should we remove it or change it to something like `krane-run-template`?

A few other complications come to mind now that I come back to thinking about this:
- We wanted to have `-f` for consistency, but does it really make sense? Maybe we didn't really think it through enough. If you have the template locally, why would you write a pod template and expand it with our util instead just writing a deployable pod spec yourself? 
- If we have both flags, are they mutually exclusive? Both repeatable? This could get confusing, especially if either has a default.
- Would it really be possible to make it repeatable? That for sure would get hairy in combination with any of the args/env-modifying flags.

My inclination (not a strong opinion at this time) is to improve the current flag's name, remove its default, and not provide `-f` for this command after all.

dturn:
I've added the 1.0 requirement label to this issue. Though 1.0 requirement need not be the `-f` as in the design doc (or current issue name), but answering the concerns raised here and likely keeping support for the functionally that `--template` provides.

jessie-JNing:
>Also its default value is kinda random (`task-runner-template`); should we remove it or change it to something like `krane-run-template`?
 My inclination (not a strong opinion at this time) is to improve the current flag's name, remove its default, and not provide `-f` for this command after all.

From the discussion above, we decide: 
- do NOT provide `-f` for `krane_run` command. 
- rename default template from `task-runner-template` to some better name like `krane-run-template`, or remove default template value. @KnVerey could you share why should remove it? I think it's a non-nullable arg.

KnVerey:
[...]

PR #579, opened 2019-10-02
Make template required in krane-run

jessie-JNing:
close: https://github.com/Shopify/kubernetes-deploy/issues/568

Make `--template` to be a required argument in krane-run.

**How is this accomplished?**
- change `template` to be required in run cli
-  add `with_processed_task_template` check in exe command, raise `OptionsError` if missing.
- fix tests
- update readme

### DIFF
--- a/exe/kubernetes-run
+++ b/exe/kubernetes-run
@@ -25,6 +25,7 @@ end
 
 namespace = ARGV[0]
 context = ARGV[1]
+logger = KubernetesDeploy::FormattedLogger.build(namespace, context, verbose_prefix: false)
 
 begin
   KubernetesDeploy::OptionsHelper.with_processed_task_template(template) do |path|
--- a/exe/kubernetes-run
+++ b/exe/kubernetes-run
@@ -5,7 +5,7 @@ require 'kubernetes-deploy/runner_task'
 require 'kubernetes-deploy/options_helper'
 require 'optparse'
 
-template = nil
+template = "task-runner-template"
 entrypoint = nil
 env_vars = []
 verify_result = true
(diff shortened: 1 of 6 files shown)


jessie-JNing:
> Isn't the validation in `RunnerTaskConfigValidator` already covering this case? If that's true, we shouldn't need the change in `OptionsHelper`. Example:

I consider about these two validations: 
`RunnerTaskConfigValidator` validate the template given, if it's an empty string to invalid path. 
The `OptionsHelper` should check if `--template` arg is given or not, so update the code.  
@dirceu could you review  👀it?

lib/kubernetes-deploy/options_helper.rb
```diff
@@ -29,6 +29,14 @@ def with_processed_template_paths(template_paths, require_explicit_path: false)
         end
       end
 
+      def with_processed_task_template(template_path)
```
dirceu: `OptionsHelper.with_processed_template_paths` both manipulates **and** validates the value(s) we pass to it, so its block interface makes sense.

In this particular case we're just validating the value so maybe we could change the name to reflect that. What do you think?

Also, we maybe don't need the block interface, but I have no strong feelings about that.

exe/kubernetes-run
```diff
@@ -5,7 +5,7 @@ require 'kubernetes-deploy/runner_task'
 require 'kubernetes-deploy/options_helper'
 require 'optparse'
 
-template = "task-runner-template"
+template = nil
```
dturn: The `exe/kubernetes-*` tasks need to remain backwards compatible.

lib/kubernetes-deploy/options_helper.rb
jessie-JNing: Thanks for the comment @dirceu , I remove the check `OptionsHelper.with_processed_template_paths` as it should be backwards compatible: https://github.com/Shopify/kubernetes-deploy/pull/579#discussion_r332127566

exe/kubernetes-run
jessie-JNing: Updated the PR.  `exe/kubernetes-run` remains as it is. The `--template` is changed to be required in CLI.

test/exe/run_test.rb
```diff
@@ -73,9 +82,10 @@ def set_krane_run_expectations(new_args: {}, run_args: {})
   end
 
   def krane_run!(flags: '')
+    cmd_flags = "#{flags} --template #{TASK_TEMPLATE}".split
```
dturn: Unconditionally adding `--template #{TASK_TEMPLATE}` seems wrong. I liked the pattern from the deploy test (` flags += ' -f /tmp' unless flags.include?('-f')`)

test/exe/run_test.rb
```diff
@@ -42,22 +44,29 @@ def test_run_parses_env_vars
     krane_run!(flags: '--env-vars SOMETHING=8000,FOO=bar')
   end
 
+  def test_run_failure_with_no_template_given
```
dturn: I don't think we need to test that thor 'require' works as expected.

test/exe/run_test.rb
jessie-JNing: Thanks Danny, that pattern is really a more correct way. Will use it

Issue #565, opened 2019-09-24
Improve toleration of daemonsets during autoscaling operations


dalehamel:
# Feature request

- [X] If the maintainers agree with the feature as described here, I intend to submit a Pull Request myself.<sup>1</sup>

**Proposal:** 

The current behavior of kubernetes-deploy is flakey when deploying daemonsets to clusters which autoscale. As nodes may be added during the deploy operation, they have little hope of succeeding within the typical timeout window.

The workaround proposed is that we consider only the set of nodes in the target cluster at the beginning of the deploy operation when we make the determining of success, failure, or timeout.

This way, we may filter the daemonset pods which we check against to only those whose node is within this initial list.

My theory is that this will allow for kubernetes-deploy to have a fair chance of succeeding, as the "goal post isn't moving" midway through. If there are flakey nodes, they should still cause a deploy failure on a subsequent deploy.

KnVerey:
Makes sense to me 👍

dturn:
I'm 👍 on this improvement. Though I'm wondering if there is a corresponding problem when nodes are being removed when autoscaling down that needs to be considered? Its fine to submit a PR for just what you wrote and we can consider the downscaling case at a later time.

dalehamel:
Yeah i agree, i'm not actually totally sure downscaling is actually an issue, but if it is we can probably just correspondingly ignore nodes that are explicitly not schedulable

dwradcliffe:
Seems marginally better so I 👍 but probably not enough to make a big difference IMO.

dalehamel:
I think also when scanning the list of nodes, any nodes that are `NotReady` should not be considered for convergence - only ready nodes should be considered. A lot of nodes that stay in the NotReady state may either be booting up or being drained to be shut down.

KnVerey:
@dalehamel can we close this?

dalehamel:
yeah, might want to open a separate issue for another daemonset toleration annotation, but the scope of this one is sufficiently covered by #580 i think.

PR #580, opened 2019-10-04
initial attempt at increased tolerance for daemonsets

dalehamel:

Provide a solution to issue #565 

**How is this accomplished?**

    Support treating daemonsets as successful if new nodes added
    
    This allows for the daemonset to be treated as successful even if
    the daemonset is not 100% converged, so long as:
    
    - The desired and scheduled number of pods match
    - All desired pods for which there was an associated node at the
    start of the deploy process are in the Ready state
    
    This should allow for better toleration of autoscaling events, which
    can currently prevent convergence on daemonsets as new nodes may be
    added in the middle of the deploy.

**What could go wrong?**

If I am missing edge cases, it is possible that deploys could be prematurely and incorrectly marked as successful, which is not desirable.

### DIFF
--- a/lib/kubernetes-deploy/kubernetes_resource/daemon_set.rb
+++ b/lib/kubernetes-deploy/kubernetes_resource/daemon_set.rb
@@ -8,6 +8,7 @@ module KubernetesDeploy
     def sync(cache)
       super
       @pods = exists? ? find_pods(cache) : []
+      @nodes = find_nodes(cache) if @nodes.blank?
     end
 
     def status
@@ -17,9 +18,9 @@ module KubernetesDeploy
 
     def deploy_succeeded?
       return false unless exists?
-      rollout_data["desiredNumberScheduled"].to_i == rollout_data["updatedNumberScheduled"].to_i &&
-      rollout_data["desiredNumberScheduled"].to_i == rollout_data["numberReady"].to_i &&
-      current_generation == observed_generation
+      current_generation == observed_generation &&
+        rollout_data["desiredNumberScheduled"].to_i == rollout_data["updatedNumberScheduled"].to_i &&
+        relevant_pods_ready?
     end
 
     def deploy_failed?
@@ -37,6 +38,34 @@ module KubernetesDeploy
     end
 
     private
+
+    class Node
+      attr_reader :name
+
+      class << self
+        def kind
+          name.demodulize
+        end
+      end
+
+      def initialize(definition:)
+        @name = definition.dig("metadata", "name").to_s
+        @definition = definition
+      end
+    end
+
+    def relevant_pods_ready?
+      return true if rollout_data["desiredNumberScheduled"].to_i == rollout_data["numberReady"].to_i # all pods ready
+      relevant_node_names = @nodes.map(&:name)
+      considered_pods = @pods.select { |p| relevant_node_names.include?(p.node_name) }
+      @logger.debug("Considered #{considered_pods.size} pods out of #{@pods.size} for #{@nodes.size} nodes")
+      considered_pods.present? && considered_pods.all?(&:deploy_succeeded?)
+    end
+
+    def find_nodes(cache)
+      all_nodes = cache.get_all(Node.kind)
+      all_nodes.map { |node_data| Node.new(definition: node_data) }
+    end
 
     def rollout_data
       return { "currentNumberScheduled" => 0 } unless exists?
--- a/lib/kubernetes-deploy/kubernetes_resource/pod.rb
+++ b/lib/kubernetes-deploy/kubernetes_resource/pod.rb
@@ -101,6 +101,10 @@ module KubernetesDeploy
       exists? && !@stream_logs # don't print them a second time
     end
 
+    def node_name
+      @instance_data.dig('spec', 'nodeName')
+    end
+
     private
 
     def failed_schedule_reason
--- a/test/fixtures/for_unit_tests/daemon_set.yml
+++ b/test/fixtures/for_unit_tests/daemon_set.yml
@@ -3,15 +3,17 @@ kind: DaemonSet
 metadata:
   name: ds-app
   generation: 2
+  uid: 'c31a9b4e-e6dd-11e9-8f47-e6322f98393a'
 spec:
   selector:
     matchLabels:
       app: ds-app
       name: ds-app
+  templateGeneration: 2
   template:
     metadata:
       labels:
-        app: hello-cloud
+        app: ds-app
         name: ds-app
     spec:
       containers:
--- a/test/fixtures/for_unit_tests/daemon_set_pod.yml
+++ b/test/fixtures/for_unit_tests/daemon_set_pod.yml
@@ -1,0 +1,105 @@
+apiVersion: v1
+kind: Pod
+metadata:
+  creationTimestamp: "2019-10-07T16:05:37Z"
+  generateName: ds-app-
+  labels:
+    app: ds-app
+    controller-revision-hash: "2703261291"
+    name: ds-app
+    pod-template-generation: "2"
+  name: ds-app-hgjhh
+  namespace: default
+  ownerReferences:
+  - apiVersion: apps/v1
+    blockOwnerDeletion: true
+    controller: true
+    kind: DaemonSet
+    name: ds-app
+    uid: c31a9b4e-e6dd-11e9-8f47-e6322f98393a
+  resourceVersion: "31010"
+  selfLink: /api/v1/namespaces/default/pods/ds-app-hgjhh
+  uid: 4cf14557-e91c-11e9-8f47-e6322f98393a
+spec:
+  containers:
+  - command:
+    - tail
+    - -f
+    - /dev/null
+    image: busybox
+    imagePullPolicy: IfNotPresent
+    name: app
+    ports:
+    - containerPort: 80
+      protocol: TCP
+    resources: {}
+    terminationMessagePath: /dev/termination-log
+    terminationMessagePolicy: File
+    volumeMounts:
+    - mountPath: /var/run/secrets/kubernetes.io/serviceaccount
+      name: default-token-bwg9f
+      readOnly: true
+  dnsPolicy: ClusterFirst
+  nodeName: minikube
+  priority: 0
+  restartPolicy: Always
+  schedulerName: default-scheduler
+  securityContext: {}
+  serviceAccount: default
+  serviceAccountName: default
+  terminationGracePeriodSeconds: 30
+  tolerations:
+  - effect: NoExecute
+    key: node.kubernetes.io/not-ready
+    operator: Exists
+  - effect: NoExecute
+    key: node.kubernetes.io/unreachable
+    operator: Exists
+  - effect: NoSchedule
+    key: node.kubernetes.io/disk-pressure
+    operator: Exists
+  - effect: NoSchedule
+    key: node.kubernetes.io/memory-pressure
+    operator: Exists
+  - effect: NoSchedule
+    key: node.kubernetes.io/unschedulable
+    operator: Exists
+  volumes:
+  - name: default-token-bwg9f
+    secret:
+      defaultMode: 420
+      secretName: default-token-bwg9f
+status:
+  conditions:
+  - lastProbeTime: null
+    lastTransitionTime: "2019-10-07T16:05:37Z"
+    status: "True"
+    type: Initialized
+  - lastProbeTime: null
+    lastTransitionTime: "2019-10-07T16:05:38Z"
+    status: "True"
+    type: Ready
+  - lastProbeTime: null
+    lastTransitionTime: null
+    status: "True"
+    type: ContainersReady
+  - lastProbeTime: null
+    lastTransitionTime: "2019-10-07T16:05:37Z"
+    status: "True"
+    type: PodScheduled
+  containerStatuses:
+  - containerID: docker://949e6b37ad1e85dfeca958bb5a54c459305ef3d87e12d03e1ba90e121701b572
+    image: busybox:latest
+    imageID: docker-pullable://busybox@sha256:fe301db49df08c384001ed752dff6d52b4305a73a7f608f21528048e8a08b51e
+    lastState: {}
+    name: app
+    ready: true
+    restartCount: 0
+    state:
+      running:
+        startedAt: "2019-10-07T16:05:38Z"
+  hostIP: 192.168.64.3
+  phase: Running
+  podIP: 172.17.0.4
+  qosClass: BestEffort
+  startTime: "2019-10-07T16:05:37Z"
--- a/test/fixtures/for_unit_tests/node.yml
+++ b/test/fixtures/for_unit_tests/node.yml
@@ -1,0 +1,136 @@
+apiVersion: v1
+kind: Node
+metadata:
+  annotations:
+    kubeadm.alpha.kubernetes.io/cri-socket: /var/run/dockershim.sock
+    node.alpha.kubernetes.io/ttl: "0"
+    volumes.kubernetes.io/controller-managed-attach-detach: "true"
+  creationTimestamp: "2019-10-04T19:28:10Z"
+  labels:
+    beta.kubernetes.io/arch: amd64
+    beta.kubernetes.io/os: linux
+    kubernetes.io/hostname: minikube
+    node-role.kubernetes.io/master: ""
+  name: minikube
+  resourceVersion: "29256"
+  selfLink: /api/v1/nodes/minikube
+  uid: 19812251-e6dd-11e9-8f47-e6322f98393a
+spec: {}
+status:
+  addresses:
+  - address: 192.168.64.3
+    type: InternalIP
+  - address: minikube
+    type: Hostname
+  allocatable:
+    cpu: "2"
+    ephemeral-storage: "15625027559"
+    hugepages-2Mi: "0"
+    memory: 1887148Ki
+    pods: "110"
+  capacity:
+    cpu: "2"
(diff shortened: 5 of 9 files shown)


dalehamel:
There is more subtlety here than I realized - we can't rely on the assumption that every daemonset is scheduled on every node, so we need to consider this based on the pods instead rather than based on the nodes.

We'll take the list of nodes at the start of the deployment as the only ones under consideration. If we have pods that aren't ready, which are scheduled on nodes that aren't in this initial list, we can safely ignore them. 

The set difference of the nodes_for_pods - original_nodes_list should be a strict superset. There should be no nodes in the original list for which there is not a ready pod that is scheduled on that node.

lib/kubernetes-deploy/kubernetes_resource/daemon_set.rb
```diff
@@ -17,8 +18,15 @@ def status
 
     def deploy_succeeded?
       return false unless exists?
-      rollout_data["desiredNumberScheduled"].to_i == rollout_data["updatedNumberScheduled"].to_i &&
-      rollout_data["desiredNumberScheduled"].to_i == rollout_data["numberReady"].to_i &&
+
+      puts pods.inspect
+      puts "TOTAL_PODS #{@pods.size}"
+      considered_pods = @pods.select { |p| @nodes.map(&:name).include?(p.node_name) }
```
dalehamel: now that I think about this more, this check should probably be gated by `rollout_data["desiredNumberScheduled"].to_i == rollout_data["updatedNumberScheduled"].to_i`

.gitignore
```diff
@@ -18,3 +18,5 @@ dev/flamegraph.svg
 dev/profile
 dev/flamegraph.pl
 .local-context
+*.swp
```
dalehamel: sorry, vim stuff :nerd_face:

lib/kubernetes-deploy/kubernetes_resource/daemon_set.rb
```diff
@@ -17,8 +18,11 @@ def status
 
     def deploy_succeeded?
       return false unless exists?
-      rollout_data["desiredNumberScheduled"].to_i == rollout_data["updatedNumberScheduled"].to_i &&
-      rollout_data["desiredNumberScheduled"].to_i == rollout_data["numberReady"].to_i &&
+
+      considered_pods = @pods.select { |p| @nodes.map(&:name).include?(p.node_name) }
+      (rollout_data["desiredNumberScheduled"].to_i == rollout_data["updatedNumberScheduled"].to_i &&
```
dalehamel: ok so i am well aware this multi-line complicated conditional isn't the prettiest thing in the world, but hey rubocop allowed it!

Open to ideas on how to make this clearer if the boolean logic is getting to sophisticated.

lib/kubernetes-deploy/kubernetes_resource/daemon_set.rb
```diff
@@ -38,6 +42,31 @@ def print_debug_logs?
...
+      end
+
+      def initialize(definition:)
+        @name = definition.dig("metadata", "name").to_s
+        @definition = definition
+      end
+    end
+
+    def find_nodes(cache)
+      # FIXME: should this also exclude tainted nodes?
```
dalehamel: NIT - remove this before merging, i don't think we care about tainted nodes

lib/kubernetes-deploy/kubernetes_resource/pod.rb
```diff
@@ -101,6 +101,16 @@ def print_debug_logs?
       exists? && !@stream_logs # don't print them a second time
     end
 
+    def ready?
```
dalehamel: I had to pull this out of `private`, so I guess i'm the first one to have a valid reason to call it. I think this is fine as a public method.

test/fixtures/for_unit_tests/daemon_set.yml
```diff
@@ -11,7 +11,7 @@ spec:
   template:
     metadata:
       labels:
-        app: hello-cloud
```
dalehamel: I think this was an existing bug

test/fixtures/for_unit_tests/node.yml
```diff
@@ -0,0 +1,136 @@
+apiVersion: v1
```
dalehamel: straight out of minikube

test/fixtures/for_unit_tests/daemon_set_pod.yml
```diff
@@ -0,0 +1,105 @@
+apiVersion: v1
```
dalehamel: Result from minikube when I applied the DS yaml.

test/unit/kubernetes-deploy/kubernetes_resource/daemon_set_test.rb
```diff
@@ -4,9 +4,11 @@
 class DaemonSetTest < KubernetesDeploy::TestCase
   include ResourceCacheTestHelper
 
+  OWNER_UID = 'c31a9b4e-e6dd-11e9-8f47-e6322f98393a'
```
KnVerey: Could we instead make the UID match in the fixtures themselves?

test/unit/kubernetes-deploy/kubernetes_resource/daemon_set_test.rb
```diff
@@ -38,22 +40,78 @@ def test_deploy_passes_when_updated_available_does_match
...
+    else
+      [base_node_manifest].flatten
+    end
   end
 
-  def build_synced_ds(template:)
-    ds = KubernetesDeploy::DaemonSet.new(namespace: "test", context: "nope", logger: logger, definition: template)
-    stub_kind_get("DaemonSet", items: [template])
-    stub_kind_get("Pod", items: [])
+  def build_synced_ds(ds_template:, pod_templates: nil, node_templates: nil)
```
KnVerey: Suggestion: since the keywords are plural, make the default values `[]` and then remove the `|| []` below

test/unit/kubernetes-deploy/kubernetes_resource/daemon_set_test.rb
```diff
@@ -38,22 +40,78 @@ def test_deploy_passes_when_updated_available_does_match
...
+
+  def test_deploy_passes_when_ready_pods_for_multiple_nodes
+    status = {
+      "desiredNumberScheduled": 3,
+      "updatedNumberScheduled": 3,
+      "numberReady": 2,
+    }
+    ds_template = build_ds_template(filename: 'daemon_set.yml', uid: OWNER_UID, generation: 2, status: status)
+    pod_template = build_pod_template(filename: 'daemon_set_pod.yml')
+    node_template = build_node_template(filename: 'node.yml')
```
KnVerey: 3 desired, 2 ready and 1 node seems like a weird state that would never be observed on a single sync cycle--you'd for sure see three nodes given that status. If we go with the node list caching approach, this test will need to multiple sync cycles. Otherwise, it will need to use multiple nodes with manipulated creation timestamps.

test/unit/kubernetes-deploy/kubernetes_resource/daemon_set_test.rb
```diff
@@ -38,22 +40,78 @@ def test_deploy_passes_when_updated_available_does_match
...
 
+  def test_deploy_fails_when_not_all_pods_updated
+    status = {
+      "desiredNumberScheduled": 3,
+      "updatedNumberScheduled": 2,
+      "numberReady": 2,
+    }
+    ds_template = build_ds_template(filename: 'daemon_set.yml', uid: OWNER_UID, generation: 2, status: status)
+    pod_template = build_pod_template(filename: 'daemon_set_pod.yml')
+    node_template = build_node_template(filename: 'node.yml')
```
KnVerey: Same comment about this test.

lib/kubernetes-deploy/kubernetes_resource/daemon_set.rb
KnVerey: Agreed, we don't care about tainted nodes. If the pod was scheduled, we should assume it was supposed to be.

lib/kubernetes-deploy/kubernetes_resource/daemon_set.rb
```diff
@@ -8,6 +8,7 @@ class DaemonSet < PodSetBase
     def sync(cache)
       super
       @pods = exists? ? find_pods(cache) : []
+      @nodes = find_nodes(cache)
```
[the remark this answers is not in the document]
dalehamel: going with the first approach + `presence` as you suggested IRL

test/unit/kubernetes-deploy/kubernetes_resource/daemon_set_test.rb
dalehamel: I agree these states are weird. the intent was to simulate where a node has been added by autoscaling. I'm going to adjust the numbers accordingly.

> If we go with the node list caching approach, this test will need to multiple sync cycles

I can add a separate test that handles the multiple-sync case

dalehamel: I think that `test_deploy_passes_when_ready_pods_but_node_added` should cover this, but i wasn't sure how I could use the same daemonset instance, though I'm not totally sure it matters.

Can you think of a way to reuse the daemonset instance and update the status transparently? Ultimately to test logic around cache I think that's needed.

dalehamel:
thanks for the review @KnVerey I think i addressed everything, can you do another pass?

test/unit/kubernetes-deploy/kubernetes_resource/daemon_set_test.rb
```diff
@@ -38,22 +38,105 @@ def test_deploy_passes_when_updated_available_does_match
...
+
+    node_added_status = {
+      "desiredNumberScheduled": 3,
+      "updatedNumberScheduled": 3,
+      "numberReady": 1,
+    }
+    ds_template = build_ds_template(filename: 'daemon_set.yml', generation: 2, status: node_added_status)
+    pod_template = build_pod_template(filename: 'daemon_set_pod.yml')
+    node_template = build_node_template(filename: 'nodes.yml')
+    ds = build_synced_ds(ds_template: ds_template, pod_templates: pod_template, node_templates: node_template)
```
KnVerey: > I think that test_deploy_passes_when_ready_pods_but_node_added should cover this, but i wasn't sure how I could use the same daemonset instance, though I'm not totally sure it matters.

> Can you think of a way to reuse the daemonset instance and update the status transparently? Ultimately to test logic around cache I think that's needed.

This is the line that needs to change--you need to manually do this method's job on your old instance:

```ruby
    stub_kind_get("DaemonSet", items: [ds_template]) # the new one from L75
    stub_kind_get("Pod", items: [ready_pod_template]) # the new one from L76
    stub_kind_get("Node", items: three_node_templates) # maybe you can cheat and return the same one three times instead of having a separate fixture for this one too?
    ds.sync(build_resource_cache) # this puts the new data into the existing instance
```

Also, this is your regression test, so when you're done with it, make sure it fails if you revert your business logic change.

test/unit/kubernetes-deploy/kubernetes_resource/daemon_set_test.rb
```diff
@@ -38,22 +38,105 @@ def test_deploy_passes_when_updated_available_does_match
...
+  end
+
+  def test_deploy_fails_when_not_all_pods_ready
+    status = {
+      "desiredNumberScheduled": 3,
+      "updatedNumberScheduled": 3,
+      "numberReady": 2,
+    }
+    ds_template = build_ds_template(filename: 'daemon_set.yml', generation: 2, status: status)
+    pod_template = build_pod_template(filename: 'daemon_set_pods.yml')
```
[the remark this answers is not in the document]
dalehamel: yeah i did miss them originally, added them back now. I'll fix the verbiage to be more consistent with expectations.

dalehamel:
@KnVerey when you have a chance I would appreciate another look :eyes:

test/unit/kubernetes-deploy/kubernetes_resource/daemon_set_test.rb
```diff
@@ -38,22 +38,99 @@ def test_deploy_passes_when_updated_available_does_match
...
+    ds = build_synced_ds(ds_template: ds_template, pod_templates: pod_templates, node_templates: node_templates)
+    refute_predicate(ds, :deploy_succeeded?)
+
+    node_added_status = {
+      "desiredNumberScheduled": 3,
+      "updatedNumberScheduled": 3,
+      "numberReady": 1,
+    }
+    ds_template = build_ds_template(filename: 'daemon_set.yml', status: node_added_status)
+    pod_templates = load_fixtures(filenames: ['daemon_set_pod.yml'])
```
KnVerey: For realism, shouldn't this return three pods, two of which are not ready? I think that right now, this test would still pass if the check were `@pods.all?(&:ready)`, unfiltered by node.

lib/kubernetes-deploy/kubernetes_resource/daemon_set.rb
```diff
@@ -38,6 +39,37 @@ def print_debug_logs?
...
+      def initialize(definition:)
+        @name = definition.dig("metadata", "name").to_s
+        @definition = definition
+      end
+    end
+
+    def relevant_pods_ready?
+      return true if rollout_data["desiredNumberScheduled"].to_i == rollout_data["numberReady"].to_i # all pods ready
+      relevant_node_names = @nodes.map(&:name)
+      considered_pods = @pods.select { |p| relevant_node_names.include?(p.node_name) }
```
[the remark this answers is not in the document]
dalehamel: i agree

test/unit/kubernetes-deploy/kubernetes_resource/daemon_set_test.rb
dalehamel: I changed numberReady to 2, and made the fixture have 2 ready pods.

dalehamel:
FYI @douglas as this might conflict with renaming in #585

douglas:
Hey Dale! I think the other PR just needs reviews to go - if this is not urgent, can we ship it first? Then I volunteer to rebase this one.

test/unit/kubernetes-deploy/kubernetes_resource/daemon_set_test.rb
```diff
@@ -38,22 +38,99 @@ def test_deploy_passes_when_updated_available_does_match
...
+  end
+
   private
 
-  def build_ds_template(status: {})
-    base_ds_maifest = YAML.load_stream(File.read(File.join(fixture_path('for_unit_tests'), 'daemon_set.yml'))).first
-    base_ds_maifest.deep_merge("status" => status)
+  def build_ds_template(filename:, status: {})
+    base_ds_manifest = YAML.load_stream(File.read(File.join(fixture_path('for_unit_tests'), filename))).first
+    base_ds_manifest.deep_merge!("status" => status)
```
dturn: why the change to `deep_merge!`

test/unit/kubernetes-deploy/kubernetes_resource/daemon_set_test.rb
```diff
@@ -38,22 +38,99 @@ def test_deploy_passes_when_updated_available_does_match
...
+    ds_template = build_ds_template(filename: 'daemon_set.yml', status: node_added_status)
+    pod_templates = load_fixtures(filenames: ['daemon_set_pods.yml'])
+
+    stub_kind_get("DaemonSet", items: [ds_template])
+    stub_kind_get("Pod", items: pod_templates)
+    ds.sync(build_resource_cache)
+    assert_predicate(ds, :deploy_succeeded?)
+  end
+
+  def test_deploy_fails_when_not_ell_pods_updated
```
dturn: typo: `test_deploy_fails_when_not_ell_pods_updated` -> `test_deploy_fails_when_not_all_pods_updated`

test/unit/kubernetes-deploy/kubernetes_resource/daemon_set_test.rb
```diff
@@ -38,22 +38,99 @@ def test_deploy_passes_when_updated_available_does_match
...
+      "updatedNumberScheduled": 1,
+    }
+    ds_template = build_ds_template(filename: 'daemon_set.yml', status: status)
+    pod_templates = load_fixtures(filenames: ['daemon_set_pod.yml'])
+    node_templates = load_fixtures(filenames: ['node.yml'])
+    ds = build_synced_ds(ds_template: ds_template, pod_templates: pod_templates, node_templates: node_templates)
+    assert_predicate(ds, :deploy_succeeded?)
+  end
+
+  def test_deploy_passes_when_ready_pods_but_node_added
```
dturn: `node_added` or `nodes_added` ?

test/unit/kubernetes-deploy/kubernetes_resource/daemon_set_test.rb
dalehamel: artifact from debugging probably, i'll fix that.

lib/kubernetes-deploy/kubernetes_resource/daemon_set.rb
```diff
@@ -38,6 +39,37 @@ def print_debug_logs?
...
+      considered_pods.present? && considered_pods.all?(&:deploy_succeeded?)
+    end
+
+    def find_nodes(cache)
+      all_nodes = cache.get_all(Node.kind)
+
+      all_nodes.each_with_object([]) do |node_data, nodes|
```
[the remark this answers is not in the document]
dalehamel: yeah that's cleaner.

dalehamel:
i double checked with @douglas and he said this is fine to merge, merging.

Issue #105, opened 2017-06-02
Better documentation


KnVerey:
- [x] Improve readme (better explanation of core functionality and options, screenshots)
- [ ] Convert all existing comment-based docs to rdoc or [yard](http://www.rubydoc.info/gems/yard/file/docs/Tags.md#List_of_Available_Tags)
- [x] Add comment-based docs for the key methods you need to use the tasks from Ruby instead of from the CLI (i.e. the `run` and `run!` methods of each `Task` class).

kirs:
We use yard in Shopify core. Maybe it would make sense to use it here too.

KnVerey:
Ideas for more readme enhancements:

- [ ] Description of a few workflows like:
  - Getting started - using stuff from an 'examples' directory
  - Walk through the steps of a deploy
- [ ] GIF screen recording of terminal running a deploy locally - put this right near the top of the readme

KnVerey:
- [ ] Readme should reference a directory of examples, even if it is just our test fixtures

dirceu:
After https://github.com/Shopify/kubernetes-deploy/pull/581 and https://github.com/Shopify/kubernetes-deploy/pull/582, do we need anything else?

I'm not sure about this:

> Convert all existing comment-based docs to rdoc or yard

There aren't many comments, and most are specific to a particular condition inside a method, etc.

KnVerey:
I'm inclined to agree now that only thing we need to document with yard are the public interfaces.

PR #581, opened 2019-10-04
Add docstrings to the public interface of all tasks

dirceu:

Solve part of https://github.com/Shopify/kubernetes-deploy/issues/105 by adding docstrings to the public interface of all `Task` implementations.

**How is this accomplished?**

- Added `yard` as a dependency and added a `dev doc` command to generate those docs
- Added docstrings to public methods of the four `*Task` classes


lib/kubernetes-deploy/deploy_task.rb
```diff
@@ -104,6 +105,21 @@ def server_version
       kubectl.server_version
     end
 
+    # Initializes the task
```
RyanBrushett:
```suggestion
    # Initializes the deploy task
```

lib/kubernetes-deploy/render_task.rb
```diff
@@ -6,7 +6,15 @@
 require 'kubernetes-deploy/template_sets'
 
 module KubernetesDeploy
+  # Render templates
   class RenderTask
+    # Initializes the task
```
RyanBrushett:
```suggestion
    # Initializes the render task
```

lib/kubernetes-deploy/restart_task.rb
```diff
@@ -21,6 +22,12 @@ def initialize(deployment_name, response)
     HTTP_OK_RANGE = 200..299
     ANNOTATION = "shipit.shopify.io/restart"
 
+    # Initializes the task
```
RyanBrushett:
```suggestion
    # Initializes the restart task
```

lib/kubernetes-deploy/runner_task.rb
```diff
@@ -11,11 +11,18 @@
 require 'kubernetes-deploy/runner_task_config_validator'
 
 module KubernetesDeploy
+  # Run a pod that exits upon completing a task
   class RunnerTask
     class TaskTemplateMissingError < TaskConfigurationError; end
 
     attr_reader :pod_name
 
+    # Initializes the task
```
RyanBrushett:
```suggestion
    # Initializes the runner task
```

lib/kubernetes-deploy/deploy_task.rb
```diff
@@ -104,6 +105,21 @@ def server_version
...
+    #
+    # @param namespace [String] Kubernetes namespace
+    # @param context [String] Kubernetes context / cluster
```
dturn:
```suggestion
    # @param context [String] Kubernetes context
```

dturn: Context = cluster + user we use them interchangeably since we always auth the same way, but they technically aren't the same.

PR #583, opened 2019-10-07
Rescue `TaskConfigurationError` on `*Task.run()`

dirceu:

Make sure that `*Task.run()` won't raise in case of a `TaskConfigurationError`, returning `false` instead. This means that we'll keep the previous behaviour, essentially.

**How is this accomplished?**

I'm just rescuing the new exception (that replaced `FatalDeploymentError` in some cases).

### DIFF
--- a/lib/kubernetes-deploy/restart_task.rb
+++ b/lib/kubernetes-deploy/restart_task.rb
@@ -42,7 +42,7 @@ module KubernetesDeploy
     def run(*args)
       perform!(*args)
       true
-    rescue FatalDeploymentError
+    rescue FatalDeploymentError, TaskConfigurationError
       false
     end
     alias_method :perform, :run
--- a/lib/kubernetes-deploy/deploy_task.rb
+++ b/lib/kubernetes-deploy/deploy_task.rb
@@ -148,7 +148,7 @@ module KubernetesDeploy
     def run(*args)
       run!(*args)
       true
-    rescue FatalDeploymentError
+    rescue FatalDeploymentError, TaskConfigurationError
       false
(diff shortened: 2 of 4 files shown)


lib/kubernetes-deploy/restart_task.rb
```diff
@@ -42,7 +42,7 @@ def initialize(context:, namespace:, logger: nil, max_watch_seconds: nil)
     def run(*args)
       perform!(*args)
       true
-    rescue FatalDeploymentError
+    rescue FatalDeploymentError, TaskConfigurationError
```
jessie-JNing: Maybe we should also resue `DeploymentTimeoutError` since it raised https://github.com/Shopify/kubernetes-deploy/pull/583/files#diff-5068ffe6b6ef4e5064ed5138b4dae265L78 ?

KnVerey: Technically you don't need to if you're going to do the same thing with the result, since `class DeploymentTimeoutError < FatalDeploymentError; end`

KnVerey: Actually, so does `class TaskConfigurationError < FatalDeploymentError; end`. @dirceu 
 are you planning on changing that?

dirceu: Ah, good catch! I didn't notice that. No, I don't plan on changing that and I think that in that case this PR won't be needed.

dirceu:
Not needed, see https://github.com/Shopify/kubernetes-deploy/pull/583#discussion_r332428310.

Issue #572, opened 2019-09-26
[Tech Debt]


dturn:
As part of [Consistent preflight validations for all tasks](https://github.com/Shopify/kubernetes-deploy/issues/344) we created a `TaskConfig` class. The high level idea of this class was that it could be passed around instead of having to pass around a combination of: namesapce, context, logger. #344 introduced to the Validation classes but stopped at that. We should push the TaskConfig into more classes to simply the code.  We may also want to consider the extent to which additional information should be stored in task specific configs (one idea what tracking what resource types  have `server-dry-run` available).

dturn:
Considering actioning https://github.com/Shopify/kubernetes-deploy/pull/567#discussion_r329718864 when PRing the change that passes TaskConfig  to KubernetesResources

PR #584, opened 2019-10-08
Task config kubectl

dturn:

Make it easy to have a global-deploy command.  #574 has shown that some objects validate too much, e.g. Kubectl validates its passes a namespace, but it shouldn't if it's only going to be used for global commands. 

By pushing TaskConfig in more places it should become easier to feel good about limiting validations in non-task classes.

This doesn't finish #572, we still need to pass it to the Resource objects. But this already hits 13+ files. 

**How is this accomplished?**
Pass TaskConfig instances around. Remove some validations in Kubectl. 

**What could go wrong?**
Test coverage is lacking ...


dturn:
>Codecov seems happy enough. Anything you could do to make yourself more comfortable here?
Did you run any of the commands locally to see how it worked out?

Taking a look at codecov it does look like everything is covered. I'm not really worried about the tasks breaking.

Issue #528, opened 2019-08-14
Krane 1.0 release


dturn:
## Problem
We're transitioning from kubernetes-deploy to krane. We need to update the gemspec, version to 1.0, and deploy to ruby gems under `krane`. This PR should be completed immediately after #527.

## Possible solution
Update the gemspec, shipit.yml, and Version file. https://github.com/Shopify/kubernetes-deploy/pull/631

## Acceptance criteria

- [x] Audit our public interfaces to ensure that arguments to each public function are sensible and consistent (same names for same thing, required vs. optional, on initialize vs on run). (https://github.com/Shopify/kubernetes-deploy/pull/619)
- [x] Delete everything "kubernetes-deploy" (https://github.com/Shopify/kubernetes-deploy/pull/622)
- [x] StatsD Rename (https://github.com/Shopify/kubernetes-deploy/pull/623)
- [x] Ensure we have documented what our public interfaces are (and possibly what they are not) (https://github.com/Shopify/kubernetes-deploy/pull/634)
- [x] Ensure that all 1.0 tasks are complete or intentionally deferred.
- [ ] Datadog dashboards are updated for `krane` and `kubernetes-deploy`
- [x] Migration guide is linked near the top of the readme. (https://github.com/Shopify/kubernetes-deploy/pull/634)
- [x] `gem 'krane', '~> 1.0.0'` works. 

## Additional context
https://github.com/Shopify/kubernetes-deploy/issues/256#issuecomment-475580770

dturn:
As we transition to Krane 1.0. We'll create 1 or more kubernetes-deploy v1.0.0-pre.X releases. The final pre-release version will be a transition version that maintains backwards compatibility with kubernetes-deploy as well as have all of the Krane commands. 

Krane 1.0.0 will not include any kubernetes-deploy modules or commands. We'll be publishing a transition guide.

dturn:
Krane 1.0.0 has been released.

PR #585, opened 2019-10-08
Rename kubernetes-deploy to krane

douglas:
## What are you trying to accomplish with this PR?

Rename `kubernetes-deploy` to `krane`.

Closes Shopify/kubernetes-deploy#527.

## TODO

- [x] Fix conflicts with master
- [x] Change occurrences of `kubernetes-deploy` in the documentation
- [x] Do local tests to ensure it is working properly
- [x] Add Update DD dashboards task to https://github.com/Shopify/kubernetes-deploy/issues/528

- [x] Squash commits

## Can be done in other moment/PRs

- Create a mail box called `krane@shopify.com` and rename all occurrences of `kubernetes-deploy@shopify.com` to use the created email.
- Verify TaskConfigValidatorTest#test_valid_configuration /usr/src/app/test/integration/task_config_validator_test.rb:40: warning: already initialized constant Krane::MIN_KUBE_VERSION
- Change README (and its screenshots) to reflect the change to `Krane`

## How is this accomplished?

- Move files in `lib/kubernetes_deploy` dir to `lib/krane`
- Merge lib/kubernetes-deploy/deploy_task.rb in lib/krane/deploy_task.rb
- Rename occurrences of `module KubernetesDeploy` to `module Krane`
- Change occurrences of `require kubernetes_deploy/` to `require krane/`
- Rename occurrences of `KubernetesDeploy::` to `Krane::`
- Rename occurrences of `KubernetesDeploy.` to `Krane.`
- Rename remaining occurrences of `KubernetesDeploy` to `Krane`
- Use top-level Krane constant for all files under `lib/krane/cli`
- Move test/unit/kubernetes_deploy to `test/unit/krane`
- Rename kubernetes_deploy_test.rb to krane_deploy_test.rb
- Rename kubernetes_deploy_test.rb to krane_deploy_test.rb
- Rename occurences of kubernetes-deploy in code
- Add DeprecatedDeploy to allow global resources with a warning
- Keep compatibility with KubernetesDeploy API
- Still use KubernetesDeploy public api for exes, errors and integration tests

## What could go wrong?

Well, I'm worried that it could contain errors that are not caught by our test suites as it moves code from `KubernetesDeploy` escope to `Krane`, which already exists.

The tests that we have now are passing, so I'm less stressed, but I would love reviews to ensure nothing breaks.

Also, is there a way to make this PR smaller or in small batches? What would be the best way go over this issue so reviews are more pleasant?


douglas:
For this change `Rename Krane::CLI::Krane to Krane::CLI::Commands`:

I had to do this because after renaming `KubernetesDeploy` to `Krane`, `lib/krane/cli/version_command.rb:9` is now `Krane::VERSION` instead of `KubernetesDeploy::VERSION`, but it fails with this error:

Screen Shot 2019-10-08 at 5 30 02 PM

It happens because in the context of any file below `lib/krane/cli/`, `Krane` refers to `Krane::CLI::Krane` instead of `Krane` module.

That is why I decided to rename `Krane::CLI::Krane` to `Krane::CLI::Commands`, I'm not sure this is the best solution, but if you have any suggestion, I would love to know ❤️

KnVerey:
You can explicitly tell Ruby you mean the top-level constant like this: `::Krane::VERSION`

douglas:
TIL, thanks a lot, Katrina! I will change the code to match your suggestion ❤️

lib/krane.rb
```diff
@@ -1,5 +1,10 @@
 # frozen_string_literal: true
-require 'kubernetes-deploy/common'
+require 'krane/common'
+
+require 'krane/deploy_task'
+require 'krane/render_task'
+require 'krane/restart_task'
+require 'krane/runner_task'
 
 module Krane
```
dturn: Is this still needed?

lib/krane.rb
```diff
@@ -1,5 +1,10 @@
 # frozen_string_literal: true
-require 'kubernetes-deploy/common'
+require 'krane/common'
```
dturn: Is this pulled in by any of the other requires?

lib/krane.rb
douglas: Seems so, without it the tests breaks.

douglas:
About the [Merge lib/kubernetes-deploy/deploy_task.rb in lib/krane/deploy_task.rb](https://github.com/Shopify/kubernetes-deploy/pull/585/commits/c241e4d84225c07cf06cf78a715f8c577f457534
) change:

After some reading, as Danny mentioned above, it is related to https://github.com/Shopify/kubernetes-deploy/pull/574, therefore, I rebased with `master` and removed this `raise "Use Krane::DeployGlobalTask to deploy global resources" if args[:allow_globals]`
in https://github.com/Shopify/kubernetes-deploy/blob/master/lib/krane/deploy_task.rb#L8, so the tests could pass.

What would be the best way to proceed with this?

jessie-JNing:
Great work on this massive change ~
I have two things not quite sure: 
- we are not going to rename the git repo, right?  So there's no need to change the name in [dev.yaml](https://github.com/Shopify/kubernetes-deploy/blob/master/dev.yml#L2)
- about the [kubernetes-deploy.gemspec](https://github.com/Shopify/kubernetes-deploy/blob/master/kubernetes-deploy.gemspec#L8), I saw a plan somewhere that we're going to do a pre-release, is that the reason we don't change the gem name here? Could anyone share more details about the plan?

dturn:
>we are not going to rename the git repo, right? So there's no need to change the name in dev.yaml

We will rename the repo but not in this issue. (Probably as part of the 1.0 release issue)

>about the kubernetes-deploy.gemspec, I saw a plan somewhere that we're going to do a pre-release, is that the reason we don't change the gem name here? Could anyone share more details about the plan?

The goal of not updating the gemspec in this issue is mostly to try and keep the scope reasonably sized.

jessie-JNing:
Thank for answering https://github.com/Shopify/kubernetes-deploy/pull/585#issuecomment-540695322 . That's pretty reasonable to me 🙂

dturn:
>  Verify TaskConfigValidatorTest#test_valid_configuration /usr/src/app/test/integration/task_config_validator_test.rb:40: warning: already initialized constant Krane::MIN_KUBE_VERSION

That warning isn't new. There might be a better way to write that test that doesn't involve redefining a constant, but I don't think you need to fix that in this PR.

CONTRIBUTING.md
```diff
@@ -143,7 +143,7 @@ Using another local cluster:
 
 To make StatsD log what it would have emitted, run a test with `STATSD_DEV=1`.
 
-To see the full-color output of a specific integration test, you can use `PRINT_LOGS=1`. For example: `PRINT_LOGS=1 bundle exec ruby -I test test/integration/kubernetes_deploy_test.rb -n/test_name/`.
+To see the full-color output of a specific integration test, you can use `PRINT_LOGS=1`. For example: `PRINT_LOGS=1 bundle exec ruby -I test test/integration/krane_test.rb -n/test_name/`.
```
dturn: This should probably be `krane_deploy_test.rb` `krane_test.rb` is something different.

CONTRIBUTING.md
```diff
@@ -163,6 +163,6 @@ Please make sure you run the tests locally before submitting your PR (see [Runni
 
 #### Employees: Triggering CI for a contributed PR
 
-Go to the [kubernetes-deploy pipeline](https://buildkite.com/shopify/kubernetes-deploy) and click "New Build". Use branch `external_contrib_ci` and the specific sha of the commit you want to build. Add `BUILDKITE_REFSPEC="refs/pull/${PR_NUM}/head"` in the Environment Variables section. Since CI is only visible to Shopify employees, you will need to provide any failing tests and output to the the contributor.
+Go to the [krane pipeline](https://buildkite.com/shopify/krane) and click "New Build". Use branch `external_contrib_ci` and the specific sha of the commit you want to build. Add `BUILDKITE_REFSPEC="refs/pull/${PR_NUM}/head"` in the Environment Variables section. Since CI is only visible to Shopify employees, you will need to provide any failing tests and output to the the contributor.
```
dturn: Are you committing to rename the pipeline?

ISSUE_TEMPLATE.md
```diff
@@ -8,7 +8,7 @@
 
 **Actual behavior:** [What actually happened]
 
-**Version(s) affected:** [run `kubernetes-deploy --version`]
+**Version(s) affected:** [run `krane --version`]
```
dturn: `krane version` its now a command not a flag

README.md
```diff
@@ -1,14 +1,14 @@
-# krane
+# kubernetes-deploy
```
dturn: Wrong direction?

README.md
```diff
@@ -1,14 +1,14 @@
...
 
-As this project approaches the v1.0 milestone, we're excited to announce that `kubernetes-deploy` will be [officially renamed as `krane`](https://github.com/Shopify/kubernetes-deploy/issues/30#issuecomment-468750341). Follow the [1.0 requirement label](https://github.com/Shopify/kubernetes-deploy/issues?q=is%3Aissue+is%3Aopen+label%3A%22%3Arocket%3A+1.0+requirement%22) to keep up with the progress.
+As this project approaches the v1.0 milestone, we're excited to announce that `krane` will be [officially renamed as `krane`](https://github.com/Shopify/kubernetes-deploy/issues/30#issuecomment-468750341). Follow the [1.0 requirement label](https://github.com/Shopify/kubernetes-deploy/issues?q=is%3Aissue+is%3Aopen+label%3A%22%3Arocket%3A+1.0+requirement%22) to keep up with the progress.
```
dturn: needs to be re-worded

README.md
```diff
@@ -147,22 +147,22 @@ You can add additional variables using the `--bindings=BINDINGS` option which ca
 
 ```
 # Comma separated string. Exposes, 'color' and 'size'
-$ kubernetes-deploy my-app cluster1 --bindings=color=blue,size=large
+$ krane my-app cluster1 --bindings=color=blue,size=large
```
dturn: These are now `krane deploy` plu to use bindings you'll need the allow erb flag

CONTRIBUTING.md
douglas: No, It slipped from the rename/replace process - I will revert the changes to the pipeline name.

README.md
douglas: Yup, good catch!

README.md
douglas: I changed the sentence for this phrase:

```
This was the previous name of this project! If you're curious about the rename to `Krane`, check [this issue](https://github.com/Shopify/kubernetes-deploy/issues/30#issuecomment-468750341) and [this issue](https://github.com/Shopify/kubernetes-deploy/issues?q=is%3Aissue+is%3Aopen+label%3A%22%3Arocket%3A+1.0+requirement%22) for context.
```

Could I improve it somehow?

douglas:
Just tested with:

Screen Shot 2019-10-15 at 6 05 43 PM

Which worked perfectly! 

Screen Shot 2019-10-15 at 6 06 50 PM

`krane` is SOOOO cool ❤️

douglas:
We will need to change the references in README of `--template-dir` to `--filenames` for `krane deploy` as it `fails`:

...

README.md
```diff
@@ -99,7 +99,7 @@ official compatibility chart below.
 
 ## Usage
 
-`kubernetes-deploy <app's namespace> <kube context>`
+`krane <app's namespace> <kube context>`
```
dturn: `krane deploy <app's namespace> <kube context>` looks like this needs to be fixed up in a few places in this file, it's a bit easier to see it in the 'rich diff' mode.

lib/krane/statsd.rb
```diff
@@ -2,11 +2,11 @@
 require 'statsd-instrument'
 require 'logger'
 
-module KubernetesDeploy
+module Krane
   class StatsD
     extend ::StatsD
 
-    PREFIX = "KubernetesDeploy"
+    PREFIX = "Krane"
```
dturn: We need to add `update DD dashboards` to the 1.0 release plan.

lib/krane.rb
dturn: Changing this file to 

```
# frozen_string_literal: true

require 'krane/deploy_task'
require 'krane/render_task'
require 'krane/restart_task'
require 'krane/runner_task'
```

worked for me locally

lib/krane/deploy_task.rb
```diff
@@ -1,12 +1,612 @@
...
+require 'krane/ejson_secret_provisioner'
+require 'krane/renderer'
+require 'krane/cluster_resource_discovery'
+require 'krane/template_sets'
+require 'krane/deploy_task_config_validator'
 
 module Krane
-  class DeployTask < KubernetesDeploy::DeployTask
-    def initialize(**args)
-      raise "Use Krane::DeployGlobalTask to deploy global resources" if args[:allow_globals]
```
dturn: This behavior is still important.

.rubocop.yml
```diff
@@ -6,8 +6,6 @@ AllCops:
 
 Naming/FileName:
   Enabled: true
-  Exclude:
-    - lib/kubernetes-deploy.rb
```
jessie-JNing: also remove or rename https://github.com/Shopify/kubernetes-deploy/pull/585/files#diff-11a0d7906801d9dea0eccb85667ad811L17  ?

jessie-JNing:
It's almost there 👍

.rubocop.yml
douglas: Yup, good catch! Thanks!

lib/krane/deploy_task.rb
dturn: I've done a poor job explaining this.

command                          | namespaced resources | global resources | both |
-------------------------|-------------------------|------------------|-----|
exe/kubernetes-deploy   | happy                               | prints a warning  | prints a warning |
krane deploy                     | happy                              | exception             | exception |
krane global-deploy         | exception                         | happy                   | exception |

douglas:
I will do these changes and fix the latest conflicts, thanks for your review, Danny!

KnVerey:
Can someone please point me to the rollout plan? After this goes in, we're effectively officially krane and will need to release under that gem instead. Once we're krane, I don't see why we need to keep any of the compatibility code--users that need the old version should be able to install the old gem too, no? If we want to have a version of kubernetes-deploy that has 100% feature overlap between the old ways and the new ways, we need to hold off the rename until the rest of the features are in. Which option we go with of course has implications for how we roll this out internally. One way or the other, we should be dogfooding the krane versions before the official 1.0 release.

Regardless, I think we need to do some more documentation work before merging this. It should remain straightfoward to start using our gem immediately after the rename, when new users will never have seen kubernetes-deploy. We also need to make sure that old users coming back to the repo will be able to have a seamless experience understanding what happened and what they need to do. This is something that could piss people off if we aren't careful. Minimally, we need some sort of prominent message about the cutover, documentation of which versions of our gem include both versions of the tasks/executables, and a guide on how to convert your existing usages (ruby and exes).

lib/krane/statsd.rb
KnVerey: I don't think this should change yet--it should match the name of the gem the user is installing

douglas:
Thanks for your questions, Katrina! I feel we have a clear vision now that we have this document created by Danny: https://docs.google.com/document/d/11i8Tk-qdnsMDci16KURFfAyf3FdZmXdyj93uqEu6Pns/edit#heading=h.2dalcahkcd6t

lib/krane/statsd.rb
douglas: Oh, this makes sense! Fixing it asap 👍

test/helpers/fixture_deploy_helper.rb
```diff
@@ -65,7 +67,7 @@ def deploy_dirs_without_profiling(dirs, wait: true, allow_protected_ns: false, p
     protected_namespaces: nil, render_erb: false, allow_globals: true)
     kubectl_instance ||= build_kubectl
 
-    deploy = KubernetesDeploy::DeployTask.new(
+    deploy = Krane::DeprecatedDeployTask.new(
```
KnVerey: This is the basis of the integration tests, so it should use one of the real public interfaces. There may be other changes associated with using `Krane::` (and having the globals command available might help), so that could be a separate PR, but I'd do that very soon.

test/integration-serial/serial_deploy_test.rb
```diff
@@ -453,18 +453,18 @@ def test_cr_failure_with_arbitrary_rollout_conditions
   end
 
   def test_deploying_crs_with_invalid_crd_conditions_fails
-    # Since CRDs are not always deployed along with their CRs and kubernetes-deploy is not the only way CRDs are
+    # Since CRDs are not always deployed along with their CRs and krane is not the only way CRDs are
     # deployed, we need to model the case where poorly configured rollout_conditions are present before deploying a CR
-    KubernetesDeploy::DeployTask.any_instance.expects(:validate_resources).returns(:true)
+    Krane::DeprecatedDeployTask.any_instance.expects(:validate_resources).returns(:true)
```
KnVerey: Why not stub on `Krane`here? The deprecated task is a very temporary implementation detail, so I'm wondering why it should be referenced by anything at all other than its (temporary) subclasses.

exe/kubernetes-deploy
```diff
@@ -68,8 +68,8 @@ end
 template_paths = [template_dir] if template_paths.empty? && template_dir
 
 begin
-  KubernetesDeploy::OptionsHelper.with_processed_template_paths(template_paths) do |paths|
-    runner = KubernetesDeploy::DeployTask.new(
+  Krane::OptionsHelper.with_processed_template_paths(template_paths) do |paths|
+    runner = Krane::DeprecatedDeployTask.new(
```
KnVerey: Wait, is `KubernetesDeploy::DeployTask` (and other tasks) still gone? We can't do that yet. That's part of the public interface, and we decided not to break that in the prereleases.

exe/kubernetes-render
```diff
@@ -22,11 +22,11 @@ ARGV.options do |opts|
...
 
 templates = ARGV
-logger = KubernetesDeploy::FormattedLogger.build(verbose_prefix: false)
+logger = Krane::FormattedLogger.build(verbose_prefix: false)
 
 begin
-  KubernetesDeploy::OptionsHelper.with_processed_template_paths(template_dir) do |dir|
-    runner = KubernetesDeploy::RenderTask.new(
+  Krane::OptionsHelper.with_processed_template_paths(template_dir) do |dir|
+    runner = Krane::RenderTask.new(
```
KnVerey: Since this is the kubernetes-deploy executable, we should probably use the `KubernetesDeploy` ruby API here (same with all the old exes). If we don't, we definitely need to do two separate tophats, one of the exe and one in a console.

exe/kubernetes-deploy
douglas: Yup, addressing this right now! Thanks again for your insights!

test/helpers/fixture_deploy_helper.rb
douglas: Just changed to address it, I agree we should still use `KubernetesDeploy` API while Krane 1.0 is not released.

exe/kubernetes-render
douglas: Done! I will 🎩 the exe's now - thanks for your help!

test/integration-serial/serial_deploy_test.rb
douglas: Hmmmm, I made it use public interface again... do you still think we should stub on `Krane` here?

KnVerey: Actually I don't think you can use `Krane` here, since the integration tests are still using `KubernetesDeploy::DeployTask` via the helper.

douglas:
I think I addressed all new comments and it is good for another review! Thank you all for the help 🙏

lib/krane/errors.rb
```diff
@@ -1,8 +1,9 @@
 # frozen_string_literal: true
+require 'kubernetes-deploy/errors'
+
 module Krane
-  class FatalDeploymentError < StandardError; end
-  class FatalKubeAPIError < FatalDeploymentError; end
```
KnVerey: Why remove the `Krane` version of this one?

lib/krane/render_task.rb
```diff
@@ -29,7 +29,7 @@ def initialize(logger: nil, current_sha:, template_dir: nil, template_paths: [],
     def run(*args)
       run!(*args)
       true
-    rescue Krane::FatalDeploymentError
+    rescue KubernetesDeploy::FatalDeploymentError
```
KnVerey: I think this `Krane` class should still be rescuing the `Krane` version.

lib/krane/errors.rb
douglas: Sorry, it was a bad diff/rebase! Thanks for catching it 👍

douglas:
I 🎩 `exe/kubernetes-deploy` with `gabi-test` repository in `tier4` and it still works, will 🎩 the other commands:

`REVISION=973693cf638f94f5ebbe1e50a73aa29b9e59d2cb be exe/kubernetes-deploy gabi-test-production tier4 --template-dir=../gabi-test/config/deploy/production --verbose-log-prefix`

Screen Shot 2019-10-25 at 11 17 51 AM

Also `exe/krane` works:

`REVISION=973693cf638f94f5ebbe1e50a73aa29b9e59d2cb be exe/krane deploy gabi-test-production tier4 --filenames=../gabi-test/config/deploy/production/ --verbose-log-prefix`

Screen Shot 2019-10-25 at 11 51 08 AM

lib/krane/errors.rb
```diff
@@ -1,8 +1,10 @@
 # frozen_string_literal: true
+require 'kubernetes-deploy/errors'
+
 module Krane
   class FatalDeploymentError < StandardError; end
   class FatalKubeAPIError < FatalDeploymentError; end
-  class KubectlError < StandardError; end
+  class KubectlError < FatalDeploymentError; end
```
KnVerey: Did you mean to change this? Kubectl errors aren't necessarily fatal (though if it raised out all the way to the user, obviously it was!)

lib/krane/restart_task.rb
```diff
@@ -42,7 +42,7 @@ def initialize(context:, namespace:, logger: nil, max_watch_seconds: nil)
     def run(*args)
       perform!(*args)
       true
-    rescue FatalDeploymentError
+    rescue Krane::FatalDeploymentError, KubernetesDeploy::FatalDeploymentError
```
KnVerey: Why do you need to add the `KuberentesDeploy` versions to these? I would expect that everything is actually raising the `Krane` versions, so things under the Krane namespace can ignore the fact that the KD versions exist.

lib/krane/errors.rb
douglas: No, it was not intended - thanks for catching this one.

lib/krane/restart_task.rb
douglas: Because of https://github.com/Shopify/kubernetes-deploy/pull/585/commits/9042b22beec335efc66601b6896c8c22ea062d81#diff-d6e404f21ee0321894c447fb753b8423L153, which was:

```ruby
    def run(*args)
      run!(*args)
      true
    rescue FatalDeploymentError
      false
    end
```

The thing is that in this context, `FatalDeploymentError` is actually `Krane::FatalDeploymentError`, and as the exception is changed to `KubernetesDeploy::FatalDeploymentError` due to the recent changes it was not being caught anymore, failing the tests related to this area.

KnVerey: Oh, why is the error conversion in the shared superclass? Can't you do it directly in `KubernetesDeploy::DeployTask` and avoid this problem?

douglas: hmmmm, let me try!

douglas: Thanks for the feedback, I think it should work now.

lib/krane/kubernetes_resource/custom_resource.rb
```diff
@@ -68,7 +68,7 @@ def validate_definition(*)
       @crd.validate_rollout_conditions
     rescue RolloutConditionsError => e
       @validation_errors << "The CRD that specifies this resource is using invalid rollout conditions. " \
-      "Kubernetes-deploy will not be able to continue until those rollout conditions are fixed.\n" \
+      "krane will not be able to continue until those rollout conditions are fixed.\n" \
```
KnVerey: This is kinda a funny case--both versions of the command will now use the new name here. I think this is not a big enough deal to fix.

lib/krane/errors.rb
```diff
@@ -0,0 +1,34 @@
...
+    attr_reader :content
+    attr_accessor :filename
+    def initialize(err, filename: nil, content: nil)
+      @filename = filename
+      @content = content
+      super(err)
+    end
+  end
+
+  class NamespaceNotFoundError < FatalDeploymentError
```
KnVerey: Not introduced by this PR, but I think this is error class is dead code?

lib/kubernetes-deploy/rescue_krane_exceptions.rb
```diff
@@ -0,0 +1,16 @@
+# frozen_string_literal: true
+require 'kubernetes-deploy/errors'
+
+module RescueKraneExceptions
```
KnVerey: Sorry for only noticing now, but this module should be namespaced (under `KubernetesDeploy` because that's where it is in the directory structure)

test/unit/krane/kubernetes_resource/custom_resource_definition_test.rb
```diff
@@ -92,15 +92,15 @@ def test_rollout_conditions_fails_validation_when_condition_is_wrong_type
...
+    cr = Krane::KubernetesResource.build(namespace: "test", context: "test",
       logger: @logger, statsd_tags: @statsd_tags, crd: crd,
       definition: {
         "kind" => "UnitTest",
         "metadata" => { "name" => "test" },
       })
     cr.validate_definition(kubectl)
     assert(cr.validation_error_msg.include?(
-      "The CRD that specifies this resource is using invalid rollout conditions. Kubernetes-deploy will not be " \
+      "The CRD that specifies this resource is using invalid rollout conditions. krane will not be " \
```
KnVerey: Actually, let's capitalize our name in this error message please.

lib/krane/errors.rb
douglas: Good catch, removed it 👍

Issue #526, opened 2019-08-14
Remove support for `kubernetes-deploy.shopify.io` annotations


dturn:
## Problem

We're transitioning from kubernetes-deploy to krane. This means that we need to stop supporting kubernetes-deploy.shopify.io annotations. This is the final step and should be done no sooner than the 1.0 release and it may be desirable to do this after the 1.0 release

## Possible solution
Remove all instances from the codebase. 

## Alternative solutions
Old annotations raise errors.

## Acceptance criteria
- `kubernetes-deploy.shopify.io` annotations no longer work.
- `krane.shopify.io` annotations continue to work.

## Additional context
https://github.com/Shopify/kubernetes-deploy/issues/256#issuecomment-475580770

dturn:
We should stop annotating ejson secrets when we action this issue. The annotations aren't being used  by any code paths, though we should be sure to check with K8s#Krane to ensure external users aren't relying on it. 

This came up because in pre.1 the annotations still use the k8s-deploy prefix and that's causing the k8s-deploy annotation depreciations warning to trigger. This sucks since users can't fix it without going to 1.0, which is a managed env like ship-it is out of the user's control.

KnVerey:
Is the warning about the ejson annotation we force actually fixed by moving 1.0? The annotation still uses `kubernetes-deploy` on master: https://github.com/Shopify/krane/blob/master/lib/krane/ejson_secret_provisioner.rb#L15

airhorns:
Is it time for this maybe?

dturn:
@airhorns the project is ready for it. Are you willing to PR it?

xaviablaza:
Could we move forward with this deprecation with #587?

dturn:
@xaviablaza Thanks for reminding us about this issue. We're ready to move forward, but it would be better if we handled all removing all of the old kubernetes-deploy annotations at once. Is that something you'd be willing to PR? If not even rebasing the PR you linked would be a big help. If you're unable we've got the work scheduled but I can't commit us to a definitive date.

PR #587, opened 2019-10-10
Fix deprecated ejson secret annotation prefix

DazWorrall:

`kubernetes-deploy` is generating ejson secrets with the deprecated annotation, this clears up noise in deploy logs which the user can't fix:

```
[INFO][2019-10-10 09:09:06 +0000]	Template warning: Secret-shopify-build-environment-secrets20191010-854063-91nkb7.yml
[INFO][2019-10-10 09:09:06 +0000]	> Warning message:
[INFO][2019-10-10 09:09:06 +0000]	    kubernetes-deploy.shopify.io as a prefix for annotations is deprecated: Use the 'krane.shopify.io' annotation prefix instead
```

As far as I can tell this annotation isn't actually _read_ or searched for by kubernetes-deploy, in which case backwards compatibility is not a concern. Am I right?

### DIFF
--- a/lib/kubernetes-deploy/ejson_secret_provisioner.rb
+++ b/lib/kubernetes-deploy/ejson_secret_provisioner.rb
@@ -12,7 +12,7 @@ module KubernetesDeploy
   end
 
   class EjsonSecretProvisioner
-    EJSON_SECRET_ANNOTATION = "kubernetes-deploy.shopify.io/ejson-secret"
+    EJSON_SECRET_ANNOTATION = "krane.shopify.io/ejson-secret"
     EJSON_SECRET_KEY = "kubernetes_secrets"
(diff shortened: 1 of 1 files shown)


KnVerey:
Yeah, this is something that we add to mark secrets we created. IIRC older versions of `kubernetes-deploy` use it to identify secrets it created for pruning purposes, but now we treat secrets like other resources and use the `last-applied` annotation for that (well, kubectl does it for us). @dturn what was the thinking in adding the deprecation?

dturn:
> @dturn what was the thinking in adding the deprecation?

Shopify doesn't use the annotation, and we're going to update it when we action  #526. The deprecation was to give external users time to update any tooling they've built that uses it.

dturn:
Fixed in https://github.com/Shopify/krane/pull/738/files

PR #594, opened 2019-10-18
Rework StatsD usage to use separate client

wvanbergen:

Get rid of the StatsD monkeypatches that cause us to be stuck on a specific version of statsd-instrument.

**How is this accomplished?**

By instantiating a separate `StatsD::Instrument::Client`, there is no need to override the singleton anymore.

**What could go wrong?**

Metrics may be broken after this change, although the test suite suggests otherwise.


lib/kubernetes-deploy/statsd.rb
```diff
@@ -4,37 +4,14 @@
...
-    def self.build
-      if ENV['STATSD_DEV'].present?
-        self.backend = ::StatsD::Instrument::Backends::LoggerBackend.new(Logger.new($stderr))
-      elsif ENV['STATSD_ADDR'].present?
-        statsd_impl = ENV['STATSD_IMPLEMENTATION'].present? ? ENV['STATSD_IMPLEMENTATION'] : "datadog"
-        self.backend = ::StatsD::Instrument::Backends::UDPBackend.new(ENV['STATSD_ADDR'], statsd_impl)
-      else
-        self.backend = ::StatsD::Instrument::Backends::NullBackend.new
-      end
-    end
```
wvanbergen: I *think* the defaults that we pick up from the environment are fine in this case. The new client uses `datadog` as default implementation.

wvanbergen:
Updated this PR. The client that gets instantiated should now reflect the legacy client we would instantiate before.

One change is that I changed `STATSD_DEV=1` to `STATD_ENV=development`, because this is a standard environment variable that the library already supports.

lib/kubernetes-deploy/deploy_task.rb
```diff
@@ -200,25 +200,28 @@ def run!(verify_result: true, allow_protected_ns: false, prune: true)
         MSG
         @logger.summary.add_paragraph(ColorizedString.new(warning).yellow)
       end
-      StatsD.event("Deployment of #{@namespace} succeeded",
+      tags = statsd_tags + ["status:success"]
+      StatsD.client.event("Deployment of #{@namespace} succeeded",
         "Successfully deployed all #{@namespace} resources to #{@context}",
-        alert_type: "success", tags: statsd_tags << "status:success")
-      StatsD.distribution('all_resources.duration', StatsD.duration(start), tags: statsd_tags << "status:success")
```
wvanbergen: I think this was a (minor) bug. `<<` modifies the array, and then returns it. This means that every time it is called, the array grows by one element. In this case, we add the same tag twice, which doesn't really cause any issues in practice, but I figured I should fix this while I am at it.

wvanbergen:
I added a changelog entry, but I still have to fix a compatibility issue in the test helpers in statsd-instrument.

lib/kubernetes-deploy/statsd.rb
```diff
@@ -4,39 +4,25 @@
...
-      if ENV['STATSD_DEV'].present?
-        self.backend = ::StatsD::Instrument::Backends::LoggerBackend.new(Logger.new($stderr))
-      elsif ENV['STATSD_ADDR'].present?
-        statsd_impl = ENV['STATSD_IMPLEMENTATION'].present? ? ENV['STATSD_IMPLEMENTATION'] : "datadog"
-        self.backend = ::StatsD::Instrument::Backends::UDPBackend.new(ENV['STATSD_ADDR'], statsd_impl)
-      else
-        self.backend = ::StatsD::Instrument::Backends::NullBackend.new
+    def self.client
+      @client ||= begin
+        sink = if ::StatsD::Instrument::Environment.current.env.fetch('STATSD_ENV', nil) == 'development'
```
timothysmith0609: It seems as though the `::StatsD::Instrument::Environment.current.env` is just a [basic wrapper around `ENV`](https://github.com/Shopify/statsd-instrument/blob/master/lib/statsd/instrument/environment.rb#L8), in this case. The other fallbacks appear to be geared more towards Rails applications, of which krane isn't one. With that in mind, it might be cleaner just to do `if ENV.fetch(X, nil) == Y`, instead.

wvanbergen:
https://github.com/Shopify/statsd-instrument/pull/234 fixes the issue that causes us not being able to parse and assert stuff on the `event` datagrams being emitted.

CHANGELOG.md
```diff
@@ -3,6 +3,9 @@
 *Important!*
 - The next release will be 1.0.0, which means that master will contain breaking changes.
 
+*Other*
+- Refactor StatsD usage so we can depend on the latest version again.
```
timothysmith0609:
```suggestion
- Refactor StatsD usage so we can depend on the latest version again. ([#594](https://github.com/Shopify/kubernetes-deploy/pull/594))
```

Issue #592, opened 2019-10-18
kubernetes-deploy fails with API version 1.16


redmcg:
# Bug report

`kubernetes-deploy` (and `krane deploy`) fail during the prune phase on deploy to a k8s cluster using v1.16 of the api. This is because the following apis:
```
extensions/v1beta1/ReplicaSet
extensions/v1beta1/DaemonSet
extensions/v1beta1/Deployment
apps/v1beta1/StatefulSet
```

have been deprecated. See https://kubernetes.io/blog/2019/07/18/api-deprecations-in-1-16/ for reference.

I have created a fork that addresses this issue for v1.16 (see https://github.com/redmcg/kubernetes-deploy/tree/fixforv1.16), but I did not submit this as a PR as I'm worried it will break older versions of the k8s API (where the new APIs may not be implemented).

**Expected behavior:**
`kubernetes-deploy` (and `krane deploy`) to successfully deploy to a k8s v1.16 cluster

**Actual behavior:**
`kubernetes-deploy` (and `krane deploy`) fail during the prune phase with the following error:
```
[INFO][2019-10-18 11:19:43 +1100]	------------------------------------------Result: FAILURE-------------------------------------------
[FATAL][2019-10-18 11:19:43 +1100]	Command failed: apply -f /tmp/d20191018-14256-1x9qqjf --prune --all --prune-whitelist\=core/v1/configmap --prune-whitelist\=core/v1/pod --prune-whitelist\=core/v1/service --prune-whitelist\=core/v1/resourcequota --prune-whitelist\=core/v1/secret --prune-whitelist\=core/v1/serviceaccount --prune-whitelist\=core/v1/podtemplate --prune-whitelist\=batch/v1/job --prune-whitelist\=extensions/v1beta1/replicaset --prune-whitelist\=extensions/v1beta1/daemonset --prune-whitelist\=extensions/v1beta1/deployment --prune-whitelist\=extensions/v1beta1/ingress --prune-whitelist\=networking.k8s.io/v1/networkpolicy --prune-whitelist\=apps/v1beta1/statefulset --prune-whitelist\=autoscaling/v1/horizontalpodautoscaler --prune-whitelist\=policy/v1beta1/poddisruptionbudget --prune-whitelist\=batch/v1beta1/cronjob --prune-whitelist\=rbac.authorization.k8s.io/v1/role --prune-whitelist\=rbac.authorization.k8s.io/v1/rolebinding
[...]

PR #595, opened 2019-10-18
Support 1.16

dturn:

Add support for 1.16
Fixes: https://github.com/Shopify/kubernetes-deploy/issues/592

**How is this accomplished?**
Update resource versions in prune white list https://github.com/redmcg/kubernetes-deploy/commit/73e03f23f3efd6813b0b0b7e0a505cc45c5ed754

**What could go wrong?**
We're already not compatible with 1.16 so unlikely this will make it worse.

### DIFF
--- a/.buildkite/pipeline.nightly.yml
+++ b/.buildkite/pipeline.nightly.yml
@@ -1,4 +1,11 @@
 steps:
+  - name: 'Run Test Suite (:kubernetes: 1.16-latest)'
+    command: bin/ci
+    agents:
+      queue: k8s-ci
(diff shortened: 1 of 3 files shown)


KnVerey:
What is the addition/deprecation sequence for each of those GVs? Is it possible that all the k8s versions we support are compatible with the newer ones (or with an intermediate GV)?

dturn:
Replaced by https://github.com/Shopify/kubernetes-deploy/pull/596 since our CI isn't ready for 1.16 yet.

PR #600, opened 2019-10-21
Fix the vast majority of Ruby warnings

wvanbergen:

When running `rake` to run the test suite using macOS system Ruby, the output is 90% Ruby warnings, because this `rake` version runs with warnings enabled. Most of the warnings are caused by uninitialized instance variables.

**How is this accomplished?**

Initializing instance variables before using them makes them go away. There are two wranings coming from dependencies that this doesn't address.

**What could go wrong?**

Unsure. It's theoretically possible that this breaks something, but it's highly unlikely. The more important question is whether we care about these warnings or not.

### DIFF
--- a/test/test_helper.rb
+++ b/test/test_helper.rb
@@ -62,6 +62,7 @@ module KubernetesDeploy
 
     def run
       ban_net_connect? ? WebMock.disable_net_connect! : WebMock.allow_net_connect!
+      @namespace = nil
       yield if block_given?
       super
     end
--- a/lib/kubernetes-deploy/container_logs.rb
+++ b/lib/kubernetes-deploy/container_logs.rb
@@ -14,6 +14,7 @@ module KubernetesDeploy
       @lines = []
       @next_print_index = 0
       @printed_latest = false
+      @last_timestamp = nil
     end
 
     def sync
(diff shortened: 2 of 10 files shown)


test/test_helper.rb
```diff
@@ -68,6 +68,7 @@ def run
 
     def setup
       Kubectl.any_instance.expects(:run).never if ban_net_connect? # can't use mocha in Minitest::Test#run
+      @namespace = nil
```
KnVerey: This needs to go in `run` instead. Right now it is overwriting the value set in the integration test helper.

wvanbergen:
Going to close this. Because of ordering issues of nested initializers, this is not super trivial to fix for somebody without context, especially because `dev uyp` does not give me a working environment in which I can run the full test suite locally.

Somebody else is welcome to pick this back up and re-open the PR.

KnVerey:
> especially because dev uyp does not give me a working environment in which I can run the full test suite locally.

All you really need is a local cluster--any one will do if the minikube installation is failing for you. To use a different one, simply `echo "name-of-the-context-you-want-to-use" > .local-context`. More information is available in our contributors' guide: https://github.com/Shopify/kubernetes-deploy/blob/master/CONTRIBUTING.md#running-the-test-suite-locally 😄

PR #602, opened 2019-10-22
Global deploy 2

dturn:
Second attempt to build a global deploy task. 

**How is this accomplished?**
Three major differences from #574.

1)  The global_deploy_task.rb doesn't inherit from deploy_task.rb. 

2) I created a concern(module) for the code that should be shared. There is still duplicate code, but what's duplicated is either commonly duplicated in the codebase (kubectl) or small snippets that might look different in the future without causing harm to the other task. 

2) New class introduced: ResourceDeployer. 

**ToDo**

- [x] Write more tests
- [x] Rebsae after the rename PR merged

**What could go wrong?**
This doesn't support pruning, that will come in a follow-up PR. 

Something subtle and untested breaks (or changes) behavior in deploy_test.rb due to it being in the new concern or class.


lib/kubernetes-deploy/deploy_task.rb
```diff
@@ -41,11 +41,16 @@
...
+require 'kubernetes-deploy/resource_deployer'
+
+require 'krane/concerns/template_reporting'
 
 module KubernetesDeploy
   # Ship resources to a namespace
   class DeployTask
     extend KubernetesDeploy::StatsD::MeasureMethods
+    include Krane::TemplateReporting
+    delegate :logger, to: :@task_config
```
dturn: `Krane::TemplateReporting` expectes to be able to access the logger as `logger` instead of `@logger`. I can convert this entire file to use `logger` over `@logger` if desired.

lib/kubernetes-deploy/resource_deployer.rb
```diff
@@ -0,0 +1,236 @@
...
+require 'kubernetes-deploy/resource_watcher'
+require 'krane/concerns/template_reporting'
+
+module KubernetesDeploy
+  class ResourceDeployer
+    extend KubernetesDeploy::StatsD::MeasureMethods
+    include Krane::TemplateReporting
+
+    delegate :logger, to: :@task_config
+    attr_reader :statsd_tags
```
dturn: Needed by `KubernetesDeploy::StatsD::MeasureMethods`

lib/kubernetes-deploy/kubernetes_resource.rb
```diff
@@ -317,7 +317,7 @@ def debug_message(cause = nil, info_hash = {})
     def fetch_events(kubectl)
       return {} unless exists?
       out, _err, st = kubectl.run("get", "events", "--output=go-template=#{Event.go_template_for(type, name)}",
-        log_failure: false)
+        log_failure: false, use_namespace: !global?)
```
dturn: `kubectl api-resouces` indicates that events are `NAMESPACED`. Running `kubectl get events` returns events about `Nodes` which are also global, so I believe this is the correct behavior.

test/integration-serial/serial_deploy_test.rb
```diff
@@ -503,6 +503,25 @@ def test_apply_failure_with_sensitive_resources_hides_template_content
...
 
+  def test_global_deploy_task
+    KubernetesDeploy::FormattedLogger.expects(:build).returns(@logger)
+    global_deploy = ::Krane::GlobalDeployTask.new(
+      context: task_config.context,
+      template_paths: [fixture_path('globals')],
+      max_watch_seconds: 300,
+      selector: KubernetesDeploy::LabelSelector.parse('app=krane')
+    )
+    global_deploy.run!(verify_result: true, prune: false)
```
dturn: This test is logging:
```
I, [2019-10-23T16:46:47.093069 #38172]  INFO -- : [StatsD]  Deployment succeeded:Successfully deployed all resources to minikube #context: #status:success
I, [2019-10-23T16:46:47.093143 #38172]  INFO -- : [StatsD] distribution all_resources.duration:1.1 #context: #status:success
```

I suspect something is configuring statsd different for the global_deploy_task than deploy_task.

lib/kubernetes-deploy/resource_cache.rb
```diff
@@ -51,9 +51,10 @@ def use_or_populate_cache(kind)
 
     def fetch_by_kind(kind)
       resource_class = KubernetesResource.class_for_kind(kind)
+      global_kind = @task_config.global_kinds.include?(kind.downcase)
```
KnVerey: Seeing this, I definitely think that `TaskConfig` should do the fetching. Otherwise we have a call ordering dependency involving three objects (this, the task and the mediating task config).

lib/kubernetes-deploy/resource_deployer.rb
```diff
@@ -0,0 +1,236 @@
...
+      command = %w(apply)
+
+      Dir.mktmpdir do |tmp_dir|
+        resources.each do |r|
+          FileUtils.symlink(r.file_path, tmp_dir)
+          r.deploy_started_at = Time.now.utc
+        end
+        command.push("-f", tmp_dir)
+
+        if prune
```
KnVerey: Maybe to be extra careful we could make this `if prune && prune_whitelist.present?`. I assume that if you pass `--prune --all --prune-whitelist=[]` kubectl will take that to mean "actually don't prune", but passing `prune=false` (aka no pruning-related args) would make me feel better.

lib/kubernetes-deploy/resource_deployer.rb
```diff
@@ -0,0 +1,236 @@
...
+      applyables, individuals = resources.partition { |r| r.deploy_method == :apply }
+      # Prunable resources should also applied so that they can  be pruned
+      pruneable_types = @prune_whitelist.map { |t| t.split("/").last }
+      applyables += individuals.select { |r| pruneable_types.include?(r.type) }
+
+      individuals.each do |r|
+        r.deploy_started_at = Time.now.utc
+        case r.deploy_method
+        when :replace
+          _, _, replace_st = kubectl.run("replace", "-f", r.file_path, log_failure: false)
```
KnVerey: Do we need to conditionally exclude the namespace arg on these?

lib/kubernetes-deploy/resource_deployer.rb
```diff
@@ -0,0 +1,236 @@
...
+          if @selector
+            command.push("--selector", @selector.to_s)
+          else
+            command.push("--all")
+          end
+          @prune_whitelist.each { |type| command.push("--prune-whitelist=#{type}") }
+        end
+
+        output_is_sensitive = resources.any?(&:sensitive_template_content?)
+        global_mode = resources.all?(&:global?)
```
KnVerey: So in the mixed-mode scenario currently possible under the old deploy task, we pass a namespace arg that is just ignored by kubectl as appropriate? I guess that's what we've always done, and it works. 🤔

lib/krane/global_deploy_task.rb
```diff
@@ -0,0 +1,209 @@
...
+      record_invalid_template(err: e.message, filename: e.filename, content: e.content)
+      raise KubernetesDeploy::FatalDeploymentError, "Failed to render and parse template"
+    end
+    measure_method(:discover_resources)
+
+    def cluster_resource_discoverer
+      @cluster_resource_discoverer ||= KubernetesDeploy::ClusterResourceDiscovery.new(task_config: @task_config)
+    end
+
+    def validate_resources(resources)
```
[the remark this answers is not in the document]
dturn: I was actually debating if this method belongs to the global_task_config_validator.

lib/kubernetes-deploy/kubernetes_resource.rb
```diff
@@ -40,7 +40,7 @@ class KubernetesResource
     SERVER_DRY_RUNNABLE = false
 
     class << self
-      def build(namespace:, context:, definition:, logger:, statsd_tags:, crd: nil, global_names: [])
+      def build(namespace: nil, context:, definition:, logger:, statsd_tags:, crd: nil, global_names: [])
```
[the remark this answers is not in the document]
dturn: Yeah, we've got a tracking issue for making build take a task_config.

lib/krane/global_deploy_task.rb
```diff
@@ -0,0 +1,209 @@
...
+      false
+    end
+
+    # Runs the task, raising exceptions in case of issues
+    #
+    # @param verify_result [Boolean] Wait for completion and verify success
+    # @param prune [Boolean] Enable deletion of resources that do not appear in the template dir
+    #
+    # @return [nil]
+    def run!(verify_result: true, prune: true)
```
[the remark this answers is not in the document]
dturn: I'm not opposed to adding a pre-deploy phase, and if someone comes up with a use case I'll add it. Till then 👍 on documenting it.

lib/krane/global_deploy_task.rb
```diff
@@ -0,0 +1,173 @@
...
+      start = Time.now.utc
+      logger.reset
+
+      logger.phase_heading("Initializing deploy")
+      validator = validate_configuration
+      resources = discover_resources
+      validator.validate_resources(resources, @selector)
```
dturn: Not 100% sold that `validate_resources` belongs to the (Global)DeployTaskConfigValidator. But I do like the idea of moving validation code out of this file.

test/exe/global_deploy_test.rb
```diff
@@ -0,0 +1,75 @@
...
+    krane_global_deploy!(flags: "--selector #{selector}")
+  end
+
+  private
+
+  def set_krane_global_deploy_expectations!(new_args: {}, run_args: {})
+    options = default_options(new_args, run_args)
+    selector_args = options[:new_args][:selector]
+    selector = mock('LabelSelector')
+    Krane::LabelSelector.expects(:parse).with(selector_args).returns(selector)
```
KnVerey: Why go to all this effort to stub out the selector parsing? Your dummy selector values are invalid right now.

test/exe/global_deploy_test.rb
```diff
@@ -0,0 +1,75 @@
...
+    krane_global_deploy!(flags: '--verify-result true')
+    set_krane_global_deploy_expectations!(run_args: { verify_result: false })
+    krane_global_deploy!(flags: '--verify-result false')
+  end
+
+  def test_deploy_passes_filename
+    set_krane_global_deploy_expectations!(new_args: { filenames: ['/my/file/path'] })
+    krane_global_deploy!(flags: '-f /my/file/path')
+    set_krane_global_deploy_expectations!(new_args: { filenames: ['/my/other/file/path'] })
+    krane_global_deploy!(flags: '--filenames /my/other/file/path')
```
KnVerey: Add a case for multiple filenames?

test/unit/krane/resource_deployer_test.rb
```diff
@@ -0,0 +1,96 @@
...
+  def test_deploy_prune_builds_whitelist
+    whitelist_kind = "fake_kind"
+    resource = build_mock_resource
+    Krane::Kubectl.any_instance.expects(:run).with do |*args|
+      args.include?("--prune-whitelist=#{whitelist_kind}")
+    end.returns(["", "", stub(success?: true)])
+    resource_deployer(kubectl_times: 0, prune_whitelist: [whitelist_kind]).deploy!([resource], false, true)
+  end
+
+  def test_deploy_no_prune_doesnt_prune
```
KnVerey: I would have expected to see almost everything in this file as an integration test, not a unit test: this is the core functionality of the task. Is there a particular reason you chose unit testing this time?

test/unit/krane/kubernetes_resource_test.rb
```diff
@@ -398,7 +398,7 @@ def test_whitespace_in_debug_message
 
   def test_disappeared_is_true_if_resource_has_been_deployed_and_404s
     dummy = DummyResource.new
-    cache = Krane::ResourceCache.new(task_config: task_config(namespace: 'test', context: 'minikube'))
+    cache = Krane::ResourceCache.new(task_config(namespace: 'test', context: 'minikube'))
```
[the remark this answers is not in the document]
dturn: ResourceCache takes a positional args not kawrgs, but until the change in this PR that cause the task_config to store the global_kinds. The use or ResourceCache in this test  didn't actually try and use the task_config, so anything passed in would have worked.

test/unit/krane/kubernetes_resource/pod_test.rb
```diff
@@ -220,7 +220,7 @@ def test_deploy_failed_is_true_for_disappeared_unmanaged_pods
     template = build_pod_template
     pod = Krane::Pod.new(namespace: 'test', context: 'nope', definition: template,
       logger: @logger, deploy_started_at: Time.now.utc)
-    cache = build_resource_cache
+    cache = build_resource_cache(global_kinds: nil)
```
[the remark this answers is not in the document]
dturn: `build_resource_cache` stubs `@task_config.global_kinds` (which gets called by `fetch_by_kind`). Because we've configured the fail on unused stubs we need an opt-out mechanism. Want me to change the name? The other option is to explicitly opt-in, but seemed odd since its the default behavior.

test/exe/global_deploy_test.rb
dturn: The `with` on line 48 is really hard to write if you want to allow any instance of a class inside of a hash, but still enforce everything else.

test/integration-serial/serial_deploy_test.rb
```diff
@@ -503,6 +501,25 @@ def test_apply_failure_with_sensitive_resources_hides_template_content
     refute_logs_match("kind: Deployment") # content of the sensitive template
   end
 
+  def test_global_deploy_task
+    Krane::FormattedLogger.expects(:build).returns(@logger)
+    global_deploy = ::Krane::GlobalDeployTask.new(
+      context: task_config.context,
+      filenames: [fixture_path('globals')],
+      global_timeout: 300,
+      selector: Krane::LabelSelector.parse('app=krane')
```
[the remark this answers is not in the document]
dturn: Hmm, I'd like to write a few more tests before we try and make them parallel. To do it right now I think we'd have to inject the selector, update the name to be unique and then have the ensure blocks delete by selector. Which I can't decide if its a good pattern or one that would be frustrating.

test/unit/krane/resource_deployer_test.rb
dturn: If I had written it from scratch I definitely would have made them integration tests. But between the existing integration tests exercising this code and having unit tests for many of the other classes I went this direction. Happy to make them integration tests.

KnVerey: Please do make them integration tests.

> having unit tests for many of the other classes

Not for the tasks themselves though, right? In general I personally tried to use integration tests for everything except edge cases other behavior that can't easily be tested by an integration test.

test/unit/krane/kubernetes_resource/pod_test.rb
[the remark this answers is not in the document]
KnVerey: Oh, I see. My suggestion would be to make the signature `def build_resource_cache(task_config: build_task_config)`, allowing you to pass in a task config with different stubbing if you need to. Looking at this again, it is a little odd to be passing the builder `global_kinds`, which is not a property of `ResourceCache`.

test/exe/global_deploy_test.rb
KnVerey: Did you try using an `instance_of` parameter matcher? https://mocha.jamesmead.org/Mocha/ParameterMatchers.html#instance_of-instance_method

test/integration-serial/serial_deploy_test.rb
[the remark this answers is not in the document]
KnVerey: What you're describing does sound like what we'll need to do (I don't see how it could change in the future), and we should have a test helper that takes care of it for future contributors to this task. In my opinion, this PR is responsible for introducing a sustainable way to write tests for its new task. What I'd like to see is a new `test/integration/global_deploy_test.rb` like we have for all the other tasks, with similar kinds of integration test coverage. I'm not sure whether it actually makes sense for the new suite to inherit from our regular integration test superclass, since the setup it needs is different (needs a unique partition key, but not a namespace)

test/unit/krane/resource_deployer_test.rb
dturn: >Not for the tasks themselves though, right? 

I'm not sure I follow, this isn't a task and now that I look only the tasks have integration tests.

KnVerey: Ah, you're right, sorry--not sure what made me think these were task tests. Does the GlobalDeployTask have integration coverage of these things? I only remember seeing one integration test for it, but maybe I missed something.

dturn: Added additional integration tests to `test/integration/global_deploy_test.rb `  (I'll add ones that test pruning in the PR that adds pruning support.)

dturn:
Pruning draft PR https://github.com/Shopify/kubernetes-deploy/pull/612

lib/krane/cluster_resource_discovery.rb
```diff
@@ -23,7 +23,8 @@ def global_resource_kinds
     private
 
     def fetch_globals
-      raw, _, st = kubectl.run("api-resources", "--namespaced=false", output: "wide", attempts: 5)
+      raw, _, st = kubectl.run("api-resources", "--namespaced=false", output: "wide", attempts: 5,
+        use_namespace: false)
```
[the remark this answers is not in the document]
dturn: I'd prefer not to push this logic into the kubectl class it would be very special purpose. As an example of a future thing we might want: to list all api-resources we'd still want `use_namespace: false` but not to add `--namespaced=`. So it would still require a second param.

lib/krane/global_deploy_task.rb
```diff
@@ -0,0 +1,207 @@
...
+require 'krane/kubernetes_resource'
+require 'krane/global_deploy_task_config_validator'
+require 'krane/concerns/template_reporting'
+
+%w(
+  custom_resource
+  custom_resource_definition
+).each do |subresource|
+  require "krane/kubernetes_resource/#{subresource}"
+end
```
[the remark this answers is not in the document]
dturn: Its future proofing for when we support more global resources.

test/unit/krane/resource_deployer_test.rb
timothysmith0609: Is-there/do-we-need a test to ensure a namespaced CR is _not_ deployed whereas a cluster-scope CR _is_?

test/fixtures/globals/storage_classes.yml
```diff
@@ -2,4 +2,6 @@ apiVersion: storage.k8s.io/v1
 kind: StorageClass
 metadata:
   name: testing-storage-class
+  labels:
+    app: krane
```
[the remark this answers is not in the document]
dturn: Makes it possible to use the fixtures in this bucket from the command line

`krane global-deploy minikube -f test/fixtures/globals/ --selector=app=krane  `

lib/krane/global_deploy_task.rb
```diff
@@ -0,0 +1,207 @@
...
+        end
+        raise FatalDeploymentError, "Template validation failed"
+      end
+    end
+    measure_method(:validate_resources)
+
+    def validate_globals(resources)
+      return unless (namespaced = resources.reject(&:global?).presence)
+      namespaced_names = namespaced.map do |resource|
+        "#{resource.name} (#{resource.type}) in #{File.basename(resource.file_path)}"
```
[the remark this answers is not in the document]
dturn: This has come up a few times so I've added a test `test_global_deploy_validation_catches_namespaced_cr`. But resource.global? works because [KubernetesResource::build]( https://github.com/Shopify/kubernetes-deploy/pull/602/files#diff-58f1540757c3323310e495f480799ebdL56) set's the resources `global` ivar based on the list of all global kinds fetched from the cluster ("api-resources", "--namespaced=false"). So there isn't actually a need to parse it out of the CRD.

lib/krane/global_deploy_task.rb
```diff
@@ -0,0 +1,207 @@
...
+
+    def discover_resources
+      logger.info("Discovering resources:")
+      resources = []
+      crds_by_kind = cluster_resource_discoverer.crds.map { |crd| [crd.name, crd] }.to_h
+      @template_sets.with_resource_definitions do |r_def|
+        crd = crds_by_kind[r_def["kind"]]&.first
+        r = KubernetesResource.build(context: context, logger: logger, definition: r_def,
+          crd: crd, global_names: global_kinds, statsd_tags: statsd_tags)
+        resources << r
```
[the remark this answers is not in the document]
dturn: https://github.com/Shopify/kubernetes-deploy/pull/602#discussion_r343237770 has an explanation of how we catch namespaced resources. I don't think the extra caution is needed and changing this flow would add complexity in order to keep our logging helpful.

test/unit/krane/resource_deployer_test.rb
```diff
@@ -0,0 +1,96 @@
...
+  end
+
+  def test_deploy_verify_false_no_failure_error
+    resource = build_mock_resource(final_status: "failure")
+    resource_deployer.deploy!([resource], false, false)
+    logger.print_summary(:done) # Force logger to flush
+    assert_logs_match_all(["Deploy result verification is disabled for this deploy."])
+  end
+
+  def test_predeploy_priority_resources_respectes_pre_deploy_list
```
timothysmith0609: nit: `respectes_pre_deploy... -> respects_pre_deploy...`

test/unit/krane/resource_deployer_test.rb
```diff
@@ -0,0 +1,96 @@
...
+    kind = "MockResource"
+    resource = build_mock_resource
+    watcher = mock("ResourceWatcher")
+    watcher.expects(:run).returns(true)
+    Krane::ResourceWatcher.expects(:new).returns(watcher)
+    priority_list = [kind]
+    resource_deployer.predeploy_priority_resources([resource], priority_list)
+  end
+
+  def test_predeploy_priority_resources_respectes_empty_pre_deploy_list
```
timothysmith0609: nit: `respectes -> respects`

timothysmith0609:
One more note is to fix the linter error being thrown in CI

test/integration/global_deploy_test.rb
```diff
@@ -0,0 +1,109 @@
...
+      %r{Deploying StorageClass\/testing-storage-class[\w-]+ \(timeout: 300s\)},
+      "Result: SUCCESS",
+      "Deployed 1 resource",
+      "Deploy result verification is disabled for this deploy.",
+      "This means the desired changes were communicated to Kubernetes, but the"\
+      " deploy did not make sure they actually succeeded.",
+    ])
+  end
+
+  def test_global_deploy_task_success_selector
```
KnVerey: Is there a test somewhere that shows attempting to pass in a blank selector fails validation? Or that a selector that doesn't match one of the resources in the set also fails validation?

test/integration/global_deploy_test.rb
```diff
@@ -0,0 +1,109 @@
+# frozen_string_literal: true
+require 'integration_test_helper'
+
+class GlobalDeployTest < Krane::IntegrationTest
+  include StatsDHelper
```
KnVerey: are you using this?

test/integration/global_deploy_test.rb
```diff
@@ -0,0 +1,109 @@
...
+      " deploy did not make sure they actually succeeded.",
+    ])
+  end
+
+  def test_global_deploy_task_success_selector
+    assert_deploy_success(deploy_global_fixtures('globals', selector: "app=krane"))
+
+    assert_logs_match_all([
+      "Phase 1: Initializing deploy",
+      "Using resource selector test=",
```
KnVerey: this doesn't match the selector you passed in

lib/krane/task_config.rb
```diff
@@ -9,8 +9,15 @@ def initialize(context, namespace, logger = nil)
       @logger = logger
     end
 
+    def global_kinds
+      @global_kinds ||= begin
+        cluster_resource_discoverer = ClusterResourceDiscovery.new(task_config: self)
+        cluster_resource_discoverer.global_resource_kinds
```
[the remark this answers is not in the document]
dturn: This changes a in the pruning PR. I'd like to defer this question till that PR.

test/unit/krane/resource_deployer_test.rb
```diff
@@ -0,0 +1,96 @@
...
+  end
+
+  def test_deploy_verify_false_no_failure_error
+    resource = build_mock_resource(final_status: "failure")
+    resource_deployer.deploy!([resource], false, false)
+    logger.print_summary(:done) # Force logger to flush
+    assert_logs_match_all(["Deploy result verification is disabled for this deploy."])
+  end
+
+  def test_predeploy_priority_resources_respects_pre_deploy_list
```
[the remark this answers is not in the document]
dturn: Added a comment in the code with an explanation.

lib/krane/global_deploy_task.rb
```diff
@@ -33,10 +33,10 @@ class GlobalDeployTask
     # @param global_timeout [Integer] Timeout in seconds
     # @param selector [Hash] Selector(s) parsed by Krane::LabelSelector
     # @param template_paths [Array<String>] An array of template paths
-    def initialize(context:, global_timeout: nil, selector: nil, filenames: [])
+    def initialize(context:, global_timeout: nil, selector: nil, filenames: [], logger: nil)
       template_paths = filenames.map { |path| File.expand_path(path) }
 
-      @task_config = TaskConfig.new(context, nil)
+      @task_config = TaskConfig.new(context, nil, logger)
```
KnVerey: Let's default this like the old task does, so they don't _have_ to pass one in.

test/integration/global_deploy_test.rb
```diff
@@ -0,0 +1,119 @@
...
+      "All required parameters and files are present",
+      "Discovering resources:",
+      "  - StorageClass/testing-storage-class",
+      "Phase 2: Checking initial resource statuses",
+      %r{StorageClass\/testing-storage-class[\w-]+\s+Not Found},
+      "Phase 3: Deploying all resources",
+      %r{Deploying StorageClass\/testing-storage-class[\w-]+ \(timeout: 300s\)},
+      "Don't know how to monitor resources of type StorageClass.",
+      %r{Assuming StorageClass\/testing-storage-class[\w-]+ deployed successfully.},
+      %r{Successfully deployed in 0.[\d]+s: StorageClass\/testing-storage-class},
```
KnVerey: Oh, this needs to be relaxed:
> Successfully deployed in 0.[\d]+s: StorageClass\/testing-storage-class)' not found in the following logs:
Because it took more than a second for whatever reason

test/integration/global_deploy_test.rb
```diff
@@ -0,0 +1,119 @@
...
```
KnVerey: Same here I imagine

Issue #591, opened 2019-10-16
ResourceCache prewarm


dturn:
The concurrency behavior of `DeployTask#check_initial_status` and `ResourceWatcher#sync_resources` is random. In the worst case all N threads could be servicing the same resource type. This would practically mean no parallelism since  n-1 threads would block on a mutex while the remaining thread is fetching the resources needed by that resource type. This behavior is especially painful for no-op deploys. 

A second related problem is that each resource type can fetch additional types in its sync loop. 

We can do better. We should add a `prewarm` function to the ResourceCache. A sketch of how this would look from @KnVerey :

```
cache = ResourceCache.new(@task_config)
cache.prewarm(resources)     
KubernetesDeploy::Concurrency.split_across_threads(resources) { |r| r.sync(cache) }
```

 The `prewarm` function would build a uniqe array from a new constant on resource:
`PREFETCH_KINDS_ON_SYNC = [Pod.kind, Deployment.kind, ReplicaSet.kind]`

```
def prewarm(resources)
   kinds = resources.flat_map { |r| r.PREFETCH_KINDS_ON_SYNC }.uniq
   KubernetesDeploy::Concurrency.split_across_threads(kinds) { |k| use_or_populate_cache(k) }
end
```

We're not intending `PREFETCH_KINDS_ON_SYNC` to be mandatory, if something is missed it can still be fetched in the `sync` function.  This is would simply be a performance improvement feature.

RyanBrushett:
@dturn any thoughts/opinions on what resources absolutely should fetch as part of the cache warming?
Pods will certainly be a common one but does anything else stand out?

KnVerey:
The prefetch should attempt to fetch everything that will be needed inline _for that particular sync_. Each class should tell you which kinds it will need, so that you can loop over the resources involved in the sync, de-dup the list of their classes, query each class for its list of required kinds, and then de-dup and fetch those (in parallel). To build the list for a class, you have to walk through its whole sync method. Note that this is going to break a fetch optimization we have on ReplicaSet (but it's probably worth it, performance-wise).

PR #603, opened 2019-10-24
Add resource cache prewarm

RyanBrushett:

Resolve https://github.com/Shopify/kubernetes-deploy/issues/591
Warm up the resource cache before running `resource.sync` by populating it with various kinds that we know we'll need.

**How is this accomplished?**

- Similar to the suggestion in the linked issue, `use_or_populate_cache(kind)` on each kind before the `sync` step.
- I came up with which kinds to prefetch and when by looking at each sync method. There's a chance I overlooked something so let me know.
- I tested it by assuring the method that does the warming was making kubectl calls to both a resource (deployment) and its listed prefetch kinds (in this case, pods).

**What could go wrong?**

- I might have missed something
- This doesn't improve things how we/I thought it would

### DIFF
--- a/lib/krane/kubernetes_resource/cloudsql.rb
+++ b/lib/krane/kubernetes_resource/cloudsql.rb
@@ -2,6 +2,7 @@
 module Krane
   class Cloudsql < KubernetesResource
     TIMEOUT = 10.minutes
+    SYNC_DEPENDENCIES = %w(Deployment Service)
 
     def sync(cache)
       super
--- a/lib/krane/kubernetes_resource/daemon_set.rb
+++ b/lib/krane/kubernetes_resource/daemon_set.rb
@@ -3,6 +3,7 @@ require "krane/kubernetes_resource/pod_set_base"
 module Krane
   class DaemonSet < PodSetBase
     TIMEOUT = 5.minutes
+    SYNC_DEPENDENCIES = %w(Pod)
     attr_reader :pods
 
     def sync(cache)
--- a/test/helpers/cluster_resource_discovery_helper.rb
+++ b/test/helpers/cluster_resource_discovery_helper.rb
@@ -1,0 +1,21 @@
+# frozen_string_literal: true
+module ClusterResourceDiscoveryHelper
+  def mocked_cluster_resource_discovery(response, success: true, namespaced: false)
+    Krane::Kubectl.any_instance.stubs(:run)
+      .with("api-resources", "--namespaced=#{namespaced}", attempts: 5, use_namespace: false, output: "wide")
+      .returns([response, "", stub(success?: success)])
+    Krane::ClusterResourceDiscovery.new(task_config: task_config, namespace_tags: [])
+  end
+
+  def api_versions_full_response
+    File.read(File.join(fixture_path('for_unit_tests'), "api_versions.txt"))
+  end
+
+  def api_resources_namespaced_full_response
+    File.read(File.join(fixture_path('for_unit_tests'), "api_resources_namespaced.txt"))
+  end
+
+  def api_resources_not_namespaced_full_response
+    File.read(File.join(fixture_path('for_unit_tests'), "api_resources_global.txt"))
+  end
+end
--- a/test/helpers/mock_resource.rb
+++ b/test/helpers/mock_resource.rb
@@ -1,6 +1,9 @@
 # frozen_string_literal: true
 
 MockResource = Struct.new(:id, :hits_to_complete, :status) do
+  self::SYNC_DEPENDENCIES = []
+  self::SENSITIVE_TEMPLATE_CONTENT = false
+
   def debug_message(*)
     @debug_message
   end
--- a/test/unit/krane/kubernetes_resource/deployment_test.rb
+++ b/test/unit/krane/kubernetes_resource/deployment_test.rb
@@ -283,6 +283,7 @@ class DeploymentTest < Krane::TestCase
   end
 
   def test_deploy_succeeded_not_fooled_by_stale_status_data
+    stub_kind_get("Pod")
     deployment_status = {
       "replicas" => 3,
       "updatedReplicas" => 3, # stale -- hasn't been updated since deploy
--- a/test/unit/krane/kubernetes_resource/replica_set_test.rb
+++ b/test/unit/krane/kubernetes_resource/replica_set_test.rb
@@ -23,26 +23,6 @@ class ReplicaSetTest < Krane::TestCase
     refute_predicate(rs, :deploy_failed?)
   end
 
-  def test_sync_does_not_request_pods_if_we_already_know_they_are_fine
-    should_fetch = build_rs_template(status: { "readyReplicas": 1, "observedGeneration": 2 })
-    should_not_fetch = build_rs_template(status: { "readyReplicas": 2, "observedGeneration": 2 })
-
-    rs = Krane::ReplicaSet.new(namespace: "test", context: "nope", logger: logger, definition: should_fetch)
-    stub_kind_get("ReplicaSet", items: [should_fetch])
-    stub_kind_get("Pod", items: [])
-    rs.sync(build_resource_cache)
-
-    stub_kind_get("ReplicaSet", items: [should_not_fetch])
-    rs.sync(build_resource_cache)
-  end
-
-  def test_sync_does_not_request_pods_if_desired_replicas_is_zero
-    template = build_rs_template
-    template["status"] = { "observedGeneration": 2 }
-    template["spec"]["replicas"] = 0
-    build_synced_rs(template: template) # does not stub pod get
-  end
-
   private
 
   def build_rs_template(status: {})
@@ -50,6 +30,7 @@ class ReplicaSetTest < Krane::TestCase
   end
 
   def build_synced_rs(template:)
+    stub_kind_get("Pod")
     rs = Krane::ReplicaSet.new(namespace: "test", context: "nope", logger: logger, definition: template)
     stub_kind_get("ReplicaSet", items: [template])
     rs.sync(build_resource_cache)
--- a/test/unit/resource_cache_test.rb
+++ b/test/unit/resource_cache_test.rb
@@ -86,6 +86,60 @@ class ResourceCacheTest < Krane::TestCase
     assert(all_resources.all?(&:synced?))
   end
 
+  def test_warming_the_resource_cache_makes_kubectl_calls_and_populates_cache
+    deployments = build_fake_deployments(2) # these also get pods
+    pods = build_fake_pods(2)
+
+    stub_kind_get("FakeDeployment", items: deployments.map(&:kubectl_response), times: 1)
+    stub_kind_get("FakePod", items: pods.map(&:kubectl_response), times: 1)
+    @cache.prewarm(deployments) # Fetches both Deployments and Pods
+    assert_equal(2, @cache.get_all("FakeDeployment").count)
+    assert_equal(2, @cache.get_all("FakePod").count)
+  end
+
+  def test_warming_the_resource_cache_means_no_more_kubectl_calls_after
+    deployments = build_fake_deployments(10) # these also get pods
+    pods = build_fake_pods(10)
+    resources = deployments + pods
+
+    stub_kind_get("FakeDeployment", items: deployments.map(&:kubectl_response), times: 1)
+    stub_kind_get("FakePod", items: pods.map(&:kubectl_response), times: 1)
+    @cache.prewarm(deployments) # Fetches both Deployments and Pods
+
+    # The following line should act as assurance that no additional calls are made to kubectl
+    # In other words, if the test passes with the call made below, the cache warming did its job
+    resources.each { |r| r.sync(@cache) }
+  end
+
+  def test_warming_the_resource_cache_works_for_custom_resources
+    spec = YAML.load_file(File.join(fixture_path('for_unit_tests'), 'crd_test.yml'))
+    spec = spec.merge(
+      "metadata" => {
+        "name" => "unittests.stable.example.io",
+      },
+    )
+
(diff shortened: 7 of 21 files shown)


lib/kubernetes-deploy/kubernetes_resource/cron_job.rb
```diff
@@ -1,6 +1,7 @@
 # frozen_string_literal: true
 module KubernetesDeploy
   class CronJob < KubernetesResource
+    PREFETCH_KINDS_ON_SYNC = %w(Pod)
```
dturn: why does this need Pod?

lib/kubernetes-deploy/kubernetes_resource/daemon_set.rb
```diff
@@ -1,8 +1,10 @@
 # frozen_string_literal: true
 require 'kubernetes-deploy/kubernetes_resource/pod_set_base'
+
 module KubernetesDeploy
   class DaemonSet < PodSetBase
     TIMEOUT = 5.minutes
+    PREFETCH_KINDS_ON_SYNC = %w(Pod)
```
dturn: Nodes too?

lib/kubernetes-deploy/kubernetes_resource/job.rb
```diff
@@ -1,6 +1,7 @@
 # frozen_string_literal: true
 module KubernetesDeploy
   class Job < KubernetesResource
+    PREFETCH_KINDS_ON_SYNC = %w(Pod)
```
dturn: why Pod?

lib/kubernetes-deploy/kubernetes_resource/daemon_set.rb
RyanBrushett: Yes oops!

lib/kubernetes-deploy/kubernetes_resource/cron_job.rb
RyanBrushett: My bad (for this and Job)

lib/kubernetes-deploy/runner_task.rb
```diff
@@ -122,6 +122,7 @@ def watch_pod(pod)
 
     def record_status_once(pod)
       cache = ResourceCache.new(@task_config)
+      cache.prewarm([pod])
```
KnVerey: I guess it doesn't hurt either, but prewarming won't change anything with a single resource.

test/unit/resource_cache_test.rb
```diff
@@ -78,6 +78,15 @@ def test_concurrently_syncing_huge_numbers_of_resources_makes_exactly_one_kubect
     assert(all_resources.all?(&:synced?))
   end
 
+  def test_warming_the_resource_cache_means_no_more_kubectl_calls_after
+    deployments = build_fake_deployments(10) # these also get pods
+    pods = build_fake_pods(10)
+
+    stub_kind_get("FakeDeployment", items: deployments.map(&:kubectl_response), times: 1)
+    stub_kind_get("FakePod", items: pods.map(&:kubectl_response), times: 1)
+    @cache.prewarm(deployments) # Fetches both Deployments and Pods
```
KnVerey: Don't you need to call `sync` after to prove no more calls are made? It seems to me that right now this test only proves that prewaming makes the calls, not that sync doesn't also make them.

lib/kubernetes-deploy/kubernetes_resource.rb
```diff
@@ -421,6 +422,10 @@ def global?
       @global || self.class::GLOBAL
     end
 
+    def prefetch_kinds_on_sync
```
KnVerey: Nit: it isn't really necessary to expose this, since you necessarily can get the classes from the resources, i.e.:

```ruby
      dependencies = resources.map(&:class).uniq.flat_map do |c|
        c.const_get(:SYNC_DEPENDENCIES)
      end
      kinds = (resources.map(&:kubectl_resource_type) + dependencies).compact.uniq
```

lib/kubernetes-deploy/kubernetes_resource.rb
```diff
@@ -38,6 +38,7 @@ class KubernetesResource
     LAST_APPLIED_ANNOTATION = "kubectl.kubernetes.io/last-applied-configuration"
     SENSITIVE_TEMPLATE_CONTENT = false
     SERVER_DRY_RUNNABLE = false
+    PREFETCH_KINDS_ON_SYNC = []
```
KnVerey: Nit: This doesn't match the method name you chose, and kinda sounds like it would make the sync do this automatically. Suggestion: `SYNC_DEPENDENCIES` or `PREWARM_KINDS_FOR_SYNC`.

lib/kubernetes-deploy/kubernetes_resource/daemon_set.rb
```diff
@@ -1,8 +1,10 @@
 # frozen_string_literal: true
 require 'kubernetes-deploy/kubernetes_resource/pod_set_base'
+
 module KubernetesDeploy
   class DaemonSet < PodSetBase
     TIMEOUT = 5.minutes
+    PREFETCH_KINDS_ON_SYNC = %w(Pod Node)
```
KnVerey: I actually think Node should not be prewarmed. It is only needed on the very first sync, so letting that one get delayed by an inline fetch seems ok to me.

lib/kubernetes-deploy/kubernetes_resource/deployment.rb
```diff
@@ -4,6 +4,7 @@
 module KubernetesDeploy
   class Deployment < KubernetesResource
```
KnVerey: This class and probably others have optimizations like `@latest_rs = exists? ? find_latest_rs(cache) : nil` that will no longer take effect. Please check all of them and either (a) remove them if they're useless now; or (b) remove the resource in question from the prewarming list (probably only for Node on DS, but I'm not sure).

lib/kubernetes-deploy/resource_cache.rb
```diff
@@ -36,6 +36,11 @@ def get_all(kind, selector = nil)
       []
     end
 
+    def prewarm(resources)
+      kinds = (resources.map(&:type) + resources.flat_map(&:prefetch_kinds_on_sync)).uniq
+      KubernetesDeploy::Concurrency.split_across_threads(kinds) { |kind| use_or_populate_cache(kind) }
```
KnVerey: I think we want to use the public `get_all` instead. If we don't, we need to handle the `KubectlError`.

KnVerey: Also, since we know this is all IO, we should probably crank up the concurrency instead of using even splitting like this. You could make `MAX_THREAD` configurable instead of constant, and set it to the number of kinds we have?

lib/kubernetes-deploy/runner_task.rb
RyanBrushett: Fair! I'll clean this up.

test/unit/resource_cache_test.rb
RyanBrushett: Yeah I originally had that in and it worked fine. The issue I encountered then is that deleting the prewarm line didn't make the test fail so it kind of like... didn't feel right.
Thoughts?

KnVerey: > deleting the prewarm line didn't make the test fail 

I see what you mean... if `prewarm` was just `return true` it would pass. Maybe we need two tests: this one as-is, and another proving that the pre-warming gets used by subsequent `sync` calls?

> Would that be a better option here?

No need to split it across threads (I think this is really fast with the dummy resources), but that's more or less what I was trying to suggest. 😄

lib/kubernetes-deploy/resource_cache.rb
```diff
@@ -36,6 +36,12 @@ def get_all(kind, selector = nil)
       []
     end
 
+    def prewarm(resources)
+      sync_dependencies = resources.map(&:class).flat_map { |c| c.const_get(:SYNC_DEPENDENCIES) }
```
dturn: Let's avoid the double map:
`resources.flat_map { |r| r.class.const_get(:SYNC_DEPENDENCIES) }`

RyanBrushett:
Rebase done. At a glance it looks like it went well, that wasn't too bad at all.
Thanks @dturn good shout. Done 👍 
@KnVerey ready for 👀

test/unit/krane/kubernetes_resource/deployment_test.rb
```diff
@@ -298,29 +298,6 @@ def test_validation_works_with_no_strategy_and_max_unavailable_annotation
     assert(deploy.validate_definition(kubectl))
   end
 
-  def test_deploy_succeeded_not_fooled_by_stale_status_data
```
KnVerey: This test still sounds legit. Can you instead update it to expect a pod fetch?

test/unit/krane/kubernetes_resource/replica_set_test.rb
```diff
@@ -5,44 +5,27 @@ class ReplicaSetTest < Krane::TestCase
...
 
   def test_deploy_succeeded_not_fooled_by_stale_status
+    stub_kind_get("Pod")
     template = build_rs_template(status: { "observedGeneration": 1 })
     rs = build_synced_rs(template: template)
     refute_predicate(rs, :deploy_succeeded?)
   end
 
   def test_deploy_failed_ensures_controller_has_observed_deploy
+    stub_kind_get("Pod")
```
KnVerey: This should go in `build_synced_rs`

RyanBrushett: Oh yeah good eye 👍

lib/krane/deprecated_deploy_task.rb
```diff
@@ -250,6 +250,7 @@ def deploy_has_priority_resources?(resources)
 
     def check_initial_status(resources)
       cache = ResourceCache.new(@task_config)
+      cache.prewarm(resources)
```
dturn: We'll need to add this to the GlobalDeployTask

lib/krane/kubernetes_resource/daemon_set.rb
```diff
@@ -3,6 +3,7 @@
 module Krane
   class DaemonSet < PodSetBase
     TIMEOUT = 5.minutes
+    SYNC_DEPENDENCIES = %w(Pod)
```
dturn: We're intentionally not capturing nodes here?

test/helpers/cluster_resource_discovery_helper.rb
```diff
@@ -0,0 +1,17 @@
+# frozen_string_literal: true
+module ClusterResourceDiscoveryHelper
+  def mocked_cluster_resource_discovery(response, success: true)
+    Krane::Kubectl.any_instance.stubs(:run)
+      .with("api-resources", "--namespaced=false", attempts: 5, use_namespace: false, output: "wide")
+      .returns([response, "", stub(success?: success)])
+    Krane::ClusterResourceDiscovery.new(task_config: task_config, namespace_tags: [])
+  end
+
+  def api_versions_full_response
```
dturn: Seems odd to have this in here but not something to mock its kubectl call

test/helpers/mock_resource.rb
```diff
@@ -1,6 +1,9 @@
 # frozen_string_literal: true
 
 MockResource = Struct.new(:id, :hits_to_complete, :status) do
+  ::SYNC_DEPENDENCIES = []
```
dturn: Is this just due to some strange behavior of how Struct works?

lib/krane/kubernetes_resource/daemon_set.rb
RyanBrushett: Yep, Katrina left a comment about it above but TLDR we only ever fetch it once anyway apparently so we've explicitly skipped it.

lib/krane/deprecated_deploy_task.rb
RyanBrushett: Oh yeah good shout, my bad.

test/helpers/cluster_resource_discovery_helper.rb
RyanBrushett: Straight copy pasta, I can iron this out better for sure. Will do.

test/helpers/mock_resource.rb
RyanBrushett: I guess so yep. I can tinker with it to see how many other ways I can get it to work (like `self::SYNC_DEPENDENCIES`) but just putting the constant there by itself is no dice in this case.

RyanBrushett: Hmm it didn't break the tests when I tried it this time around so sticking with `self::` 👍

tylermercier:
excited

lib/krane/kubernetes_resource/cloudsql.rb
```diff
@@ -2,6 +2,7 @@
 module Krane
   class Cloudsql < KubernetesResource
     TIMEOUT = 10.minutes
+    SYNC_DEPENDENCIES = %w(Deployment Service)
```
jessie-JNing: How do you resolve the dependencies of those types of resource?

RyanBrushett: We go through the sync loop for that resource and have a look at what ones it's fetching. In the case of cloudsql we see [Service and Deployment](https://github.com/Shopify/kubernetes-deploy/blob/master/lib/krane/kubernetes_resource/cloudsql.rb#L8-L9) so we prewarm those! 👍 
Does that answer your question?

jessie-JNing: Yes, that answers the question 😄 , I thought you looked for some k8s document

jessie-JNing: The PR looks great, :shipit:

PR #606, opened 2019-10-28
Support multiple filenames in krane deploy

dturn:

Currently the `-f` flag in for `krane deploy` only supports a single filename or directly. This appears to be an oversight from waiting for thor to handle repeated flags.  `Krane render` supports the multiple files/directories. 

**How is this accomplished?**
Change thor type to an array. 

**What could go wrong?**
We don't support csv only white space separated.

### DIFF
--- a/lib/krane/cli/deploy_command.rb
+++ b/lib/krane/cli/deploy_command.rb
@@ -13,7 +13,7 @@ module Krane
         "bindings" => { type: :array, banner: "foo=bar abc=def",
                         desc: "Expose additional variables to ERB templates (format: k1=v1 k2=v2, JSON string or file "\
                           "(JSON or YAML) path prefixed by '@')" },
-        "filenames" => { type: :string, banner: '/tmp/my-resource.yml', aliases: :f, required: true,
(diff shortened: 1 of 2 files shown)


lib/krane/cli/deploy_command.rb
```diff
@@ -13,8 +13,9 @@ class DeployCommand
         "bindings" => { type: :array, banner: "foo=bar abc=def",
                         desc: "Expose additional variables to ERB templates (format: k1=v1 k2=v2, JSON string or file "\
                           "(JSON or YAML) path prefixed by '@')" },
-        "filenames" => { type: :string, banner: '/tmp/my-resource.yml', aliases: :f, required: true,
-                         desc: "Path to file or directory that contains the configuration to apply" },
+        "filenames" => { type: :array, banner: 'config/deploy/production config/deploy/my-extra-resource.yml',
+                         aliases: :f, required: true,
+                         desc: "Directories and files that contains the configuration to apply" },
```
timothysmith0609: For consistency let's add this to `render` as well

PR #607, opened 2019-10-28
Add first version of the migration guide to Krane 1.0

dirceu:

Add a migration guide to help people migrate off of `kubernetes-deploy` and into `krane`.

This can still change before 1.0, but it might be easier to maintain this guide if we merge it and update as needed (e.g. we still need to send a PR to update flag names: https://github.com/Shopify/kubernetes-deploy/pull/602#discussion_r339247531).

### DIFF
--- a/1.0-Upgrade.md
+++ b/1.0-Upgrade.md
@@ -1,8 +1,8 @@
-# kubernetes-deploy 0.30.0 -> krane 1.0 migration guide
+# kubernetes-deploy 0.30.0 -> krane 1.0.0 migration guide
 
-**THIS MIGRATION GUIDE IS A WORK IN PROGRESS. THINGS CAN STILL CHANGE BEFORE 1.0**
+**THIS MIGRATION GUIDE IS A WORK IN PROGRESS. THINGS CAN STILL CHANGE BEFORE 1.0.0**
 
-`kubernetes-deploy` was renamed `krane` when version 1.0.0 was released. Version 1.0.0 introduced new features and [breaking changes](CHANGELOG.md). This guide will help you transition to version 1.0 as smoothly as possible.
+`kubernetes-deploy` was renamed `krane` when version 1.0.0 was released. Version 1.0.0 introduced new features and [breaking changes](CHANGELOG.md). This guide will help you transition to version 1.0.0 as smoothly as possible.
 
 **TL;DR**:
 * The command-line interface was redesigned; besides the name change, there are breaking changes in several flags.
--- a/1.0-Upgrade.md
+++ b/1.0-Upgrade.md
@@ -2,7 +2,7 @@
 
 **THIS MIGRATION GUIDE IS A WORK IN PROGRESS. THINGS CAN STILL CHANGE BEFORE 1.0.0**
 
-`kubernetes-deploy` was renamed `krane` when version 1.0.0 was released. Version 1.0.0 introduced new features and [breaking changes](CHANGELOG.md). This guide will help you transition to version 1.0.0 as smoothly as possible.
+`kubernetes-deploy` was renamed `krane` when version 1.0.0 was released. Version 1.0.0 introduced [new features and breaking changes](CHANGELOG.md). This guide will help you transition to version 1.0.0 as smoothly as possible.
 
 **TL;DR**:
 * The command-line interface was redesigned; alongside the name change there are breaking changes in several flags.
@@ -18,7 +18,7 @@ gem install krane
 
 ## Public API changes
 
-The only breaking change in the public API between version 0.31.0 and 1.0.0 is the renaming of the `KubernetesDeploy` namespace to `Krane`. Besides that, the API of the major public classes (`DeployTask`, `RenderTask`, `RunnerTask`, and `RestartTask`) didn't change at all between 0.31.0 and 1.0.0. If you're curious about this API, check the comment-based docs on these classes or the [rendered documentation at RubyGems.org](https://www.rubydoc.info/gems/kubernetes-deploy/1.0.0/KubernetesDeploy/DeployTask).
+The only breaking change in the public API between version 0.31.0 and 1.0.0 is the renaming of the `KubernetesDeploy` namespace to `Krane`. Otherwise, the APIs of the major public classes (`DeployTask`, `RenderTask`, `RunnerTask`, and `RestartTask`) have not changed between 0.31.0 and 1.0.0. If you're curious about this API, check the comment-based docs on these classes or the [rendered documentation at RubyGems.org](https://www.rubydoc.info/gems/kubernetes-deploy/1.0.0/KubernetesDeploy/DeployTask).
 
 ## Command-line interface changes
 
@@ -61,7 +61,7 @@ Old flag | New flag | Comments
 --- | --- | ---
 --deployments=LIST | --deployments=LIST |
 --max-watch-seconds=seconds | --global-timeout=300s | Changed flag name and default value to be a duration (expressed using strings like "300s" or "1h")
-[none] | --verify-result=true | Define whether it should wait for results or exit immediately after validation
+[none] | --verify-result=true | Defines whether it should wait for results or exit immediately after validation
 
 #### krane run
 
@@ -79,7 +79,7 @@ Old flag | New flag | Comments
 Old flag | New flag | Comments
 --- | --- | ---
 --bindings=BINDINGS | --bindings=BINDINGS |
---template-dir | -f, --filename | Changed to be more aligned with `kubectl apply` and our other tasks
+--template-dir | -f, --filename | Changed to be more aligned with `kubectl apply` and other krane tasks
 $REVISION | --current-sha | The environment variable REVISION was dropped in favour of an explicit flag
 
 ## Running `kubernetes-deploy` and `krane` side by side
--- a/1.0-Upgrade.md
+++ b/1.0-Upgrade.md
@@ -18,12 +18,12 @@ gem install krane
 
 ## Public API changes
 
-The only breaking change in the public API is the renaming of the `KubernetesDeploy` namespace to `Krane`. Besides that, the API of the major public classes (`DeployTask`, `RenderTask`, `RunTask`, and `RestartTask`) didn't change at all between 0.30.0 and 1.0.0. If you're curious about this API, check the comment-based docs on these classes or the [rendered documentation at RubyGems.org](https://www.rubydoc.info/gems/kubernetes-deploy/1.0.0/KubernetesDeploy/DeployTask).
+The only breaking change in the public API is the renaming of the `KubernetesDeploy` namespace to `Krane`. Besides that, the API of the major public classes (`DeployTask`, `RenderTask`, `RunnerTask`, and `RestartTask`) didn't change at all between 0.30.0 and 1.0.0. If you're curious about this API, check the comment-based docs on these classes or the [rendered documentation at RubyGems.org](https://www.rubydoc.info/gems/kubernetes-deploy/1.0.0/KubernetesDeploy/DeployTask).
 
 ## Command-line interface changes
 
 Old command | New command
---- | --- | ---
+--- | ---
 `kubernetes-deploy` | `krane deploy`
 `kubernetes-deploy -v` | `krane version`
 `kubernetes-render` | `krane render`
@@ -33,7 +33,7 @@ Old command | New command
 
 ### Flag changes
 
-The following tables provide a mapping of the previous flags we supported and their new version (if applicable).
+The following tables provide a mapping of the flags previously supported in `kubernetes-deploy` and their new version in `krane` (if applicable).
 
 #### krane deploy
 
--- a/1.0-Upgrade.md
+++ b/1.0-Upgrade.md
@@ -9,12 +9,7 @@
 * There are breaking changes in the public API (such as the renaming of the `KubernetesDeploy` namespace to `Krane`, and the change in default values for different arguments of the public interface).
 * StatsD metrics will now be generated with the `krane` prefix.
 * `kubernetes-deploy` (now `krane deploy`) / `DeployTask` can no longer deploy global (non-namespaced) resources. A new command called `krane global-deploy` and a related class called `GlobalDeployTask` were added to replace that feature.
-* If you attempt to install two gems that have conflicting executables, `gem install` will warn you but the most recently installed one will win. This means that you can run both `kubernetes-deploy` 0.31.1 and `krane` 1.0.0 side by side by doing:
-
-```bash
-gem install kubernetes-deploy
-gem install krane
-```
+* If you attempt to install two gems that have conflicting executables, `gem install` will warn you but the most recently installed one will win.
 
 ## Public API changes
 
--- a/1.0-Upgrade.md
+++ b/1.0-Upgrade.md
@@ -103,6 +103,8 @@ $ cat my-template.yml
     kind: StorageClass
     metadata:
       name: testing-storage-class
+      labels:
+        app: krane
     provisioner: kubernetes.io/no-provisioner
 $ krane global-deploy my-k8s-context -f my-template.yml --selector app=krane
 ```
--- a/1.0-Upgrade.md
+++ b/1.0-Upgrade.md
@@ -2,11 +2,11 @@
 
 **THIS MIGRATION GUIDE IS A WORK IN PROGRESS. THINGS CAN STILL CHANGE BEFORE 1.0**
 
-The `kubernetes-deploy` became `krane` in its 1.0 version, and besides the name change we also introduced new features and [breaking changes](CHANGELOG.md). This guide will help you transition to version 1.0 as smoothly as possible.
+`kubernetes-deploy` was renamed `krane` when version 1.0.0 was released. Version 1.0.0 introduced new features and [breaking changes](CHANGELOG.md). This guide will help you transition to version 1.0 as smoothly as possible.
 
 **TL;DR**:
 * The command-line interface was redesigned; besides the name change, there are breaking changes in several flags.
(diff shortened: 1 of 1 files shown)


1.0-Upgrade.md
```diff
@@ -0,0 +1,106 @@
...
+```
+
+## Public API changes
+
+The only breaking change in the public API is the renaming of the `KubernetesDeploy` namespace to `Krane`. Besides that, the API of the major public classes (`DeployTask`, `RenderTask`, `RunTask`, and `RestartTask`) didn't change at all between 0.30.0 and 1.0.0. If you're curious about this API, check the comment-based docs on these classes or the [rendered documentation at RubyGems.org](https://www.rubydoc.info/gems/kubernetes-deploy/1.0.0/KubernetesDeploy/DeployTask).
+
+## Command-line interface changes
+
+Old command | New command
+--- | --- | ---
```
randy-abson: The 3 columns is causing display issues for the table I think

1.0-Upgrade.md
```diff
@@ -0,0 +1,106 @@
...
+```bash
+gem install kubernetes-deploy -v 0.30.0
+gem install krane -v 1.0.0
+```
+
+This can help you incrementally port scripts that use the old CLI to the new one.
+
+## New task: `krane global-deploy`
+
+`krane global-deploy` (accessible through the Ruby API as `Krane::GlobalDeployTask`) can deploy global (non-namespaced) resources such as `PersistentVolume`, `Namespace`, and `CustomResourceDefinition`. Its interface is very similar to `krane deploy`. Example usage:
```
randy-abson: Should mention that `krane deploy` does not support global resources? Maybe add the same comment in the `krane deploy` as well?

1.0-Upgrade.md
```diff
@@ -0,0 +1,106 @@
...
+`kubernetes-deploy` | `krane deploy`
+`kubernetes-deploy -v` | `krane version`
+`kubernetes-render` | `krane render`
+`kubernetes-run` | `krane run`
+`kubernetes-restart` | `krane restart`
+`[kubernetes-deploy with global resources in templates]` | `krane global-deploy`
+
+### Flag changes
+
+The following tables provide a mapping of the previous flags we supported and their new version (if applicable).
```
randy-abson: Should not use language `we`. Should refer to `supported in kubernetes-deploy v.X`

1.0-Upgrade.md
dirceu: Whoops, good catch! This rendered correctly in VSCode.

1.0-Upgrade.md
```diff
@@ -0,0 +1,106 @@
+# kubernetes-deploy 0.30.0 -> krane 1.0 migration guide
+
+**THIS MIGRATION GUIDE IS A WORK IN PROGRESS. THINGS CAN STILL CHANGE BEFORE 1.0**
+
+The `kubernetes-deploy` became `krane` in its 1.0 version, and besides the name change we also introduced new features and [breaking changes](CHANGELOG.md). This guide will help you transition to version 1.0 as smoothly as possible.
```
dturn:
```suggestion
`kubernetes-deploy` was renamed `krane` when version 1.0.0 was released. Verison 1.0.0 introduced new features and [breaking changes](CHANGELOG.md). This guide will help you transition to version 1.0 as smoothly as possible.
```

1.0-Upgrade.md
```diff
@@ -0,0 +1,106 @@
...
+
+**TL;DR**:
+* The command-line interface was redesigned; besides the name change, there are breaking changes in several flags.
+* The only breaking change in the public API is the renaming of the `KubernetesDeploy` namespace to `Krane`.
```
dturn: This isn't true, the `template_dir`  is for sure gone, default values have changed too.

1.0-Upgrade.md
```diff
@@ -0,0 +1,106 @@
...
+Old command | New command
+--- | ---
+`kubernetes-deploy` | `krane deploy`
+`kubernetes-deploy -v` | `krane version`
+`kubernetes-render` | `krane render`
+`kubernetes-run` | `krane run`
+`kubernetes-restart` | `krane restart`
+`[kubernetes-deploy with global resources in templates]` | `krane global-deploy`
+
+### Flag changes
```
dturn: We should highlight that you can't repeat flags instead you provide a list of space separated options

1.0-Upgrade.md
```diff
@@ -0,0 +1,106 @@
...
+--no-prune | --prune=true |
+--template-dir | -f, --filename | Makes all our commands accept this arg, which is now required for the deploy task
+--verbose-log-prefix | --verbose-log-prefix |
+--max-watch-seconds=seconds | --global-timeout=300s | Changed flag name and default value
+--selector | --selector |
+-h, --help | -h, --help |
+-v, --version | [none] | Replaced with `krane version`
+$ENVIRONMENT | [none] | Dropped in favour of `-f`
+$REVISION | --current-sha | The environment variable REVISION was dropped in favour of an explicit flag
+[none] | --render-erb | **Important:** the new CLI doesn't render ERB by default
```
dturn: We should make sure we're committed to opt-in vs totally removing it.

1.0-Upgrade.md
```diff
@@ -0,0 +1,106 @@
...
+--template-dir | -f, --filename | Changed to be more aligned with `kubectl apply` and our other tasks
+$REVISION | --current-sha | The environment variable REVISION was dropped in favour of an explicit flag
+
+## Running `kubernetes-deploy` and `krane` side by side
+
+If you attempt to install two gems that have conflicting executables (as is the case here), `gem install` will warn you but the most recently installed one will win. This means that you can run both `kubernetes-deploy` 0.30.0 and `krane` 1.0.0 side by side by doing:
+
+```bash
+gem install kubernetes-deploy -v 0.30.0
+gem install krane -v 1.0.0
```
dturn: This will prompt for user input unless they add `-f `

JennaBlack:
Is there a difference between `version 1.0` and `version 1.0.0` (1 decimal vs 2)  it seems to be used interchangeably throughout this doc but not really consistent

dirceu:
There shouldn't be a difference, but in any case we should make it consistent. Good shout!

1.0-Upgrade.md
```diff
@@ -0,0 +1,108 @@
...
+`kubernetes-render` | `krane render`
+`kubernetes-run` | `krane run`
+`kubernetes-restart` | `krane restart`
+`[kubernetes-deploy with global resources in templates]` | `krane global-deploy`
+
+### Flag changes
+
+The following tables provide a mapping of the flags previously supported in `kubernetes-deploy` and their new version in `krane` (if applicable).
+
+Important: you can't repeat flags. If you need to provide multiple arguments for a flag, use a space-separated list (e.g. `-f file1.txt file2.txt`) unless specified otherwise.
```
dturn: file1.yml (we don't actually allow other sufixes from what I remember)

1.0-Upgrade.md
```diff
@@ -0,0 +1,108 @@
...
+* If you attempt to install two gems that have conflicting executables, `gem install` will warn you but the most recently installed one will win. This means that you can run both `kubernetes-deploy` 0.30.0 and `krane` 1.0.0 side by side by doing:
+
+```bash
+gem install kubernetes-deploy
+gem install krane
+```
+
+## Public API changes
+
+The only breaking change in the public API is the renaming of the `KubernetesDeploy` namespace to `Krane`. Besides that, the API of the major public classes (`DeployTask`, `RenderTask`, `RunnerTask`, and `RestartTask`) didn't change at all between 0.30.0 and 1.0.0. If you're curious about this API, check the comment-based docs on these classes or the [rendered documentation at RubyGems.org](https://www.rubydoc.info/gems/kubernetes-deploy/1.0.0/KubernetesDeploy/DeployTask).
```
dturn: I don't think this is true, args and defaults changed. We did also change functionality by  removing the ability to deploy global resources)

1.0-Upgrade.md
```diff
@@ -0,0 +1,108 @@
...
+
+Old flag | New flag | Comments
+--- | --- | ---
+--bindings=BINDINGS | --bindings=BINDINGS |
+--skip-wait | --verify-result=true |
+--allow-protected-ns | --protected-namespaces=default,kube-system,kube-public | Added the ability to specify which namespaces are protected
+--no-prune | --prune=true |
+--template-dir | -f, --filename | Makes all our commands accept this arg, which is now required for the deploy task
+--verbose-log-prefix | --verbose-log-prefix |
+--max-watch-seconds=seconds | --global-timeout=300s | Changed flag name and default value
```
dturn: I think we should highlight we require a duration, e.g. `300s` not `300`

1.0-Upgrade.md
dirceu: Did args really change between `kubernetes-deploy` 0.30.0 and `krane 1.0.0`? I mean, their names will change but the default values are already different now aren't they?

dturn: I think the goal of the guide it help someone move from  `KubernetesDeploy::XTask` - > `Krane::XTask` as well as `exe/kubernetes-x` -> `exe/krane x` . An example of a default value change would be erb rendering.

dirceu: That's an example of a _flag_ default value change; the default value in the code is still `false`:

https://github.com/Shopify/kubernetes-deploy/blob/6c6f930cecf3fda21b6fb57cca4f7ab39455d0c9/lib/krane/deprecated_deploy_task.rb#L128-L130

My point is that people moving from `KubernetesDeploy::DeployTask` to `Krane::DeployTask` will not actually have to change much **if they are on a recent version**. For instance, global deploys were disabled by default in 0.30.0.

1.0-Upgrade.md
```diff
@@ -0,0 +1,108 @@
+# kubernetes-deploy 0.31.0 -> krane 1.0.0 migration guide
+
+**THIS MIGRATION GUIDE IS A WORK IN PROGRESS. THINGS CAN STILL CHANGE BEFORE 1.0.0**
+
+`kubernetes-deploy` was renamed `krane` when version 1.0.0 was released. Version 1.0.0 introduced new features and [breaking changes](CHANGELOG.md). This guide will help you transition to version 1.0.0 as smoothly as possible.
+
+**TL;DR**:
+* The command-line interface was redesigned; besides the name change, there are breaking changes in several flags.
```
lei-lo: ⛏ `s/besides/alongside` and remove the comma

1.0-Upgrade.md
```diff
@@ -0,0 +1,108 @@
...
+* If you attempt to install two gems that have conflicting executables, `gem install` will warn you but the most recently installed one will win. This means that you can run both `kubernetes-deploy` 0.31.0 and `krane` 1.0.0 side by side by doing:
+
+```bash
+gem install kubernetes-deploy
+gem install krane
+```
+
+## Public API changes
+
+The only breaking change in the public API between version 0.31.0 and 1.0.0 is the renaming of the `KubernetesDeploy` namespace to `Krane`. Besides that, the API of the major public classes (`DeployTask`, `RenderTask`, `RunnerTask`, and `RestartTask`) didn't change at all between 0.31.0 and 1.0.0. If you're curious about this API, check the comment-based docs on these classes or the [rendered documentation at RubyGems.org](https://www.rubydoc.info/gems/kubernetes-deploy/1.0.0/KubernetesDeploy/DeployTask).
```
lei-lo: ⛏ This sounds smoother: `Otherwise, the APIs ...[keep the same]... have not changed between 0.31.0 and 1.0.0`

I didn't immediately understand what `comment-based docs` were, so maybe something more descriptive  like `check out the comments on the classes themselves` would be more clear? It could also just be me so feel free to ignore.

1.0-Upgrade.md
```diff
@@ -0,0 +1,108 @@
...
+Important: you can't repeat flags. If you need to provide multiple arguments for a flag, use a space-separated list (e.g. `-f file1.yml file2.yml`) unless specified otherwise.
+
+#### krane deploy
+
+Old flag | New flag | Comments
+--- | --- | ---
+--bindings=BINDINGS | --bindings=BINDINGS |
+--skip-wait | --verify-result=true |
+--allow-protected-ns | --protected-namespaces=default,kube-system,kube-public | Added the ability to specify which namespaces are protected
+--no-prune | --prune=true |
```
lei-lo: Shouldn't `--no-prune` be `--prune=false`? Or am I missing something 🤔

1.0-Upgrade.md
```diff
@@ -0,0 +1,108 @@
...
+--template-dir | -f, --filename | Makes all our commands accept this arg, which is now required for the deploy task
```
lei-lo: Can we expand `arg` out to `argument`? Also, I'd use `all krane commands`, `our` sounds personal.

1.0-Upgrade.md
```diff
@@ -0,0 +1,108 @@
...
+$REVISION | --current-sha | The environment variable REVISION was dropped in favour of an explicit flag
+[none] | --render-erb | **Important:** the new CLI doesn't render ERB by default
+
+#### krane restart
+
+Old flag | New flag | Comments
+--- | --- | ---
+--deployments=LIST | --deployments=LIST |
+--max-watch-seconds=seconds | --global-timeout=300s | Changed flag name and default value to be a duration (expressed using strings like "300s" or "1h")
+[none] | --verify-result=true | Define whether it should wait for results or exit immediately after validation
```
lei-lo: `s/Define/Defines` because "the flag defines something"

1.0-Upgrade.md
```diff
@@ -0,0 +1,108 @@
...
+--template | --template | Changed to be required
+[it is positional now] | --arguments | Optional flag, as `command` or the template might already specify the required arguments
+--env-vars=ENV_VARS | --env-vars=ENV_VARS |
+
+#### krane render
+
+Old flag | New flag | Comments
+--- | --- | ---
+--bindings=BINDINGS | --bindings=BINDINGS |
+--template-dir | -f, --filename | Changed to be more aligned with `kubectl apply` and our other tasks
```
lei-lo: Same comment about `our`, replace with `krane` or `the`.

1.0-Upgrade.md
dirceu: No, it was called like that: https://github.com/Shopify/kubernetes-deploy/blob/6c6f930cecf3fda21b6fb57cca4f7ab39455d0c9/exe/kubernetes-deploy#L30

1.0-Upgrade.md
dturn: So it's true right now because we've intentionally forced ourselves into backwards compatibility. But, once we cut 1.0.0-pre.1 it wont be true. Are you thinking the release of 1.0.0-pre.1 will change the wording? Or the first PR to actually make a change will be responsible for updating this?

dirceu: > Or the first PR to actually make a change will be responsible for updating this?

That's what I'm thinking. I plan to have this merged and then rebase https://github.com/Shopify/kubernetes-deploy/pull/619 and have that PR document particular changes. It might be easier to keep track of all the changes we're doing that way. Does that make sense?

dturn: How about: `The only breaking change in the public API, so far, is` . I don't want someone to read this today and see the 1.0 and thing ok cool. Not realizing this is a living doc.

dirceu: Makes sense. Updated.

1.0-Upgrade.md
```diff
@@ -0,0 +1,108 @@
...
+## New task: `krane global-deploy`
+
+`krane global-deploy` (accessible through the Ruby API as `Krane::GlobalDeployTask`) can deploy global (non-namespaced) resources such as `PersistentVolume`, `Namespace`, and `CustomResourceDefinition`. Its interface is very similar to `krane deploy`. Example usage:
+
+```bash
+$ cat my-template.yml
+    apiVersion: storage.k8s.io/v1
+    kind: StorageClass
+    metadata:
+      name: testing-storage-class
```
dturn: This actually needs a 
```
label: 
  app: krane
```

or else the command will fail

1.0-Upgrade.md
```diff
@@ -0,0 +1,110 @@
...
+`kubernetes-deploy` was renamed `krane` when version 1.0.0 was released. Version 1.0.0 introduced [new features and breaking changes](CHANGELOG.md). This guide will help you transition to version 1.0.0 as smoothly as possible.
+
+**TL;DR**:
+* The command-line interface was redesigned; alongside the name change there are breaking changes in several flags.
+* There are breaking changes in the public API (such as the renaming of the `KubernetesDeploy` namespace to `Krane`, and the change in default values for different arguments of the public interface).
+* StatsD metrics will now be generated with the `krane` prefix.
+* `kubernetes-deploy` (now `krane deploy`) / `DeployTask` can no longer deploy global (non-namespaced) resources. A new command called `krane global-deploy` and a related class called `GlobalDeployTask` were added to replace that feature.
+* If you attempt to install two gems that have conflicting executables, `gem install` will warn you but the most recently installed one will win. This means that you can run both `kubernetes-deploy` 0.31.1 and `krane` 1.0.0 side by side by doing:
+
+```bash
```
jessie-JNing: Personally, I feel there's no need to provide the bash command in TL;DR, and the gem command here has no version arg, which is inconsistent with those [here]( https://github.com/Shopify/kubernetes-deploy/pull/607/files#diff-18e872f2ce48f8abe3267c4bd07d46d7R89), might confuse users

jessie-JNing:
Just a nit. It's a great migration guide, thanks @dirceu

PR #608, opened 2019-10-29
Allow Krane to deploy resources which use generateName

RyanBrushett:

~Fix bug~ Allow Krane to deploy resources which use generateName.
In short, Kubernetes doesn't support using `apply` on templates that contain `generateName`, One must specify a `name`. [See here](https://kubernetes.io/docs/reference/generated/kubectl/kubectl-commands#apply).

**How is this accomplished?**

Support `create` when `name` is blank and `generateName` is present!
This involves some hackery though, notably setting `@file` to `nil` and setting `@name` explicitly.
Had to also update our `dry-run` validations to use `create` rather than `apply`.

**What could go wrong?**

Could break deploys in general maybe?

Feedback would be greatly appreciated!

EDIT I had this written as if this fixes a bug but this has been the behaviour forever so maybe that's not fair.

### DIFF
--- a/lib/krane/deprecated_deploy_task.rb
+++ b/lib/krane/deprecated_deploy_task.rb
@@ -414,27 +414,26 @@ module Krane
       pruneable_types = prune_whitelist.map { |t| t.split("/").last }
       applyables += individuals.select { |r| pruneable_types.include?(r.type) }
 
-      individuals.each do |r|
-        r.deploy_started_at = Time.now.utc
-        case r.deploy_method
+      individuals.each do |individual_resource|
+        individual_resource.deploy_started_at = Time.now.utc
+
+        case individual_resource.deploy_method
+        when :create
+          err, status = create_resource(individual_resource)
         when :replace
-          _, _, replace_st = kubectl.run("replace", "-f", r.file_path, log_failure: false)
+          err, status = replace_or_create_resource(individual_resource)
         when :replace_force
-          _, _, replace_st = kubectl.run("replace", "--force", "--cascade", "-f", r.file_path,
-            log_failure: false)
+          err, status = replace_or_create_resource(individual_resource, force: true)
         else
           # Fail Fast! This is a programmer mistake.
-          raise ArgumentError, "Unexpected deploy method! (#{r.deploy_method.inspect})"
+          raise ArgumentError, "Unexpected deploy method! (#{individual_resource.deploy_method.inspect})"
         end
 
-        next if replace_st.success?
-        # it doesn't exist so we can't replace it
-        _, err, create_st = kubectl.run("create", "-f", r.file_path, log_failure: false)
+        next if status.success?
 
-        next if create_st.success?
         raise FatalDeploymentError, <<~MSG
-          Failed to replace or create resource: #{r.id}
-          #{err}
+          Failed to replace or create resource: #{individual_resource.id}
+          #{individual_resource.sensitive_template_content? ? '<suppressed sensitive output>' : err}
         MSG
       end
 
@@ -445,6 +444,35 @@ module Krane
           timeout: @max_watch_seconds, task_config: @task_config, sha: @current_sha)
         watcher.run(record_summary: record_summary)
       end
+    end
+
+    def replace_or_create_resource(resource, force: false)
+      args = if force
+        ["replace", "--force", "--cascade", "-f", resource.file_path]
+      else
+        ["replace", "-f", resource.file_path]
+      end
+
+      _, err, status = kubectl.run(*args, log_failure: false, output_is_sensitive: resource.sensitive_template_content?,
+        raise_if_not_found: true)
+
+      [err, status]
+    rescue Krane::Kubectl::ResourceNotFoundError
+      # it doesn't exist so we can't replace it, we try to create it
+      create_resource(resource)
+    end
+
+    def create_resource(resource)
+      out, err, status = kubectl.run("create", "-f", resource.file_path, log_failure: false, output: 'json',
+        output_is_sensitive: resource.sensitive_template_content?)
+
+      # For resources that rely on a generateName attribute, we get the `name` from the result of the call to `create`
+      # We must explicitly set this name value so that the `apply` step for pruning can run successfully
+      if status.success? && resource.uses_generate_name?
+        resource.use_generated_name(JSON.parse(out))
+      end
+
+      [err, status]
     end
 
     def deploy_all_resources(resources, prune: false, verify:, record_summary: true)
--- a/lib/krane/kubernetes_resource.rb
+++ b/lib/krane/kubernetes_resource.rb
@@ -87,8 +87,9 @@ module Krane
           raise InvalidTemplateError.new("Template is missing required field 'kind'", content: debug_content)
         end
 
-        if definition.dig('metadata', 'name').blank?
-          raise InvalidTemplateError.new("Template is missing required field 'metadata.name'", content: debug_content)
+        if definition.dig('metadata', 'name').blank? && definition.dig('metadata', 'generateName').blank?
+          raise InvalidTemplateError.new("Template must specify one of 'metadata.name' or 'metadata.generateName'",
+            content: debug_content)
         end
       end
     end
@@ -112,7 +113,7 @@ module Krane
 
     def initialize(namespace:, context:, definition:, logger:, statsd_tags: [])
       # subclasses must also set these if they define their own initializer
-      @name = definition.dig("metadata", "name").to_s
+      @name = (definition.dig("metadata", "name") || definition.dig("metadata", "generateName")).to_s
       @optional_statsd_tags = statsd_tags
       @namespace = namespace
       @context = context
@@ -235,9 +236,13 @@ module Krane
       !deploy_succeeded? && !deploy_failed? && (Time.now.utc - @deploy_started_at > timeout)
     end
 
-    # Expected values: :apply, :replace, :replace_force
+    # Expected values: :apply, :create, :replace, :replace_force
     def deploy_method
-      :apply
+      if @definition.dig("metadata", "name").blank? && uses_generate_name?
+        :create
+      else
+        :apply
+      end
     end
 
     def sync_debug_info(kubectl)
@@ -350,11 +355,27 @@ module Krane
     end
 
     def server_dry_runnable_resource?
-      self.class::SERVER_DRY_RUNNABLE
+      # generateName and server-side dry run are incompatible because the former only works with `create`
+      # and the latter only works with `apply`
+      self.class::SERVER_DRY_RUNNABLE && !uses_generate_name?
+    end
+
+    def uses_generate_name?
+      @definition.dig('metadata', 'generateName').present?
     end
 
     def server_dry_run_validated?
       @server_dry_run_validated
+    end
+
+    # If a resource uses generateName, we don't know the full name of the resource until it's deployed to the cluster.
+    # In this case, we need to update our local definition with the realized name in order to accurately track the
+    # resource during deploy
+    def use_generated_name(instance_data)
+      @name = instance_data.dig('metadata', 'name')
+      @definition['metadata']['name'] = @name
+      @definition['metadata'].delete('generateName')
+      @file = create_definition_tempfile
     end
 
     class Event
@@ -480,13 +501,13 @@ module Krane
     def validate_spec_with_kubectl(kubectl)
       err = ""
       if kubectl.server_dry_run_enabled? && server_dry_runnable_resource?
-        _, err, st = validate_with_dry_run_option(kubectl, "--server-dry-run")
+        _, err, st = validate_with_server_side_dry_run(kubectl)
         @server_dry_run_validated = st.success?
         return true if st.success?
       end
 
       if err.empty? || err.match(SERVER_DRY_RUN_DISABLED_ERROR)
-        _, err, st = validate_with_dry_run_option(kubectl, "--dry-run")
+        _, err, st = validate_with_local_dry_run(kubectl)
       end
 
       return true if st.success?
@@ -497,10 +518,21 @@ module Krane
       end
     end
 
-    def validate_with_dry_run_option(kubectl, dry_run_option)
-      command = ["apply", "-f", file_path, dry_run_option, "--output=name"]
+    # Server side dry run is only supported on apply
+    def validate_with_server_side_dry_run(kubectl)
+      command = ["apply", "-f", file_path, "--server-dry-run", "--output=name"]
       kubectl.run(*command, log_failure: false, output_is_sensitive: sensitive_template_content?,
-                               retry_whitelist: [:client_timeout], attempts: 3)
+        retry_whitelist: [:client_timeout], attempts: 3)
+    end
+
+    # Local dry run is supported on only create and apply
+    # If the deploy method is create, validating with apply will fail
+    # If the resource template uses generateName, validating with apply will fail
+    def validate_with_local_dry_run(kubectl)
+      verb = deploy_method == :apply ? "apply" : "create"
+      command = [verb, "-f", file_path, "--dry-run", "--output=name"]
+      kubectl.run(*command, log_failure: false, output_is_sensitive: sensitive_template_content?,
+        retry_whitelist: [:client_timeout], attempts: 3)
     end
 
     def labels
--- a/lib/krane/kubernetes_resource/custom_resource_definition.rb
+++ b/lib/krane/kubernetes_resource/custom_resource_definition.rb
@@ -46,10 +46,6 @@ module Krane
       @definition.dig("spec", "names", "kind")
     end
 
-    def name
-      @definition.dig("metadata", "name")
-    end
-
     def prunable?
       prunable = krane_annotation_value("prunable")
       prunable == "true"
--- a/test/integration/krane_deploy_test.rb
+++ b/test/integration/krane_deploy_test.rb
@@ -1015,7 +1015,7 @@ unknown field \"myKey\" in io.k8s.apimachinery.pkg.apis.meta.v1.ObjectMeta",
 
     assert_logs_match_all([
       "Result: FAILURE",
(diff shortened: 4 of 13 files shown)


lib/krane/deprecated_deploy_task.rb
```diff
@@ -447,6 +438,31 @@ def deploy_resources(resources, prune: false, verify:, record_summary: true)
...
+    def create_resource!(resource)
+      out, err, status = kubectl.run("create", "-f", resource.file_path, log_failure: false, output: 'json')
+
+      resource.use_generate_name(out) if resource.deploy_method == :create
+      return if status.success?
+
+      raise FatalDeploymentError, <<~MSG
+        Failed to replace or create resource: #{resource.id}
+        #{err}
+      MSG
```
timothysmith0609: I feel like the caller should be responsible for catching this error (e.g. in`deploy_resources`)

lib/krane/kubernetes_resource.rb
```diff
@@ -87,7 +87,7 @@ def validate_definition_essentials(definition)
           raise InvalidTemplateError.new("Template is missing required field 'kind'", content: debug_content)
         end
 
-        if definition.dig('metadata', 'name').blank?
+        if definition.dig('metadata', 'name').blank? && definition.dig('metadata', 'generateName').blank?
           raise InvalidTemplateError.new("Template is missing required field 'metadata.name'", content: debug_content)
```
timothysmith0609: This error message is now out of date, it should be something like:
`Template has neither metadata.name nor metadata.generateName, but must specify exactly 1`

Also, what happens when _both_ name and generateName are specified: that's also an error case, no?

lib/krane/deprecated_deploy_task.rb
```diff
@@ -447,6 +438,31 @@ def deploy_resources(resources, prune: false, verify:, record_summary: true)
...
+      _, _, status = kubectl.run(args, log_failure: false)
+      return if status.success?
+      # it doesn't exist so we can't replace it, we try to create it
+      create_resource!(resource)
+    end
+
+    def create_resource!(resource)
+      out, err, status = kubectl.run("create", "-f", resource.file_path, log_failure: false, output: 'json')
+
+      resource.use_generate_name(out) if resource.deploy_method == :create
```
timothysmith0609: This should have an explanatory comment, it's hard to understand what/why it's doing this without context

lib/krane/kubernetes_resource.rb
RyanBrushett: Yes that's an error 👍 Good call.

lib/krane/deprecated_deploy_task.rb
RyanBrushett: Can do!

lib/krane/kubernetes_resource.rb
```diff
@@ -237,7 +238,11 @@ def deploy_timed_out?
 
     # Expected values: :apply, :replace, :replace_force
     def deploy_method
-      :apply
+      if @definition.dig("metadata", "name").blank? && @definition.dig("metadata", "generateName").present?
```
dturn: Can both be set?

lib/krane/kubernetes_resource.rb
```diff
@@ -357,6 +362,14 @@ def server_dry_run_validated?
       @server_dry_run_validated
     end
 
+    def use_generate_name(json)
```
dturn: should this function take the name as a string?

lib/krane/kubernetes_resource.rb
```diff
@@ -498,7 +511,7 @@ def validate_spec_with_kubectl(kubectl)
     end
 
     def validate_with_dry_run_option(kubectl, dry_run_option)
-      command = ["apply", "-f", file_path, dry_run_option, "--output=name"]
+      command = ["create", "-f", file_path, dry_run_option, "--output=name"]
```
dturn: Can we use the result of `deploy_method` here?

lib/krane/deprecated_deploy_task.rb
```diff
@@ -447,6 +445,32 @@ def deploy_resources(resources, prune: false, verify:, record_summary: true)
...
+      create_resource(resource)
+    end
+
+    def create_resource(resource)
+      out, err, status = kubectl.run("create", "-f", resource.file_path, log_failure: false, output: 'json')
+
+      # For resources that have a deploy_method of :create, we get the `name` from the result of the call to `create`
+      # We must explicitly set this name value so that the `apply` step for pruning can run successfully
+      if status.success? && resource.deploy_method == :create
+        resource.use_generate_name(out)
```
KnVerey: nit `use_generated_name` (generateName is the name of the field, and we want the name the server already generated from it, not the value of that field)

lib/krane/kubernetes_resource.rb
KnVerey: Unfortunately I think we need to switch these validation methods around a bit more, because `create` doesn't accept the `--server-dry-run` feature, at least as of 1.12 (that's a possible value of `dry_run_option` here). Adding a raw `Secret` using `generateName` to your new fixture should showcase the problem.

lib/krane/kubernetes_resource/custom_resource_definition.rb
```diff
@@ -47,7 +47,7 @@ def kind
     end
 
     def name
-      @definition.dig("metadata", "name")
+      @definition.dig("metadata", "name") || @definition.dig("metadata", "generateName")
```
KnVerey: Hm, I wonder why we have this as a method in this one class only, and why the method isn't just an accessor for `@name`. It's not like this name source is unusual. Can you try removing this? The superclass already has `attr_reader :name`

test/integration/krane_deploy_test.rb
```diff
@@ -1687,6 +1687,15 @@ def test_deploy_globals_without_allow_globals_fails
     ], in_order: true)
   end
 
+  def test_deploy_resource_using_generate_name_succeeds
+    assert_deploy_success(deploy_fixtures("generateName"))
+    pod_name_matcher = /pod-using-generate-name-[a-z0-9]{5}/
+    pods = kubeclient.get_pods(namespace: @namespace)
+    assert_equal(1, pods.count)
+    assert_match(pod_name_matcher, pods.first.dig('metadata', 'name')) # the name contains a generated ID
+    assert_logs_match_all([%r{Pod/pod-using-generate-name-[a-z0-9]{5}\s+Succeeded}])
```
KnVerey: Are there a few more times where we log the name earlier in the logs? Let's assert on more of them, since we're changing the value of the name field midway into the deploy.

test/unit/krane/ejson_secret_provisioner_test.rb
```diff
@@ -113,7 +113,7 @@ def stub_dry_run_validation_request
   end
 
   def stub_server_dry_run_validation_request
-    stub_kubectl_response("apply", "-f", anything, "--server-dry-run", "--output=name",
+    stub_kubectl_response("create", "-f", anything, "--server-dry-run", "--output=name",
```
KnVerey: This is the call that won't actually work 😞

lib/krane/kubernetes_resource.rb
KnVerey: > GenerateName is an optional prefix, used by the server, to generate a unique name ONLY IF the Name field has not been provided. 

Tested it out, and if you have both, `generateName` is ignored.

lib/krane/kubernetes_resource.rb
KnVerey: Not directly--`replace_force` isn't a real verb, and more importantly it looks like `replace` doesn't have a dry run flag at all.

lib/krane/deprecated_deploy_task.rb
```diff
@@ -404,8 +404,8 @@ def deploy_resources(resources, prune: false, verify:, record_summary: true)
           @logger.info("- #{r.id} (#{r.pretty_timeout_type})")
         end
       else
-        resource = resources.first
-        @logger.info("Deploying #{resource.id} (#{resource.pretty_timeout_type})")
+        r = resources.first
```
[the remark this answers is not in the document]
RyanBrushett: It ends up being a rubocop warning for the variable getting shadowed later hahah. I'll pick better names 👍

lib/krane/kubernetes_resource.rb
RyanBrushett: Nope

test/unit/krane/ejson_secret_provisioner_test.rb
RyanBrushett: Noted OK

lib/krane/kubernetes_resource/custom_resource_definition.rb
RyanBrushett: I tried removing it to see what would happen. There are other tests failing so I'm actually not sure. It didn't cause any _extra_ tests to fail so maybe you're right, it can 🔥. I'll leave it for now and come back to it when I'm able to get the rest of my tests passing.

RyanBrushett:
That codecov data is a bit off, I didn't change most of those classes 😅 Will keep an eye and try to get that greener. Maybe a webhook issue?

lib/krane/kubernetes_resource.rb
```diff
@@ -497,10 +509,21 @@ def validate_spec_with_kubectl(kubectl)
...
+      command = ["apply", "-f", file_path, "--server-dry-run", "--output=name"]
+      kubectl.run(*command, log_failure: false, output_is_sensitive: sensitive_template_content?,
+        retry_whitelist: [:client_timeout], attempts: 3)
+    end
+
+    # Local dry run is supported on only create and apply
+    # If the deploy method is create, validating with apply will fail
+    # If the resource template uses generateName, validating with apply will fail
+    def validate_with_local_dry_run(kubectl)
+      verb = deploy_method == :apply ? "apply" : "create"
```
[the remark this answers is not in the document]
RyanBrushett: From testing, we can't validate using `replace` + `generateName` either. So I switched it up to `create` since that supports `generateName` and works.

test/unit/krane/ejson_secret_provisioner_test.rb
```diff
@@ -101,8 +101,8 @@ def test_run_with_selector_does_not_raise_exception
 
   private
 
-  def stub_dry_run_validation_request
-    stub_kubectl_response("apply", "-f", anything, "--dry-run", "--output=name",
+  def stub_dry_run_validation_request(verb: "apply")
```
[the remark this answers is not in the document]
RyanBrushett: Good eye, no. 🔥

lib/krane/deprecated_deploy_task.rb
```diff
@@ -447,6 +446,33 @@ def deploy_resources(resources, prune: false, verify:, record_summary: true)
...
 
+    def replace_or_create_resource(resource, force: false)
+      args = if force
+        ["replace", "--force", "--cascade", "-f", resource.file_path]
+      else
+        ["replace", "-f", resource.file_path]
+      end
+
+      _, _, status = kubectl.run(*args, log_failure: false)
+      return [nil, status] if status.success?
```
timothysmith0609: Do we need to check if this actually a `Not Found` error, as opposed to, say, a timeout?

lib/krane/kubernetes_resource.rb
```diff
@@ -350,13 +355,20 @@ def sensitive_template_content?
...
 
     def server_dry_run_validated?
       @server_dry_run_validated
     end
 
+    def use_generated_name(instance_data)
+      @name = instance_data.dig('metadata', 'name')
+      @definition['metadata']['name'] = @name
+      @definition['metadata'].delete('generateName')
+      @file = nil
```
timothysmith0609: Is this to force a new write of the updated definition on subsequent calls to file_path? If so, why not `@file = create_definition_tempfile` right here inline?

lib/krane/deprecated_deploy_task.rb
KnVerey: We weren't doing that before, but that's a great idea. FYI there's a `raise_if_not_found` option on `run` that could be helpful for that (specifically, it raises `Krane::Kubectl::ResourceNotFoundError`).

lib/krane/deprecated_deploy_task.rb
```diff
@@ -447,6 +446,33 @@ def deploy_resources(resources, prune: false, verify:, record_summary: true)
...
     end
 
+    def replace_or_create_resource(resource, force: false)
+      args = if force
+        ["replace", "--force", "--cascade", "-f", resource.file_path]
+      else
+        ["replace", "-f", resource.file_path]
+      end
+
+      _, _, status = kubectl.run(*args, log_failure: false)
```
KnVerey: We should also pass `output_is_sensitive: resource.sensitive_template_content?` in this version (we didn't before, but that's great fix you've made below!)

lib/krane/deprecated_deploy_task.rb
```diff
@@ -447,6 +446,33 @@ def deploy_resources(resources, prune: false, verify:, record_summary: true)
...
+      create_resource(resource)
+    end
+
+    def create_resource(resource)
+      out, err, status = kubectl.run("create", "-f", resource.file_path, log_failure: false, output: 'json',
+        output_is_sensitive: resource.sensitive_template_content?)
+
+      # For resources that rely on a generateName attribute, we get the `name` from the result of the call to `create`
+      # We must explicitly set this name value so that the `apply` step for pruning can run successfully
+      if status.success? && resource.deploy_method != :apply
```
KnVerey: If we're here, don't we already know `resource.deploy_method != :apply`? This will make us set the generated name in a lot of cases we don't need to. Why not directly check whether the resource used generateName? Two options come to mind:
A - expose a `uses_generate_name?` method on the resource that looks at `@definition`
B - parse the JSON here and dig for `generateName`. Then also dig out the real name and pass that in instead of the full JSON.

B probably looks nicer and reduces the API surface of resources... but then again it would make us parse JSON when we don't need to. So I'm inclined to go with A.

lib/krane/kubernetes_resource.rb
```diff
@@ -237,7 +238,11 @@ def deploy_timed_out?
 
     # Expected values: :apply, :replace, :replace_force
```
KnVerey: nit: update this comment

lib/krane/kubernetes_resource.rb
```diff
@@ -350,13 +355,20 @@ def sensitive_template_content?
     end
 
     def server_dry_runnable_resource?
-      self.class::SERVER_DRY_RUNNABLE
+      self.class::SERVER_DRY_RUNNABLE && @definition.dig('metadata', 'generateName').blank?
```
KnVerey:
```suggestion
      # generateName and server-side dry run are incompatible because the former only works with `create` and the latter only works with `apply`
      self.class::SERVER_DRY_RUNNABLE && @definition.dig('metadata', 'generateName').blank?
```

lib/krane/kubernetes_resource.rb
KnVerey: Good idea, let's do that

lib/krane/kubernetes_resource.rb
KnVerey: Sidenote: if you make a `uses_generate_name?` method, this (and maybe other places?) could also use it

lib/krane/deprecated_deploy_task.rb
RyanBrushett: Good eye, I missed this here.

lib/krane/deprecated_deploy_task.rb
RyanBrushett: Is what I've done in my most recent commit what you had in mind?

lib/krane/kubernetes_resource.rb
```diff
@@ -350,13 +355,26 @@ def sensitive_template_content?
...
+      @definition.dig('metadata', 'generateName').present?
     end
 
     def server_dry_run_validated?
       @server_dry_run_validated
     end
 
+    def use_generated_name(instance_data)
+      @name = instance_data.dig('metadata', 'name')
+      @definition['metadata']['name'] = @name
```
timothysmith0609: This could use a comment:
```
If a resource uses generateName, we don't know the full name of the resource until it's deployed to the cluster.
In this case, we need to update our local definition with the realized name in order to accurately
track the resource during deploy
```

lib/krane/deprecated_deploy_task.rb
```diff
@@ -453,9 +453,11 @@ def replace_or_create_resource(resource, force: false)
...
       end
 
-      _, _, status = kubectl.run(*args, log_failure: false)
+      _, _, status = kubectl.run(*args, log_failure: false, output_is_sensitive: resource.sensitive_template_content?,
+        raise_if_not_found: true)
       return [nil, status] if status.success?
 
+    rescue Krane::Kubectl::ResourceNotFoundError
       # it doesn't exist so we can't replace it, we try to create it
       create_resource(resource)
```
KnVerey: I think you need to explicitly `return` here--that's probably what's causing your CI failure

lib/krane/deprecated_deploy_task.rb
```diff
@@ -453,9 +453,11 @@ def replace_or_create_resource(resource, force: false)
         ["replace", "-f", resource.file_path]
       end
 
-      _, _, status = kubectl.run(*args, log_failure: false)
+      _, _, status = kubectl.run(*args, log_failure: false, output_is_sensitive: resource.sensitive_template_content?,
+        raise_if_not_found: true)
       return [nil, status] if status.success?
```
KnVerey: actually, now that we're doing the `create` inside the `rescue`, we _always_ want to return `[err, status]`; otherwise we return `nil` when the command fails with a non-404, which is not a valid return value for this method.

RyanBrushett:
Rebased and added a changelog entry. Once CI goes 🍏 I'll merge.

Thank you everyone. Your feedback and input is greatly appreciated 😃

PR #612, opened 2019-11-04
Global prune 2

dturn:
Add pruning by default to global deploy

**How is this accomplished?**
Fetch a list of global resources that are pruneable (everything that supports delete) exclude node and namespace since I can't image we'd actually want to ever prune these and then pass them as the list of resources to prune.

**What could go wrong?**
Worst case is that we are including something in the prune list we shouldn't


KnVerey:
> Worst case is that we are including something in the prune list we shouldn't

For sanity checking purposes, can you please post what this list actually is in the oldest and newest k8s versions we currently support?

dturn:
We intentionally black list Node and Namespace. I can't really see a reason where pruning a Node is a good idea.  Pruning a Namespace is more reasonable but really dangerous. 

### 1.11.0
/v1/PersistentVolumeClaim
/v1/PersistentVolume
admissionregistration.k8s.io/v1beta1/MutatingWebhookConfiguration
admissionregistration.k8s.io/v1beta1/ValidatingWebhookConfiguration
apiextensions.k8s.io/v1beta1/CustomResourceDefinition
apiregistration.k8s.io/v1beta1/APIService
certificates.k8s.io/v1beta1/CertificateSigningRequest
extensions/v1beta1/PodSecurityPolicy
policy/v1beta1/PodSecurityPolicy
rbac.authorization.k8s.io/v1beta1/ClusterRoleBinding
rbac.authorization.k8s.io/v1beta1/ClusterRole
scheduling.k8s.io/v1beta1/PriorityClass
storage.k8s.io/v1beta1/StorageClass
storage.k8s.io/v1beta1/VolumeAttachment

# 1.16.1

/v1/PersistentVolumeClaim
/v1/PersistentVolume
admissionregistration.k8s.io/v1beta1/MutatingWebhookConfiguration
admissionregistration.k8s.io/v1beta1/ValidatingWebhookConfiguration
apiextensions.k8s.io/v1beta1/CustomResourceDefinition
apiregistration.k8s.io/v1beta1/APIService
certificates.k8s.io/v1beta1/CertificateSigningRequest
node.k8s.io/v1beta1/RuntimeClass # not in 1.11
policy/v1beta1/PodSecurityPolicy
rbac.authorization.k8s.io/v1beta1/ClusterRoleBinding
rbac.authorization.k8s.io/v1beta1/ClusterRole
scheduling.k8s.io/v1beta1/PriorityClass
storage.k8s.io/v1beta1/CSIDriver # not in 1.11
storage.k8s.io/v1beta1/StorageClass
storage.k8s.io/v1beta1/VolumeAttachment

test/unit/cluster_resource_discovery_test.rb
```diff
@@ -17,6 +17,15 @@ def test_global_resource_kinds_success
     end
   end
 
+  def test_prunable_resources
```
RyanBrushett: Maybe it's not necessary but one could also refute_includes Node and Namespace yeah? Seems like an unlikely regression but might be nice to be explicit about.

lib/krane/cluster_resource_discovery.rb
```diff
@@ -17,13 +17,38 @@ def crds
...
+        version = api_versions.fetch(resource['apigroup'], ['v1']).last
+        [resource['apigroup'], version, resource['kind']].compact.join("/")
+      end.compact
     end
 
     private
 
-    def fetch_globals
-      raw, _, st = kubectl.run("api-resources", "--namespaced=false", output: "wide", attempts: 5,
+    def fetch_api_versions
```
KnVerey: Could you add a comment documenting an example of the format/fields returned by this and `fetch_resources`? I think it would make it easier to review this class.

lib/krane/cluster_resource_discovery.rb
```diff
@@ -17,13 +17,38 @@ def crds
...
     def global_resource_kinds
-      @globals ||= fetch_globals.map { |g| g["kind"] }
+      @globals ||= fetch_resources(only_globals: true).map { |g| g["kind"] }
+    end
+
+    def prunable_resources
+      api_versions = fetch_api_versions
+      fetch_resources.map do |resource|
+        next unless resource['verbs'].include?("delete")
+        version = api_versions.fetch(resource['apigroup'], ['v1']).last
```
KnVerey: I'm a bit confused by this--it seems to imply that the api versions list can be missing a resource's API group. Isn't the logic actually "if the resource has no api group, then the version is v1; otherwise, grab it from the api versions list"? 

```ruby
version = resource["apigroup"].blank? ? 'v1' : api_versions["apigroup"].last
```

Also, why is the last one the correct choice? They seem to be alphabetical, so this would make us pick beta over alpha, but also beta over GA. Does it matter?

lib/krane/cluster_resource_discovery.rb
```diff
@@ -17,13 +17,38 @@ def crds
...
     private
 
-    def fetch_globals
-      raw, _, st = kubectl.run("api-resources", "--namespaced=false", output: "wide", attempts: 5,
+    def fetch_api_versions
+      raw, _, st = kubectl.run("api-versions", attempts: 5, use_namespace: false)
+      if st.success?
+        rows = raw.split("\n")
+        rows.each_with_object({}) do |group_version, hash|
+          group, version = group_version.split("/")
```
KnVerey: When I run this locally, I see core/v1 is listed as `v1`, which will give us a `v1 => []` entry in this array--it's actually the opposite, since v1 is of course a version not a group. Actually, if we handle that here, I think we can remove the special case above, since there'd be an entry in the hash for the blank api group?

test/integration-serial/serial_deploy_test.rb
```diff
@@ -603,6 +603,24 @@ def test_global_deploy_validation_catches_namespaced_cr
     wait_for_all_crd_deletion
   end
 
+  def test_global_deploy_prune_black_box_success
+    setup_template_dir("globals") do |target_dir|
```
KnVerey: Let's augment this fixture set with one of the blacklisted kinds to show it doesn't get pruned.

test/integration-serial/serial_deploy_test.rb
```diff
@@ -603,6 +603,24 @@ def test_global_deploy_validation_catches_namespaced_cr
...
+      flags = "-f #{target_dir} --selector app=krane"
+      out, err, status = krane_black_box("global-deploy", "#{KubeclientHelper::TEST_CONTEXT} #{flags}")
+      assert_empty(out)
+      assert_match("Success", err)
+      assert_predicate(status, :success?)
+
```
KnVerey: flush the logger

test/integration/global_deploy_test.rb
```diff
@@ -116,4 +126,54 @@ def test_global_deploy_task_failure
       "Template validation failed",
     ])
   end
+
+  def test_global_deploy_prune_black_box_success
```
KnVerey: nit: this and the next test are not black-box tests

test/integration/global_deploy_test.rb
```diff
@@ -14,14 +14,17 @@ def test_global_deploy_task_success
...
+      "Deploying resources:",
+      %r{StorageClass\/testing-storage-class[\w-]+ \(timeout: 300s\)},
+      %r{PriorityClass/testing-priority-class[\w-]+ \(timeout: 300s\)},
       "Don't know how to monitor resources of type StorageClass.",
       %r{Assuming StorageClass\/testing-storage-class[\w-]+ deployed successfully.},
-      %r{Successfully deployed in [\d.]+s: StorageClass\/testing-storage-class},
+      %r{Successfully deployed in [\d.]+s: PriorityClass/testing-priority-class[\w-]+, StorageClass\/testing-storage-},
       "Result: SUCCESS",
-      "Successfully deployed 1 resource",
+      "Successfully deployed 2 resource",
```
KnVerey: nit: this should say `resources` (and probably does)

test/integration/global_deploy_test.rb
```diff
@@ -77,7 +84,7 @@ def test_global_deploy_task_empty_selector_validation_failure
   end
 
   def test_global_deploy_task_success_selector
-    selector = "app=krane"
+    selector = "app=krane2"
```
KnVerey: Why this change? Should we be using both this _and_ a generated selector?

test/integration/global_deploy_test.rb
```diff
@@ -89,14 +96,17 @@ def test_global_deploy_task_success_selector
...
+      "Deploying resources:",
+      %r{PriorityClass/testing-priority-class[\w-]+ \(timeout: 300s\)},
+      %r{StorageClass\/testing-storage-class[\w-]+ \(timeout: 300s\)},
       "Don't know how to monitor resources of type StorageClass.",
       %r{Assuming StorageClass\/testing-storage-class[\w-]+ deployed successfully.},
-      %r{Successfully deployed in [\d.]+s: StorageClass\/testing-storage-class},
+      /Successfully deployed in [\d.]+s/,
       "Result: SUCCESS",
-      "Successfully deployed 1 resource",
+      "Successfully deployed 2 resource",
```
KnVerey:
```suggestion
      "Successfully deployed 2 resources",
```

test/integration/global_deploy_test.rb
```diff
@@ -54,12 +57,16 @@ def test_global_deploy_task_success_verify_false
...
       %r{StorageClass\/testing-storage-class[\w-]+\s+Not Found},
+      %r{PriorityClass/testing-priority-class[\w-]+\s+Not Found},
       "Phase 3: Deploying all resources",
-      %r{Deploying StorageClass\/testing-storage-class[\w-]+ \(timeout: 300s\)},
+      "Deploying resources:",
+      %r{StorageClass\/testing-storage-class[\w-]+ \(timeout: 300s\)},
+      %r{PriorityClass/testing-priority-class[\w-]+ \(timeout: 300s\)},
       "Result: SUCCESS",
-      "Deployed 1 resource",
+      "Deployed 2 resource",
```
KnVerey:
```suggestion
      "Deployed 2 resources",
```

test/integration/global_deploy_test.rb
```diff
@@ -116,4 +126,54 @@ def test_global_deploy_task_failure
       "Template validation failed",
     ])
   end
+
+  def test_global_deploy_prune_black_box_success
+    assert_deploy_success(deploy_global_fixtures('globals', clean_up: false, selector: 'test=prune1'))
```
KnVerey: same question, should we be adding this selector to the generated one so we don't have to hope future contributors won't reuse this manual selector?

test/integration/global_deploy_test.rb
```diff
@@ -116,4 +126,54 @@ def test_global_deploy_task_failure
...
+      "Phase 1: Initializing deploy",
+      "Using resource selector test=prune1",
+      "All required parameters and files are present",
+      "Discovering resources:",
+      "  - StorageClass/testing-storage-class",
+      "Phase 2: Checking initial resource statuses",
+      %r{StorageClass\/testing-storage-class[\w-]+\s+Exists},
+      "Phase 3: Deploying all resources",
+      %r{Deploying StorageClass\/testing-storage-class[\w-]+ \(timeout: 300s\)},
+      "Don't know how to monitor resources of type StorageClass.",
```
KnVerey: I'd remove these next two lines--the fact that we don't yet support this class properly isn't relevant to this test

test/integration/global_deploy_test.rb
```diff
@@ -116,4 +126,54 @@ def test_global_deploy_task_failure
...
+      %r{Assuming StorageClass\/testing-storage-class[\w-]+ deployed successfully.},
+      %r{Successfully deployed in [\d.]+s: StorageClass\/testing-storage-class},
+      "Result: SUCCESS",
+      "Pruned 1 resource and successfully deployed 1 resource",
```
KnVerey: let's also assert on the line that specifies which resource was pruned

test/integration/global_deploy_test.rb
```diff
@@ -116,4 +126,54 @@ def test_global_deploy_task_failure
...
+      "Phase 1: Initializing deploy",
+      "Using resource selector test=prune2",
+      "All required parameters and files are present",
+      "Discovering resources:",
+      "  - StorageClass/testing-storage-class",
+      "Phase 2: Checking initial resource statuses",
+      %r{StorageClass\/testing-storage-class[\w-]+\s+Exists},
+      "Phase 3: Deploying all resources",
+      %r{Deploying StorageClass\/testing-storage-class[\w-]+ \(timeout: 300s\)},
+      "Don't know how to monitor resources of type StorageClass.",
```
KnVerey: 🔥

test/integration/global_deploy_test.rb
```diff
@@ -116,4 +126,54 @@ def test_global_deploy_task_failure
...
+      %r{Assuming StorageClass\/testing-storage-class[\w-]+ deployed successfully.},
+      %r{Successfully deployed in [\d.]+s: StorageClass\/testing-storage-class},
+      "Result: SUCCESS",
+      "Successfully deployed 1 resource",
+      "Successful resources",
+      "StorageClass/testing-storage-class",
+    ])
+    assert_deploy_success(deploy_global_fixtures('globals', selector: 'test=prune2'))
```
KnVerey: we should refute that the logs include "[pP]rune[d]"

lib/krane/cluster_resource_discovery.rb
dturn: > Also, why is the last one the correct choice? They seem to be alphabetical, so this would make us pick beta over alpha, but also beta over GA. Does it matter?

We get lucky and for global, last just happens to work. The PR to add this to namespaced resources actually parses the versions, pickets the latest and has overides for a resource isn't supported in the latest version of the group.
https://github.com/Shopify/kubernetes-deploy/pull/616/files#diff-3c59dc6581aa611cc9f2c1c012d0f539R28

lib/krane/cluster_resource_discovery.rb
```diff
@@ -34,6 +59,7 @@ def fetch_globals
         fields = full_width_field_names.each_with_object({}) do |name, hash|
           start = cursor
           cursor = start + name.length
+          cursor = 0 if full_width_field_names.last == name.strip
```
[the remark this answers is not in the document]
dturn: I can probably solve this in a more elegant manner, but the reason is that the header row doesn't include whitespace after the last field name, even though the last field can be much longer than the fieldname so we were accidentally truncating the last field. This wasn't noticed before because we didn't expose `VERBS` to the public interface.

```
NAME                              SHORTNAMES   APIGROUP                       NAMESPACED   KIND                             VERBS
bindings                                                                      true         Binding                          [create]
```

test/integration/global_deploy_test.rb
dturn: Before we always generated a selector, this PR allows us to pass the selector (needed for pruning reasons). But if the selector matches a different test's selector the pruning process conflicts.

lib/krane/cluster_resource_discovery.rb
KnVerey: I think the answer is yes, but to confirm: if we pick a version that doesn't work, do we 100% for sure find out in our CI, even for resources that aren't in our test suite? (otherwise should we be making an API call during the deploy to validate our choice?)

dturn: The answer is yes. `kubectl apply` fails so we get lots of failures for global-deploy and deploy.

test/integration-serial/serial_deploy_test.rb
dturn: krane_black_box uses `Open3.capture3` so each call get's a unique stream

test/integration/global_deploy_test.rb
dturn: We use the test name to modify the fixture name, so if we refute 'prune' we can't use 'prune' in the test name. Instead I'm refuting 'pruned'.

lib/krane/kubeclient_builder.rb
```diff
@@ -103,6 +103,14 @@ def build_storage_v1_kubeclient(context)
       )
     end
 
+    def build_scheduling_v1beta1_kubeclient(context)
```
dirceu: Is this actually necessary? I tried removing this and running tests but they keep passing; I tested with the tests modified/added in this commit:

```bash
$ alias rt="bundle exec ruby -I test"
$ rt test/integration-serial/serial_deploy_test.rb --name test_global_deploy_prune_black_box_success
$ rt test/integration/global_deploy_test.rb
$ rt test/unit/cluster_resource_discovery_test.rb
```

Am I missing anything?

dturn: `build_scheduling_v1beta1_kubeclient` are mostly used in the tests to allow for manual interaction with resources, usually to delete them in an ensure block. They aren't being used but seemed less than ideal  to add a new kind to the fixtures but not add its helper methods.

Issue #605, opened 2019-10-28
Replace `ENV['REVISION']` with `--revision`


dirceu:
We missed replacing `ENV['REVISION']` with `--revision`, which was defined in [Krane's design doc](https://docs.google.com/document/d/1oInUsKplYGNWTymPY48xtDCx3x4Cwc-phcEKUkt2v4U/view). It's important to do so before 1.0 because it's technically a breaking change.

dturn:
We marked this as optional in the design doc. How would people feel about just adding a `--revision` flag to the commands that takes priority but still allow the `ENV["REVISION"]` fallback?

PR #613, opened 2019-11-05
Add optional revision flag

dturn:

As part of the [krane 1.0 migration](https://docs.google.com/document/d/1oInUsKplYGNWTymPY48xtDCx3x4Cwc-phcEKUkt2v4U/view) we decided to add a ~`--revision`~ `--current-sha` flag. 

~This PR adds the flag with higher precedence than `ENV["REVISION"]`.~

I've added it to the old tasks so that transitioning to `krane` is easier.
 
closes: https://github.com/Shopify/kubernetes-deploy/issues/605

**How is this accomplished?**
~Add a `--revision` flag that is used over `ENV["REVISION"]`~

Add a `--current-sha` flag and remove usage of `ENV["REVISION"]` in Krane (we still use that as a fallback in `kubernetes-deploy` and `kubernetes-render`)

**What could go wrong?**
~People get confused  about the priority.~

### DIFF
--- a/lib/krane/deprecated_deploy_task.rb
+++ b/lib/krane/deprecated_deploy_task.rb
@@ -597,7 +597,8 @@ module Krane
     end
 
     def statsd_tags
-      %W(namespace:#{@namespace} sha:#{@current_sha} context:#{@context}) | @namespace_tags
+      tags = %W(namespace:#{@namespace} context:#{@context}) | @namespace_tags
+      @current_sha.nil? ? tags : %W(sha:#{@current_sha}) | tags
     end
 
     def with_retries(limit)
--- a/test/integration/krane_test.rb
+++ b/test/integration/krane_test.rb
@@ -2,8 +2,6 @@
 require 'integration_test_helper'
 
 class KraneTest < Krane::IntegrationTest
-  include EnvTestHelper
-
   def test_restart_black_box
     assert_deploy_success(deploy_fixtures("hello-cloud", subset: ["configmap-data.yml", "web.yml.erb", "redis.yml"]))
     refute(fetch_restarted_at("web"), "RESTARTED_AT env on fresh deployment")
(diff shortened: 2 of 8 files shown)


KnVerey:
> I find it confusing that a flag called revision ends up generating an internal template variable named current_sha: can we switch current_sha to revision to maintain a 1:1 correspondence?

We should make sure the flag and arg in our Ruby API match, at the very least. Switching the ERB variable would be a huge PITA for us, if not also other companies. Maybe we could expose both and switch our documentation to refer to `revision`?

> it, generally, controls an extremely important field (the actual image you're going to run). 

That's what we use it for internally, but technically it doesn't have to be used at all in the ERB (and I'm ok with throwing the default rendering error if you try to use the variable without giving us a revision). IIRC the one thing krane itself uses this value for is to emit statsD tags. And naturally the prefix is just `sha`, so a third name 🤦‍♀. That one would be easier to change to `revision` when we swap to the `Krate` prefix on our metrics, if we want to.

> It seems like the reasons we use an env var to pass this information are largely historic (e.g. predating bindings).

You're likely correct that it predates bindings, but the specific historic reason is that we started as a Shipit integration, and Shipit gives you that env var by default.

> My preference would be to drop support for the implicit ENV var assignment so there's only one, canonical, place to declare it

I'm also leaning this way. I don't *think* it causes us much toil? Like, internally we just have to switch our (still rather few) individual krane invocations to include `--revision=ENV["REVISION"]`, right? Or am I missing something?

dturn:
> We should make sure the flag and arg in our Ruby API match, at the very least. Switching the ERB variable would be a huge PITA for us, if not also other companies. Maybe we could expose both and switch our documentation to refer to revision?

Both seems like a compromise in the wrong direction. I'd rather call the flag `--current-sha`

dirceu:
Updated the PR with the following changes:
1. Change flag name to `--current-sha`
2. Remove the fallback to `ENV["REVISION"]`

test/integration/krane_test.rb
```diff
@@ -38,19 +36,15 @@ def test_render_black_box
     bindings = "data=#{data_value}"
     test_sha = rand(10_000).to_s
 
-    out, err, status = nil
-    with_env("REVISION", test_sha) do
-      out, err, status = krane_black_box("render", "-f #{paths.join(' ')} --bindings #{bindings}")
-    end
+    out, err, status = krane_black_box("render", "-f #{paths.join(' ')} --bindings #{bindings} "\
```
dturn: should we just stick the second arg on its own line rather than split the string

lib/krane/deprecated_deploy_task.rb
```diff
@@ -597,7 +597,11 @@ def ejson_keys_secret
     end
 
     def statsd_tags
-      %W(namespace:#{@namespace} sha:#{@current_sha} context:#{@context}) | @namespace_tags
+      tags = %W(namespace:#{@namespace} context:#{@context}) | @namespace_tags
+      unless @current_sha.nil?
+        tags = %W(sha:#{@current_sha}) | tags
+      end
+      tags
```
timothysmith0609: You can one-line this as `@current_sha.present? ? %W(sha:#{@current_sha}) : tags`

lib/krane/deprecated_deploy_task.rb
```diff
@@ -597,7 +597,9 @@ def ejson_keys_secret
     end
 
     def statsd_tags
-      %W(namespace:#{@namespace} sha:#{@current_sha} context:#{@context}) | @namespace_tags
+      tags = %W(namespace:#{@namespace} context:#{@context}) | @namespace_tags
+      @current_sha.nil? ? tags : %W(sha:#{@current_sha}) | tags
+      tags
```
timothysmith0609: The ternary operator will return the evaluated expression so no need for the explicit `tags` line

PR #616, opened 2019-11-07
Switch namespaced deploy to pruning blacklist

dturn:
Prune all namespaced kinds by default instead of maintaining a whitelist. 

**How is this accomplished?**
This mirrors  #612. However, the latest group/versions doesn't always include all kinds, so I've added an override for the 5 kinds that need it. As far as I can tell there is no kubectl call to get all resources supported version. There is a call to get a resource's supported version making a call per resource seems extreme. 

**What could go wrong?**
This will prune things we didn't before, full list is below.. 
We may have to keep the override list up-to-date. That being said the list currently contains five items, which is much shorter than the white list it's replacing.


dturn:
### Namespaced pruneable resources in k8s v1.11.10

"/v1/ConfigMap",
 "/v1/Endpoints",
 "/v1/Event",
 "/v1/LimitRange",
 "/v1/PersistentVolumeClaim",
 "/v1/Pod",
 "/v1/PodTemplate",
 "/v1/ReplicationController",
 "/v1/ResourceQuota",
 "/v1/Secret",
 "/v1/ServiceAccount",
 "/v1/Service",
 "apiregistration.k8s.io/v1beta1/APIService",
 "apps/v1beta2/ControllerRevision",
 "apps/v1beta2/DaemonSet",
 "apps/v1beta2/Deployment",
 "apps/v1beta2/ReplicaSet",
 "apps/v1beta2/StatefulSet",
 "autoscaling/v2beta1/HorizontalPodAutoscaler",
 "batch/v1beta1/CronJob",
 "batch/v1beta1/Job", # The group/version is wrong, investigating 
 "events.k8s.io/v1beta1/Event",
 "extensions/v1beta1/DaemonSet",
 "extensions/v1beta1/Deployment",
 "extensions/v1beta1/Ingress",
 "extensions/v1beta1/NetworkPolicy",
 "extensions/v1beta1/PodSecurityPolicy",
 "extensions/v1beta1/ReplicaSet",
 "networking.k8s.io/v1/NetworkPolicy",
 "policy/v1beta1/PodDisruptionBudget",
 "policy/v1beta1/PodSecurityPolicy",
 "rbac.authorization.k8s.io/v1beta1/ClusterRoleBinding",
 "rbac.authorization.k8s.io/v1beta1/ClusterRole",
 "rbac.authorization.k8s.io/v1beta1/RoleBinding",
 "rbac.authorization.k8s.io/v1beta1/Role"

lib/krane/deploy_task.rb
```diff
@@ -8,5 +8,9 @@ def initialize(**args)
       raise "Use Krane::DeployGlobalTask to deploy global resources" if args[:allow_globals]
       super(args.merge(allow_globals: false))
     end
+
+    def prune_whitelist
```
dirceu: are we adding this here (instead of updating `Krane::DeprecatedDeployTask`) to avoid changing `KubernetesDeploy::DeployTask`?

dturn: Yes

dturn:
This PR and the global-deploy pruning PR no longer respect the `krane.shopify.io/prunable` annotation. I think that's the right behavior, but we'll need to clean up the readme & migration guide, in a subsequent PR depending if this merges before 1.0.0-pre.1

test/unit/cluster_resource_discovery_test.rb
```diff
@@ -2,41 +2,62 @@
...
       .returns([api_versions_full_response, "", stub(success?: true)])
-    crd = mocked_cluster_resource_discovery(api_resources_full_response)
+    crd = mocked_cluster_resource_discovery(api_resources_not_namespaced_full_response)
     kinds = crd.prunable_resources(namespaced: false)
 
     assert_equal(kinds.length, 13)
-    %w(scheduling.k8s.io/v1beta1/PriorityClass storage.k8s.io/v1beta1/StorageClass).each do |kind|
-      assert_includes(kinds, kind)
+    %w(PriorityClass StorageClass).each do |expected_kind|
+      assert kinds.one? { |k| k.include?(expected_kind) }
```
timothysmith0609: Is this just a sanity check or is there a reason for these 2 in particular?

dturn: Just a sanity check, the `assert_equal(kinds.length, 13)` is the check to ensure list has the right number of items.

test/helpers/fixture_deploy_helper.rb
```diff
@@ -185,8 +185,7 @@ def namespace_globals(fixtures, selector)
     fixtures.each do |_, kinds_map|
       kinds_map.each do |_, resources|
         resources.each do |resource|
-          resource["metadata"]["name"] = (resource["metadata"]["name"] + @namespace)[0..63]
-          resource["metadata"]["name"] += "0" if resource["metadata"]["name"].end_with?("-")
+          resource["metadata"]["name"] = (resource["metadata"]["name"] + Digest::MD5.hexdigest(@namespace))[0..63]
```
dturn: This is to fix a bug that was causing resources to have the same name across different tests. Its triggered when the real name was long as the start of two tests were the same. We need the value appended to the name to be stable across tests hence the hashing of @namespace.

lib/krane/cluster_resource_discovery.rb
```diff
@@ -85,6 +62,37 @@ def fetch_resources(namespaced: false)
...
+      version_override = { "CronJob" => "v1beta1", "VolumeAttachment" => "v1beta1",
+                           "CSIDriver" => "v1beta1", "Ingress" => "v1beta1", "CSINode" => "v1beta1" }
+
+      pattern = /v(?<major>\d+)(beta)?(?<minor>\d+)?/
+      latest = versions.sort_by do |version|
+        match = version.match(pattern)
+        [match[:major].to_i, (match[:minor] || 999).to_i]
+      end.last
+      version_override.fetch(kind, latest)
+    end
```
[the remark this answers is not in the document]
dturn: Good point, I've also tried to fix up names and added a comment to make things clearer.

PR #617, opened 2019-11-07
Ensure instance data is synced with pod state before reporting DS success

timothysmith0609:
#580 added more detailed handling of daemon set rollouts. In particular, the addition of new nodes during a deployment should not interfere with convergence logic. However, this has resulted in a flaky test where the DS deploy is considered successful, but is reporting:
`1 updatedNumberScheduled, 1 desiredNumberScheduled, 0 numberReady`

After some digging, it appears that we fail to ensure the state of the target DS pods is reflected in the instance data of the DS itself when checking `deploy_succeeded?`, leading to the situation where a deploy is successful (e.g. the pod(s) is/are up), but the DS itself hasn't converged to this reality.

**How is this accomplished?**
Add an extra check in `deploy_succeeded?` to ensure `considered_pods.length == rollout_data["numberReady"]

**What could go wrong?**
In the worst case, this adds additional sync loops while we wait for the DS to converge, but that's the real definition of success, so it was an error to ignore this in the first place 

It's also impossible to test this, since it relies on timing of sync loops between pods and the DS :(

### DIFF
--- a/lib/krane/kubernetes_resource/daemon_set.rb
+++ b/lib/krane/kubernetes_resource/daemon_set.rb
@@ -58,8 +58,8 @@ module Krane
       return true if rollout_data["desiredNumberScheduled"].to_i == rollout_data["numberReady"].to_i # all pods ready
       relevant_node_names = @nodes.map(&:name)
       considered_pods = @pods.select { |p| relevant_node_names.include?(p.node_name) }
-      @logger.debug("DaemonSet is reporting #{rollout_data['numberReady']} pods ready")
-      @logger.debug("Considered #{considered_pods.size} pods out of #{@pods.size} for #{@nodes.size} nodes.")
+      @logger.debug("DaemonSet is reporting #{rollout_data['numberReady']} pods ready." \
+        " Considered #{considered_pods.size} pods out of #{@pods.size} for #{@nodes.size} nodes.")
       considered_pods.present? &&
         considered_pods.all?(&:deploy_succeeded?) &&
         rollout_data["numberReady"].to_i >= considered_pods.length
--- a/lib/krane/kubernetes_resource/daemon_set.rb
+++ b/lib/krane/kubernetes_resource/daemon_set.rb
@@ -58,7 +58,8 @@ module Krane
       return true if rollout_data["desiredNumberScheduled"].to_i == rollout_data["numberReady"].to_i # all pods ready
       relevant_node_names = @nodes.map(&:name)
       considered_pods = @pods.select { |p| relevant_node_names.include?(p.node_name) }
-      @logger.debug("Considered #{considered_pods.size} pods out of #{@pods.size} for #{@nodes.size} nodes")
+      @logger.debug("DaemonSet is reporting #{rollout_data['numberReady']} pods ready")
+      @logger.debug("Considered #{considered_pods.size} pods out of #{@pods.size} for #{@nodes.size} nodes.")
       considered_pods.present? &&
         considered_pods.all?(&:deploy_succeeded?) &&
         rollout_data["numberReady"].to_i >= considered_pods.length
--- a/lib/krane/kubernetes_resource/daemon_set.rb
+++ b/lib/krane/kubernetes_resource/daemon_set.rb
@@ -61,7 +61,7 @@ module Krane
       @logger.debug("Considered #{considered_pods.size} pods out of #{@pods.size} for #{@nodes.size} nodes")
       considered_pods.present? &&
         considered_pods.all?(&:deploy_succeeded?) &&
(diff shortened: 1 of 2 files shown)


dalehamel:
> It's also impossible to test this, since it relies on timing of sync loops between pods and the DS :(

In our test can't we mock the daemonset being out of sync with the pods?

dalehamel:
> Add an extra check in deploy_succeeded? to ensure `considered_pods.length == rollout_data["numberReady"]

As discussed in slack, (and now changed in your PR) i think this should be >=, as I think that this should always be equal to or greater than the number of pods considered, since we are only considering pods for potentially a subset of nodes.

test/unit/krane/kubernetes_resource/daemon_set_test.rb
```diff
@@ -107,6 +107,19 @@ def test_deploy_fails_when_not_all_pods_ready
...
+      "desiredNumberScheduled": 1,
+      "updatedNumberScheduled": 1,
+      "numberReady": 0,
+    }
+    ds_template = build_ds_template(filename: 'daemon_set.yml', status: status)
+    ready_pod_template = load_fixtures(filenames: ['daemon_set_pods.yml']).first # should be a pod in `Ready` state
+    node_templates = load_fixtures(filenames: ['nodes.yml'])
+    ds = build_synced_ds(ds_template: ds_template, pod_templates: [ready_pod_template], node_templates: node_templates)
+    refute_predicate(ds, :deploy_succeeded?)
+  end
```
timothysmith0609: Testing this against master yields a failure, proving my hypothesis that the divergent state between pods and the DS were the root of the error

lib/krane/kubernetes_resource/daemon_set.rb
```diff
@@ -59,7 +59,9 @@ def relevant_pods_ready?
       relevant_node_names = @nodes.map(&:name)
       considered_pods = @pods.select { |p| relevant_node_names.include?(p.node_name) }
       @logger.debug("Considered #{considered_pods.size} pods out of #{@pods.size} for #{@nodes.size} nodes")
```
dalehamel: I know you didn't touch this line, but maybe we should add `rollout_data["numberReady"]` to this debug line so that if this causes an unexpected regression for daemonset tolerance, we can debug it more easily

test/unit/krane/kubernetes_resource/daemon_set_test.rb
```diff
@@ -107,6 +107,26 @@ def test_deploy_fails_when_not_all_pods_ready
...
+    }
+    not_ready_ds_template = build_ds_template(filename: 'daemon_set.yml', status: status)
+    ready_pod_template = load_fixtures(filenames: ['daemon_set_pods.yml']).first # should be a pod in `Ready` state
+    node_templates = load_fixtures(filenames: ['nodes.yml'])
+    ds = build_synced_ds(ds_template: not_ready_ds_template, pod_templates: [ready_pod_template],
+      node_templates: node_templates)
+    refute_predicate(ds, :deploy_succeeded?)
+
+    status[:numberReady] = 1
+    ready_ds_template = build_ds_template(filename: 'daemon_set.yml', status: status)
```
dalehamel: you are basically building a new daemonset here, but i think the behavior is better tested if you take the approach from https://github.com/Shopify/kubernetes-deploy/blob/218a82c3a9b8cad2d3594c1e38ab48af673ee18c/test/unit/krane/kubernetes_resource/daemon_set_test.rb#L58, and just stub and forcefully resync the original one (see https://github.com/Shopify/kubernetes-deploy/blob/218a82c3a9b8cad2d3594c1e38ab48af673ee18c/test/unit/krane/kubernetes_resource/daemon_set_test.rb#L78-L80)

lib/krane/kubernetes_resource/daemon_set.rb
```diff
@@ -58,8 +58,11 @@ def relevant_pods_ready?
       return true if rollout_data["desiredNumberScheduled"].to_i == rollout_data["numberReady"].to_i # all pods ready
       relevant_node_names = @nodes.map(&:name)
       considered_pods = @pods.select { |p| relevant_node_names.include?(p.node_name) }
-      @logger.debug("Considered #{considered_pods.size} pods out of #{@pods.size} for #{@nodes.size} nodes")
-      considered_pods.present? && considered_pods.all?(&:deploy_succeeded?)
+      @logger.debug("DaemonSet is reporting #{rollout_data['numberReady']} pods ready")
```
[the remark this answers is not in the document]
timothysmith0609: Yeah, originally I didn't want to do that to avoid an overly-long log line, but since it's debug that's not really a big concern. Fixed to make a single statement

PR #619, opened 2019-11-11
Make task arguments match the new flag names

dirceu:

- Make sure task arguments match the new flag names so we have a consistent interface both in the API and in the CLI (https://github.com/Shopify/kubernetes-deploy/pull/602#discussion_r339247531).
- Be explicit about which arguments are required.

**How is this accomplished?**

By changing the public task interfaces so their argument names would match the respective flag names, and updating the test setup accordingly.

### DIFF
--- a/lib/krane/cli/deploy_command.rb
+++ b/lib/krane/cli/deploy_command.rb
@@ -68,10 +68,10 @@ module Krane
             namespace: namespace,
             context: context,
             current_sha: options['current-sha'],
-            template_paths: paths,
+            filenames: paths,
             bindings: bindings_parser.parse,
             logger: logger,
-            max_watch_seconds: ::Krane::DurationParser.new(options["global-timeout"]).parse!.to_i,
+            global_timeout: ::Krane::DurationParser.new(options["global-timeout"]).parse!.to_i,
             selector: selector,
             protected_namespaces: protected_namespaces,
           )
--- a/lib/krane/cli/deploy_command.rb
+++ b/lib/krane/cli/deploy_command.rb
@@ -68,7 +68,7 @@ module Krane
             namespace: namespace,
             context: context,
             current_sha: options['current-sha'],
-            filenames: paths,
+            filenames: filenames,
             bindings: bindings_parser.parse,
             logger: logger,
             global_timeout: ::Krane::DurationParser.new(options["global-timeout"]).parse!.to_i,
--- a/lib/krane/cli/deploy_command.rb
+++ b/lib/krane/cli/deploy_command.rb
@@ -68,7 +68,7 @@ module Krane
             namespace: namespace,
             context: context,
             current_sha: options['current-sha'],
-            filenames: filenames,
+            filenames: paths,
             bindings: bindings_parser.parse,
             logger: logger,
             global_timeout: ::Krane::DurationParser.new(options["global-timeout"]).parse!.to_i,
--- a/lib/krane/cli/restart_command.rb
+++ b/lib/krane/cli/restart_command.rb
@@ -21,7 +21,7 @@ module Krane
         restart = ::Krane::RestartTask.new(
           namespace: namespace,
           context: context,
-          max_watch_seconds: ::Krane::DurationParser.new(options["global-timeout"]).parse!.to_i,
+          global_timeout: ::Krane::DurationParser.new(options["global-timeout"]).parse!.to_i,
         )
         restart.run!(
           options[:deployments],
--- a/lib/krane/cli/restart_command.rb
+++ b/lib/krane/cli/restart_command.rb
@@ -24,7 +24,7 @@ module Krane
           global_timeout: ::Krane::DurationParser.new(options["global-timeout"]).parse!.to_i,
         )
         restart.run!(
-          options[:deployments],
+          deployments: options[:deployments],
           selector: selector,
           verify_result: options["verify-result"]
         )
--- a/lib/krane/cli/run_command.rb
+++ b/lib/krane/cli/run_command.rb
@@ -37,14 +37,14 @@ module Krane
         runner = ::Krane::RunnerTask.new(
           namespace: namespace,
           context: context,
-          max_watch_seconds: ::Krane::DurationParser.new(options["global-timeout"]).parse!.to_i,
+          global_timeout: ::Krane::DurationParser.new(options["global-timeout"]).parse!.to_i,
         )
 
         runner.run!(
           verify_result: options['verify-result'],
-          task_template: options['template'],
-          entrypoint: options['command'],
-          args: options['arguments']&.split(" "),
+          template: options['template'],
+          command: options['command'],
+          arguments: options['arguments']&.split(" "),
           env_vars: options['env-vars'].split(','),
         )
       end
--- a/lib/krane/cli/run_command.rb
+++ b/lib/krane/cli/run_command.rb
@@ -19,7 +19,7 @@ module Krane
         },
         "verify-result" => { type: :boolean, desc: "Wait for completion and verify pod success", default: true },
         "command" => { type: :array, desc: "Override the default command in the container image" },
-        "filename" => {
+        "template" => {
           type: :string,
           desc: "The template file you'll be rendering",
           required: true,
@@ -43,7 +43,7 @@ module Krane
 
         runner.run!(
           verify_result: options['verify-result'],
-          filename: options['filename'],
+          template: options['template'],
           command: options['command'],
           arguments: options['arguments']&.split(" "),
           env_vars: options['env-vars'].split(','),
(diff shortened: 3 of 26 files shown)


exe/kubernetes-restart
```diff
@@ -23,7 +23,7 @@ namespace = ARGV[0]
 context = ARGV[1]
 
 restart = KubernetesDeploy::RestartTask.new(namespace: namespace, context: context,
-   max_watch_seconds: max_watch_seconds)
+   global_timeout: global_timeout)
```
lei-lo: Doesn't this need to be `global_timeout: max_watch_seconds`?

And do we need to update this line as well? If yes, it's also still present in the `deploy` and `run` exes. https://github.com/Shopify/kubernetes-deploy/blob/6c6f930cecf3fda21b6fb57cca4f7ab39455d0c9/exe/kubernetes-restart#L15

dirceu: We don't need to update the flag name here, as this will be removed in https://github.com/Shopify/kubernetes-deploy/pull/622.

dirceu: Good catch on the wrong variable name!

lib/krane/deprecated_deploy_task.rb
```diff
@@ -112,24 +112,24 @@ def server_version
...
+    # @param context [String] Kubernetes context (*required*)
     # @param current_sha [String] The SHA of the commit
     # @param logger [Object] Logger object (defaults to an instance of Krane::FormattedLogger)
     # @param kubectl_instance [Kubectl] Kubectl instance
     # @param bindings [Hash] Bindings parsed by Krane::BindingsParser
-    # @param max_watch_seconds [Integer] Timeout in seconds
+    # @param global_timeout [Integer] Timeout in seconds
     # @param selector [Hash] Selector(s) parsed by Krane::LabelSelector
-    # @param template_paths [Array<String>] An array of template paths
+    # @param filenames [Array<String>] An array of template paths (*required*)
```
timothysmith0609:
```suggestion
# @param filenames [Array<String>] An array of filenames and/or directories containing templates (*required*)
```

lib/krane/global_deploy_task.rb
```diff
@@ -29,10 +29,10 @@ class GlobalDeployTask
 
     # Initializes the deploy task
     #
-    # @param context [String] Kubernetes context
+    # @param context [String] Kubernetes context (*required*)
     # @param global_timeout [Integer] Timeout in seconds
-    # @param selector [Hash] Selector(s) parsed by Krane::LabelSelector
-    # @param template_paths [Array<String>] An array of template paths
+    # @param selector [Hash] Selector(s) parsed by Krane::LabelSelector (*required*)
+    # @param filenames [Array<String>] An array of template paths (*required*)
```
timothysmith0609:
```suggestion
# @param filenames [Array<String>] An array of filenames and/or directories containing templates (*required*)
```

lib/krane/cli/deploy_command.rb
```diff
@@ -60,10 +60,10 @@ def self.from_options(namespace, context, options)
             namespace: namespace,
             context: context,
             current_sha: options['current-sha'],
-            template_paths: paths,
+            filenames: paths,
```
dturn: Might as well rename `paths` to filenames too

lib/krane/cli/deploy_command.rb
```diff
@@ -60,10 +60,10 @@ def self.from_options(namespace, context, options)
...
             bindings: bindings_parser.parse,
             logger: logger,
-            max_watch_seconds: ::Krane::DurationParser.new(options["global-timeout"]).parse!.to_i,
+            global_timeout: ::Krane::DurationParser.new(options["global-timeout"]).parse!.to_i,
             selector: selector,
             protected_namespaces: protected_namespaces,
           )
```
dturn: I can't comment on line 73. But should `allow_protected_ns: !protected_namespaces.empty?,` be renamed too?

lib/krane/cli/restart_command.rb
```diff
@@ -21,7 +21,7 @@ def self.from_options(namespace, context, options)
         restart = ::Krane::RestartTask.new(
           namespace: namespace,
           context: context,
-          max_watch_seconds: ::Krane::DurationParser.new(options["global-timeout"]).parse!.to_i,
+          global_timeout: ::Krane::DurationParser.new(options["global-timeout"]).parse!.to_i,
         )
         restart.run!(
           options[:deployments],
```
dturn: should this be a keyword?

lib/krane/cli/run_command.rb
```diff
@@ -37,14 +37,14 @@ def self.from_options(namespace, context, options)
...
-          max_watch_seconds: ::Krane::DurationParser.new(options["global-timeout"]).parse!.to_i,
+          global_timeout: ::Krane::DurationParser.new(options["global-timeout"]).parse!.to_i,
         )
 
         runner.run!(
           verify_result: options['verify-result'],
-          task_template: options['template'],
-          entrypoint: options['command'],
-          args: options['arguments']&.split(" "),
+          template: options['template'],
```
dturn: `filename`?

dirceu:
Added a commit with changes to the migration guide.

lib/krane/cli/deploy_command.rb
dirceu: This actually could be removed from the public interface, right? This is maintained as an argument mostly for backwards compatibility with `kubernetes-deploy`.

dturn: `allow_protected_ns` can be removed.

dirceu: OK, will remove on https://github.com/Shopify/kubernetes-deploy/pull/622.

lib/krane/restart_task.rb
```diff
@@ -49,18 +49,18 @@ def run(*args)
...
     # Runs the task, raising exceptions in case of issues
     #
-    # @param deployments_names [Array<String>] Array of workload names to restart
+    # @param deployments [Array<String>] Array of workload names to restart
     # @param selector [Hash] Selector(s) parsed by Krane::LabelSelector
     # @param verify_result [Boolean] Wait for completion and verify success
     #
     # @return [nil]
-    def run!(deployments_names = nil, selector: nil, verify_result: true)
+    def run!(deployments: nil, selector: nil, verify_result: true)
```
dturn: I think deployments is the right name here. There was talk about supporting DS's as well, under a new keyword. I think that's better than a generic `resources` where we have to manually fetch both and check to see what exists.

lib/krane/render_task.rb
```diff
@@ -12,13 +12,13 @@ class RenderTask
     #
     # @param logger [Object] Logger object (defaults to an instance of Krane::FormattedLogger)
     # @param current_sha [String] The SHA of the commit
-    # @param template_dir [String] Path to a directory with templates to render (deprecated)
-    # @param template_paths [Array<String>] An array of template paths to render
+    # @param template_dir [String] Path to a directory with templates to render (*deprecated*)
```
dturn: Are you 🔥 this in a different PR?

dirceu: Yes, on https://github.com/Shopify/kubernetes-deploy/pull/622

tsontario:
WDYT about the fact we still have `OptionsHelper::with_process_template_paths` and `TemplateSets`. Do these terms still make sense with the `filename(s)` semantics?

dirceu:
Thanks for the rebase! I think it makes sense; I'll just send a separate commit bringing back the information about the removed arguments in the tables we're using to describe the renames.

> WDYT about the fact we still have OptionsHelper::with_process_template_paths and TemplateSets. Do these terms still make sense with the filename(s) semantics?

Hmmm... well, the filename is the name of the file containing a template, so I think that TemplateSets still make sense. We could arguably change the OptionsHelper method but I wouldn't consider this a blocker for this issue, as our main goal here is to update the public interface of the Task classes. Unless you don't feel strongly that we should change this in this PR, I'd rather make this change afterwards.

dturn:
> We could arguably change the OptionsHelper method but I wouldn't consider this a blocker for this issue, as our main goal here is to update the public interface of the Task classes. Unless you feel strongly that we should change this in this PR, I'd rather make this change afterwards.

Added to the tech debt tracking issue: 
https://github.com/Shopify/kubernetes-deploy/issues/637

lib/krane/cli/run_command.rb
```diff
@@ -19,10 +19,11 @@ class RunCommand
         },
         "verify-result" => { type: :boolean, desc: "Wait for completion and verify pod success", default: true },
         "command" => { type: :array, desc: "Override the default command in the container image" },
-        "template" => {
+        "filename" => {
```
dturn: I thought we decided against this https://github.com/Shopify/kubernetes-deploy/issues/568

dirceu: Good catch, thank you! I wasn't aware of that. I'll update this PR undoing this particular change.

dirceu: Done. This is ready for another review.

PR #626, opened 2019-11-13
Remove rendering from krane deploy

dturn:
Remove the ability of `krane deploy` to render erb

**How is this accomplished?**
Remove the flags, remove the args to new/run, fix the tests.

**What could go wrong?**
- current-sha gets fed to statsd tags, we could leave the flag just for that, but I'd rather add an explicit flag for extra statsd tags if desired. 

- Make the migration to krane 1.0 harder.

### DIFF
--- a/lib/krane/cli/deploy_command.rb
+++ b/lib/krane/cli/deploy_command.rb
@@ -10,7 +10,7 @@ module Krane
         kube-public
       )
       OPTIONS = {
-        "filenames" => { type: :array, banner: 'config/deploy/production config/deploy/my-extra-resource.yml',
+        "filenames" => { type: :string, banner: 'config/deploy/production config/deploy/my-extra-resource.yml',
                          aliases: :f, required: true,
                          desc: "Directories and files that contains the configuration to apply" },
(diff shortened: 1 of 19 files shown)


lib/krane/cli/deploy_command.rb
```diff
@@ -10,10 +10,7 @@ class DeployCommand
         kube-public
       )
       OPTIONS = {
-        "bindings" => { type: :array, banner: "foo=bar abc=def",
-                        desc: "Expose additional variables to ERB templates (format: k1=v1 k2=v2, JSON string or file "\
-                          "(JSON or YAML) path prefixed by '@')" },
-        "filenames" => { type: :array, banner: 'config/deploy/production config/deploy/my-extra-resource.yml',
+        "filenames" => { type: :string, banner: 'config/deploy/production config/deploy/my-extra-resource.yml',
```
dturn: We didn't have a black box test for parsing `-f -`. Turns out that array input type can't have leading `-` or thor assumes its the start of a new flag :(

dturn:
This didn't make it into v1.0

Issue #628, opened 2019-11-14
`--std-in` flag


dturn:
We need to support std-in for both render and deploy. 

Thor can't parse `-` in an array type. We could switch `filenames` back to a string, but it can't have any whitespace in it. E.g. "file1.yml,file2.yml", this would be different from all other commands (its possible quoting might work, but that's ugly too).

To keep the ability to support std-in we'll add a new flag `--std-in` or possibly something with a better name. This will add the `-` directly to the list of templates passed to `with_processed_template_paths`.  This would be allowed with `--filenames` but one or the other must be passed. (The idea of implicitly using std-in when no --filenames is passed seemed like a bad idea. We also didn't like making up a new magic stdin character). 

This flag needs to be added to krane render, krane deploy, kubernetes-deploy, kubernetes-render. 

I'm open to alternative ideas, but this seems like the one with the least future restrictions on how `--filenames` will work.

dturn:
Done in #630

PR #630, opened 2019-11-14
Allow template filenames given from stdin stream

jessie-JNing:
work on: https://github.com/Shopify/kubernetes-deploy/issues/628

Allow template paths given from user std-in

**How is this accomplished?**
- Add `--stdin` flag to `krane deploy`, `krane global-deploy` and `krane render`
    - `--stdin` is a boolean type optional flag, whose default value is `false`.
    - If `--stdin` is given,  "-" will be added to template_paths

- For `-f/--filenames` and `--stdin`, at least one of them should be given, otherwise `RequiredArgumentMissingError` will be raised.

**Side note from the PR**
- Thor option should be **READ-ONLY**, never mutate it. We were blocked because of this, as we mutate the value of `option[:filenames]` by appending `-` to it, and this change was passed to option that we would use after.

### DIFF
--- a/lib/krane/cli/deploy_command.rb
+++ b/lib/krane/cli/deploy_command.rb
@@ -14,10 +14,10 @@ module Krane
                         desc: "Expose additional variables to ERB templates (format: k1=v1 k2=v2, JSON string or file "\
                           "(JSON or YAML) path prefixed by '@')" },
         "filenames" => { type: :array, banner: 'config/deploy/production config/deploy/my-extra-resource.yml',
-                         aliases: :f, required: true,
+                         aliases: :f, required: false, default: [],
                          desc: "Directories and files that contains the configuration to apply" },
         "std-in" => { type: :boolean, default: false,
-                     desc: "Have directories and files to render from stdin" },
+                      desc: "Read resources from stdin" },
         "global-timeout" => { type: :string, banner: "duration", default: DEFAULT_DEPLOY_TIMEOUT,
                               desc: "Max duration to monitor workloads correctly deployed" },
         "protected-namespaces" => { type: :array, banner: "namespace1 namespace2 namespaceN",
@@ -55,7 +55,12 @@ module Krane
         if options['protected-namespaces'].size == 1 && %w('' "").include?(options['protected-namespaces'][0])
           protected_namespaces = []
         end
+
         options[:filenames] << "-" if options['std-in']
+        if options[:filenames].empty?
+          raise Thor::RequiredArgumentMissingError, 'Must provied a value for --filenames or --std-in'
+        end
+
         ::Krane::OptionsHelper.with_processed_template_paths(options[:filenames],
           require_explicit_path: true) do |paths|
           deploy = ::Krane::DeployTask.new(
--- a/lib/krane/cli/deploy_command.rb
+++ b/lib/krane/cli/deploy_command.rb
@@ -56,12 +56,14 @@ module Krane
           protected_namespaces = []
         end
 
-        options[:filenames] << "-" if options[:stdin]
-        if options[:filenames].empty?
+        # never mutate options directly
+        filenames = options[:filenames].dup
+        filenames << "-" if options[:stdin]
+        if filenames.empty?
           raise Thor::RequiredArgumentMissingError, 'Must provied a value for --filenames or --stdin'
         end
 
-        ::Krane::OptionsHelper.with_processed_template_paths(options[:filenames],
+        ::Krane::OptionsHelper.with_processed_template_paths(filenames,
           require_explicit_path: true) do |paths|
           deploy = ::Krane::DeployTask.new(
             namespace: namespace,
--- a/lib/krane/cli/deploy_command.rb
+++ b/lib/krane/cli/deploy_command.rb
@@ -16,8 +16,8 @@ module Krane
         "filenames" => { type: :array, banner: 'config/deploy/production config/deploy/my-extra-resource.yml',
                          aliases: :f, required: false, default: [],
                          desc: "Directories and files that contains the configuration to apply" },
-        "std-in" => { type: :boolean, default: false,
-                      desc: "Read resources from stdin" },
+        "stdin" => { type: :boolean, default: false,
+                     desc: "Read resources from stdin" },
         "global-timeout" => { type: :string, banner: "duration", default: DEFAULT_DEPLOY_TIMEOUT,
                               desc: "Max duration to monitor workloads correctly deployed" },
         "protected-namespaces" => { type: :array, banner: "namespace1 namespace2 namespaceN",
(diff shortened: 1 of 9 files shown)


lib/krane/cli/deploy_command.rb
```diff
@@ -16,6 +16,8 @@ class DeployCommand
         "filenames" => { type: :array, banner: 'config/deploy/production config/deploy/my-extra-resource.yml',
                          aliases: :f, required: true,
                          desc: "Directories and files that contains the configuration to apply" },
+        "std-in" => { type: :boolean, default: false,
```
timothysmith0609: my preference is for the flag to be `--stdin` but I don't mind if I get outvoted

lib/krane/cli/deploy_command.rb
```diff
@@ -16,6 +16,8 @@ class DeployCommand
...
+                     desc: "Have directories and files to render from stdin" },
```
timothysmith0609:
```suggestion
                     desc: Read templates from stdin. May be combined with -f },
```

lib/krane/cli/deploy_command.rb
```diff
@@ -16,6 +16,8 @@ class DeployCommand
         "filenames" => { type: :array, banner: 'config/deploy/production config/deploy/my-extra-resource.yml',
                          aliases: :f, required: true,
```
timothysmith0609: This is no longer required if we can accept `STDIN`. However, we have to ensure that it is not empty after parsing options

lib/krane/cli/deploy_command.rb
```diff
@@ -53,7 +55,7 @@ def self.from_options(namespace, context, options)
         if options['protected-namespaces'].size == 1 && %w('' "").include?(options['protected-namespaces'][0])
           protected_namespaces = []
         end
-
+        options[:filenames] << "-" if options['std-in']
```
timothysmith0609: If/until we manage to fix the underlying issue with Thor forcing us to do this, this might be worth an explanatory comment. At first blush, this won't make sense to people unfamiliar with why we were compelled to do this.

test/exe/deploy_test.rb
```diff
@@ -16,6 +16,16 @@ def test_deploy_parses_global_timeout
...
   end
 
+  def test_deploy_parses_std_in
+    set_krane_deploy_expectations(new_args: { template_paths: ['/my/file/path', '-'] })
+    krane_deploy!(flags: '-f /my/file/path --std-in')
+  end
+
+  def test_deploy_parses_std_in_without_filenames
+    set_krane_deploy_expectations(new_args: { template_paths: ['-'] })
+    krane_deploy!(flags: '--std-in')
```
timothysmith0609: This will currently fail since `-f` is still marked as a required field, yielding the following err:
```bash
$ bundle exec krane deploy --std-in
No value provided for required options '--filenames'
```

test/exe/render_test.rb
```diff
@@ -17,6 +17,17 @@ def test_render_parses_paths
...
 
+  def test_render_parses_std_in
+    file_path = "/dev/null"
+    install_krane_render_expectations(template_paths: [file_path, "-"])
+    krane_render!("--filenames #{file_path} --std-in")
+  end
+
+  def test_render_parses_std_in_without_filenames
+    install_krane_render_expectations(template_paths: ["-"])
+    krane_render!("--std-in")
```
timothysmith0609: `-f` is likewise a required arg here too, still, so this will fail

test/exe/render_test.rb
```diff
@@ -17,6 +17,33 @@ def test_render_parses_paths
...
+  def test_render_parses_std_in_without_filenames
+    Dir.mktmpdir do |tmp_path|
+      $stdin.expects("read").returns("")
+      Dir.expects(:mktmpdir).with("krane").yields(tmp_path).once
+      install_krane_render_expectations(template_paths: [tmp_path])
+      krane_render!("--std-in")
+    end
+  end
+
+  def test_render_fails_without_filename_and_std_in
```
dturn: There is something flakey or not working about this test `rt test/exe/render_test.rb --seed 58960` reproduces it. 

Somehow render task is getting called with these options `{"filenames"=>["-"], "std-in"=>false, "current-sha"=>"1"}`

lib/krane/cli/deploy_command.rb
jessie-JNing: I'll change to use `--stdin`, since https://en.wikipedia.org/wiki/Standard_streams

lib/krane/cli/deploy_command.rb
jessie-JNing: an empty check `options[:filenames].empty?` is given, we'll raise Thor RequiredArgumentMissingError if it's empty

test/exe/global_deploy_test.rb
```diff
@@ -23,13 +23,38 @@ def test_deploy_passes_verify_result
...
   end
 
+  def test_deploy_parses_std_in_without_filenames
+    Dir.mktmpdir do |tmp_path|
+      $stdin.expects("read").returns("")
+      Dir.expects(:mktmpdir).with("krane").yields(tmp_path)
+      set_krane_global_deploy_expectations!(new_args: { filenames: [tmp_path] })
+      krane_global_deploy!(flags: '--stdin')
+    end
+  end
```
timothysmith0609: We need an integration test that actually reads something from STDIN (not just `""`) and check to make sure we output what we expect.

lib/krane/cli/deploy_command.rb
```diff
@@ -54,6 +56,11 @@ def self.from_options(namespace, context, options)
           protected_namespaces = []
         end
 
+        options[:filenames] << "-" if options[:stdin]
```
timothysmith0609: As discussed irl, we __must__ treat the `options` hash as `READ_ONLY`. Thus, maintain a local array for filenames instead of writing directly into `options[:filenames]`

test/exe/global_deploy_test.rb
timothysmith0609: Actually, we don't need an integration test. This, along with optionshelper tests are enough to prove this works as intended

lib/krane/cli/render_command.rb
```diff
@@ -18,7 +19,14 @@ def self.from_options(options)
         bindings_parser = ::Krane::BindingsParser.new
         options[:bindings]&.each { |b| bindings_parser.add(b) }
 
-        ::Krane::OptionsHelper.with_processed_template_paths(options[:filenames]) do |paths|
+        # never mutate options directly
+        filenames = options[:filenames].dup
+        filenames << "-" if options[:stdin]
+        if filenames.empty?
+          raise Thor::RequiredArgumentMissingError, 'Must provied a value for --filenames or --stdin'
```
timothysmith0609: nit: Bring this in line with the deploy_command

jessie-JNing: Oh, definitely should be changed. It will come in the next commit, together with lint fix :)

PR #631, opened 2019-11-14
Update the gemspec for Krane 1.0

dturn:
One of the last steps for releasing Krane 1.0.0 is updating the gemspec.

**How is this accomplished?**
Update the gemspec and use the existing description from https://rubygems.org/gems/krane

There is no rubygem rename ability, its just create a new one: https://help.rubygems.org/discussions/questions/104-how-should-i-handle-the-rename-of-a-gem

Shipit will automatically figure out the new gemspec. 

**What could go wrong?**
Not much, worst case we quickly yank 1.0.0 while we fix the issue and create it again.

### DIFF
--- a/krane.gemspec
+++ b/krane.gemspec
@@ -5,14 +5,15 @@ $LOAD_PATH.unshift(lib) unless $LOAD_PATH.include?(lib)
 require 'krane/version'
 
 Gem::Specification.new do |spec|
-  spec.name          = "kubernetes-deploy"
+  spec.name          = "krane"
   spec.version       = Krane::VERSION
-  spec.authors       = ["Katrina Verey", "Kir Shatrov"]
+  spec.authors       = ["Katrina Verey", "Daniel Turner", "Kir Shatrov"]
   spec.email         = ["ops-accounts+shipit@shopify.com"]
 
-  spec.summary       = 'Kubernetes deploy scripts'
+  spec.summary       = 'A command line tool that helps you ship changes to a Kubernetes' \
+    ' namespace and understand the result'
   spec.description   = spec.summary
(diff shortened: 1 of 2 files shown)


krane.gemspec
```diff
@@ -5,14 +5,15 @@ $LOAD_PATH.unshift(lib) unless $LOAD_PATH.include?(lib)
...
-  spec.authors       = ["Katrina Verey", "Kir Shatrov"]
+  spec.authors       = ["Katrina Verey", "Daniel Turner", "Kir Shatrov"]
   spec.email         = ["ops-accounts+shipit@shopify.com"]
 
-  spec.summary       = 'Kubernetes deploy scripts'
+  spec.summary       = 'A command line tool that helps you ship changes to a Kubernetes' \
+    ' namespace and understand the result'
   spec.description   = spec.summary
-  spec.homepage      = "https://github.com/Shopify/kubernetes-deploy"
+  spec.homepage      = "https://github.com/Shopify/krane"
```
douglas: Just to confirm, after the release are we going to rename the `kubernetes-deploy` to `krane` but are going to keep `kubernetes-deploy` repo around just with a README pointing to `krane` repo?

dturn: We're going to rename the repo which should cause github to automatically redirect people. We'll also make sure the top of the readme points mentions the rename and points to the migration guide

tsontario:
Can/do we document the yanking process anywhere? We've had a few releases go out that should be yanked (0.31.0 most recently) but I have no clue how to do this

dturn:
> Can/do we document the yanking process anywhere? 

We don't. And as far as I know we don't have any automation around it either.

PR #632, opened 2019-11-15
Remove rendering from krane deploy cli

dturn:
Remove the ability of krane deploy cli to render erb

Ideally the deploy task will not render erb either, #626 is a start down that path, but its a large change and one we likely don't have time to ship before 1.0 this is a stop-gap idea. 

**How is this accomplished?**
Remove the flags, allow current_sha to be nil

**What could go wrong?**

* current-sha gets fed to statsd tags, we could leave the flag just for that, but I'd rather add an explicit flag for extra statsd tags if desired.

* We have to keep the rendering in the task, though removing it later is less difficult than removing it from the cli tool later.

### DIFF
--- a/lib/krane/cli/deploy_command.rb
+++ b/lib/krane/cli/deploy_command.rb
@@ -10,9 +10,6 @@ module Krane
         kube-public
       )
       OPTIONS = {
-        "bindings" => { type: :array, banner: "foo=bar abc=def",
-                        desc: "Expose additional variables to ERB templates (format: k1=v1 k2=v2, JSON string or file "\
-                          "(JSON or YAML) path prefixed by '@')" },
         "filenames" => { type: :array, banner: 'config/deploy/production config/deploy/my-extra-resource.yml',
                          aliases: :f, required: false, default: [],
                          desc: "Directories and files that contains the configuration to apply" },
@@ -26,25 +23,18 @@ module Krane
                                     default: PROTECTED_NAMESPACES },
         "prune" => { type: :boolean, desc: "Enable deletion of resources that do not appear in the template dir",
                      default: true },
-        "render-erb" => { type: :boolean, desc: "Enable ERB rendering", default: false },
         "selector" => { type: :string, banner: "'label=value'",
                         desc: "Select workloads by selector(s)" },
         "verbose-log-prefix" => { type: :boolean, desc: "Add [context][namespace] to the log prefix",
                                   default: true },
         "verify-result" => { type: :boolean, default: true,
                              desc: "Verify workloads correctly deployed" },
-        "current-sha" => { type: :string, banner: "SHA", desc: "Expose SHA `current_sha` in ERB bindings" },
-
       }
 
       def self.from_options(namespace, context, options)
         require 'krane/deploy_task'
         require 'krane/options_helper'
-        require 'krane/bindings_parser'
         require 'krane/label_selector'
-
-        bindings_parser = ::Krane::BindingsParser.new
-        options[:bindings]&.each { |binding_pair| bindings_parser.add(binding_pair) }
 
         selector = ::Krane::LabelSelector.parse(options[:selector]) if options[:selector]
 
@@ -67,9 +57,7 @@ module Krane
           deploy = ::Krane::DeployTask.new(
             namespace: namespace,
             context: context,
-            current_sha: options['current-sha'],
             filenames: paths,
-            bindings: bindings_parser.parse,
             logger: logger,
             global_timeout: ::Krane::DurationParser.new(options["global-timeout"]).parse!.to_i,
             selector: selector,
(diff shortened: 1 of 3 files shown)


lib/krane/deprecated_deploy_task.rb
```diff
@@ -125,7 +125,7 @@ def server_version
     # @param protected_namespaces [Array<String>] Array of protected Kubernetes namespaces (defaults
     #   to Krane::DeployTask::PROTECTED_NAMESPACES)
     # @param render_erb [Boolean] Enable ERB rendering
-    def initialize(namespace:, context:, current_sha:, logger: nil, kubectl_instance: nil, bindings: {},
+    def initialize(namespace:, context:, current_sha: nil, logger: nil, kubectl_instance: nil, bindings: {},
       max_watch_seconds: nil, selector: nil, template_paths: [], template_dir: nil, protected_namespaces: nil,
       render_erb: true, allow_globals: false)
       template_dir = File.expand_path(template_dir) if template_dir
```
timothysmith0609: At some point we call `TemplateSets::from_dirs_and_files` on the input, what do we do if there are `.erb` files included in the set? We can either ignore (I think that's wrong) or raise (fast, before waiting for a downstream error). Unfortunately, the library code still supports `current_sha` and `bindings`, so it _is_ still valid to pass in ERB... do we need to scan at the CLI command level for appropriate templates, at least in the short term?

Now that `deploy` doesn't render, there's no longer a 1-1 correspondence between it and the `render` task as far as `TemplateSets` goes.

test/exe/deploy_test.rb
```diff
@@ -79,12 +68,6 @@ def test_deploy_fails_without_filename
     end
   end
 
-  def test_deploy_uses_current_sha
-    test_sha = "TEST"
-    set_krane_deploy_expectations(new_args: { current_sha: test_sha })
-    krane_deploy!(flags: "--current-sha #{test_sha}")
-  end
```
timothysmith0609: Do we need a test to figure out what happens when we pass an ERB template to deploy (e.g. extension `.yml.erb`)?

lib/krane/cli/deploy_command.rb
```diff
@@ -24,26 +21,19 @@ class DeployCommand
...
                      default: true },
-        "render-erb" => { type: :boolean, desc: "Enable ERB rendering", default: false },
         "selector" => { type: :string, banner: "'label=value'",
                         desc: "Select workloads by selector(s)" },
         "verbose-log-prefix" => { type: :boolean, desc: "Add [context][namespace] to the log prefix",
                                   default: true },
         "verify-result" => { type: :boolean, default: true,
                              desc: "Verify workloads correctly deployed" },
-        "current-sha" => { type: :string, banner: "SHA", desc: "Expose SHA `current_sha` in ERB bindings" },
-
```
timothysmith0609: As mentioned IRL, I think we should remove `current-sha` from `DeprecatedDeployTask#statsd_tags` and `ResourceWatcher#statsd_tags`, due to cardinality concerns.

lib/krane/deprecated_deploy_task.rb
dturn: So this isn't a new failure case, you can set `render_erb: false` and pass in erb templates. The don't get rendered and will fail likely fail on apply unless somehow your erb is valid yaml&k8s. 

I agree this is something we should fail fast on, the right path looks like passing `render_erb` to the TemplateSet#validtate  and raising there. 🤞 that doesn't cause test failures.

test/exe/deploy_test.rb
dturn: Covered in https://github.com/Shopify/kubernetes-deploy/pull/635/files#diff-ab79bd0231cddd4a971c2848fad61e1cL1647

lib/krane/deprecated_deploy_task.rb
dturn: Error messaging is coming in https://github.com/Shopify/kubernetes-deploy/pull/635/files

test/integration/krane_test.rb
```diff
@@ -53,8 +53,8 @@ def test_render_black_box
   end
 
   def test_deploy_black_box_success
-    setup_template_dir("hello-cloud") do |target_dir|
-      flags = "-f #{target_dir} --render-erb --bindings deployment_id=1 current_sha=123"
+    setup_template_dir("hello-cloud", subset: %w(bare_replica_set.yml)) do |target_dir|
```
douglas: Out of curiosity, why the change to `subset:`?

dturn: hello-cloud has several fixtures that require erb.

douglas: Cool! Thanks for the clarification, Danny!

dturn:
> Should #635 rebase on this branch and then we just merge that to master? Either way I'm ok with this

I think they should stay as 2 separate PRs. #635 is something we'd want even if we don't ship this PR.

PR #635, opened 2019-11-15
Fail fast when deploying with erb

RyanBrushett:

Fail fast when one tries to deploy with ERB

**How is this accomplished?**

Check if they're trying to deploy with ERB and if they are, fail the deploy.

### DIFF
--- a/lib/krane/template_sets.rb
+++ b/lib/krane/template_sets.rb
@@ -9,14 +9,15 @@ module Krane
     # private inner class
     class TemplateSet
       include DelayedExceptions
-      def initialize(template_dir:, file_whitelist: [], logger:)
+      def initialize(template_dir:, file_whitelist: [], logger:, render_erb: true)
         @template_dir = template_dir
         @files = file_whitelist
         @logger = logger
+        @render_erb = render_erb
       end
 
-      def with_resource_definitions_and_filename(render_erb: false, current_sha: nil, bindings: nil, raw: false)
-        if render_erb
+      def with_resource_definitions_and_filename(current_sha: nil, bindings: nil, raw: false)
+        if @render_erb
           @renderer = Renderer.new(
             template_dir: @template_dir,
             logger: @logger,
@@ -39,27 +40,43 @@ module Krane
 
       def validate
         errors = []
+        supported_extensions = if @render_erb
+          TemplateSets::VALID_TEMPLATES
+        else
+          TemplateSets::VALID_TEMPLATES.reject { |extension| extension.include?('erb') }
+        end
+
         if Dir.entries(@template_dir).none? do |filename|
-             filename.end_with?(*TemplateSets::VALID_TEMPLATES) ||
+             filename.end_with?(*supported_extensions) ||
              filename.end_with?(EjsonSecretProvisioner::EJSON_SECRETS_FILE)
            end
-          return errors << "Template directory #{@template_dir} does not contain any valid templates"
+          return errors << "Template directory #{@template_dir} does not contain any valid templates (supported " \
+              "suffixes: #{supported_extensions.join(', ')}, or #{EjsonSecretProvisioner::EJSON_SECRETS_FILE})"
+        end
+
+        if !@render_erb && deploying_with_erb_files?
+          errors << "ERB template discovered with rendering disabled. If you were trying to render ERB and " \
+            "deploy the result, try piping the output of `krane render` to `krane-deploy` with the --stdin flag"
         end
 
         @files.each do |filename|
           filename = File.join(@template_dir, filename)
           if !File.exist?(filename)
             errors << "File #{filename} does not exist"
-          elsif !filename.end_with?(*TemplateSets::VALID_TEMPLATES) &&
+          elsif !filename.end_with?(*supported_extensions) &&
                 !filename.end_with?(EjsonSecretProvisioner::EJSON_SECRETS_FILE)
             errors << "File #{filename} does not have valid suffix (supported suffixes: " \
-              "#{TemplateSets::VALID_TEMPLATES.join(', ')}, or #{EjsonSecretProvisioner::EJSON_SECRETS_FILE})"
+              "#{supported_extensions.join(', ')}, or #{EjsonSecretProvisioner::EJSON_SECRETS_FILE})"
           end
         end
         errors
       end
 
       private
+
+      def deploying_with_erb_files?
+        @files.any? { |file| file.end_with?("erb") }
+      end
 
       def templates(filename:, raw:)
         file_content = File.read(File.join(@template_dir, filename))
@@ -83,7 +100,7 @@ module Krane
     private_constant :TemplateSet
 
     class << self
-      def from_dirs_and_files(paths:, logger:)
+      def from_dirs_and_files(paths:, logger:, render_erb: true)
         resource_templates = {}
         dir_paths, file_paths = paths.partition { |path| File.directory?(path) }
 
@@ -103,16 +120,16 @@ module Krane
 
         template_sets = []
         resource_templates.each do |template_dir, files|
-          template_sets << TemplateSet.new(template_dir: template_dir, file_whitelist: files, logger: logger)
+          template_sets << TemplateSet.new(template_dir: template_dir, file_whitelist: files, logger: logger,
+            render_erb: render_erb)
         end
         TemplateSets.new(template_sets: template_sets)
       end
     end
 
-    def with_resource_definitions_and_filename(render_erb: false, current_sha: nil, bindings: nil, raw: false)
+    def with_resource_definitions_and_filename(current_sha: nil, bindings: nil, raw: false)
       with_delayed_exceptions(@template_sets, Krane::InvalidTemplateError) do |template_set|
         template_set.with_resource_definitions_and_filename(
-          render_erb: render_erb,
           current_sha: current_sha,
           bindings: bindings,
           raw: raw
@@ -122,9 +139,8 @@ module Krane
       end
     end
 
-    def with_resource_definitions(render_erb: false, current_sha: nil, bindings: nil, raw: false)
-      with_resource_definitions_and_filename(render_erb: render_erb,
-        current_sha: current_sha, bindings: bindings, raw: raw) do |r_def, _|
+    def with_resource_definitions(current_sha: nil, bindings: nil, raw: false)
+      with_resource_definitions_and_filename(current_sha: current_sha, bindings: bindings, raw: raw) do |r_def, _|
         yield r_def
       end
     end
--- a/lib/krane/template_sets.rb
+++ b/lib/krane/template_sets.rb
@@ -9,6 +9,7 @@ module Krane
     # private inner class
     class TemplateSet
       include DelayedExceptions
+      attr_reader :render_erb
       def initialize(template_dir:, file_whitelist: [], logger:, render_erb: true)
         @template_dir = template_dir
         @files = file_whitelist
@@ -54,11 +55,6 @@ module Krane
               "suffixes: #{supported_extensions.join(', ')}, or #{EjsonSecretProvisioner::EJSON_SECRETS_FILE})"
         end
 
-        if !@render_erb && deploying_with_erb_files?
-          errors << "ERB template discovered with rendering disabled. If you were trying to render ERB and " \
-            "deploy the result, try piping the output of `krane render` to `krane-deploy` with the --stdin flag"
-        end
-
         @files.each do |filename|
           filename = File.join(@template_dir, filename)
           if !File.exist?(filename)
@@ -72,11 +68,11 @@ module Krane
         errors
       end
 
-      private
-
       def deploying_with_erb_files?
         @files.any? { |file| file.end_with?("erb") }
       end
+
+      private
 
       def templates(filename:, raw:)
         file_content = File.read(File.join(@template_dir, filename))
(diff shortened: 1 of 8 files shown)


test/integration/krane_deploy_test.rb
```diff
@@ -1712,6 +1712,18 @@ def test_deploy_resource_using_generate_name_succeeds
     ])
   end
 
+  def test_deploy_fails_fast_when_erb_is_present
```
dturn: How is this different from https://github.com/Shopify/kubernetes-deploy/pull/635/files#diff-ab79bd0231cddd4a971c2848fad61e1cL1647 ?

lib/krane/template_sets.rb
```diff
@@ -37,23 +37,30 @@ def ejson_secrets_file
         end
       end
 
-      def validate
+      def validate(render_erb: true)
         errors = []
+        supported_extensions = if render_erb
+          TemplateSets::VALID_TEMPLATES
+        else
+          TemplateSets::VALID_TEMPLATES.select { |extension| !extension.include?('erb') }
```
douglas: Maybe we can use this instead?

```suggestion
          TemplateSets::VALID_TEMPLATES.reject { |extension| extension.include?('erb') }
```

test/integration/krane_deploy_test.rb
douglas: hmmm, `test-wise` they seems to test the same thing, if I'm not misreading it I feel we should delete this test.

lib/krane/template_sets.rb
```diff
@@ -37,23 +37,30 @@ def ejson_secrets_file
...
           filename = File.join(@template_dir, filename)
           if !File.exist?(filename)
             errors << "File #{filename} does not exist"
-          elsif !filename.end_with?(*TemplateSets::VALID_TEMPLATES) &&
+          elsif !filename.end_with?(*supported_extensions) &&
                 !filename.end_with?(EjsonSecretProvisioner::EJSON_SECRETS_FILE)
             errors << "File #{filename} does not have valid suffix (supported suffixes: " \
-              "#{TemplateSets::VALID_TEMPLATES.join(', ')}, or #{EjsonSecretProvisioner::EJSON_SECRETS_FILE})"
+              "#{supported_extensions.join(', ')}, or #{EjsonSecretProvisioner::EJSON_SECRETS_FILE})"
           end
```
tsontario: What do you think about explicitly handling the case when `render_erb: false` and we find `.erb` files? This would be a good place to give a hint to use `krane render | krane deploy`

test/integration/krane_test.rb
```diff
@@ -3,7 +3,9 @@
 
 class KraneTest < Krane::IntegrationTest
   def test_restart_black_box
-    assert_deploy_success(deploy_fixtures("hello-cloud", subset: ["configmap-data.yml", "web.yml.erb", "redis.yml"]))
+    assert_deploy_success(
+      deploy_fixtures("hello-cloud", subset: ["configmap-data.yml", "web.yml.erb", "redis.yml"], render_erb: true)
```
tsontario: @dturn is there an open issue to refactor tests to reflect the new 'deploy can't render' world? It feels wrong to be able to set `render_erb` here (though I understand now is probably not the time to fix this)

test/integration/krane_test.rb
```diff
@@ -53,8 +55,8 @@ def test_render_black_box
   end
 
   def test_deploy_black_box_success
-    setup_template_dir("hello-cloud") do |target_dir|
-      flags = "-f #{target_dir} --render-erb --bindings deployment_id=1 current_sha=123"
+    setup_template_dir("hello-cloud", subset: %w(bare_replica_set.yml)) do |target_dir|
+      flags = "-f #{target_dir} --bindings deployment_id=1 current_sha=123"
```
tsontario: #632 is removing `bindings / current-sha` flag from deploy and the subset you specify doesn't need them anyway /

test/integration/krane_test.rb
```diff
@@ -71,8 +73,8 @@ def test_deploy_black_box_failure
   end
 
   def test_deploy_black_box_timeout
-    setup_template_dir("hello-cloud") do |target_dir|
-      flags = "-f #{target_dir} --render-erb --bindings deployment_id=1 current_sha=123 --global-timeout=1s"
+    setup_template_dir("hello-cloud", subset: %w(bare_replica_set.yml)) do |target_dir|
+      flags = "-f #{target_dir} --bindings deployment_id=1 current_sha=123 --global-timeout=0.1s"
```
tsontario: Same as above

test/unit/krane/template_sets_test.rb
```diff
@@ -16,7 +16,8 @@ def test_empty_template_sets_directory_is_invalid
     Dir.mktmpdir("empty_dir") do |dir|
       template_sets = template_sets_from_paths(dir)
       expected = [
-        "Template directory #{dir} does not contain any valid templates",
+        "Template directory #{dir} does not contain any valid templates " \
+          "(supported suffixes: .yml.erb, .yml, .yaml, .yaml.erb, or secrets.ejson)",
       ]
       assert_equal(template_sets.validate, expected)
```
tsontario: if `render_erb: false`, do we still say that *.erb suffixes are supported?

lib/krane/deprecated_deploy_task.rb
```diff
@@ -290,7 +290,7 @@ def validate_configuration(allow_protected_ns:, prune:)
         @task_config, kubectl, kubeclient_builder)
       errors = []
       errors += task_config_validator.errors
-      errors += @template_sets.validate
+      errors += @template_sets.validate(render_erb: @render_erb)
```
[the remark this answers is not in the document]
RyanBrushett: That's a great point. Will look.

test/integration/krane_deploy_test.rb
RyanBrushett: Yeah oops! I figured I'd add a test for my change, didn't check if there was one there already. My bad!

test/integration/krane_test.rb
RyanBrushett: No but I had planned on opening one.

test/unit/krane/template_sets_test.rb
RyanBrushett: Nope, we adjust the list of supported suffixes as needed.

test/unit/krane/template_sets_test.rb
```diff
@@ -32,6 +33,16 @@ def test_template_sets_with_invalid_suffix_is_invalid
...
 
+  def erb_is_not_considered_valid_when_render_erb_is_false
+    bad_filepath = File.join(fixture_path("for_unit_tests"), "partials_test.yaml.erb")
+    template_sets = template_sets_from_paths(bad_filepath)
+    expected = [
+      "File #{bad_filepath} does not have valid suffix (supported suffixes: " \
+        ".yml, .yaml, or secrets.ejson)",
+    ]
+    assert_equal(template_sets.validate, expected)
+  end
```
timothysmith0609: Functions that don't start with `test_` are not actually run in the test suite. Furthermore, this test currently fails, since `render_erb` defaults to true, there are no errors when rendering this. 

Furthermore, when we do explicitly set `render_erb: false`, the expected error doesn't match actual:

```
Failure:
TemplateSetsTest#test_erb_is_not_considered_valid_when_render_erb_is_false [/Users/timothysmith/src/github.com/Shopify/kubernetes-deploy/test/unit/krane/template_sets_test.rb:44]
Minitest::Assertion: --- expected
+++ actual
@@ -1 +1 @@
-["Rendering erb during deploy is not supported. Try piping the output of `krane render` to `krane-deploy` with the --stdin flag"]
+["File /Users/timothysmith/src/github.com/Shopify/kubernetes-deploy/test/fixtures/for_unit_tests/partials_test.yaml.erb does not have valid suffix (supported suffixes: .yml, .yaml, or secrets.ejson)"]
```

RyanBrushett: :facepalm: Will fix this as soon as I have a coffee.

test/unit/krane/template_sets_test.rb
```diff
@@ -32,6 +33,16 @@ def test_template_sets_with_invalid_suffix_is_invalid
...
   end
 
+  def test_template_sets_with_erb_files_are_considered_invalid_when_render_erb_is_false
+    bad_filepath = File.join(fixture_path("for_unit_tests"), "partials_test.yaml.erb")
+    template_sets = template_sets_from_paths(bad_filepath, render_erb: false)
+    expected = [
+      "Rendering erb during deploy is not supported. " \
+        "Try piping the output of `krane render` to `krane-deploy` with the --stdin flag",
+    ]
+    assert_equal(template_sets.validate, expected)
```
timothysmith0609: I'm ok with this for now, but would prefer if, in the near future, we change this to be more in line with the 
```
File #{filename} does not have valid suffix (supported suffixes: ...
```
error (along with the hint to use `krane render`, to maintain consistency and also to enumerate the 'bad' files to the user

RyanBrushett: I'm here now, I'll just do it now!

lib/krane/template_sets.rb
```diff
@@ -39,21 +40,31 @@ def ejson_secrets_file
...
+              "suffixes: #{supported_extensions.join(', ')}, or #{EjsonSecretProvisioner::EJSON_SECRETS_FILE})"
         end
 
         @files.each do |filename|
           filename = File.join(@template_dir, filename)
           if !File.exist?(filename)
             errors << "File #{filename} does not exist"
-          elsif !filename.end_with?(*TemplateSets::VALID_TEMPLATES) &&
+          elsif filename.end_with?("erb") && !@render_erb
+            errors << "Rendering erb during deploy is not supported. Try piping the output of `krane render` "\
```
dturn: Can we ensure we only emit the error once? I'm also worried this error is a bit to specific, e.g. if its triggered from a different situation it would be rather confusing. Maybe something slightly more generic: 
"ERB template discovered with rendering disabled. If you were trying to deploy try piping..."

Screen Shot 2019-11-17 at 9 30 48 AM

(^ doesn't happen right now since erb is enabled from deploy, but it would with 632)

RyanBrushett: Yep on it now.

lib/krane/template_sets.rb
```diff
@@ -39,28 +40,44 @@ def ejson_secrets_file
...
              filename.end_with?(EjsonSecretProvisioner::EJSON_SECRETS_FILE)
            end
-          return errors << "Template directory #{@template_dir} does not contain any valid templates"
+          return errors << "Template directory #{@template_dir} does not contain any valid templates (supported " \
+              "suffixes: #{supported_extensions.join(', ')}, or #{EjsonSecretProvisioner::EJSON_SECRETS_FILE})"
+        end
+
+        if !@render_erb && deploying_with_erb_files?
+          errors << "ERB template discovered with rendering disabled. If you were trying to render ERB and " \
+            "deploy the result, try piping the output of `krane render` to `krane-deploy` with the --stdin flag"
```
timothysmith0609: To truly only raise this message once, we need to move the check out of `TemplateSet#validate` and into `TemplateSets#validate`, no?

RyanBrushett: Pushed up what I think is a fix. Have a look please!

lib/krane/template_sets.rb
```diff
@@ -72,12 +68,12 @@ def validate
         errors
       end
 
-      private
-
       def deploying_with_erb_files?
         @files.any? { |file| file.end_with?("erb") }
       end
 
+      private
```
RyanBrushett: Why did I add this?

RyanBrushett: Never mind it's git showing the diff weirdly

lib/krane/template_sets.rb
```diff
@@ -134,11 +146,26 @@ def ejson_secrets_files
...
     def validate
-      @template_sets.flat_map(&:validate)
+      errors = @template_sets.flat_map(&:validate)
+
+      if rendering_erb_disabled? && deploying_with_erb_files?
+        errors << "ERB template discovered with rendering disabled. If you were trying to render ERB and " \
+          "deploy the result, try piping the output of `krane render` to `krane-deploy` with the --stdin flag"
+      end
+
+      errors
```
timothysmith0609: ❤️

lib/krane/template_sets.rb
```diff
@@ -134,11 +146,26 @@ def ejson_secrets_files
...
     private
 
+    def deploying_with_erb_files?
+      @template_sets.any?(&:deploying_with_erb_files?)
+    end
+
+    def rendering_erb_disabled?
+      !@template_sets.any?(&:render_erb)
+    end
+
```
timothysmith0609: I feel the `TemplateSets` object should already know whether or not it allows its constituent template sets to render erb, it feels somewhat backwards to make it inspect them. Can you think of a case where we have some template sets that allow rendering while at the same time disallowing others?

RyanBrushett: No not at all.

dturn: In my experience when you have a collection class that owns an inner class you want one of them to be as stateless as possible, otherwise reasoning becomes really hard. We've picked the outer class to be the simple one so far. We definitely could have TemplateSets own `@render_erb` but it would be passing it to 2/4 of TemplateSet's public methods which feels a bit off.

timothysmith0609: For a given template set, we can ask “does this have a @renderer?”. If no, it can’t render anything. Would this be a sufficient check?

Issue #637, opened 2019-11-16
Krane 1.0 Release tech debt


dturn:
- [ ] PVC issue in tests
- [ ] Deploy tests that require erb rendering

airhorns:
Just for the record it'd be great to get that gem version pulled, I was unknowingly stuck with it in my bundle for a while and it doesn't include the `krane` executable which was very confusing. It also will be the gem version that anyone using `fog-google` gets stuck with until `fog-google` gets a release and something around https://github.com/Shopify/krane/pull/669 gets done methinks!

KnVerey:
Thanks for the reminder @airhorns . krane 0.27.0.pre has now been yanked:

![image](https://user-images.githubusercontent.com/4789493/72743173-2705e200-3b79-11ea-9fc4-9aa6657be9dc.png)

PR #638, opened 2019-11-18
[bug fix] Render corner cases

dturn:

1. Disable rendering by default in DeployTask.

2. Fix bugs with the render task:

(a) Thor defaults the value of a flag with no values passes as the flag name, e.g. `--current-sha ` sets the value of of `options['current-sha'] == "current-sha"`
(b) `krane render --std-in` doesn't actually render

**How is this accomplished?**
1. flip the default to false
2.a set a lazy default
2.b pass information `OptionsHelper` that the file name for data from stdin needs to end in `.erb`

**What could go wrong?**

Not much with the bug fixes. The default flip is part of the 1.0 breaking changes.

### DIFF
--- a/lib/krane/render_task.rb
+++ b/lib/krane/render_task.rb
@@ -91,6 +91,10 @@ module Krane
       if @filenames.blank?
         errors << "filenames must be set"
       end
+
+      if !@current_sha.nil? && @current_sha.empty?
+        errors << "current-sha is optional but can not be blank"
+      end
       errors += template_sets.validate
 
       unless errors.empty?
--- a/test/test_helper.rb
+++ b/test/test_helper.rb
@@ -227,9 +227,9 @@ module Krane
       @task_config ||= Krane::TaskConfig.new(context, namespace, logger)
     end
 
-    def krane_black_box(command, args = "")
+    def krane_black_box(command, args = "", stdin: nil)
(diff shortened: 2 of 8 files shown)


lib/krane/render_task.rb
```diff
@@ -91,6 +91,10 @@ def validate_configuration(template_sets)
       if @filenames.blank?
         errors << "filenames must be set"
       end
+
+      if !@current_sha.nil? && @current_sha.empty?
+        errors << "current-sha is optional but can not be blank"
+      end
```
dirceu: This (and the `@filenames` check right before it) might be better suited to a `TaskConfigValidator` subclass, if we want to follow the same pattern we used with the others tasks; I think it's totally fine to move that in another PR though.

test/test_helper.rb
```diff
@@ -227,9 +227,9 @@ def task_config(context: KubeclientHelper::TEST_CONTEXT, namespace: @namespace,
       @task_config ||= Krane::TaskConfig.new(context, namespace, logger)
     end
 
-    def krane_black_box(command, args = "")
+    def krane_black_box(command, args = "", stdin: nil)
```
timothysmith0609: :heart:

lib/krane/render_task.rb
dturn: Tracking in #637

PR #640, opened 2019-11-18
Update repository name in the codebase

douglas:

- Update repository name in the codebase as `kubernetes-deploy` was renamed to `krane`.
- Removed the  `shipit.yml` as I removed all the content that was there:

```yaml
ci:
  hide:
    -  buildkite/kubernetes-deploy-gem-nightly
```

**How is this accomplished?**
Renamed repository entries and removed `shipit.yml`.

**What could go wrong?**
Nothing that I could think of.

### DIFF
--- a/shipit.yml
+++ b/shipit.yml
@@ -1,4 +1,0 @@
-ci:
-  hide:
-    -  buildkite/kubernetes-deploy-gem-nightly
-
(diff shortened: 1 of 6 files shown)


shipit.yml
```diff
@@ -1,4 +1,4 @@
 ci:
   hide:
-    -  buildkite/kubernetes-deploy-gem-nightly
+    -  buildkite/krane-gem-nightly
```
dturn: Should we kill this? I'm not convinced we're getting value out of it.

douglas: I feel we don't have a lot of changes happening daily that justifies it, so 👍 for 🔥 it.

shipit.yml
```diff
@@ -1,4 +0,0 @@
-ci:
-  hide:
-    -  buildkite/kubernetes-deploy-gem-nightly
```
dturn: 🔥  `.buildkite/pipeline.nightly.yml `  , you'll need to delete the pipeline and probably kill the webhook that triggers it

PR #645, opened 2019-11-27
Remove some rendering from some tests that don't need it

RyanBrushett:

Remove some rendering from some tests that don't need it / shouldn't use it.

- This is by no means all of the examples, just a few. Notably, I haven't tackled hello cloud at all.
- Some tests depend on two versions of the same thing with e.g. slightly different labels. In one example here, I've duplicated the fixture with a different label. That's one approach but maybe we could add a render-and-pipe-to-deploy helper for tests where that is convenient?

**How is this accomplished?**

In most cases so far, very minor refactors.

**What could go wrong?**

I could have invalidated some tests by misunderstanding their intention. Good 👀 please.


test/integration/krane_deploy_test.rb
```diff
@@ -537,17 +532,6 @@ def test_wait_false_ignores_non_priority_resource_failures
     ], in_order: true)
   end
 
-  def test_extra_bindings_should_be_rendered
```
dirceu: Was this removed because this is already covered by `test_dynamic_erb_collection_works`?

test/integration/krane_deploy_test.rb
```diff
@@ -1467,19 +1448,6 @@ def test_validation_failure_on_sensitive_resources_does_not_print_template
     refute_logs_match("YWRtaW4=")
   end
 
-  def test_render_failure_on_sensitive_resource_does_not_print_template
```
dirceu: Is this particular case (render failure not showing secrets) being handled in any other test?

RyanBrushett: I think in this case, I forgot to recreate this in the render tests 😬 

👀

test/integration/krane_deploy_test.rb
RyanBrushett: Yep 👍

test/integration/krane_deploy_test.rb
RyanBrushett: @dirceu so the reason I removed this is because the test on line 1439 tests that validation failure on a sensitive resource doesn't print the template.
This test tests that a render failure on a sensitive resource doesn't print the template

This isn't tested anywhere else so I'm going to add this back and get this minimally moved over to the render task test in a different PR.

Make sense?

dirceu: Yes it does, thank you!

PR #646, opened 2019-11-27
Mark kubectl-sourced strings with invalid encoding as UTF-8

### LINKED ISSUE #395: Invalid byte sequence in US-ASCII exception when locales are not set as expected
# Bug report

When using `kubernetes-deploy`, if the locales is not set accordingly it may thrown an exception in cases where the log output from a pod contains non US-ASCII characters.

The environment where the process has been executed was a `ruby:2.3.3` container without any special packages or configs. It was pretty much `docker run -it --rm --name deployer bash`.

From the `bash` session `kubernetes-deploy ...`

**Expected behavior:**

The output of the `provision` container which contains pure text lines as well as some json formatted messages to show up on the `kubernetes-deploy` logs.

**Actual behavior:** 

The following message is thrown and the deployment process breaks.

```
[INFO][2018-12-05 16:07:04 +0000]	Deploying Pod/provision-e03421f6-e9aae851 (timeout: 600s)
/usr/local/lib/ruby/2.3.0/shellwords.rb:142:in `gsub!': invalid byte sequence in US-ASCII (ArgumentError)
	from /usr/local/lib/ruby/2.3.0/shellwords.rb:142:in `shellescape'
	from /usr/local/lib/ruby/2.3.0/shellwords.rb:208:in `shellescape'
	from /usr/local/bundle/gems/kubernetes-deploy-0.20.5/lib/kubernetes-deploy/kubectl.rb:28:in `run'
	from /usr/local/bundle/gems/kubernetes-deploy-0.20.5/lib/kubernetes-deploy/kubernetes_resource/pod.rb:98:in `block in fetch_logs'
	from /usr/local/bundle/gems/kubernetes-deploy-0.20.5/lib/kubernetes-deploy/kubernetes_resource/pod.rb:90:in `each'
	from /usr/local/bundle/gems/kubernetes-deploy-0.20.5/lib/kubernetes-deploy/kubernetes_resource/pod.rb:90:in `each_with_object'
	from /usr/local/bundle/gems/kubernetes-deploy-0.20.5/lib/kubernetes-deploy/kubernetes_resource/pod.rb:90:in `fetch_logs'
	from /usr/local/bundle/gems/kubernetes-deploy-0.20.5/lib/kubernetes-deploy/kubernetes_resource/pod.rb:143:in `display_logs'
	from /usr/local/bundle/gems/kubernetes-deploy-0.20.5/lib/kubernetes-deploy/kubernetes_resource/pod.rb:36:in `sync'
	from /usr/local/bundle/gems/kubernetes-deploy-0.20.5/lib/kubernetes-deploy/sync_mediator.rb:44:in `block in sync'
	from /usr/local/bundle/gems/kubernetes-deploy-0.20.5/lib/kubernetes-deploy/concurrency.rb:13:in `each'
	from /usr/local/bundle/gems/kubernetes-deploy-0.20.5/lib/kubernetes-deploy/concurrency.rb:13:in `block (2 levels) in split_across_threads'
```

**Version(s) affected:** 
0.22.0

### Steps to Reproduce

1. Execute `kubernetes-deploy --template-dir="${TEMPLATE_DIR}" --bindings=current_sha="${REVISION}",prev_sha="${prev_commit:-$REVISION}",environment="${ENVIRONMENT}" "${NAMESPACE}" "${CONTEXT}"`
2. Wait for pod output. It is important the your pod should output a message similar to:
```
{"message":"Any message.","context":[],"level":200,"level_name":"INFO","channel":"k8s","datetime":{"date":"2018-12-05 16:12:36.239316","timezone_type":3,"timezone":"UTC"},"extra":[],"timestamp":"2018-12-05T16:12:36.239Z"}
```

**Workaround**

The following commands worked for setting up the right locales and suppress the exception:

```
apt-get update && apt-get install locales
sed -i -e 's/# en_US.UTF-8 UTF-8/en_US.UTF-8 UTF-8/' /etc/locale.gen && locale-gen
export LANG=en_US.UTF-8
export LANGUAGE=en_US:en  
export LC_ALL=en_US.UTF-8 
```
and then
```
kubernetes-deploy ...
```

KnVerey:

Fixes #395 for good.

**How is this accomplished?**
* We know that kubectl is returning utf-8 encoded strings, so if Ruby has interpreted them otherwise, correct it. 
* Only do the gsub when we're going to use the result.

**What could go wrong?**
* I don't have much experience working with encodings, so if this solution is weird, call me out!

### DIFF
--- a/lib/krane/kubectl.rb
+++ b/lib/krane/kubectl.rb
@@ -34,7 +34,16 @@ module Krane
       (1..attempts).to_a.each do |current_attempt|
         logger.debug("Running command (attempt #{current_attempt}): #{cmd.join(' ')}")
         out, err, st = Open3.capture3(*cmd)
-        logger.debug("Kubectl out: " + out.gsub(/\s+/, ' ')) unless output_is_sensitive
+
+        # https://github.com/Shopify/krane/issues/395
+        if out.encoding != Encoding::UTF_8
+          out = out.dup.force_encoding(Encoding::UTF_8)
+        end
+
+        if logger.debug? && !output_is_sensitive
+          # don't do the gsub unless we're going to print this
+          logger.debug("Kubectl out: " + out.gsub(/\s+/, ' '))
+        end
 
         break if st.success?
         raise(ResourceNotFoundError, err) if err.match(ERROR_MATCHERS[:not_found]) && raise_if_not_found
--- a/lib/krane/kubectl.rb
+++ b/lib/krane/kubectl.rb
@@ -36,7 +36,7 @@ module Krane
         out, err, st = Open3.capture3(*cmd)
 
         # https://github.com/Shopify/krane/issues/395
-        if out.encoding != Encoding::UTF_8
+        unless out.valid_encoding?
           out = out.dup.force_encoding(Encoding::UTF_8)
         end
 
--- a/test/unit/krane/kubectl_test.rb
+++ b/test/unit/krane/kubectl_test.rb
@@ -348,6 +348,16 @@ class KubectlTest < Krane::TestCase
     end
   end
 
+  def test_debug_level_output_log_uses_correct_encoding
+    logger.level = ::Logger::DEBUG
+    good = "hélas"
+    bad = good.dup.force_encoding(Encoding::US_ASCII)
+
+    stub_open3(%W(kubectl get pods --namespace=testn --context=testc --request-timeout=#{timeout}), resp: bad)
+    out, _err, _st = build_kubectl.run("get", "pods")
+    assert_equal good, out
+  end
+
   private
 
   def timeout
--- a/test/unit/krane/kubectl_test.rb
+++ b/test/unit/krane/kubectl_test.rb
@@ -348,14 +348,27 @@ class KubectlTest < Krane::TestCase
     end
   end
 
-  def test_debug_level_output_log_uses_correct_encoding
+  def test_kubectl_run_fixes_encoding_when_locales_set_to_non_utf8
+    ext_before = Encoding.default_external
+    int_before = Encoding.default_internal
     logger.level = ::Logger::DEBUG
-    good = "hélas"
-    bad = good.dup.force_encoding(Encoding::US_ASCII)
+    utf8 = "こんにちは！hélas!"
 
-    stub_open3(%W(kubectl get pods --namespace=testn --context=testc --request-timeout=#{timeout}), resp: bad)
+    # This is how setting the env from https://github.com/Shopify/krane/issues/395 manifests internally
+    Encoding.default_external = Encoding::US_ASCII
+    Encoding.default_internal = nil
+
+    # put the string through the default external encoder
+    data = %x(echo #{utf8})
+    assert_equal(Encoding::US_ASCII, data.encoding)
+
+    stub_open3(%W(kubectl get pods --namespace=testn --context=testc --request-timeout=#{timeout}), resp: data)
     out, _err, _st = build_kubectl.run("get", "pods")
-    assert_equal good, out
+    assert_equal(utf8, out)
+    assert_equal(Encoding::UTF_8, out.encoding)
+  ensure
+    Encoding.default_external = ext_before
+    Encoding.default_internal = int_before
   end
 
   private
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 ## next
 
+*Bug Fixes*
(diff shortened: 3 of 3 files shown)


test/unit/krane/kubectl_test.rb
```diff
@@ -348,6 +348,16 @@ def test_retry_delay_backoff
     end
   end
 
+  def test_debug_level_output_log_uses_correct_encoding
+    logger.level = ::Logger::DEBUG
+    good = "hélas"
+    bad = good.dup.force_encoding(Encoding::US_ASCII)
```
KnVerey: This obviously isn't how the encoding goes bad in the wild, but this test does reproduce the stack trace from the bug report. Note that we have to `dup` because string literals are frozen, and `force_encoding` is considered a modification even though it doesn't touch the characters themselves.

lib/krane/kubectl.rb
```diff
@@ -34,7 +34,16 @@ def run(*args, log_failure: nil, use_context: true, use_namespace: true, output:
       (1..attempts).to_a.each do |current_attempt|
         logger.debug("Running command (attempt #{current_attempt}): #{cmd.join(' ')}")
         out, err, st = Open3.capture3(*cmd)
-        logger.debug("Kubectl out: " + out.gsub(/\s+/, ' ')) unless output_is_sensitive
+
+        # https://github.com/Shopify/krane/issues/395
+        if out.encoding != Encoding::UTF_8
```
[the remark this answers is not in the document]
KnVerey: > it seems like in other cases (e.g. utf-16) this might be wrong?

What are you thinking would cause the string we get back from kubectl to actually have a different encoding?

dalehamel:
I wonder what the consequences of ignoring the locales could be...

but if the code assumes UTF-8 it is probably better to assert this.

lib/krane/kubectl.rb
```diff
@@ -34,7 +34,16 @@ def run(*args, log_failure: nil, use_context: true, use_namespace: true, output:
       (1..attempts).to_a.each do |current_attempt|
         logger.debug("Running command (attempt #{current_attempt}): #{cmd.join(' ')}")
         out, err, st = Open3.capture3(*cmd)
-        logger.debug("Kubectl out: " + out.gsub(/\s+/, ' ')) unless output_is_sensitive
+
+        # https://github.com/Shopify/krane/issues/395
+        unless out.valid_encoding?
```
KnVerey: This returns false when the string uses characters that aren't in the character set for the encoding the string is claiming to have. I think this is more strictly correct than my previous solution of checking if the original tag is utf8 and always changing it if so. The trade-off is potentially performance: in this version we have to check the encoding on every single string, but we only have to dup the string and change it when a non-ASCII character is found. In the previous version, we never had to check the encoding, but if your locales are set wrong, we're dup'ing and forcing encoding on every single string.

Note that setting `Encoding.default_external` would also fix the bug, but I don't think we should do that, because as a gem, we should not set globals (not to mention that the docs say "don't set this yourself").

lib/krane/kubectl.rb
```diff
@@ -34,7 +34,16 @@ def run(*args, log_failure: nil, use_context: true, use_namespace: true, output:
...
+          out = out.dup.force_encoding(Encoding::UTF_8)
```
KnVerey: After further thought and reading about encodings, I am comfortable saying this is safe to do. The string we're tagging here isn't arbitrary--it's coming from kubectl. I'd be shocked if kubectl decided to suddenly start outputting e.g. UTF-16 encoding. Putting that aside, the worst case is that a user gets a non-UTF8 string and still sees the original bug. If we want to be super paranoid, we can add this code after this line:

```ruby
          unless out.valid_encoding?
            out.encode(invalid: :replace, undef: :replace)
          end
```

That will _force_ the string to be UTF8, with invalid characters replaced with "?". But I just don't think it's worth it. I spent a few minutes trying to figure out how to test that, and couldn't figure out how.

test/unit/krane/kubectl_test.rb
```diff
@@ -348,6 +348,29 @@ def test_retry_delay_backoff
     end
   end
 
+  def test_kubectl_run_fixes_encoding_when_locales_set_to_non_utf8
```
KnVerey: This is now a much better version of the test that reproduces the Ruby environment from the bug report (to confirm, start a console session with the wrong local envs set and check the `Encoding` values below). Implementation inspired by [this patch on Pysch](https://github.com/ruby/psych/commit/fe65329ce9ece399d61dadf1610e342ff456654e), which I found via [a bug report on RubyGems](https://github.com/rubygems/rubygems/issues/139).

Note that I confirmed this is a true regression test that files if I remove my new code.

PR #648, opened 2019-11-29
Dev dependency best practices / fix linter

KnVerey:
1 - Follow best practices for dev dependency listing. I noticed that we have some dev deps listed in the Gemfile and others in the Gemspec, so I looked up what we're supposed to be doing conventionally. The answer is gemspec: https://yehudakatz.com/2010/12/16/clarifying-the-roles-of-the-gemspec-and-gemfile/.
2 - Follow best practices for rubocop. If you look at the readme for the ruby styleguide repo we inherit from, you'll see that the recommendation is to commit the cache file. This will let us control when we update our rubocop rules (no more failing on random PRs) and see which changes we're importing in the diff.
3 - Make rubocop pass on CI.

**How is this accomplished?**
1 - Move all dev dependencies to the gemspec.
2 - Commit the rubocop cache file.
3 - Pin rubocop to a version that works with the upstream style guide. The latest version, release a couple days ago, breaks it, and we're in code freeze right now. 

**What could go wrong?**
1 - I'm pretty sure this is functionally equivalent. I could have missed a dep?
2 - We never update it and fall behind upstream rules.
3 - We have to unpin this at some point in the future.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -3,14 +3,3 @@ source 'https://rubygems.org'
 
 # Specify your gem's dependencies in krane.gemspec
 gemspec
-
-gem 'pry'
-gem 'pry-byebug'
-gem 'rubocop'
-gem 'timecop'
-gem 'byebug'
-gem 'codecov', require: false
-gem 'ruby-prof', require: false
-gem 'ruby-prof-flamegraph', require: false
-gem 'minitest-reporters'
-gem 'yard', require: false
--- a/.gitignore
+++ b/.gitignore
@@ -8,9 +8,6 @@
 /spec/reports/
 /tmp/
 
-# Ignore RuboCop cache
-.rubocop-http---shopify-github-io-ruby-style-guide-rubocop-yml
-
 .byebug_history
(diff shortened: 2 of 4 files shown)


Gemfile
```diff
@@ -3,14 +3,3 @@ source 'https://rubygems.org'
 
 # Specify your gem's dependencies in krane.gemspec
 gemspec
-
-gem 'pry'
-gem 'pry-byebug'
-gem 'rubocop'
-gem 'timecop'
-gem 'byebug'
-gem 'codecov', require: false
```
dturn: Can/should we drop `require 'X'`s  in test_helper.rb

KnVerey: Good question. I don't have a single authoritative source to point you to, but reading various things on the internet indicates that:
- Nothing from the gemspec is required by default.
- Even things in the gemfile aren't "magically" required unless you do `Bundler.require` somewhere, which we don't. Since our test helper does the `require`s individually anyway, our old `require: false` wasn't doing anything for us.

PR #649, opened 2019-12-02
Fix broken link in readme.

shane-lawrence:
Fix a broken link in the readme.

**How is this accomplished?**
Change the hash in the link to match a location in the readme.

**What could go wrong?**
Very little.

### DIFF
--- a/test/unit/krane/resource_deployer_test.rb
+++ b/test/unit/krane/resource_deployer_test.rb
@@ -86,7 +86,8 @@ class ResourceDeployerTest < Krane::TestCase
 
   def resource_deployer(kubectl_times: 1, prune_whitelist: [])
     unless kubectl_times == 0
-      Krane::Kubectl.expects(:new).returns(build_runless_kubectl).times(kubectl_times)
+      runless = build_runless_kubectl
+      Krane::Kubectl.expects(:new).returns(runless).times(kubectl_times)
     end
     @deployer = Krane::ResourceDeployer.new(current_sha: 'test-sha',
       statsd_tags: [], task_config: task_config, prune_whitelist: prune_whitelist,
--- a/test/unit/krane/resource_deployer_test.rb
+++ b/test/unit/krane/resource_deployer_test.rb
@@ -84,7 +84,7 @@ class ResourceDeployerTest < Krane::TestCase
 
   private
 
-  def resource_deployer(kubectl_times: 2, prune_whitelist: [])
+  def resource_deployer(kubectl_times: 1, prune_whitelist: [])
     unless kubectl_times == 0
       Krane::Kubectl.expects(:new).returns(build_runless_kubectl).times(kubectl_times)
     end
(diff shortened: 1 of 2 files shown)


test/unit/krane/resource_deployer_test.rb
```diff
@@ -84,7 +84,7 @@ def test_predeploy_priority_resources_respects_empty_pre_deploy_list
 
   private
 
-  def resource_deployer(kubectl_times: 2, prune_whitelist: [])
+  def resource_deployer(kubectl_times: 1, prune_whitelist: [])
```
jessie-JNing: `Krane::Kubectl` should only be initialized once, not sure why they were OK before

jessie-JNing:
cc @Shopify/production-excellence

test/unit/krane/resource_deployer_test.rb
KnVerey: This is very curious. I'm still investigating, but what I've found so far is that it was caused by `mocha`. If you use version 1.9, the tests pass with 2 here. If you use version 1.10, they pass with 1. There's a lot in the release notes: https://github.com/freerange/mocha/blob/master/RELEASE.md.

KnVerey: Ok, so I still can't pinpoint exactly why this changed, but we _are_ calling `Kubectl::new` twice! Here's how you can check:

```ruby
  def resource_deployer(kubectl_times: 2, prune_whitelist: [])
    unless kubectl_times == 0
      Krane::Kubectl.expects(:new).with do |actual|
        binding.pry
        true
      end.returns(build_runless_kubectl).times(kubectl_times)
    end
    @deployer = Krane::ResourceDeployer.new(current_sha: 'test-sha',
      statsd_tags: [], task_config: task_config, prune_whitelist: prune_whitelist,
      global_timeout: 1, selector: nil)
  end
```

You'll see you get thrown into the block twice when you run an affected test (e.g. `test_deploy_verify_false_no_failure_error`), regardless of your mocha version.  You can then use `backtrace` to see where the invocations are coming from... and the second one is on the expectation line itself! It's `build_runless_kubectl`.

Conclusion: somehow mocha 1.10 is not counting the invocation that happens as part of the expectation chain anymore, even though it does capture it. Regardless, invoking the constructor we're stubbing as part of the expectation chain is sketchy! Let's use the following code, which passes regardless of the mocha version:

```ruby
  def resource_deployer(kubectl_times: 1, prune_whitelist: [])
    unless kubectl_times == 0
      runless = build_runless_kubectl
      Krane::Kubectl.expects(:new).returns(runless).times(kubectl_times)
    end
```

jessie-JNing: Thanks @KnVerey for digging it, very detailed reasoning prove. 
Dragging the stub from explanation chain is the best solution so far, without pinning mocha version. 
I'll update the PR as suggested above.

PR #653, opened 2019-12-04
Scope all global fixture deploys

KnVerey:

Make it safer and easier to write tests that modify the global namespace.

Related to https://github.com/Shopify/krane/issues/637.

**How is this accomplished?**

- [x] Make `deploy_global_fixtures` automatically rescope all resources correctly, including CRDs.
- [x] Make the test suite automatically clean up everything deployed by `deploy_global_fixtures` _after_ the test is finished (the same time we clean up namespaces).
- [x] Get rid of all usage of `deploy_global_fixtures_non_namespaced` and the `clean_up: false` option on `global_deploy_dirs_without_profiling`
- [x] Scope all CRD name fields to the test to make them truly unique. Adjust a boatload of tests.
- [x] Move any tests that are now concurrency-safe out of the serial test file and make sure the reasons the others can't move are well documented.

Note that unfortunately any test that creates CRDs still isn't concurrency-safe, because it effectively changes our dynamic pruning whitelist. If a CRD is removed in the middle of a namespaced deploy, that deploy will fail with an `apply` error because the pruning whitelist will reference a non-existent kind. That said, CRD tests are still improved by this PR in that the suite now cleans up after them properly with no additional effort on the test writer's part, and we no longer need to wait for the CRDs to fully go away between tests because they are actually unique (and won't interfere with each other).

**What could go wrong?**
Tests could be invalidated or become flakey.

### DIFF
--- a/.rubocop-http---shopify-github-io-ruby-style-guide-rubocop-yml
+++ b/.rubocop-http---shopify-github-io-ruby-style-guide-rubocop-yml
@@ -20,7 +20,7 @@ Style/Alias:
   - prefer_alias
   - prefer_alias_method
 
-Layout/AlignHash:
+Layout/HashAlignment:
   EnforcedHashRocketStyle: key
   EnforcedColonStyle: key
   EnforcedLastArgumentHashStyle: ignore_implicit
@@ -30,7 +30,7 @@ Layout/AlignHash:
   - ignore_implicit
   - ignore_explicit
 
-Layout/AlignParameters:
+Layout/ParameterAlignment:
   EnforcedStyle: with_fixed_indentation
   SupportedStyles:
   - with_first_parameter
@@ -172,7 +172,7 @@ Naming/FileName:
   Regex:
   IgnoreExecutableScripts: true
 
-Layout/IndentFirstArgument:
+Layout/FirstArgumentIndentation:
   EnforcedStyle: consistent
   SupportedStyles:
   - consistent
@@ -225,7 +225,7 @@ Layout/IndentationConsistency:
 Layout/IndentationWidth:
   Width: 2
 
-Layout/IndentFirstArrayElement:
+Layout/FirstArrayElementIndentation:
   EnforcedStyle: consistent
   SupportedStyles:
   - special_inside_parentheses
@@ -233,10 +233,10 @@ Layout/IndentFirstArrayElement:
   - align_brackets
   IndentationWidth:
 
-Layout/IndentAssignment:
+Layout/AssignmentIndentation:
   IndentationWidth:
 
-Layout/IndentFirstHashElement:
+Layout/FirstHashElementIndentation:
   EnforcedStyle: consistent
   SupportedStyles:
   - special_inside_parentheses
@@ -340,9 +340,9 @@ Style/PercentQLiterals:
 Naming/PredicateName:
   NamePrefix:
   - is_
-  NamePrefixBlacklist:
+  ForbiddenPrefixes:
   - is_
-  NameWhitelist:
+  AllowedMethods:
   - is_a?
   Exclude:
   - 'spec/**/*'
@@ -467,7 +467,7 @@ Style/TernaryParentheses:
   - require_no_parentheses
   AllowSafeAssignment: true
 
-Layout/TrailingBlankLines:
+Layout/TrailingEmptyLines:
   EnforcedStyle: final_newline
   SupportedStyles:
   - final_newline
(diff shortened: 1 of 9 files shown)


test/integration/krane_deploy_test.rb
```diff
@@ -199,6 +199,24 @@ def test_refuses_deploy_to_protected_namespace_with_override_if_pruning_enabled
     @namespace = generated_ns
   end
 
+  def test_deploying_to_protected_namespace_does_not_prune
```
KnVerey: The `krane` version of this didn't need to be serial, because it didn't need to stub protected namespaces list anymore (it used to be a constant)

test/integration/krane_deploy_test.rb
```diff
@@ -1575,6 +1593,175 @@ def test_deploy_resource_using_generate_name_succeeds
...
+    assert_logs_match_all([
+      "Failed to deploy 1 priority resource",
+      "Pod/pvc: TIMED OUT (timeout: 10s)",
+      %r{Pod could not be scheduled because 0/\d+ nodes are available:},
+      /\d+ node[(]s[)] didn't find available persistent volumes to bind./,
+    ], in_order: true)
+  end
+
+  def test_create_secrets_from_ejson
+    logger.level = ::Logger::DEBUG # for assertions that we don't log secret data
```
KnVerey: These logging tests are not changing anything global. I think we made these concurrency-safe long ago, e.g. we originally were setting `ENV[DEBUG]` or something and when we stopped we didn't move them.

test/integration_test_helper.rb
```diff
@@ -15,9 +15,20 @@ class IntegrationTest < Krane::TestCase
     end
 
     def run
-      super { @namespace = TestProvisioner.claim_namespace(name) }
+      super do
+        @namespace = TestProvisioner.claim_namespace(name)
+        @global_fixtures_deployed = []
```
dturn: Perhaps `global_fixtures_paths_deployed`  it wasn't clear at first read this was a list of files vs. names or raw yaml.

.rubocop-http---shopify-github-io-ruby-style-guide-rubocop-yml
```diff
@@ -20,7 +20,7 @@ Style/Alias:
   - prefer_alias
   - prefer_alias_method
 
-Layout/AlignHash:
+Layout/HashAlignment:
```
dturn: When I run  `bundle exec rubocop` locally I get 
`Warning: unrecognized cop Sorbet/ConstantsFromStrings found in .rubocop.yml` plus 23 failures all but 1 in serial_deploy_test.rb. CI has a bunch of warnings but no failures. Any idea what's up?

test/helpers/fixture_deploy_helper.rb
```diff
@@ -183,22 +171,29 @@ def build_kubectl(log_failure_by_default: true, timeout: '5s')
...
+          if resource["kind"] == "CustomResourceDefinition"
+            %w(kind listKind plural singular).each do |field|
+              if (original_name = resource.dig("spec", "names", field))
+                resource["spec"]["names"][field] = add_unique_prefix_for_test(original_name)
+              end
+            end
+            # metadata.name has to be composed this way for CRDs
+            resource["metadata"]["name"] = "#{resource['spec']['names']['plural']}.#{resource['spec']['group']}"
+          else
+            resource["metadata"]["name"] = add_unique_prefix_for_test(resource["metadata"]["name"])[0..63]
```
dturn: names can be 253 chars long https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#names. Its only labels that are limited to 63 https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/#syntax-and-character-set

.rubocop-http---shopify-github-io-ruby-style-guide-rubocop-yml
KnVerey: I think it's a mismatch between the rubocop config and the rubocop version. I'm not sure how to prevent that. Right now we pin the rubocop gem version to make sure our CI doesn't leap ahead of the style guide, but the reverse can happen: the styleguide gets updates in a way that's not compatible with our (now-old) pinned version. I was hoping we might be able to change force our CI invocation to use the committed copy of the styleguide, but it looks like that isn't possible. Rubocop will try to refresh the cache if it is > 24 hours old.

In any case, I'll solve the immediate problem by bumping rubocop, updating the committed file again, and removing some rules we didn't need anymore while I'm at it.

Issue #658, opened 2019-12-16
Use Thor 1.0


timothysmith0609:
### Why upgrade Thor?

The update to Thor contains 2 desirable changes:
__1.__ [Accept `-` as an argument, not a flag](https://github.com/erikhuda/thor/pull/691)
__2.__ [Allow options to be repeatable](https://github.com/erikhuda/thor/pull/674) __NOTE: This is dependent on [Thor#713](https://github.com/erikhuda/thor/pull/713)__

__1__ will allow us to deprecate the `--stdin` flag in favour of allowing `-` to be passed as an argument to `-f`
__2__ allows a closer semblance to `kubectl` semantics by allowing flags to be repeated in a given invocation

### TODO

- [x] Flesh out CLI tests

dturn:
I believe it also addresses the issues raised in https://github.com/Shopify/krane/pull/559#issuecomment-561857195

timothysmith0609:
Closed by #684

timothysmith0609:
We have set the lower bar of Thor to 1.0, however we still need to add the repeatability of flags (as mentioned above, this is blocked on [Thor#713](https://github.com/erikhuda/thor/pull/674) __NOTE: This is dependent on [Thor#713](https://github.com/erikhuda/thor/pull/713)

PR #659, opened 2019-12-17
fix the --filename option`

shivnagarajan:
```$ krane help render
Usage:
  krane render

Options:
      [--bindings=foo=bar abc=def]                                               # Bindings for erb
  f, [--filenames=config/deploy/production config/deploy/my-extra-resource.yml]  # Directories and files to render
      [--stdin], [--no-stdin]                                                    # Read resources from stdin
      [--current-sha=SHA]                                                        # Expose SHA `current_sha` in ERB bindings

Render templates
```

the `--filename` option needs to be `--filenames` to be consistent with the code/usage description.

...

**How is this accomplished?**
...

**What could go wrong?**
...

### DIFF
--- a/1.0-Upgrade.md
+++ b/1.0-Upgrade.md
@@ -153,7 +153,7 @@ Old flag | New flag | Comments
 Old flag | New flag | Comments
 --- | --- | ---
 --bindings=BINDINGS | --bindings=BINDINGS |
---template-dir | -f, --filename | Changed to be more aligned with `kubectl apply` and other krane tasks
+--template-dir | -f, --filenames | Changed to be more aligned with `kubectl apply` and other krane tasks
 $REVISION | --current-sha | The environment variable REVISION was dropped in favour of an explicit flag
 [none] | --stdin | Allow template filenames given from stdin stream
 


RyanBrushett:
See https://github.com/Shopify/krane/blob/master/lib/krane/cli/render_command.rb#L8

KnVerey:
Actually the [design doc](https://docs.google.com/document/d/1oInUsKplYGNWTymPY48xtDCx3x4Cwc-phcEKUkt2v4U/view#) has it as singular, which is mirroring kubectl's arg:

```
 -f, --filename=[]: Filename, directory, or URL to files identifying the resource to get from a server.
 -f, --filename=[]: that contains the configuration to apply
 -f, --filename=[]: Filename, directory, or URL to files to use to create the resource
```

That makes more sense when the way to add multiple files is to repeat the arg, which we couldn't do because of the Thor bug... but that's fixed upsteam, right?

RyanBrushett:
Yeah you're right, something to fix when we do the Thor bump.
As of today though, `--filenames` is what the render command actually takes as an arg tho so the docs should probably reflect that. Worth making note of fixing this as part of https://github.com/Shopify/krane/issues/658?

shivnagarajan:
> Yeah you're right, something to fix when we do the Thor bump.
As of today though, --filenames is what the render command actually takes as an arg tho so the docs should probably reflect that. Worth making note of fixing this as part of #658?

How would a future fix handle the case for clients that already use `--filenames` (plural) instead of `--filename` 

e.g. https://github.com/Shopify/shopify-core-mcrouter/pull/84, I could just use `-f` instead for now, but I am thinking of any one else that could already have versions that use `--filenames`

RyanBrushett:
Hmm indeed, cutting that over would be a breaking change. Maybe an alias could be an option?

shivnagarajan:
Should I merge this? or just close it?

PR #662, opened 2020-01-06
Relax googleauth version

ramontayag:
Makes it easier to install in apps that have other versions of googleauth

**How is this accomplished?**
Relax googleauth version to be greater than or equal to the version it used to require

**What could go wrong?**
Deployments fail if there's a new version of googleauth with a breaking change (any version of 0.x does not have any compatibility promises according to semver)

### DIFF
--- a/krane.gemspec
+++ b/krane.gemspec
@@ -26,7 +26,7 @@ Gem::Specification.new do |spec|
   spec.required_ruby_version = '>= 2.4.0'
   spec.add_dependency("activesupport", ">= 5.0")
   spec.add_dependency("kubeclient", "~> 4.3")
-  spec.add_dependency("googleauth", "~> 0.8")
+  spec.add_dependency("googleauth", [">= 0.8.0", "< 1"])
   spec.add_dependency("ejson", "~> 1.0")
   spec.add_dependency("colorize", "~> 0.8")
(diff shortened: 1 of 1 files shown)


ramontayag:
I've signed the CLA

krane.gemspec
```diff
@@ -26,7 +26,7 @@ Gem::Specification.new do |spec|
   spec.required_ruby_version = '>= 2.4.0'
   spec.add_dependency("activesupport", ">= 5.0")
   spec.add_dependency("kubeclient", "~> 4.3")
-  spec.add_dependency("googleauth", "~> 0.8.0")
+  spec.add_dependency("googleauth", ">= 0.8.0")
```
dturn: Can we do something like `['>= 0.8.0', "< 1"]`? I'd prefer not to risk someone installing a version with a breaking change.

ramontayag: I missed this somehow. Sure! Updated the branch

dturn:
I believe this was fixed in https://github.com/Shopify/krane/pull/667/files#diff-384f457f2f67f3b057c1d511575b72ccR29 and was part of the 1.1.1 release.

ramontayag:
Ok, that works too. Thanks!

PR #668, opened 2020-01-17
`--template` is required to use `krane run`

cjoudrey:
When reading the usage of `krane run` in the `README.md`, I noticed the `--template` option was documented as being optional.

This doesn't match the current behaviour of `krane run`:

```
$ krane run
No value provided for required options '--template'
```

After chatting with @dturn, it sounds like this was a typo in the `README.md`, hence this PR. 😄

### DIFF
--- a/README.md
+++ b/README.md
@@ -503,7 +503,7 @@ Based on this specification `krane run` will create a new pod with the entrypoin
 
 *Options:*
 
-* `--template=TEMPLATE`:  Specifies the name of the PodTemplate to use (default is `task-runner-template` if this option is not set).
+* `--template=TEMPLATE`: Specifies the name of the PodTemplate to use.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -503,7 +503,7 @@ Based on this specification `krane run` will create a new pod with the entrypoin
 
 *Options:*
 
-* `--template=TEMPLATE`:  Specifies the name of the PodTemplate to use (default is `task-runner-template` if this option is not set).
+* `--template=TEMPLATE`: Specifies the name of the PodTemplate to use.
```
cjoudrey: Wasn't sure what the convention was to mention the option is required.

I was going to put:

```suggestion
* `--template=TEMPLATE` (required): Specifies the name of the PodTemplate to use.
```

dturn: Not a convention we're currently using.

PR #669, opened 2020-01-18
Relax googleauth dependency to allow 0.9 series versions

airhorns:
Allow compatibility with `fog-google` and recent `google-api-client` gems that require a newer version of `googleauth`

**How is this accomplished?**

Relax version dependency in gemspec 

**What could go wrong?**

New `googleauth` could break google functionality in krane, which is actually functionality in `kubeclient`.  I don't see anything breaking in the changelog here: https://github.com/googleapis/google-auth-library-ruby/blob/master/CHANGELOG.md , but, I'm not familiar enough with krane's test suite to say if the `kubeclient` bits that use `googleauth` are implicitly tested anywhere. So, what do y'all think?

### DIFF
--- a/krane.gemspec
+++ b/krane.gemspec
@@ -26,7 +26,7 @@ Gem::Specification.new do |spec|
   spec.required_ruby_version = '>= 2.4.0'
   spec.add_dependency("activesupport", ">= 5.0")
   spec.add_dependency("kubeclient", "~> 4.3")
-  spec.add_dependency("googleauth", "~> 0.8.0")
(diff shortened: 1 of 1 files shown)


dturn:
This is slated for the 1.1.1 release https://github.com/Shopify/krane/pull/667/files#diff-384f457f2f67f3b057c1d511575b72ccR29  . I'm aiming for the release to happen this week.

airhorns:
<3 thanks @dturn !

PR #671, opened 2020-01-20
Use metadata.generation instead of deprecated attributes

### LINKED ISSUE #627: Review usage of deprecated field in DaemonSet parent_of_pod? method
# Improvement Analysis / Request

During the work on https://github.com/Shopify/kubernetes-deploy/pull/598 to support Kubernetes 1.16 (and move from `extensions/v1beta1`, we discovered that the `templateGeneration` field was removed in `DaemonSet v1` and was deprecated in this PR:

https://github.com/kubernetes/kubernetes/pull/49454
**Related issue:** https://github.com/kubernetes/kubernetes/issues/49336

In order to ship Kubernetes 1.16 support sooner we decided to use `deprecated.daemonset.template.generation` as `templateGeneration` field was removed in `DaemonSet` `apps/v1` API.

# Suggested approach

After https://github.com/Shopify/kubernetes-deploy/pull/598 ships we need to review [this code](https://github.com/Shopify/kubernetes-deploy/pull/598/commits/95fbabc789bdfa724c4c0eb3d1725ef99300f1c0#diff-ceac72c3f3a0ea8637a801fc800c3f24R81-R82) to check if there is a way to avoid the usage of a deprecated field.

## DaemonSetSpec v1

Screen Shot 2019-11-14 at 12 31 47 PM

## DaemonSetSpec v1beta1

Screen Shot 2019-11-14 at 12 31 25 PM

douglas:

Avoid usage of `spec.templateGeneration` as it was deprecated in `k8s 1.7` and `deprecated.daemonset.template.generation` in favor or the standard api `metadata.generation`.

For context: https://github.com/kubernetes/kubernetes/issues/49336 and https://github.com/kubernetes/kubernetes/pull/49454.

Fixes #627.

**How is this accomplished?**

Changing:

```ruby
(...)
template_generation = @instance_data.dig("spec", "templateGeneration") ||
        @instance_data.dig("metadata", "annotations", "deprecated.daemonset.template.generation")
(...)
```

To:

```ruby
template_generation = @instance_data.dig("metadata", "generation")
```

**What could go wrong?**

The only scenario that I could think of is people trying to use `k8s 1.7` with a new release of `krane` after this commit gets merged *but* it would break in so many other scenarios that I don't think it is something that we have to consider, as our users can use an older version of the gem if they want to use an old `k8s` version.

Reviewers more experienced with `k8s`: Can you see other problems that I couldn't see?

### DIFF
--- a/lib/krane/kubernetes_resource/daemon_set.rb
+++ b/lib/krane/kubernetes_resource/daemon_set.rb
@@ -80,8 +80,7 @@ module Krane
     def parent_of_pod?(pod_data)
       return false unless pod_data.dig("metadata", "ownerReferences")
 
-      template_generation = @instance_data.dig("spec", "templateGeneration") ||
-        @instance_data.dig("metadata", "annotations", "deprecated.daemonset.template.generation")
+      template_generation = @instance_data.dig("metadata", "generation")
       return false unless template_generation.present?
 
       pod_data["metadata"]["ownerReferences"].any? { |ref| ref["uid"] == @instance_data["metadata"]["uid"] } &&


KnVerey:
`metadata.generation` is not equivalent to the deprecated attribute: it is bumped every time anything in the desired state (i.e. outside `.status`) changes, not just the template. So it's not going to match that label on the pod (even if that label isn't deprecated too, which I'm not sure about). If you write a test that runs a second deploy that updates DS spec or metadata other than the template, it should fail.

The issues you referenced show the deprecation happening upstream, but what we need to know is why they considered this information removable and what they replaced it with internally. After all, the daemonset controller must still need to know which pods belong to each generation of the template in order to implement rolling updates correctly. This issue I found seems to indicate that the replacement is the ControllerRevisions object: https://github.com/kubernetes/kubernetes/pull/45924. 

Interestingly, that object looks very generic, which made me wonder if it is the official new way to implement revision history for other objects too. A cursory look at our live clusters indicates that it is used for StatefulSet too. We should confirm if/when our logic for other objects should be converted.

> The only scenario that I could think of is people trying to use k8s 1.7 with a new release of krane

Don't worry about that at all. It's been a long time since we supported 1.7: https://github.com/Shopify/krane#prerequisites.

douglas:
Thanks a lot for the context, Katrina! I will read about the issue you mentioned and will try to come up with something.

timothysmith0609:
Closing as will be fixed in a different PR

PR #672, opened 2020-01-20
Deploy CustomResources after dependencies

### LINKED ISSUE #670: Custom Resources often need secrets/config maps
# Bug report

Krane deploys priority resources in an order that doesn't make sense, given that CRs are likely to depend on other resource types such as Secrets or RBAC resources in order to transition to Ready.

**Expected behavior:** Krane should not deploy CRs until after Secrets, ConfigMaps, etc. The only thing that needs to come after CRs is Pods.

**Actual behavior:** Krane deploys CRs before the things they quite possibly depend on.

**Version(s) affected:** 1.1.0

### Steps to Reproduce

1. Deploy templates which include a Secret and a CR (eg usage of the percona-operator).

benlangfeld:

CustomResources are quite likely to depend on things like Secrets, RBAC resources or even PVCs. They should be available before blocking on monitoring the CustomResources for success.

Fixes #670

**How is this accomplished?**

This change ensures that CustomResources are pre-deployed after the other pre-deployed resource types.

**What could go wrong?**

WW3?

### DIFF
--- a/lib/krane/deploy_task.rb
+++ b/lib/krane/deploy_task.rb
@@ -60,14 +60,14 @@ module Krane
       before_crs = %w(
         ResourceQuota
         NetworkPolicy
-      )
-      after_crs = %w(
         ConfigMap
         PersistentVolumeClaim
         ServiceAccount
         Role
         RoleBinding
         Secret
+      )
+      after_crs = %w(
         Pod
       )
 


benlangfeld:
Could someone please paste the test failure?

dturn:
Tests passed, it shows failure, because of https://github.com/Shopify/krane/issues/506

benlangfeld:
@dturn Might this be included in this week's release?

dturn:
> Might this be included in this week's release?

Yes

PR #673, opened 2020-01-21
[bugfix] Restore default value for verbose-log-prefix

KnVerey:
Fix the default value of --verbose-log-prefix on the deploy command. This was accidentally flipped with the move to Thor.

**How is this accomplished?**
Flipping the default.

**What could go wrong?**
If someone liked that it was flipped, they might be upset?


KnVerey:
> Can you write a change log entry?

Since the version bump PR is already open, I'll push a commit directly on that branch.

PR #678, opened 2020-01-28
Add code owners

KnVerey:
Make it easy (automatic) to tag all the interested people on Krane PRs, even though we're not on the same team.

(I also moved the issue and PR templates to the .github folder we already had)

**How is this accomplished?**
Adding [code owners](https://help.github.com/en/github/creating-cloning-and-archiving-repositories/about-code-owners). I wanted to make a Github team, but it isn't possible to make teams inherit members. That would have meant needing to keep the krane team in sync with the Prod Excellence team, which seemed unacceptable.

@DazWorrall I put you on the list because we've agreed in the past that Shipping Pipeline theoretically wants to be more involved in / aware of this project. If you'd rather not be included or have someone I could add in your place, let me know.

**What could go wrong?**
This makes us all _always_ get pinged, which will add noise for people who were not being pinged on everything before.

@Shopify/production-excellence


DazWorrall:
> @DazWorrall I put you on the list because we've agreed in the past that Shipping Pipeline theoretically wants to be more involved in / aware of this project. If you'd rather not be included or have someone I could add in your place, let me know.

This has maybe changed with the creation + direction of production registry, but we can stick with me for now.

PR #684, opened 2020-02-06
Set Thor min version to 1.0

### LINKED ISSUE #658: Use Thor 1.0

timothysmith0609:
Component of #658 
I originally attempted to add the repeatable option to `--filenames` in this PR, but unfortunately there appears to be a bug (I suspect in Thor itself) that precludes that. I'm going to take a look at that after this PR merges.

Sets the lower bound of the Thor version we use to 1.0. This version includes parsing support for `-` as flag arguments (required to get rid of `--stdin` flag in favour of `--filenames -`).

**How is this accomplished?**
- Bump the lower bound for Thor in the gemspec
- Remove references in README for stdin, edit references to `--filenames` to indicate how to read from STDIN
- Update tests
- Fix code references

**What could go wrong?**
Small change across a module of the system (the CLI layer), so this should be a fairly safe and testable change. More difficult might be ensuring we clean up all references to the now-obsolete flags.


timothysmith0609:
> I though we'd just stop documenting it.

You're right, I was overzealous, here. I've added back the `stdin` flag to deploy, global-deploy, and render. The documentation still has all references to `--stdin` removed, so it's undocumented now. I've also made sure the tests exercise cases were we use either `--filenames`, `--stdin`, and their combination

I've also added a deprecation warning to the logs if we end up parsing the stdin option.

lib/krane/cli/global_deploy_command.rb
```diff
@@ -8,7 +8,8 @@ class GlobalDeployCommand
         "filenames" => { type: :array, banner: 'config/deploy/production config/deploy/my-extra-resource.yml',
                          aliases: :f, required: false, default: [],
                          desc: "Directories and files that contains the configuration to apply" },
-        "stdin" => { type: :boolean, default: false, desc: "Read resources from stdin" },
+        "stdin" => { type: :boolean, default: false,
+                     desc: "Read resources from stdin" },
```
dturn: Let's add `[DEPRECATED]` to the descriptions

PR #688, opened 2020-02-10
Document passing multiple files to --bindings

sco11morgan:
Alleviate confusion on how to use multiple bindings files.  This was confusing when migrating from kubernetes-deploy that used multiple entries like `--bindings=@file1.yml --bindings=@file2.yml`

**How is this accomplished?**
Document an example of using multiple bindings files.

**What could go wrong?**
Nothing.  Its markdown!

### DIFF
--- a/README.md
+++ b/README.md
@@ -565,6 +565,9 @@ $ krane render --bindings='@config/production.json'
 
 # Load YAML file from ./config (.yaml or yml supported)
 $ krane render --bindings='@config/production.yaml'
+
+# Load multiple files via a space separated string
+$ krane render --bindings='@config/production.yaml' '@config/common.yaml'
 ```
 
 #### Using partials


dturn:
I've tagged Tim to review this, the good news is we have work in flight to restore the ability to allow passing flags multiple times.

README.md
```diff
@@ -565,6 +565,9 @@ $ krane render --bindings='@config/production.json'
 
 # Load YAML file from ./config (.yaml or yml supported)
 $ krane render --bindings='@config/production.yaml'
+
+# Load multiple files via a space separated string
+$ krane render --bindings='@config/production.yaml' '@config/common.yaml'
```
timothysmith0609: :+1: This looks ok to me. As @dturn mentioned we'll be adding back the ability to specify certain flags multiple times but that is blocked on an upstream PR in [thor](https://github.com/erikhuda/thor/pull/713). Once that merges I might remove this in favour of the repeatable method, but that's a consideration for the future.

PR #691, opened 2020-02-11
Ignore empty dry run errors

benlangfeld:

Fix https://github.com/Shopify/krane/issues/683#issuecomment-583858118

**How is this accomplished?**

Ignoring dry run failure with an empty error.

**What could go wrong?**

We might apply something that's broken. That's the best case scenario, since it's the only time we'd find out why it's broken.

### DIFF
--- a/lib/krane/kubectl.rb
+++ b/lib/krane/kubectl.rb
@@ -57,9 +57,9 @@ module Krane
             "The following command failed and cannot be retried"
           end
           logger.warn("#{warning}: #{Shellwords.join(cmd)}")
-          logger.warn(err) unless output_is_sensitive
+          logger.warn(err.inspect) unless output_is_sensitive
         else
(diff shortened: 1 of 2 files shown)


benlangfeld:
I'm unable to get minikube booted right now, so I can't deal with the tests for this.

benlangfeld:
@dturn If that's the case, we should stop using dry-run, because it cannot be trusted. This is causing us regular deployment failures.

dturn:
> we should stop using dry-run, because it cannot be trusted. 

Given the failures appear random, do you think retrying the command would be a viable fix?

benlangfeld:
@dturn I hope that #705 is acceptable. It seems like it may be a sufficient solution for our case.

benlangfeld:
Closing in favour of #705.

PR #693, opened 2020-02-11
Drop Ruby 2.4, Add Ruby 2.6, Add k8s 1.17

dturn:
Add tests for K8s 1.17
Add tests for Ruby 2.6
Ruby 2.4 is EOL 3/31 might as well drop it 6 weeks early as long as I'm in this code. 

**How is this accomplished?**
Update our pipeline file

**What could go wrong?**
Upset people who were still using 2.4. Though it's possible we don't cut a release before the 3/31 EoL date.

Sadly ruby 2.7 fails due to kuebclient https://github.com/abonas/kubeclient/issues/432

### DIFF
--- a/.shopify-build/krane.yml
+++ b/.shopify-build/krane.yml
@@ -9,7 +9,7 @@ steps:
       - bundle exec rubocop
     dependencies:
       - bundler
-  - label: 'Run Test Suite (:kubernetes: 1.17-latest :ruby: 2.7)'
+  - label: 'Run Test Suite (:kubernetes: 1.17-latest :ruby: 2.6)'
     command: bin/ci
     agents:
       queue: k8s-ci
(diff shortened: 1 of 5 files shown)


.shopify-build/krane.yml
```diff
@@ -9,6 +9,14 @@ steps:
...
     dependencies:
       - bundler
+  - label: 'Run Test Suite (:kubernetes: 1.17-latest :ruby: 2.6)'
+    command: bin/ci
+    agents:
+      queue: k8s-ci
+    env:
+      LOGGING_LEVEL: "4"
+      KUBERNETES_VERSION: v1.17-latest
+      RUBY_VERSION: "2.6"
```
KnVerey: How does this work? Isn't the container still `circleci/ruby:2.5.7-node-browsers`?

dturn: It installs ruby 2.6 https://buildkite.com/shopify/krane/builds/156#7815f427-a96f-4346-9448-4e35686c70a5/209-211

.shopify-build/krane.yml
```diff
@@ -1,6 +1,6 @@
 containers:
   default:
-    docker: circleci/ruby:2.4.6-node-browsers
+    docker: circleci/ruby:2.5.7-node-browsers
```
KnVerey: I wonder why we're using the `-node-browsers` image, especially since we're running our tests in their own container anyway.

dturn: probably copy/paste from a doc that assumes rails.

PR #696, opened 2020-02-14
fix: add an override for Job in cluster resource discovery

sco11morgan:
...
fix the error `no matches for kind "Job" in version "batch/v2alpha1"` https://github.com/Shopify/krane/issues/687

**How is this accomplished?**
adding an override for Job in cluster resource discovery

**What could go wrong?**
...

### DIFF
--- a/test/unit/cluster_resource_discovery_test.rb
+++ b/test/unit/cluster_resource_discovery_test.rb
@@ -50,7 +50,18 @@ class ClusterResourceDiscoveryTest < Krane::TestCase
     kinds = crd.prunable_resources(namespaced: true)
 
     assert_equal(kinds.length, 25)
-    %w(ConfigMap CronJob Deployment batch/v1/Job).each do |expected_kind|
+    %w(ConfigMap CronJob Deployment).each do |expected_kind|
+      assert kinds.one? { |k| k.include?(expected_kind) }
+    end
+  end
+
+  def test_prunable_namespaced_resources_apply_group_version_kind_overrides
+    Krane::Kubectl.any_instance.stubs(:run).with("api-versions", attempts: 5, use_namespace: false)
+      .returns([api_versions_full_response, "", stub(success?: true)])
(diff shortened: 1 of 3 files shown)


sco11morgan:
@dturn any idea why buildkite isn't kicking off for my PR?

dturn:
> @dturn any idea why buildkite isn't kicking off for my PR?

For security reasons, a Shopify employee has to manually trigger it on PRs from external contributors.  Thanks for the ping, I'll kick it off shortly.

dturn:
Test passed, the red x is due to a CI limitation on our side.

test/unit/cluster_resource_discovery_test.rb
```diff
@@ -50,7 +50,7 @@ def test_prunable_namespaced_resources
     kinds = crd.prunable_resources(namespaced: true)
 
     assert_equal(kinds.length, 25)
-    %w(ConfigMap CronJob Deployment).each do |expected_kind|
+    %w(ConfigMap CronJob Deployment batch/v1/Job).each do |expected_kind|
```
KnVerey: Shouldn't this just be "Job"? Otherwise of course L54 will only ever find one matching entry, before or after this PR. I guess the matcher needs to get more specific than just `include?` since it will also match "CronJob".

sco11morgan: @KnVerey the point of the test is to assert that the version override is working and a specific Job API is returned ("v1" not v1beta1 or v2alpha1)

dturn: @sco11morgan  The name of this test is  vague, it was originally designed to ensure that a few obvious Kinds were present. How would you feel about adding an additional block to test for full group/version/kind matches:

```
%w(batch/v1/Job batch/v1beta1/cronjob).each do |expected_gvk|
  assert kinds.include?(expected_gvk)
end

```

sco11morgan: 👍 moved the assertions to its own test

PR #709, opened 2020-03-24
Set the Ruby version in dev.yml according to #693

MrMarvin:
Solving version conflict between dev and bundler, making `dev up` complete successfully again.

**How is this accomplished?**
Setting the Ruby version in dev.yml. This seems to be an oversight in https://github.com/Shopify/krane/pull/693.

**What could go wrong?**
/

### DIFF
--- a/dev.yml
+++ b/dev.yml
@@ -1,7 +1,7 @@
 ---
 name: krane
 up:
-  - ruby: 2.4.6 # Matches gemspec
+  - ruby: 2.6.5 # Matches gemspec
   - bundler
   - homebrew:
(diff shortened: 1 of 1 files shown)


dev.yml
```diff
@@ -1,7 +1,7 @@
 ---
 name: krane
 up:
-  - ruby: 2.4.6 # Matches gemspec
+  - ruby: 2.6.5 # Matches gemspec
```
dturn: Can you set this 2.5.7 to match the gemspec.

dturn:
Fixed in https://github.com/Shopify/krane/pull/782

PR #710, opened 2020-03-31
fix various Ruby deprecation warnings

breunigs:

Remove Ruby deprecation warnings, even during testsuite run.

**How is this accomplished?**

Mostly just using backwards compatible syntax that doesn't emit warnings in newer Rubies. I had to add an additional check where Ruby's behaviour changed: https://github.com/Shopify/krane/pull/710/files#diff-590b908e9e2a650e8308c57e958bdb1aR194

**What could go wrong?**
I don't see what could, apart from funny edge cases in old Ruby versions. 

**Notes**
With https://github.com/Shopify/krane/pull/675 , the tip of https://github.com/abonas/kubeclient , and recursive-open-struct 1.1.1 the testsuite passes for me and only emits one deprecation warning. It's in mocha though and a doozy to fix.  I will test a full deployment using that version combination shortly.

The CLA has not been signed yet, but it's in progress.

### DIFF
--- a/lib/krane/kubernetes_resource/deployment.rb
+++ b/lib/krane/kubernetes_resource/deployment.rb
@@ -97,7 +97,7 @@ module Krane
       progress_condition.present? ? deploy_failing_to_progress? : super
     end
 
-    def validate_definition(*)
+    def validate_definition(*, **)
       super
 
       unless REQUIRED_ROLLOUT_TYPES.include?(required_rollout) || percent?(required_rollout)
--- a/lib/krane/kubernetes_resource/deployment.rb
+++ b/lib/krane/kubernetes_resource/deployment.rb
@@ -191,7 +191,7 @@ module Krane
     def min_available_replicas
       if percent?(required_rollout)
         (desired_replicas * required_rollout.to_i / 100.0).ceil
-      elsif max_unavailable =~ /%/
+      elsif max_unavailable.is_a?(String) && max_unavailable =~ /%/
         (desired_replicas * (100 - max_unavailable.to_i) / 100.0).ceil
       else
         desired_replicas - max_unavailable.to_i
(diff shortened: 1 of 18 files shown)


lib/krane/kubernetes_resource/deployment.rb
```diff
@@ -191,7 +191,7 @@ def find_latest_rs(cache)
     def min_available_replicas
       if percent?(required_rollout)
         (desired_replicas * required_rollout.to_i / 100.0).ceil
-      elsif max_unavailable =~ /%/
+      elsif max_unavailable.is_a?(String) && max_unavailable =~ /%/
```
dturn: Why the `.is_a?(String)` ?

breunigs: To get rid of this warning:
```
[1] pry(main)> obj = 5; obj =~ /%/
(pry):1: warning: deprecated Object#=~ is called on Integer; it always returns nil
```
It shows up in the test suite, and it seems likely it can be either an integer or string (e.g. `"25%"`) in production.

dturn:
Two depreciations warnings I noticed https://github.com/Shopify/krane/pull/711#issuecomment-606756418

breunigs:
@dturn I cherry picked the patches into this branch. The CLI has been signed, or at least submitted (it's a corporate one, and while I'm not the person signing it, I am included in the list of people).

test/test_helper.rb
```diff
@@ -201,7 +201,7 @@ def stub_kubectl_response(*args, kwargs: {}, resp:, err: "", success: true, json
     def build_runless_kubectl
       obj = Krane::Kubectl.new(task_config: task_config(namespace: 'test'),
         log_failure_by_default: false)
-      def obj.run(*)
+      def obj.run(*,**)
```
dturn, line 204:
```
test/test_helper.rb:204:20: C: Layout/SpaceAfterComma: Space missing after comma.
      def obj.run(*,**)

```

PR #675, opened 2020-01-24
Fix some warnings on Ruby 2.7.0

airhorns:
**What are you trying to accomplish with this PR?**

Fix a few warnings coming out during deploys using ruby 2.7.0 like so:

```
[INFO][2020-01-24 11:17:20 -0500]	------------------------------------Phase 1: Initializing deploy------------------------------------
/Users/airhorns/.rbenv/versions/2.7.0/lib/ruby/gems/2.7.0/gems/krane-1.1.1/lib/krane/statsd.rb:41: warning: Using the last argument as keyword parameters is deprecated; maybe ** should be added to the call
/Users/airhorns/.rbenv/versions/2.7.0/lib/ruby/gems/2.7.0/gems/krane-1.1.1/lib/krane/deploy_task.rb:254: warning: The called method `validate_configuration' is defined here
[INFO][2020-01-24 11:17:25 -0500]	All required parameters and files are present
[INFO][2020-01-24 11:17:25 -0500]	Discovering resources:
[INFO][2020-01-24 11:17:28 -0500]	  - Ingress/tunnel-server-ingress
[INFO][2020-01-24 11:17:28 -0500]	  - Service/tunnel-server
[INFO][2020-01-24 11:17:28 -0500]	  - Deployment/tunnel-server
[INFO][2020-01-24 11:17:28 -0500]	  - Certificate/localtunnel.gadget.dev-production
/Users/airhorns/.rbenv/versions/2.7.0/lib/ruby/gems/2.7.0/gems/krane-1.1.1/lib/krane/kubernetes_resource/custom_resource.rb:66: warning: Using the last argument as keyword parameters is deprecated; maybe ** should be added to the call
/Users/airhorns/.rbenv/versions/2.7.0/lib/ruby/gems/2.7.0/gems/krane-1.1.1/lib/krane/kubernetes_resource.rb:135: warning: The called method `validate_definition' is defined here
/Users/airhorns/.rbenv/versions/2.7.0/lib/ruby/gems/2.7.0/gems/krane-1.1.1/lib/krane/kubernetes_resource/deployment.rb:101: warning: Using the last argument as keyword parameters is deprecated; maybe ** should be added to the call
/Users/airhorns/.rbenv/versions/2.7.0/lib/ruby/gems/2.7.0/gems/krane-1.1.1/lib/krane/kubernetes_resource.rb:135: warning: The called method `validate_definition' is defined here
<snip>
[...]

PR #711, opened 2020-03-31
Fix more ruby2 7 deprecation warnings

dturn:
This is https://github.com/Shopify/krane/pull/710 + an extra commit for adding a ruby 2.7 test in ci

### DIFF
--- a/.rubocop-http---shopify-github-io-ruby-style-guide-rubocop-yml
+++ b/.rubocop-http---shopify-github-io-ruby-style-guide-rubocop-yml
@@ -1,5 +1,3 @@
-# Recommended rubocop version: ~> 0.78.0
-
 AllCops:
   Exclude:
   - 'db/schema.rb'
@@ -77,13 +75,6 @@ Style/BlockDelimiters:
   - lambda
   - proc
   - it
-
-Style/BracesAroundHashParameters:
-  EnforcedStyle: no_braces
-  SupportedStyles:
-  - braces
-  - no_braces
-  - context_dependent
 
 Layout/CaseIndentation:
   EnforcedStyle: end
--- a/.shopify-build/krane.yml
+++ b/.shopify-build/krane.yml
@@ -9,14 +9,21 @@ steps:
       - bundle exec rubocop
     dependencies:
       - bundler
-  - label: 'Run Test Suite (:kubernetes: 1.17-latest :ruby: 2.6)'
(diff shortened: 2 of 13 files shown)


dturn:
The tests pass but there are still some warnings:

```ruby
GlobalDeployTest#test_global_deploy_task_empty_selector_validation_failure /usr/src/app/lib/krane/statsd.rb:41: warning: Using the last argument as keyword parameters is deprecated; maybe ** should be added to the call
--
  | /usr/src/app/lib/krane/resource_deployer.rb:70: warning: The called method `deploy_all_resources' is defined here
 
```


```ruby
KraneTest#test_render_black_box /usr/src/app/lib/krane/kubernetes_resource/deployment.rb:101: warning: Using the last argument as keyword parameters is deprecated; maybe ** should be added to the call
--
  | /usr/src/app/lib/krane/kubernetes_resource.rb:135: warning: The called method `validate_definition' is defined here
```


```ruby
/usr/local/bundle/gems/rake-10.5.0/lib/rake/application.rb:381: warning: deprecated Object#=~ is called on Proc; it always returns nil
```
 This appears to be fixed in ruby 12.3.2 but that would require forcing users off of ruby 2.5 and I'm not sure we're there yet https://github.com/ruby/rake/issues/308. A compormise is probably relaxing the rake requirement to just be >= 11

breunigs:
Ah, sorry, I forgot to mention rake in my original description. Yes, I didn't fix those.

As for the other ones you mentioned, I think those are fixed through https://github.com/Shopify/krane/pull/675 – at least the source lines match and I don't see them locally.

How do you want to proceed here? I can incorporate them into my branch, likewise your 2.7 CI patch, and/or give you access to my branch so we don't spread the discussion over multiple PRs. Either way, since the tests pass locally for me using my patch + #675 I'm not sure what is broken. I'm testing locally against minikube 1.9 with kube 1.14, if that makes a difference.

dturn:
If you'd like you can pull my changes and 675 into your branch. Once you've done that and signed the CLA, ping me.

The other option, is just sign the CLA ping me, I'll merge your branch and then I'll rebase this branch and pull in 675. Both work.

breunigs:
Thank you, sounds doable. I'll see to the CLA because that's the blocker here.

dturn:
Extra code in this PR was added back to  https://github.com/Shopify/krane/pull/710

PR #719, opened 2020-06-12
Allow to set image tag for task runner

GustavoCaso:
When running `activefailover` we usually use the predefined value of the image that is calculated at the deploy time on `Shopify/shopify`. This is great, but when you want to test custom code with having to deploy `Shopify/shopify` is getting very painfully, we have to update multiple `task-runner` templates using `kbctl` and that operation is tedious and error-prone.

By allowing to pass the image to `Krane` we can programmatically make the task of testing changes in core much easier.

This is the PR in `activefailover` that utilize this feature https://github.com/Shopify/activefailover/pull/1353

**How is this accomplished?**
Passing a new keyword argument to `image` with a default value of `nil` so it is backward compatible.

I had to refactor `build_pod` because rubocop was complaining that there were too many positional arguments. I extracted the logic to a new collaborator object `ContainerOverrides`

I think there is room for more improvements, but I rather get some opinions from the maintainers first 😄 

@Shopify/job-patterns
@Shopify/pod-failovers

### DIFF
--- a/lib/krane/container_overrides.rb
+++ b/lib/krane/container_overrides.rb
@@ -1,0 +1,33 @@
+# frozen_string_literal: true
+module Krane
+  class ContainerOverrides
+    attr_reader :command, :arguments, :env_vars, :image_tag
+
+    def initialize(command: nil, arguments: nil, env_vars: [], image_tag: nil)
+      @command = command
+      @arguments = arguments
+      @env_vars = env_vars
+      @image_tag = image_tag
+    end
+
+    def run!(container)
+      container.command = command if command
+      container.args = arguments if arguments
+
+      if image_tag
+        image = container.image
+        base_image, _old_tag = image.split(':')
+        new_image = "#{base_image}:#{image_tag}"
+
+        container.image = new_image
+      end
+
+      env_args = env_vars.map do |env|
+        key, value = env.split('=', 2)
+        { name: key, value: value }
+      end
+      container.env ||= []
+      container.env = container.env.map(&:to_h) + env_args
+    end
+  end
+end
--- a/lib/krane/container_overrides.rb
+++ b/lib/krane/container_overrides.rb
@@ -10,7 +10,7 @@ module Krane
       @image_tag = image_tag
     end
 
-    def run!(container)
+    def apply!(container)
       container.command = command if command
       container.args = arguments if arguments
 
--- a/lib/krane/runner_task.rb
+++ b/lib/krane/runner_task.rb
@@ -108,7 +108,8 @@ module Krane
       task_template = get_template(template_name)
       @logger.info("Using template '#{template_name}'")
       pod_template = build_pod_definition(task_template)
-      set_container_overrides!(pod_template, container_overrides)
+      container = extract_task_runner_container(pod_template)
+      container_overrides.apply!(container)
       ensure_valid_restart_policy!(pod_template, verify_result)
       Pod.new(namespace: @namespace, context: @context, logger: @logger, stream_logs: true,
                     definition: pod_template.to_hash.deep_stringify_keys, statsd_tags: [])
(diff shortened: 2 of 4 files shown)


test/integration/runner_task_test.rb
```diff
@@ -284,6 +284,24 @@ def test_run_adds_env_vars_provided_to_the_task_container
...
   end
 
+  def test_run_adds_custom_image_provided_to_the_task_container
+    deploy_task_template
+
+    task_runner = build_task_runner
+    result = task_runner.run(
+      template: 'hello-cloud-template-runner',
+      command: ['/bin/sh', '-c'],
+      arguments: ['echo "The value is: $MY_CUSTOM_VARIABLE"'],
```
dturn, line 294: Does this do the right thing w/o actually setting a env_var called `MY_CUSTOM_VARIABLE` ?

GustavoCaso: Good point. I'll remove the `env_var` part from `arguments`

lib/krane/runner_task.rb
```diff
@@ -98,11 +104,11 @@ def create_pod(pod)
       raise FatalDeploymentError, msg
     end
 
-    def build_pod(template_name, command, args, env_vars, verify_result)
+    def build_pod(template_name, container_overrides, verify_result)
       task_template = get_template(template_name)
       @logger.info("Using template '#{template_name}'")
       pod_template = build_pod_definition(task_template)
-      set_container_overrides!(pod_template, command, args, env_vars)
+      set_container_overrides!(pod_template, container_overrides)
```
dturn, line 111: Should we move this entire function into the ContainerOverrides class and make this line something like
`pod_template = container_overrides.run!(pod_template)`

test/unit/krane/container_overrides_test.rb
```diff
@@ -0,0 +1,72 @@
+# frozen_string_literal: true
+require 'test_helper'
+
+module Krane
+  class ContainerOverridesTest < Krane::TestCase
+    def setup
+      super
+      @container = Kubeclient::Resource.new(
+        name: "task-runner",
+        image: "gcc:3.2",
```
dturn, line 10: We're using `busybox` everywhere else, why `gcc` here?

GustavoCaso: Nothing in particular. I can change if you want 😄

lib/krane/runner_task.rb
GustavoCaso: I not entirely. If we do that we create a dependency that `container_overrides` can only by run by `task_runner`

Since we have a validation before overriding the container.

```ruby
def set_container_overrides!(pod_definition, container_overrides)
  container = pod_definition.spec.containers.find { |cont| cont.name == 'task-runner' }
  if container.nil?
    message = "Pod spec does not contain a template container called 'task-runner'"
    @logger.summary.add_paragraph(message)
    raise TaskConfigurationError, message
  end

  container_overrides.run!(container)
end
```

I think is better if we leave the validation inside `task_runner`

GustavoCaso: I can modify the method `build_pod` create a new method called `extract_task_runner_container` which only takes care of the validation from above.

And finally, we can simply run `container_overrides.run!(container)`

lib/krane/container_overrides.rb
```diff
@@ -0,0 +1,33 @@
...
+    attr_reader :command, :arguments, :env_vars, :image_tag
+
+    def initialize(command: nil, arguments: nil, env_vars: [], image_tag: nil)
+      @command = command
+      @arguments = arguments
+      @env_vars = env_vars
+      @image_tag = image_tag
+    end
+
+    def run!(container)
```
KnVerey, line 13: nit: how about `apply!(container)`?

lib/krane/runner_task.rb
KnVerey: I like your proposal, though it isn't a blocker for me

PR #723, opened 2020-06-22
Add 1.18 to CI

dturn:
K8s 1.18 is out (and we have a bug filed against it). So let's add it to CI

**How is this accomplished?**
Update our pipeline file

**What could go wrong?**
Its one more CI pipeline, so worst case is CI is red.

### DIFF
--- a/.shopify-build/krane.yml
+++ b/.shopify-build/krane.yml
@@ -9,13 +9,13 @@ steps:
       - bundle exec rubocop
     dependencies:
       - bundler
-  - label: 'Run Test Suite (:kubernetes: 1.17-latest :ruby: 2.7)'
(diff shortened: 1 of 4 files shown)


dturn:
k8s-ci appears not to handle v.1.18 last time the fix was https://github.com/Shopify/k8s-ci/pull/39

.shopify-build/krane.yml
```diff
@@ -6,16 +6,15 @@ steps:
   - label: Lint
     timeout: 5m
     run:
+      - bundle: ~
```
timothysmith0609, line 9: what does this syntax do?

dturn: I believe it says run bundler and cache the results. I followed the instructions at the bottom of https://shopify-build.docs.shopify.io/troubleshooting/deprecations/

PR #730, opened 2020-07-19
Always set a deployment_id, even if the current_sha isn't set

airhorns:
The deployment_id seems like it should always be present, even if current_sha isn't set. Before this change, deployment_id would be nil if current_sha was not set, which means predeployed pods like migrations and whatnot collide. It doesn't seem like a stretch to just use a random string as a deployment ID if the current deployment ID scheme is just using randomness too, so this adds that as a fallback, and adds a test!

**What could go wrong?**

Pods using `deployment_id` in their name would now start having unique names instead of colliding when `current_sha` is unset. I kind of struggle to contrive a situation where that would matter, and especially at Shopify `TASK_ID` is set by Shipit so it seems unlikely Shopify would be relying on this behaviour. But, it is a change in behaviour. I don't think many folks are running with both `current_sha` and `TASK_ID` unset.

### DIFF
--- a/test/unit/krane/renderer_test.rb
+++ b/test/unit/krane/renderer_test.rb
@@ -4,6 +4,7 @@ require 'test_helper'
 class RendererTest < Krane::TestCase
   def setup
     super
+    SecureRandom.stubs(:hex).returns("aaaa")
     @renderer = Krane::Renderer.new(
       current_sha: "12345678",
       template_dir: fixture_path('for_unit_tests'),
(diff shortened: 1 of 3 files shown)


dturn:
@airhorns Have you looked at using the `generateName: ` field for pods (https://github.com/Shopify/krane/pull/608)? It might be a more reliable solution to your problem. If it wont work let me know and I'll 👀  this PR again.

dturn:
Two CI failures:


Failure:
--
  | RendererTest#test_deployment_id [/usr/src/app/test/unit/krane/renderer_test.rb:92]
  | Minitest::Assertion: --- expected
  | actual | 0s
  | @@ -2,7 +2,7 @@
  | apiVersion: v1
  | kind: Pod
  | metadata:
  | -  name: migrate-12345678-aaaa
  | +  name: migrate-12345678-f99f8185
  | spec:
  | containers:
  | - name: migrate
  |  
  | Failure:
  | RendererTest#test_renderer_without_current_sha_still_has_a_deployment_id [/usr/src/app/test/unit/krane/renderer_test.rb:114]
  | Minitest::Assertion: --- expected
  | actual | 1s
  | @@ -2,7 +2,7 @@
  | apiVersion: v1
  | kind: Pod
  | metadata:
  | -  name: migrate-aaaa
  | +  name: migrate-b971b0afef89cc61
  | spec:
  | containers:
  | - name: migrate

airhorns:
Does buildkite set `TASK_ID` in the environment?

test/unit/krane/renderer_test.rb
```diff
@@ -69,10 +69,48 @@ def test_nesting_fields
...
+      template_dir: fixture_path("for_unit_tests"),
+      logger: logger,
+      bindings: { "a" => "1", "b" => "2" },
+    )
+    expected = <<~EOY
+      ---
+      apiVersion: v1
+      kind: Pod
+      metadata:
+        name: migrate-aaaa
```
dturn, line 107: Without  a task_id or a a current_sha set you're going to get random bits afte the `-`. Did this pass locally for you?

airhorns: Oh oops I messed up and missed committing the mocha call I was making locally. Was having some trouble getting my editor to agree with the committed code format and was selectively staging, force pushed the correct code.

dturn:
https://github.com/Shopify/krane/pull/430/files yep this was an oversight when we made current-sha optional. We're having CI issues on 1.16 going to merge without the ✅

PR #733, opened 2020-07-22
Pin k8s 1.16 CI to 1.16.12

RyanBrushett:

There appears to be an issue specific to our CI and k8s v1.16.13 so I'm pinning CI to 1.16.12 in the meantime while we investigate.

**How is this accomplished?**

Edit the shopify-build config.

### DIFF
--- a/.shopify-build/krane.yml
+++ b/.shopify-build/krane.yml
@@ -23,13 +23,14 @@ steps:
     env:
       LOGGING_LEVEL: "4"
       KUBERNETES_VERSION: v1.17-latest
-  - label: 'Run Test Suite (:kubernetes: 1.16-latest)'
+  - label: 'Run Test Suite (:kubernetes: 1.16.12)'
     command: bin/ci
     agents:
       queue: k8s-ci
     env:
       LOGGING_LEVEL: "4"
-      KUBERNETES_VERSION: v1.16-latest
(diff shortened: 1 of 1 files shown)


dturn:
Can you recreate the failure locally with minikube?

RyanBrushett:
@dturn I haven't been able to locally as of right now. The tests themselves just timeout (and take forever) and I haven't yet had the chance to recreate the exact scenarios the tests are testing. It's top on my to-do list.
I'll do that and report back. Meanwhile, this remains an option but I won't merge until we're confident things aren't hella bad.

RyanBrushett:
I found a test that I can easily make fail locally when run against k8s 1.16.13:

```
bundle exec ruby -I test test/integration/runner_task_test.rb -n test_run_with_verify_result_fails_quickly_if_the_pod_is_deleted_out_of_band
```

I wrote a little shell script to fire up the same conditions as the test and it doesn't fail 🤷 
I run it by hand and it also doesn't fail.

I stuck a `sleep 20` between `validate_pod` and `create_pod` in `RunnerTask#Run!` and the test started to pass and then kept passing once the sleep was removed (the SA is there now and things are OK).

I'll revisit this in another PR but I think we're OK to pin CI on .12 in the meantime.

PR #734, opened 2020-07-23
Stop core service from being deployed a priority resource

beckygowland:
closes https://github.com/Shopify/krane/issues/728
Creating a custom resource def that has the same `kind` type as a core type can cause deploying a core type to be considered a priority resource. 

**How is this accomplished?**
We should also consider `group` when deciding if resources are crd's. The predeploy_sequence array list was converted to a hash keyed by kind

**What could go wrong?**
Could this mess up priority decision logic for other resources? Tests don't seem to be broken for crs so 👍 sign

### DIFF
--- a/lib/krane/deploy_task.rb
+++ b/lib/krane/deploy_task.rb
@@ -56,6 +56,7 @@ module Krane
     )
 
     def predeploy_sequence
+      default_group = { :group => nil }
       before_crs = %w(
         ResourceQuota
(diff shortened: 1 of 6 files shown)


beckygowland:
Spoke too soon, seems like a couple tests are failing, will 👀

beckygowland:
Failing tests are now the same ones as master

PR #738, opened 2020-08-03
Remove kubernetes deploy annotation prefix

### LINKED ISSUE #526: Remove support for `kubernetes-deploy.shopify.io` annotations

kddeisz:

Fixes #526.

**How is this accomplished?**

Effectively removes all instances of reference to the `kubernetes-deploy.shopify.io` annotation namespace. I've added a `Krane::Annotation::for` method that handles prefixing everything so that that's consolidated into one place. This also removes annotations for ejson secrets.

**What could go wrong?**

If there was some kind of Ruby dependency on certain constants that would be bad.


lib/krane/kubernetes_resource.rb
```diff
@@ -134,22 +131,12 @@ def to_kubeclient_resource
...
       validate_spec_with_kubectl(kubectl)
       @validation_errors.present?
     end
 
-    def validation_warning_msg
-      @validation_warnings.join("\n")
-    end
-
-    def has_warnings?
-      @validation_warnings.present?
```
timothysmith0609, line 150: So the only validation warnings we ever generated are from deprecated kubernetes.shopify.io annotations? Are we missing anything else or is that really it?

kddeisz: Yeah that's it. I went based on this commit: https://github.com/Shopify/krane/commit/d237e0c80ab36db6d77577858a9de76b6f781b40 where the annotations were first deprecated. The warnings were added to support that.

xaviablaza:
Thank you @kddeisz for this PR!

PR #741, opened 2020-08-21
Ensure YAML serialization is Kubernetes-compatible in regards to string scalars with scientific number format

jpfourny:
Fixing issue: https://github.com/Shopify/krane/issues/740 (see for details)
TLDR; `"1e3"` renders as `1e3` (unquoted), which is parsed as `1000` by K8s (or anything YAML 1.2 compliant).

**How is this accomplished?**
- Created K8sCompatibleYAMLTree visitor to handle scientific/e-notation string case.
- Replaced serialization calls to use the new visitor.

**What could go wrong?**
Hopefully nothing ;)


lib/krane/psych_k8s_patch.rb
```diff
@@ -0,0 +1,60 @@
+# frozen_string_literal: true
+
+require 'psych'
+
+# PsychK8sPatch applies patching to Psych.dump/dump_stream with an alternative string serializer that is more compatible
+# with Kubernetes (and other Go-based tooling). Related issue: https://github.com/Shopify/krane/issues/740
+module PsychK8sPatch
+  # Activate will apply the patch after creating backup aliases for the original methods.
+  def self.activate
+    class << Psych
```
timothysmith0609, line 10: See https://github.com/Shopify/krane/pull/742 for an alternative implementation

kddeisz: Interesting! One second let me put together some thoughts

kddeisz: Okay, so lots of thoughts on this.

First of all, this solution definitely works. I don't think there's anything particularly wrong with it at all, other than it _feels_ tough to stomach since it's a monkey-patch.

I think what Tim is getting at with his solution is scoping the monkey-patch locally with a refinement. I think a refinement could also work here, but since there's only one place where we're truly dumping out to YAML I think we can be even more explicit by just defining the behavior we want.

Fundamentally it seems like (and please correct my if I'm wrong here) we're trying to change the behavior so that it uses double quotes when we're visiting a string that represents some scientific notation. I think a more clear way to achieve that would be to explicitly define our own visitor and then use it. As in:

```ruby
module Psych
  module Visitors
    class ScientificYAMLTree < YAMLTree
      def visit_String(string)
        return super unless string.match?(/\A[+-]?\d+(?:\.\d+)?[eE][+-]?\d+\z/)
        @emitter.scalar(string, nil, nil, true, true, Nodes::Scalar::DOUBLE_QUOTED)
      end
    end
  end

  def self.dump_scientific(object)
    visitor = Visitors::ScientificYAMLTree.create
    visitor << object
    visitor.tree.yaml
  end
end
```

Then with that in place, calling something like `YAML.dump_scientific('123e4')` will yield `"--- \"123e4\"\n"`. I like this a little better personally because it touches slightly less internals in `Psych`, it's a locally scoped change, and it's explicit about the behavior that we want changed.

jpfourny:
Updated as per Kevin's suggestions.

lib/krane/version.rb
```diff
@@ -1,4 +1,4 @@
 # frozen_string_literal: true
 module Krane
-  VERSION = "1.1.4"
+  VERSION = "1.1.5"
```
timothysmith0609, line 3: Generally we do version bumps atomically (that is, a PR with _only_ the version change and no logic changes). Let's set this back to 1.1.4 and once this merges we can make the version bump PR

jpfourny: Will do!

jpfourny: I assume this also includes the CHANGELOG?

timothysmith0609: >I assume this also includes the CHANGELOG?

You can keep the changelog entries (it makes it easier since we won't have to backfill them during the version bump)

jpfourny: 👍

PR #745, opened 2020-09-07
Update minikube cask location in dev.yml

DazWorrall:

Had errors running `dev up`:

> Error: caskroom/cask was moved. Tap homebrew/cask instead.

**How is this accomplished?**

Updating the cask location.

**What could go wrong?**

Not much: it's only a dev.yml. I was able to dev up after this change, please give it a try yourselves.

### DIFF
--- a/dev.yml
+++ b/dev.yml
@@ -4,7 +4,7 @@ up:
   - ruby: 2.5.7 # Matches gemspec
   - bundler
   - homebrew:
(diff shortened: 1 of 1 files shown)


timothysmith0609 (approved):
Since this is only internal to Shopify I don't see a problem with this.

gigr (approved):
@rdji123 and I also ran into this.

PR #746, opened 2020-09-07
Allow specifying a kubeconfig per task in the internal API

DazWorrall:

Being primarily a cli tool, Krane currently relies on some global state in the environment - notable for this issue is the kubeconfig path. Deploying our main monolith at Shopify is a bit of a pathological case - we use Krane as a library, and deploy to multiple targets simultaneously using the internal API and threads.

Occasionally we are seeing some errors I believe are ultimately thread safety issues - e.g. apparent corruption of the kubeconfig as contexts are 'missing' (when we check the files, they are not), or base64 certificate data is invalid and un-decodeable. I have not been able to determine the precise cause here (e.g. fd leaks) as reproducing is very difficult, but would like to at least work around the symptom we're seeing by giving each deploy task its own kubeconfig instead of sharing one everywhere. This PR alters the internal API to allow this, while not changing the behaviour for cli users, and in doing so remove a little of this global state.

**How is this accomplished?**

Having the various `Task` classes take `kubeconfig` as a parameter, and ensuring that is passed through to `Kubectl` and `Kubeclient` instances. If `kubeconfig` is nil, the value of `KUBECONFIG` in the environment is used, preserving the current behaviour.

There is some prior art here. An explicit `--kubeconfig` option was added in #428 and then reverted in #468 as the cli flag did not support parsing multiple files. This approach overrides the `KUBECONFIG` envvar in the kubectl child process, which should avoid that issue.

I have removed a bunch of memoizing from the test suite as some tests currently rely on mutating global state, notably the multiple config file integration test. It's likely we can keep some of this optimisation by reworking these tests, but it could use someone with better knowledge of the test suite than me.

The various `Task` classes share a common set of options and setup, it might be worth refactoring these to share a base class and dedup some of the changes here, but I consider that out of scope for this PR.

**What could go wrong?**

As I noted above a change similar to this was attempted before and abandoned. I believe this approach of modifying the subprocess env avoids the previous issue, but maybe there's some other problem. I hope to test this branch by deploying our monolith to check for other issues, and that it actually resolves this issue we're seeing.

I have done some sanity 🎩 locally with minikube and fixtures.

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -7,6 +7,7 @@ Rake::TestTask.new(:integration_test) do |t|
   t.libs << "test"
   t.libs << "lib"
   t.test_files = FileList['test/integration/**/*_test.rb']
+  t.warning = false
 end
 
 desc("Run integration tests that CANNOT be run in parallel")
@@ -14,6 +15,7 @@ Rake::TestTask.new(:serial_integration_test) do |t|
   t.libs << "test"
   t.libs << "lib"
   t.test_files = FileList['test/integration-serial/**/*_test.rb']
+  t.warning = false
 end
 
 desc("Run unit tests")
@@ -21,6 +23,7 @@ Rake::TestTask.new(:unit_test) do |t|
   t.libs << "test"
   t.libs << "lib"
   t.test_files = FileList['test/unit/**/*_test.rb']
+  t.warning = false
 end
 
 desc("Run cli tests")
@@ -28,6 +31,7 @@ Rake::TestTask.new(:cli_test) do |t|
   t.libs << "test"
   t.libs << "lib"
   t.test_files = FileList['test/exe/**/*_test.rb']
+  t.warning = false
 end
 
 desc("Run all tests")
--- a/Rakefile
+++ b/Rakefile
@@ -7,7 +7,6 @@ Rake::TestTask.new(:integration_test) do |t|
   t.libs << "test"
   t.libs << "lib"
   t.test_files = FileList['test/integration/**/*_test.rb']
-  t.warning = false
 end
 
 desc("Run integration tests that CANNOT be run in parallel")
@@ -15,7 +14,6 @@ Rake::TestTask.new(:serial_integration_test) do |t|
   t.libs << "test"
   t.libs << "lib"
   t.test_files = FileList['test/integration-serial/**/*_test.rb']
-  t.warning = false
 end
 
 desc("Run unit tests")
@@ -23,7 +21,6 @@ Rake::TestTask.new(:unit_test) do |t|
   t.libs << "test"
   t.libs << "lib"
   t.test_files = FileList['test/unit/**/*_test.rb']
-  t.warning = false
 end
 
 desc("Run cli tests")
@@ -31,7 +28,6 @@ Rake::TestTask.new(:cli_test) do |t|
   t.libs << "test"
   t.libs << "lib"
   t.test_files = FileList['test/exe/**/*_test.rb']
-  t.warning = false
 end
 
 desc("Run all tests")
--- a/lib/krane/deploy_task.rb
+++ b/lib/krane/deploy_task.rb
@@ -101,10 +101,10 @@ module Krane
     # @param render_erb [Boolean] Enable ERB rendering
     def initialize(namespace:, context:, current_sha: nil, logger: nil, kubectl_instance: nil, bindings: {},
       global_timeout: nil, selector: nil, filenames: [], protected_namespaces: nil,
-      render_erb: false)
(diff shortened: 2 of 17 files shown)


Rakefile
```diff
@@ -7,27 +7,31 @@ Rake::TestTask.new(:integration_test) do |t|
...
   t.test_files = FileList['test/unit/**/*_test.rb']
+  t.warning = false
 end
 
 desc("Run cli tests")
 Rake::TestTask.new(:cli_test) do |t|
   t.libs << "test"
   t.libs << "lib"
   t.test_files = FileList['test/exe/**/*_test.rb']
+  t.warning = false
```
DazWorrall, line 34: I was really struggling to crawl through the test output with the warnings on. Happy to move this elsewhere though if there's more to do.

jpfourny:
This looks reasonable to me, but folks like @timothysmith0609 and @kddeisz may have stronger opinions.

Rakefile
```diff
@@ -7,27 +7,31 @@ Rake::TestTask.new(:integration_test) do |t|
   t.libs << "test"
   t.libs << "lib"
   t.test_files = FileList['test/integration/**/*_test.rb']
+  t.warning = false
```
kddeisz, line 10: Don't want to miss anything important here, but maybe we can silence the gems that are causing these warnings that are outside of our control. Either way I don't think this should be part of this PR.

lib/krane/deploy_task.rb
```diff
@@ -191,7 +193,7 @@ def resource_deployer
     end
 
     def kubeclient_builder
-      @kubeclient_builder ||= KubeclientBuilder.new
+      @kubeclient_builder ||= KubeclientBuilder.new(kubeconfig: @task_config.kubeconfig)
```
kddeisz, line 196: This is now in a lot of places, so don't love that. Seems like every time we're creating a kubeclientbuilder object from a task config, so why don't we make that a method on the task config that will return a kubeclientbuilder. That way we don't have to pass a task_config around too often.

test/unit/krane/deploy_task_test.rb
```diff
@@ -23,4 +23,17 @@ def test_initializer_without_valid_file
...
+    task = Krane::DeployTask.new(
+      namespace: "something",
+      context: KubeclientHelper::TEST_CONTEXT,
+      logger: logger,
+      current_sha: "",
+      filenames: ["unknown"],
+      kubeconfig: '/some/path.yml',
+    )
+    assert_equal('/some/path.yml', task.task_config.kubeconfig)
+    assert_equal(['/some/path.yml'], task.send(:kubeclient_builder).kubeconfig_files)
```
kddeisz, line 37: Really don't love this here since we're testing a private method (and in the other couple of test files that do this). This smells that we're not exposing an API that we should be. I think making task_config the source of truth for kubeclient_builder obejcts as mentioned above will solve this.

kddeisz (changes requested):
Generally +1 on this PR, just a couple of architectural changes I'd like to see first.

Rakefile
ajshepley: What warnings were you seeing? When I ran the tests, the only output I got was the test suite summaries, and a large log output from a failed assertion upon the log string.

DazWorrall: I don't remember exactly, it's possible it was just in CI.

ajshepley: Alrighty. I'll open a separate draft PR (targeting this one) with the original rakefile and I'll see if I can repro.

test/unit/krane/deploy_task_test.rb
ajshepley: After moving the method to kubeconfig I added the test to the kubeconfig test, sans the send call. I wasn't sure whether to fully remove the other tests as it feels like there's a small bit of value in instantiating and checking the config, but I might have been overthinking things.

ajshepley:
Since deploying this last week, we haven't seen any of the corruption issues crop up in core.

kddeisz (changes requested):
Yeah this is looking good. Last thing for me is don't delegate directly to the ivar. i.e., change:

```ruby
delegate :kubeclient_builder, to: :@task_config
```

to

```ruby
delegate :kubeclient_builder, to: :task_config
```

ajshepley:
Ah, I missed that and only thought about the task config internally. Whoops. Will fix, thanks.

kddeisz (approved):
Looks good!

ajshepley:
I accidentally overlooked some lint errors (the buildkite job spits out a looooot of stuff and the error is in the middle somewhere) but it should be good now :)

PR #751, opened 2020-09-24
Refactor/update attempts kubectl apply

rdji123:
Addressing context deadline issues associated with the [apply command](https://sh.splunk.shopifysvc.com/en-GB/app/search/causes_of_deploy_failures?form.time_field.earliest=-14d%40d&form.time_field.latest=now&form.repo=*)

Increasing the number of attempts on the apply command will not interfere with the syncing of resources I believe, as the resource watcher is initiated after the apply command and begins syncing [here.](https://github.com/Shopify/krane/blob/master/lib/krane/resource_deployer.rb#L128) As well any apply failures will be caught [here](https://github.com/Shopify/krane/blob/master/lib/krane/resource_deployer.rb#L161) reaching a DeploymentError before a sync. 

The Kubectl apply command is also idempotent, should be safe to retry.

### DIFF
--- a/.rubocop-http---shopify-github-io-ruby-style-guide-rubocop-yml
+++ b/.rubocop-http---shopify-github-io-ruby-style-guide-rubocop-yml
@@ -264,7 +264,7 @@ Style/MethodCallWithArgsParentheses:
   - raise
   - puts
   Exclude:
-  - Gemfile
+  - '**/Gemfile'
 
 Style/MethodDefParentheses:
   EnforcedStyle: require_parentheses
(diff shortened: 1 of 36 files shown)


timothysmith0609:
This looks good to me except for the linter error in CI. Are you able to investigate that problem?

timothysmith0609:
Ah, ok I'll take a look at fixing that, then. Sorry about that

rdji123:
cc @Shopify/pipeline

ayatsynych (approved):
👏 let's see if with 2 reties we still see occurrences of this problem

timothysmith0609 (approved):

ajshepley:
If you rebase with master, you'll get the rubocop changes/fixes.

rdji123:
I rebased it! just waiting for to pass CI 🤞

PR #752, opened 2020-09-25
Updated rubocop using bundlelock.

ajshepley:
Update rubocop after recent changes and ran `rubocop -a` to fix some issues.

**How is this accomplished?**
I ran `bundle lock --update=rubocop` as prescribed in slack. Alternatively, there are other options but those may take a more involved PR.

**What could go wrong?**
Rubocop lint CI may fail, but it's already failing.


jpfourny (approved):
Looks good to me.

PR #753, opened 2020-10-01
Deploy method override annotation

timothysmith0609:
Concerns https://github.com/Shopify/shopify-core-redis-proxy/pull/170

We are reaching the `annotation` size limit for certain `ConfigMap` resources. Note, this is _not_ the 1MiB limit of the `ConfigMap` data, itself, but simply the annotation. One possible strategy to get around this is to preclude the use of the `last-applied-annotation` (which is causing the bloat in size). To do this, we need to stop using `apply` when deploying this resource and instead use `replace`.

To that end, this PR introduces an annotation-driven feature, `krane.shopify.io/alpha/deploy-method-override`. This override accepts one of `apply, create, replace, replace-force` and is used to control the deploy method _at the individual resource level_. I chose this approach because:
- There is already a Krane convention of describing custom behaviour via the use of resource-level annotations (e.g. timeout override, required-rollout).
- The actual code changes are quite light and easily removed should we decide to remove this feature in the future and/or when a more long-term solution to the underlying problem becomes available.

### DIFF
--- a/lib/krane/kubernetes_resource.rb
+++ b/lib/krane/kubernetes_resource.rb
@@ -35,7 +35,7 @@ module Krane
       If you have reason to believe it will succeed, retry the deploy to continue to monitor the rollout.
       MSG
 
-    ALLOWED_DEPLOY_METHOD_OVERRIDES = %w(apply create replace replace-force)
+    ALLOWED_DEPLOY_METHOD_OVERRIDES = %w(create replace replace-force)
     DEPLOY_METHOD_OVERRIDE_ANNOTATION = "deploy-method-override"
     TIMEOUT_OVERRIDE_ANNOTATION = "timeout-override"
     LAST_APPLIED_ANNOTATION = "kubectl.kubernetes.io/last-applied-configuration"
@@ -138,6 +138,7 @@ module Krane
       @validation_errors = []
       validate_selector(selector) if selector
       validate_timeout_annotation
+      validate_deploy_method_override_annotation
       validate_spec_with_kubectl(kubectl)
       @validation_errors.present?
     end
@@ -246,10 +247,6 @@ module Krane
     def deploy_method_override
       override = krane_annotation_value(DEPLOY_METHOD_OVERRIDE_ANNOTATION)
       return unless override
-      unless ALLOWED_DEPLOY_METHOD_OVERRIDES.include?(override)
-        raise InvalidTemplateError, "unknown deploy override method: #{override}." \
-          "Accepted values are #{ALLOWED_DEPLOY_METHOD_OVERRIDES.join(', ')}"
-      end
       override.to_sym
     end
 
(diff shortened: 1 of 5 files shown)


tanner-bruce (approved):
LGTM

lib/krane/resource_deployer.rb
```diff
@@ -95,11 +95,10 @@ def deploy_resources(resources, prune: false, verify:, record_summary: true)
       applyables, individuals = resources.partition { |r| r.deploy_method == :apply }
       # Prunable resources should also applied so that they can  be pruned
       pruneable_types = @prune_whitelist.map { |t| t.split("/").last }
-      applyables += individuals.select { |r| pruneable_types.include?(r.type) }
+      applyables += individuals.select { |r| pruneable_types.include?(r.type) && !r.deploy_method_override }
```
timothysmith0609, line 98: We need to fight against default behaviour, which makes me loath to ever merge this into master. Previously, we would run `apply` here to enforce pruning, but since the goal right now is to _avoid_ pruning, we need to add this workaround. 

The good news is that this _only_ affects resources that are purposefully overridden, but I think it's enough of a landmine to give me some pause

lib/krane/kubernetes_resource.rb
```diff
@@ -35,6 +35,8 @@ class KubernetesResource
       If you have reason to believe it will succeed, retry the deploy to continue to monitor the rollout.
       MSG
 
+    ALLOWED_DEPLOY_METHOD_OVERRIDES = %w(create replace replace-force)
```
timothysmith0609, line 38: I needed to get rid of `apply` here because
- It's the default method, anyway
- https://github.com/Shopify/krane/blob/master/lib/krane/resource_deployer.rb#L98: If you try to override the deploy method of a prunable type, it will still get applied, regardless of whichever override method you choose.

Note that the latter point means overriding will prevent default pruning behaviour __for the resource(s) whose deploy methods are being overridden__. This isn't bad, in and of itself, but it's worth pointing out.

lib/krane/resource_deployer.rb
beckygowland: Would It be helpful to ever see a log warning when a pruneable resource is being overridden?   I think if it's purposely overridden, then they know they're fighting the default. but wondering if it would give less pause and help to debug if that knowledge assumption is wrong.

beckygowland (approved):
Makes sense to me!

lib/krane/kubernetes_resource.rb
RyanBrushett: > Note that the latter point means overriding will prevent default pruning behaviour for the resource(s) whose deploy methods are being overridden. This isn't bad, in and of itself, but it's worth pointing out.

Worth making a note about this in a readme somewhere or nah?

RyanBrushett (approved):

timothysmith0609:
>Worth making a note about this in a readme somewhere or nah?

Added a note to the changelog, which should be sufficient until we decide whether to keep/toss this feature longterm

RyanBrushett:
Thanks Tim. ⛵

dturn:
Code looks reasonable and on a positive note this will get resolved with server side apply (https://github.com/kubernetes/kubectl/issues/712). I definitely think we need to leave a warning somewhere (possibly in the readme / changelog) that this disables pruning. A different approach would be to automatically detect large objects instead of using an annotation, that would make it clear this is a workaround for large objects not an generic annotation to chagne the deploy method.

PR #759, opened 2020-10-21
Fix the way environment variable is passed to ejson

casperisfine:
We just noticed this with @DazWorrall. In our environment `EJSON_KEYDIR` is set, and the way this code works right now doesn't allow to override existing env:

```
>> out, err, st = Open3.capture3({ 'EJSON_KEYDIR' => 'foo' }, "EJSON_KEYDIR=bar echo $EJSON_KEYDIR")
=> ["foo\n", "", #<Process::Status: pid 11963 exit 0>]
```

It's much better to pass the env as a hash, and the command arguments as variadic arguments anyway, as it avoid having to spawn a shell to parse the command.

### DIFF
--- a/lib/krane/ejson_secret_provisioner.rb
+++ b/lib/krane/ejson_secret_provisioner.rb
@@ -133,7 +133,7 @@ module Krane
     end
 
     def decrypt_ejson(key_dir)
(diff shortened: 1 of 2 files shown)


timothysmith0609 (approved):
🚢

RyanBrushett (approved):
⛵

PR #764, opened 2020-11-17
Fast fail metric

rdji123:
Adding statsd data to get a better understanding of fast fail occurrences.

cc @Shopify/pipeline

### DIFF
--- a/lib/krane/kubernetes_resource/deployment.rb
+++ b/lib/krane/kubernetes_resource/deployment.rb
@@ -214,8 +214,13 @@ module Krane
     end
 
     def statsd_tags
-      %W(context:#{context} namespace:#{namespace} type:#{type} failing_to_progress: #{deploy_failing_to_progress?}
-         progress_deadline: #{progress_deadline.present? ? progress_deadline : nil})
+      {
+        context: context,
+        namespace: namespace,
+        failing_to_progress: deploy_failing_to_progress?,
+        progress_deadline: progress_deadline.present? ? progress_deadline : nil,
+        type: type,
+      }
     end
   end
 end
--- a/lib/krane/kubernetes_resource/deployment.rb
+++ b/lib/krane/kubernetes_resource/deployment.rb
@@ -92,7 +92,7 @@ module Krane
       return super if timeout_override
 
       if progress_condition.present?
-        StatsD.client.increment('fail_fast', 1, tags: statsd_tags)
+        StatsD.client.increment('kubectl.fail_fast', 1, tags: statsd_tags)
       end
 
       # Do not use the hard timeout if progress deadline is set
--- a/lib/krane/kubernetes_resource/deployment.rb
+++ b/lib/krane/kubernetes_resource/deployment.rb
@@ -92,7 +92,7 @@ module Krane
       return super if timeout_override
 
       if progress_condition.present?
-        StatsD.client.increment('fail.fast', 1, tags: statsd_tags)
+        StatsD.client.increment('fail_fast', 1, tags: statsd_tags)
       end
 
       # Do not use the hard timeout if progress deadline is set
--- a/lib/krane/kubernetes_resource/deployment.rb
+++ b/lib/krane/kubernetes_resource/deployment.rb
@@ -91,9 +91,11 @@ module Krane
       return false if deploy_failed?
       return super if timeout_override
 
-      if deploy_failing_to_progress?
-        @logger.info("Deploy failing to progress")
-    
+      if progress_condition.present?
+        StatsD.client.increment('kubectl.error', 1, tags: { context: context, namespace: namespace,
+                                                            progress_condition: deploy_failing_to_progress? })
+      end
+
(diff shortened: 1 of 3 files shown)


lib/krane/kubernetes_resource/deployment.rb
```diff
@@ -91,6 +91,11 @@ def deploy_timed_out?
       return false if deploy_failed?
       return super if timeout_override
 
+      if progress_condition.present?
+        StatsD.client.increment('kubectl.error', 1, tags: { context: context, namespace: namespace,
+                                                            progress_condition: deploy_failing_to_progress? })
+      end
```
timothysmith0609, line 97: I don't really like the idea of instrumenting the resource models, themselves, and would prefer if we could move that to the actual task runner, where such instrumentation makes more sense. It's worth noting we already capture timeout errors and publish them via StatsD (see https://shopify.datadoghq.com/dashboard/5kc-557-amd/krane--kubernetes-deploy-dash?from_ts=1605624831373&live=true&to_ts=1605628431373).

Alternatively, since we are constrained by the `KubernetesResource` interface, could we add a `statsd_tag` in `Deployment#deploy_timed_out?` when `deploy_failing_to_progress?` is true to indicate an actual progressing failure? 🤔 It would be hard to know if it's a fail-fast or an initial progressing failure, though.

lib/krane/kubernetes_resource/deployment.rb
```diff
@@ -91,6 +91,10 @@ def deploy_timed_out?
       return false if deploy_failed?
       return super if timeout_override
 
+      if progress_condition.present?
+        StatsD.client.increment('kubectl.error', 1, tags: statsd_tags)
+      end
```
timothysmith0609, line 96: I'm confused, why are we incrementing `kubectl.error`?

rdji123: I wanted to use an existing metric in Krane. Do you have any suggestions as to which metric we can use? We can also create a new metric for this

timothysmith0609: I don't think there's a metric we have that's quite suitable. Perhaps I'm ignorant, but is there some reason against using a bespoke metric?

rdji123: No reason against it! I've updated the metric to a new one 👍

lib/krane/kubernetes_resource/deployment.rb
```diff
@@ -91,6 +91,10 @@ def deploy_timed_out?
       return false if deploy_failed?
       return super if timeout_override
 
+      if progress_condition.present?
+        StatsD.client.increment('fail_fast', 1, tags: statsd_tags)
```
ayatsynych, line 95: might be a good idea to keep the `kubectl` prefix, so it's clear that this metric is specific to kubectl and it's easier to discover

lib/krane/version.rb
```diff
@@ -1,4 +1,4 @@
 # frozen_string_literal: true
 module Krane
-  VERSION = "2.1.2"
+  VERSION = "2.1.2.a"
```
timothysmith0609, line 3: In general, we bump the version in a dedicated PR (easier history, gives us a moment to ensure the CHANGELOG, etc is up to date), and I would recommend simply bumping the patch version to 3 instead of `.a`.

lib/krane/kubernetes_resource/deployment.rb
```diff
@@ -91,6 +91,10 @@ def deploy_timed_out?
       return false if deploy_failed?
       return super if timeout_override
 
+      if progress_condition.present?
+        StatsD.client.increment('kubectl.fail_fast', 1, tags: statsd_tags)
+      end
+
```
timothysmith0609, line 97: This will be evaluated on every sync loop, which happens X times per deployment. is that what you want?
Also, when we hit this code, we're not necessarily failing fast, no? We still need to observe that on line 99

timothysmith0609 (commented):
A few questions and comments. Will #765 obviate the need to track this? Or are you going to use it to measure changes?

rdji123:
> A few questions and comments. Will #765 obviate the need to track this? Or are you going to use it to measure changes?


These are temporary changes for testing - we're using this branch to help us collect fast fail data on [shipit-next](https://github.com/Shopify/shipit-next/pull/1825). The data will help us learn more/confirm if the errors are fail fasts before we ship https://github.com/Shopify/krane/pull/765, as well as monitor the change after. :) @timothysmith0609

lib/krane/kubernetes_resource/deployment.rb
timothysmith0609: 🤦 Sorry I realize you are capturing the specifics in the actual statsd_tag call

dturn:
Ping: Is this still in progress can we close it now that https://github.com/Shopify/krane/pull/765 has merged?

rdji123:
Yep, good to close

PR #765, opened 2020-11-18
Evaluate progress condition after progress deadline seconds have passed

rdji123:
This change checks for the progress condition of a k8s deployment type after the progress deadline seconds have passed.  
ie. when a deployment has been marked as failed to progress, we will wait for at least PDS seconds after a deploy/retry to mark it failed, instead of failing it right away

Note this assumes that `@deploy_started_at` is reset during retries as the DeployTask creates a new instance of [resource_deployer ](https://github.com/Shopify/krane/blob/master/lib/krane/deploy_task.rb#L192) each time.  
 
cc @Shopify/pipeline

### DIFF
--- a/lib/krane/kubernetes_resource/deployment.rb
+++ b/lib/krane/kubernetes_resource/deployment.rb
@@ -160,6 +160,7 @@ module Krane
 
     def deploy_failing_to_progress?
       return false unless progress_condition.present?
+      return false unless Time.now.utc - @deploy_started_at > progress_deadline
 
       # This assumes that when the controller bumps the observed generation, it also updates/clears all the status
       # conditions. Specifically, it assumes the progress condition is immediately set to True if a rollout is starting.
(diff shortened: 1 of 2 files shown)


lib/krane/kubernetes_resource/deployment.rb
```diff
@@ -155,6 +155,7 @@ def rollout_annotation_err_msg
 
     def deploy_failing_to_progress?
       return false unless progress_condition.present?
+      return false if deploy_started? && Time.now.utc - @deploy_started_at < progress_deadline
```
timothysmith0609, line 158: Worth a comment here, since the deploy is indeed failing to progress, we're just ignoring that for now :)

timothysmith0609 (approved):

lib/krane/kubernetes_resource/deployment.rb
```diff
@@ -155,6 +155,8 @@ def rollout_annotation_err_msg
 
     def deploy_failing_to_progress?
       return false unless progress_condition.present?
+      # ignore progress_condition status until after the progress_deadline is up
```
dturn, line 158:
```suggestion
      # Ensure at least progress_deadline wall clock time has passed before before examining progress_condition
```

dturn (approved):
Code looks good, can you update the readme and changelog before merging

PR #766, opened 2020-11-19
Use v1.15.12 of k8s when starting minikube

nishanthooda:
- `dev up` was failing, so bumping to with `Kubernetes 1.11.10 is not supported by this release of minikube`
- this pr bumps the version to `1.15.xx`

**What could go wrong?**
- bumping the version has some other unintended effects

cc @Shopify/pipeline

### DIFF
--- a/dev.yml
+++ b/dev.yml
@@ -13,7 +13,7 @@ up:
   - custom:
       name: Minikube Cluster
       met?: test $(minikube status | grep Running | wc -l) -ge 2 && $(minikube status | grep -q 'Configured')
-      meet: minikube start --kubernetes-version=v1.11.10 --vm-driver=hyperkit
+      meet: minikube start --kubernetes-version=v1.15.xx --vm-driver=hyperkit
       down: minikube stop
 commands:
   reset-minikube: minikube delete && rm -rf ~/.minikube
--- a/dev.yml
+++ b/dev.yml
@@ -13,7 +13,7 @@ up:
   - custom:
       name: Minikube Cluster
       met?: test $(minikube status | grep Running | wc -l) -ge 2 && $(minikube status | grep -q 'Configured')
-      meet: minikube start --kubernetes-version=v1.15.xx --vm-driver=hyperkit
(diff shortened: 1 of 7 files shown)


ayatsynych (approved):

dev.yml
```diff
@@ -13,7 +13,7 @@ up:
   - custom:
       name: Minikube Cluster
       met?: test $(minikube status | grep Running | wc -l) -ge 2 && $(minikube status | grep -q 'Configured')
-      meet: minikube start --kubernetes-version=v1.11.10 --vm-driver=hyperkit
+      meet: minikube start --kubernetes-version=v1.13.0 --vm-driver=hyperkit
```
dturn, line 16: We can bump this to 1.15.xx (don't use 0 inevitably its missing some bug fix), but we'll need to bump our minimum supported version. https://github.com/Shopify/krane/pull/546 was the last time we bumped our min supported version.

nishanthooda: updated to 1.15.xx, but as far as i can tell the PR you linked already bumps the min supported version to 1.15-latest 🤔

dev.yml
```diff
@@ -13,7 +13,7 @@ up:
   - custom:
       name: Minikube Cluster
       met?: test $(minikube status | grep Running | wc -l) -ge 2 && $(minikube status | grep -q 'Configured')
-      meet: minikube start --kubernetes-version=v1.11.10 --vm-driver=hyperkit
+      meet: minikube start --kubernetes-version=v1.15.xx --vm-driver=hyperkit
```
dturn, line 16: 1.15.12 is the most recent release https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG/CHANGELOG-1.15.md#v11512. 

I meant `xx` as a placeholder.

dturn (approved):
One minor change and then this will be ready to ship.

RyanBrushett (approved):

RyanBrushett (approved):

PR #767, opened 2020-11-30
Add servicenetworkendpointgroup to blacklist

timflapper:
Fix a pruning error when trying to prune (GKE) ServiceNetworkEndpointGroup resources on k8s clusters with version 1.18.

**How is this accomplished?**
By blacklisting the ServiceNetworkEndpointGroup resource from pruning.

**What could go wrong?**
I'm unaware of any issues related to the not pruning of this resource as it does not appear to be prunable.


dturn:
>I'm unaware of any issues related to the not pruning of this resource as it does not appear to be prunable.

What do you mean by its not prunable? The verb list includes delete?

> servicenetworkendpointgroups      svcneg       networking.gke.io                 true         ServiceNetworkEndpointGroup      [delete deletecollection get list patch create update watch]

timflapper:
> What do you mean by its not prunable? The verb list includes delete.

Yep, you're right. Thanks for that. I found out it needs to be added to `version_override` in the `version_for_kind` method. There are actually three versions in api-versions. v1, v1beta1 and v1beta2. This resource runs on v1beta1 and not v1. I'll close this one and create a new PR with the correct fix and tests.

PR #768, opened 2020-11-30
Add version override for ServiceNetworkEndpointGroup

timflapper:
Fix pruning of ServiceNetworkEndpointGroup resource.

**How is this accomplished?**
By forcing a version override to make sure the resource pruning does not fail with the error:

`error: no matches for kind "servicenetworkendpointgroup" in version "networking.gke.io/v1"`

**What could go wrong?**
I don't see anything specific that can go wrong with this change.


test/fixtures/for_unit_tests/api_versions.txt
```diff
@@ -22,6 +22,9 @@ events.k8s.io/v1beta1
 extensions/v1beta1
 networking.k8s.io/v1
 networking.k8s.io/v1beta1
+networking.gke.io/v1
```
dturn, line 25: I'm hesitant to add gke versions here. This isn't something everyone will be able replicate.

@timothysmith0609 thoughts?

dturn (commented):
Thanks for opening this PR. I have one question for Tim, but overall this will be accepted unless you're interested in tackling: https://github.com/Shopify/krane/issues/699

test/fixtures/for_unit_tests/api_versions.txt
timflapper: Since I didn't add a test for the pruning of ServiceNetworkEndpointGroup specifically I can just remove changes in the tests. It doesn't really add anything except make the expected count go from 25 to 26.

timothysmith0609: >I'm hesitant to add gke versions here

I agree, I think omission is ok, here

timothysmith0609 (commented):
I'm in agreement with @dturn: this is really the only way to do it until/unless we obviate the need for the override list in the first place, but that's longer term work

timflapper:
@timothysmith0609 @dturn I've removed the GKE services from the fixtures.

dturn (approved):
lgtm I'll merge once CI passes

PR #774, opened 2020-12-16
[Bug Fix] predeploy_sequence is wrong when there are conflicts between built-ins and CRDs

dturn:
E.g. If a CRD defines a Kind that is also in our hard coded predeploy_sequence, but with a different group, only one will be pre-deployed. That's a bug. 

Fix: https://github.com/Shopify/krane/issues/773

**How is this accomplished?**
The predeploy sequence now supports multiple groups per Kind.

**What could go wrong?**
This actually isn't the fix, it has a sublte bug where we'll always deploy something with that name ...

### DIFF
--- a/.rubocop-http---shopify-github-io-ruby-style-guide-rubocop-yml
+++ b/.rubocop-http---shopify-github-io-ruby-style-guide-rubocop-yml
@@ -195,6 +195,7 @@ Style/FrozenStringLiteralComment:
   SupportedStyles:
     - always
     - never
+  SafeAutoCorrect: true
 
 Style/GlobalVars:
   AllowedVariables: []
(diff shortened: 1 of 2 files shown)


lib/krane/kubernetes_resource/daemon_set.rb
```diff
@@ -3,7 +3,7 @@
 module Krane
   class DaemonSet < PodSetBase
     TIMEOUT = 5.minutes
-    SYNC_DEPENDENCIES = %w(Pod)
+    SYNC_DEPENDENCIES = %w(Pod.apps)
```
dturn, line 6: Having to add a group here sucks. I'm need to look at the behavior for kubectl and see what group it picks if there are multiple for a kind 🤞 its not the one a crd defines...

PR #777, opened 2020-12-17
Support k8s 1.19 and 1.20

dturn:
Add support for k8s versions 1.19 and 1.20

**How is this accomplished?**
Add them to CI

**What could go wrong?**
We're adding to new versions so CI times will go up.

### DIFF
--- a/lib/krane/cluster_resource_discovery.rb
+++ b/lib/krane/cluster_resource_discovery.rb
@@ -23,9 +23,19 @@ module Krane
       fetch_resources(namespaced: namespaced).uniq { |r| r['kind'] }.map do |resource|
         next unless resource['verbs'].one? { |v| v == "delete" }
         next if black_list.include?(resource['kind'])
-        group_versions = api_versions[resource['apigroup'].to_s]
-        version = version_for_kind(group_versions, resource['kind'])
-        [resource['apigroup'], version, resource['kind']].compact.join("/")
+        group = resource['apigroup'].to_s
+        version = ''
+        if group.empty?
+          version = resource['apiversion'].to_s
+        end
+        if version.empty?
+          group_versions = api_versions[group.to_s]
+          version = version_for_kind(group_versions, resource['kind'])
+          [group, version, resource['kind']].compact.join("/")
+        else
+          version = 'core/v1' if version == 'v1'
+          [version, resource['kind']].compact.join("/")
+        end
       end.compact
     end
 
--- a/lib/krane/cluster_resource_discovery.rb
+++ b/lib/krane/cluster_resource_discovery.rb
@@ -102,6 +102,8 @@ module Krane
     def get_resource_api_path(api_versions, resource)
       apiversion = resource['apiversion'].to_s
       apiversion = 'core/v1' if apiversion == 'v1'
+
+      ## In kubectl 1.20 APIGroups was replaced by APIVersions
       if apiversion.empty?
         apigroup = resource['apigroup'].to_s
         group_versions = api_versions[apigroup]
--- a/lib/krane/cluster_resource_discovery.rb
+++ b/lib/krane/cluster_resource_discovery.rb
@@ -23,12 +23,7 @@ module Krane
       fetch_resources(namespaced: namespaced).uniq { |r| r['kind'] }.map do |resource|
         next unless resource['verbs'].one? { |v| v == "delete" }
         next if black_list.include?(resource['kind'])
-        group = resource['apigroup'].to_s
-        group_versions = api_versions[group]
-        version = version_for_kind(group_versions, resource['kind'])
-
-        group = 'core' if group.empty? && version == 'v1'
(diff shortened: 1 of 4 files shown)


lib/krane/cluster_resource_discovery.rb
```diff
@@ -23,9 +23,19 @@ def prunable_resources(namespaced:)
       fetch_resources(namespaced: namespaced).uniq { |r| r['kind'] }.map do |resource|
         next unless resource['verbs'].one? { |v| v == "delete" }
         next if black_list.include?(resource['kind'])
-        group_versions = api_versions[resource['apigroup'].to_s]
-        version = version_for_kind(group_versions, resource['kind'])
-        [resource['apigroup'], version, resource['kind']].compact.join("/")
+        group = resource['apigroup'].to_s
+        version = ''
+        if group.empty?
```
dturn, line 28: I think this logic should be pushed into the version_for_kind function or at least not this function.

epk (approved):

karanthukral:
cc/ @dturn based on your comment, I moved the parsing logic out into a self contained function. CI for both 1.19 and 1.20 are passing now. Would you be able to take another look at this

lib/krane/cluster_resource_discovery.rb
```diff
@@ -101,6 +99,22 @@ def version_for_kind(versions, kind)
       version_override.fetch(kind, latest)
     end
 
+    def get_resource_api_path(api_versions, resource)
```
dturn, line 102: `def gvk_string(...` ?

I think of the api path as thing you'd pass to `kubectl get --raw ` (though maybe that's just me).

lib/krane/cluster_resource_discovery.rb
```diff
@@ -101,6 +99,22 @@ def version_for_kind(versions, kind)
...
+      apiversion = resource['apiversion'].to_s
+      apiversion = 'core/v1' if apiversion == 'v1'
```
dturn, line 104: I think this logic be: `apiversion = "core/#{apiversion} unless apiversion.include?("/")` (though maybe it will never matter since I don't think anything in the core group will ever get bumped)

dturn (commented):
This lgtm, but I can't approve since I'm an author. I'll tag Tim.

lib/krane/cluster_resource_discovery.rb
timothysmith0609: `def gvk` also works, though I'm ok with any of these options, really

lib/krane/cluster_resource_discovery.rb
karanthukral: That should work as long as we check for the empty too otherwise that unless would be try prior to 1.20

timothysmith0609: >I don't think anything in the core group will ever get bumped

Is that written down somewhere? I'd be tempted to future-proof here since the additional logic is sensible enough

lib/krane/cluster_resource_discovery.rb
```diff
@@ -101,6 +99,23 @@ def version_for_kind(versions, kind)
...
+    def gvk_string(api_versions, resource)
+      apiversion = resource['apiversion'].to_s
+
+      ## In kubectl 1.20 APIGroups was replaced by APIVersions
+      if apiversion.empty?
+        apigroup = resource['apigroup'].to_s
+        group_versions = api_versions[apigroup]
+
+        version = version_for_kind(group_versions, resource['kind'])
+        apigroup = 'core' if apigroup.empty?
```
karanthukral, line 111: cc/ @dturn I change this as well based on your comment https://github.com/Shopify/krane/pull/777#discussion_r552277061

lib/krane/cluster_resource_discovery.rb
```diff
@@ -101,6 +99,22 @@ def version_for_kind(versions, kind)
...
+      apiversion = resource['apiversion'].to_s
+      apiversion = 'core/v1' if apiversion == 'v1'
+
+      ## In kubectl 1.20 APIGroups was replaced by APIVersions
+      if apiversion.empty?
+        apigroup = resource['apigroup'].to_s
+        group_versions = api_versions[apigroup]
+
+        version = version_for_kind(group_versions, resource['kind'])
+        apigroup = 'core' if apigroup.empty? && version == 'v1'
```
timothysmith0609: Loosely related to [comment above](https://github.com/Shopify/krane/pull/777/files#r552277061), do we strictly require `&& version == "v1"` to determine `apigroup` is core? Is the absence of `apigroup` not enough?

timothysmith0609 (commented):
A few small questions, but nothing major

timothysmith0609 (approved):

lib/krane/cluster_resource_discovery.rb
karanthukral: Based on the above comment I removed the version check and basing it on the api group missing :)

PR #778, opened 2020-12-18
Update cluster discovery

dturn:
Remove the need for a hard coded GVK overide list: https://github.com/Shopify/krane/issues/699
Fix v.19 & 1.20 compatiblity

**How is this accomplished?**
Query the raw api to determine group/versin membership per kind

**What could go wrong?**
The code is wrong and everything blows up because apply to told to prune the wrong things or things that doesn't exist.

** Notes **

How to generate new fixtures: 
`kubectl get --raw / > test/fixtures/for_unit_tests/api_raw.txt`

```
require 'json'
raw_response = `kubectl get --raw /`
paths = JSON.parse(raw_response)["paths"].select { |p| %r{^\/api.*\/v.*$}.match(p) }
paths.each { |p| `kubectl get --raw #{p} >  test/fixtures/for_unit_tests/apis/#{p.gsub('/','_')}.txt` }
```

How kubeclient does this https://github.com/abonas/kubeclient/blob/master/lib/kubeclient/common.rb#L600


lib/krane/cluster_resource_discovery.rb
```diff
@@ -18,87 +19,61 @@ def crds
...
-          resource = fields.map { |k, (s, e)| [k.strip, resource[s..e].strip] }.to_h
-          # Manually parse verbs: "[get list]" into %w(get list)
-          resource["verbs"] = resource["verbs"][1..-2].split
-          resource
-        end
-      else
-        raise FatalKubeAPIError, "Error retrieving api-resources: #{err}"
+      paths = api_paths
+      results = []
+      Krane::Concurrency.split_across_threads(paths) do |path|
```
dturn, line 32: This isn't thread safe and only saved a minute or two on the tests compared to the 10 minutes this PR adds

lib/krane/cluster_resource_discovery.rb
```diff
@@ -18,87 +19,61 @@ def crds
...
-          # Manually parse verbs: "[get list]" into %w(get list)
-          resource["verbs"] = resource["verbs"][1..-2].split
-          resource
-        end
-      else
-        raise FatalKubeAPIError, "Error retrieving api-resources: #{err}"
+      responses = {}
+      Krane::Concurrency.split_across_threads(api_paths) do |path|
+        responses[path] = fetch_api_path(path)["resources"] || []
       end
```
timothysmith0609, line 33: The only concrete reading I found regarding this is https://ivoanjo.me/blog/2019/08/26/quick-tip-unsafe-concurrent-ruby-hash-access/
While the above bug doesn't to happen in MRI (due to GIL), it's still making me nervous about having multiple threads assign to a single hash.
I wonder if it isn't worth the effort of implementing `collect_across_threads` (rough example below).

```ruby
    def self.collect_across_threads(all_work, max_threads: MAX_THREADS, &block)
      return [] if all_work.empty?
      raise ArgumentError, "Block of work is required" unless block_given?

      slice_size = ((all_work.length + max_threads - 1) / max_threads)
      threads = []
      all_work.each_slice(slice_size).each_with_index do |slice, index|
        threads << Thread.new do
          slice.map { |s| yield s }
        end
      end
      collection = []
      threads.each do |t|
        collection << t.value
      end
      collection
    end
```

Semantically, it seems reasonable to differentiate between `split_across_threads` and `collect_across_threads` since they imply per-object side effects in the former and aggregation of data in the latter.

timothysmith0609 (commented):
I think it's worth considering a bit more the threadsafetiness of fetching api paths

lib/krane/cluster_resource_discovery.rb
dturn: I did some digging. It sounds like the code I've got is fine in MRI https://bugs.ruby-lang.org/issues/14735#note-1 If that makes you nervious, we could always use the thread-safe hash in `concurrent-ruby` which we pull in because we grab activesupport...

timothysmith0609: I'm leaning towards concurrent-ruby. As your linked article suggests, the fact that Hash is thread-safe is dependent on using MRI and using specific classes for the hash key. It's also suggested that these are implementation details and we shouldn't rest our assumptions on them remaining constant. 

However, if we observe too large a time penalty spent synchronizing we can always take a look again

timothysmith0609 (approved):

lib/krane/cluster_resource_discovery.rb
```diff
@@ -18,87 +20,61 @@ def crds
...
-      version_override = { "CronJob" => "v1beta1", "VolumeAttachment" => "v1beta1",
-                           "CSIDriver" => "v1beta1", "Ingress" => "v1beta1",
-                           "CSINode" => "v1beta1", "Job" => "v1",
-                           "IngressClass" => "v1beta1", "FrontendConfig" => "v1beta1",
-                           "ServiceNetworkEndpointGroup" => "v1beta1",
-                           "EnvoyFilter" => "v1alpha3",
-                           "TCPIngress" => "v1beta1" }
+    def fetch_api_path(path)
+      @api_path_cache[path] ||= begin
+        raw_json, err, st = kubectl.run("get", "--raw", path, attempts: 1, use_namespace: false)
```
dturn, line 56: Flipping this from `attempts: 2` to `attempts: 1` didn't speed up the tests.

karanthukral (approved):
The approach makes a lot of sense! 👏

dturn:
I've squashed and rebased this PR on master.

For 🎩 :

I manually deployed the hello-cloud fixtures on v1.15.10 minikube and then removed resources to test pruning.

I did the same thing with `minikube start --network-plugin=cni --cni=calico` which gives you a ` v1.20.0 ` cluster.

sjagoe:
This (potentially) requires extra permissions on the cluster in order to be able to deploy with krane. Could this requirement be documented?

```
kind: ClusterRole
apiVersion: rbac.authorization.k8s.io/v1
metadata:
  name: krane-deploy
rules:
- nonResourceURLs: ["/"]
  verbs: ["get"]
```

airhorns:
WOOOOO!!!!!!

PR #779, opened 2021-01-08
Only call Oj.mimic_JSON when in a CLI run

casperisfine:
### Context

We have a bunch of tests in our CI suite that require krane to test a few deploy stuffs. The problem is that requiring `krane` calls `Oj.mimic_JSON` which cause various issues to us.

### Change

I think krane should only patch JSON when it is ran as CLI, as it has better control of what's running. When used as a library, that's not a good idea to patch core classes like this.

NB: my knowledge of crane and Thor is very limited, so this PR might be totally incorrect.

cc @paracycle @Shopify/test-infra

### DIFF
--- a/lib/krane/cli/krane.rb
+++ b/lib/krane/cli/krane.rb
@@ -1,6 +1,6 @@
 # frozen_string_literal: true
-
 require 'krane'
+require 'krane/oj'
 require 'thor'
 require 'krane/cli/version_command'
 require 'krane/cli/restart_command'
--- a/lib/krane/common.rb
+++ b/lib/krane/common.rb
@@ -11,7 +11,6 @@ require 'active_support/core_ext/array/conversions'
 require 'colorized_string'
 
 require 'krane/version'
-require 'krane/oj'
 require 'krane/errors'
 require 'krane/formatted_logger'
 require 'krane/statsd'


lib/krane.rb
```diff
@@ -1,5 +1,6 @@
 # frozen_string_literal: true
 
+require 'oj'
```
paracycle, line 3: Do we really need this? Do the tasks that are exported below really depend on `Oj` in any way?

lib/krane/cli/krane.rb
```diff
@@ -17,6 +17,11 @@ class Krane < Thor
 
       package_name "Krane"
 
+      def self.start(*)
+        Oj.mimic_JSON
+        super
+      end
+
```
paracycle, line 24: Since this file is only required by the CLI binary, there should be no problems with leaking to the requiring context. So I think we can just do:
```
require 'oj'
Oj.mimic_JSON
```
at the top of this file. I don't think we need to use the `start` call for this.

casperisfine:
@paracycle all good points. I've updated the PR.

paracycle (approved):

douglas (approved):
This looks good for me from what I remember working in Krane long time ago, but I defer to @dturn since I may be missing some context.

dturn (commented):
`Oj.mimic_JSON` brought a noticeable perf improvement (on the order of seconds) since this gem does a ton json processing. I'm ok with only including it in the CLI but please restore the lib/krane/oj.rb file.

casperisfine:
> but please restore the lib/krane/oj.rb file

I don't understand. I simply moved it's content inside `krane/cli.rb`. I'm fine restoring it and requiring it from `krane/cli.rb` but I don't really see the point.

dturn:
> but I don't really see the point.

Its a minor thing but, I think its better code organization, it also makes it a one liner to restore the behavior for people using Krane as a library.

casperisfine:
> but please restore the lib/krane/oj.rb file

Done.

PR #781, opened 2021-01-13
Batch dry-run apply validation

timothysmith0609:

Perf testing 1.18 highlighted the fact that running `kubectl` _per-resource_ (as we do when running dry-run for resource validation in Phase 1) is markedly slower than simply batching all resources and dry-run applying _en masse_. To that end, this PR tweaks the resource validation of `DeployTask` by running kubectl validation on all resources at once.

**How is this accomplished?**

Add a `dry_run` flag to `ResourceDeployer#apply_all` to hook into available logic and leverage it in `DeployTask` (this PR is only limited to that task). In the event the batch apply fails, we fall back to previous behaviour of checking resource-by-resource. I tried using only the batch apply, and while it does do a good job of suppressing sensitive information (except in a small number of test cases), it is too much of a yak-shave to fix all the tests for the differences in log-matching used by the tests. If, over time, we want to move forward to exclusive batch applying, we can do a dedicated PR. 

**What could go wrong**
- The only real danger would be batch apply is successful but there's an error that the resource-by-resource apply would catch; this doesn't seem possible, though. In the case of failure, we simply fallback to current behaviour so there's no risk there, either.

- Would appreciate some feedback on the tests I had to write, since they are serial (by virtue of containing expectations). This also led me to think that we should separate out unit/serial/parallel tests into separate (parallel) buildkite steps.

### DIFF
--- a/lib/krane/deploy_task.rb
+++ b/lib/krane/deploy_task.rb
@@ -144,7 +144,7 @@ module Krane
       @logger.phase_heading("Initializing deploy")
       validate_configuration(prune: prune)
       resources = discover_resources
-      validate_resources(resources, prune: prune)
+      validate_resources(resources)
 
       @logger.phase_heading("Checking initial resource statuses")
       check_initial_status(resources)
@@ -277,9 +277,9 @@ module Krane
     end
     measure_method(:validate_configuration)
 
-    def validate_resources(resources, prune: false)
+    def validate_resources(resources)
       validate_globals(resources)
-      dry_run_success = validate_dry_run(resources, prune)
+      dry_run_success = validate_dry_run(resources)
       Krane::Concurrency.split_across_threads(resources) do |r|
         if dry_run_success
           r.validate_definition(kubectl: nil, selector: @selector)
@@ -311,8 +311,8 @@ module Krane
         "Use GlobalDeployTask instead."
     end
 
-    def validate_dry_run(resources, prune)
-      resource_deployer.dry_run(resources, prune)
+    def validate_dry_run(resources)
+      resource_deployer.dry_run(resources)
     end
 
     def namespace_definition
--- a/lib/krane/deploy_task.rb
+++ b/lib/krane/deploy_task.rb
@@ -144,7 +144,7 @@ module Krane
       @logger.phase_heading("Initializing deploy")
       validate_configuration(prune: prune)
       resources = discover_resources
-      validate_resources(resources, prune)
+      validate_resources(resources, prune: prune)
 
       @logger.phase_heading("Checking initial resource statuses")
       check_initial_status(resources)
@@ -277,7 +277,7 @@ module Krane
     end
     measure_method(:validate_configuration)
 
-    def validate_resources(resources, prune = true)
+    def validate_resources(resources, prune: false)
       validate_globals(resources)
       dry_run_success = validate_dry_run(resources, prune)
       Krane::Concurrency.split_across_threads(resources) do |r|
--- a/lib/krane/deploy_task.rb
+++ b/lib/krane/deploy_task.rb
@@ -282,9 +282,9 @@ module Krane
       dry_run_success = validate_dry_run(resources, prune)
       Krane::Concurrency.split_across_threads(resources) do |r|
         if dry_run_success
-          r.validate_definition(nil, selector: @selector)
+          r.validate_definition(kubectl: nil, selector: @selector)
         else
-          r.validate_definition(kubectl, selector: @selector)
+          r.validate_definition(kubectl: kubectl, selector: @selector)
         end
       end
       failed_resources = resources.select(&:validation_failed?)
--- a/lib/krane/deploy_task.rb
+++ b/lib/krane/deploy_task.rb
@@ -279,11 +279,11 @@ module Krane
 
     def validate_resources(resources)
       validate_globals(resources)
-      dry_run_success = validate_dry_run(resources)
+      batch_dry_run_success = validate_dry_run(resources)
       Krane::Concurrency.split_across_threads(resources) do |r|
         # No need to pass in kubectl (and do per-resource dry run apply) if batch dry run succeeded
-        if dry_run_success
-          r.validate_definition(kubectl: nil, selector: @selector)
+        if batch_dry_run_success
+          r.validate_definition(kubectl: nil, selector: @selector, dry_run: false)
         else
           r.validate_definition(kubectl: kubectl, selector: @selector)
         end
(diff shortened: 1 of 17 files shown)


lib/krane/kubectl.rb
```diff
@@ -100,6 +100,14 @@ def server_dry_run_enabled?
       server_version >= Gem::Version.new(SERVER_DRY_RUN_MIN_VERSION)
     end
 
+    def dry_run_flag
+      if server_version >= Gem::Version.new("1.18")
+        "--dry-run=server"
+      else
+        "--server-dry-run"
+      end
```
timothysmith0609, line 108: We actually need to handle this with the given kubectl client version as well, which might bite us

lib/krane/deploy_task.rb
```diff
@@ -144,7 +144,7 @@ def run!(verify_result: true, prune: true)
       @logger.phase_heading("Initializing deploy")
       validate_configuration(prune: prune)
       resources = discover_resources
-      validate_resources(resources)
+      validate_resources(resources, prune)
```
timothysmith0609, line 147: `prune` probably not really necessary for just dry-running...

karanthukral: we could always remove this here and have `validate_dry_run` pass false for the prune flag automatically as an idea

karanthukral (approved):
approach looks reasonable to me :)

lib/krane/deploy_task.rb
timothysmith0609: After thinking about it, I don't think we need the prune flag at all, since no human will see the output. wdyt @dturn ?

lib/krane/deploy_task.rb
```diff
@@ -277,15 +277,18 @@ def validate_configuration(prune:)
     end
     measure_method(:validate_configuration)
 
-    def validate_resources(resources)
+    def validate_resources(resources, prune: false)
```
dturn, line 280: Does the dry run behavior actually change with prune?

lib/krane/resource_deployer.rb
```diff
@@ -149,11 +151,11 @@ def apply_all(resources, prune)
           @prune_whitelist.each { |type| command.push("--prune-whitelist=#{type}") }
         end
 
+        command.push(kubectl.dry_run_flag) if dry_run
         output_is_sensitive = resources.any?(&:sensitive_template_content?)
         global_mode = resources.all?(&:global?)
         out, err, st = kubectl.run(*command, log_failure: false, output_is_sensitive: output_is_sensitive,
           attempts: 2, use_namespace: !global_mode)
-
```
dturn, line 156: should we be skipping `record_apply_failure` during dry run?

dturn (commented):
Overall this is a lot less code than I expected, a few questions / style thoughts added

lib/krane/resource_deployer.rb
```diff
@@ -128,17 +132,15 @@ def deploy_resources(resources, prune: false, verify:, record_summary: true)
       end
     end
 
-    def apply_all(resources, prune)
+    def apply_all(resources, prune, dry_run: false)
```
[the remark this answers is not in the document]
timothysmith0609: They are simply included in the `apply_all` call. From what I observed, it works as expected, and the actual `dry-run=false` invocation will use the appropriate deploy method. This seemed like the best way to capture all resources in batch validation. One thing this won't catch is resources that exclusively use `:create` (without, e.g. a fallback to `:replace`), since that will pass validation but fail on deploy if the resource already exists in cluster. IMO that's a bug in their config, not Krane, though.

lib/krane/deploy_task.rb
timothysmith0609: It seems it really only changes the logging (for our concerns, at least). E.g. if `prune: true` and a resource in-cluster is not part of the provided manifests, the log output will show:

```
configmap/hello-cloud-configmap-data pruned (server dry run
```

Otherwise, it will simply be ignored. Since we swallow this part of the logs, I'm thinking it's best to drop the `prune` flag.

lib/krane/resource_deployer.rb
```diff
@@ -128,17 +135,15 @@ def deploy_resources(resources, prune: false, verify:, record_summary: true)
...
```
timothysmith0609, line 138: Hmm, this is going to ruin the `apply_all` measurements since StatsD will now record for both dry run and actual deployments :/ I think I'm going to have to get rid of the `measure_method` helper and deal with that manually

lib/krane/deploy_task.rb
```diff
@@ -279,13 +279,17 @@ def validate_configuration(prune:)
 
     def validate_resources(resources)
       validate_globals(resources)
+      dry_run_success = validate_dry_run(resources)
       Krane::Concurrency.split_across_threads(resources) do |r|
-        r.validate_definition(kubectl, selector: @selector)
+        # No need to pass in kubectl (and do per-resource dry run apply) if batch dry run succeeded
+        if dry_run_success
+          r.validate_definition(kubectl: nil, selector: @selector)
```
dturn, line 286: I'd rather add a param than have passing kubectl: nil imply that dry run isn't needed

lib/krane/resource_deployer.rb
```diff
@@ -149,20 +156,22 @@ def apply_all(resources, prune)
           @prune_whitelist.each { |type| command.push("--prune-whitelist=#{type}") }
         end
 
+        command.push(kubectl.dry_run_flag) if dry_run
         output_is_sensitive = resources.any?(&:sensitive_template_content?)
         global_mode = resources.all?(&:global?)
         out, err, st = kubectl.run(*command, log_failure: false, output_is_sensitive: output_is_sensitive,
           attempts: 2, use_namespace: !global_mode)
 
+        tags = statsd_tags + (dry_run ? ['dry_run:true'] : ['dry_run:false'])
```
dturn, line 165: We're going to need to remember to update dashboards

lib/krane/resource_deployer.rb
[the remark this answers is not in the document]
dturn: How hard is it to special case this e.g. do the batch and then dry run the create only resources too? This seems like a nice to catch (though we'd only catch that today for secrets).

dturn (approved):
1 style thing and one question, everything else lgtm

lib/krane/resource_deployer.rb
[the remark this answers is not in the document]
timothysmith0609: Opened #787 to keep track of this refinement.

PR #782, opened 2021-01-13
Drop support for Ruby 2.5 as its close to EOL

karanthukral:
- Dropping support for ruby 2.5 in favour of switching the default to 2.6.6

**How is this accomplished?**
- Update gemspec, CI and local dev

**What could go wrong?**
- Unexpected issues with ruby upgrade but the version is being tested on CI

### DIFF
--- a/.shopify-build/krane.yml
+++ b/.shopify-build/krane.yml
@@ -1,6 +1,6 @@
 containers:
   default:
-    docker: circleci/ruby:2.5.7
+    docker: circleci/ruby:2.6.6
 
 steps:
   - label: Lint
--- a/.shopify-build/krane.yml
+++ b/.shopify-build/krane.yml
@@ -8,21 +8,22 @@ steps:
     run:
       - bundle: ~
       - bundle exec rubocop
-  - label: 'Run Test Suite (:kubernetes: 1.20-latest :ruby: 2.7)'
+  - label: 'Run Test Suite (:kubernetes: 1.20-latest :ruby: 3.0)'
     command: bin/ci
     agents:
       queue: k8s-ci
     env:
       LOGGING_LEVEL: "4"
       KUBERNETES_VERSION: v1.20-latest
-      RUBY_VERSION: "2.7"
-  - label: 'Run Test Suite (:kubernetes: 1.19-latest)'
+      RUBY_VERSION: "3.0"
+  - label: 'Run Test Suite (:kubernetes: 1.19-latest :ruby: 2.7)'
     command: bin/ci
     agents:
       queue: k8s-ci
     env:
       LOGGING_LEVEL: "4"
       KUBERNETES_VERSION: v1.19-latest
+      RUBY_VERSION: "2.7"
   - label: 'Run Test Suite (:kubernetes: 1.18-latest)'
     command: bin/ci
     agents:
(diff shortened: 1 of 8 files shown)


timothysmith0609 (approved):

.shopify-build/krane.yml
```diff
@@ -1,28 +1,29 @@
...
-  - label: 'Run Test Suite (:kubernetes: 1.19-latest)'
+      RUBY_VERSION: "3.0"
+  - label: 'Run Test Suite (:kubernetes: 1.19-latest :ruby: 2.7)'
     command: bin/ci
     agents:
       queue: k8s-ci
     env:
       LOGGING_LEVEL: "4"
       KUBERNETES_VERSION: v1.19-latest
+      RUBY_VERSION: "2.7"
```
dturn, line 26: 👍  (Assuming the idea is to test each version of ruby w/o building an entire matrix)

CHANGELOG.md
```diff
@@ -1,5 +1,8 @@
 ## next
 
+*Other*
+- Dropped support for Ruby 2.5 since it will be EoL shortly. [#782](https://github.com/Shopify/krane/pull/782).
```
dturn, line 4:
```suggestion
- Dropped support for Ruby 2.5 due to EoL. [#782](https://github.com/Shopify/krane/pull/782).
```

dev.yml
```diff
@@ -1,7 +1,7 @@
 ---
 name: krane
 up:
-  - ruby: 2.5.7 # Matches gemspec
+  - ruby: 3.0.0 # Matches gemspec
```
dturn, line 4: This should be 2.6.6

dturn (approved):
Just some minor stuff

.shopify-build/krane.yml
karanthukral: yup! 2.6 default and then 2.7 and 3 on single k8s versions

dev.yml
karanthukral: yes.. was using it for testing. Will set it to 2.6.6

test/unit/krane/kubernetes_resource_test.rb
```diff
@@ -177,11 +177,16 @@ def test_validate_definition_doesnt_log_raw_output_for_sensitive_resources
     resource = DummySensitiveResource.new
     kubectl.expects(:server_version).returns(Gem::Version.new('1.20'))
 
-    kubectl.expects(:run).with { |*_args, **kwargs| kwargs[:output_is_sensitive] == true }.returns([
-      "Some Raw Output",
-      "Error from kubectl: something went wrong and by the way here's your secret: S3CR3T",
-      stub(success?: false),
-    ])
+    kubectl.expects(:run)
```
timothysmith0609, line 180: That's strange that this changed just from the Ruby version :/ I would have that it specific to Mocha. Regardless, I think it's ok for now, though I'm not a huge fan of introducing so much static info into the stub :(

timothysmith0609 (approved):

test/unit/krane/kubernetes_resource_test.rb
dturn: Ruby 2.7 deprecated some stuff with keyword args I haven't fully read up on it https://www.ruby-lang.org/en/news/2019/12/12/separation-of-positional-and-keyword-arguments-in-ruby-3-0/ but maybe this smells like it could be related.

PR #793, opened 2021-01-27
Use client, not server version for server dry run flag

timothysmith0609:
Perf testing a new 1.18 core cluster exposed a sleeping error inside the logic we use to determine the appropriate `kubectl` flag for specifying server-side dry-running. Prior to this PR, the __server__ version is checked against 1.18+, when it should be the __client__ version. Unfortunately this issue didn't appear in CI (since client/server are locked together there) and was missed on the 2.1.4 release. 

**How is this accomplished?**
I verified the client/server error by attempting a deploy using various configurations of kubectl client and server versions
Screen Shot 2021-01-26 at 7 40 48 PM

As expected, the client version determines (somewhat obviously) the format of the flag we should pass in. 

**What could go wrong?**
It already went wrong, unfortunately.

### DIFF
--- a/test/unit/krane/ejson_secret_provisioner_test.rb
+++ b/test/unit/krane/ejson_secret_provisioner_test.rb
@@ -193,7 +193,9 @@ class EjsonSecretProvisionerTest < Krane::TestCase
   end
 
   def dummy_version
-    'Server: version.Info{Major:"1", Minor:"13", GitVersion:"v1.13.6", GitCommit:"a6a8ec"}'
+    'Server: version.Info{Major:"1", Minor:"13", GitVersion:"v1.13.6", GitCommit:"a6a8ec"}' \
+      "\n" \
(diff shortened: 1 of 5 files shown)


karanthukral (approved):
Thanks for the fix.

test/unit/krane/ejson_secret_provisioner_test.rb
```diff
@@ -193,7 +193,9 @@ def dummy_secret_hash(name: SecureRandom.hex(4), data: {})
   end
 
   def dummy_version
-    'Server: version.Info{Major:"1", Minor:"13", GitVersion:"v1.13.6", GitCommit:"a6a8ec"}'
+    'Server: version.Info{Major:"1", Minor:"13", GitVersion:"v1.13.6", GitCommit:"a6a8ec"}' \
+      "\n" \
+      'Client:: version.Info{Major:"1", Minor:"13", GitVersion:"v1.17.0", GitCommit:"70132b"'
```
dturn, line 198: Should the Minor version be 17 here?

dturn (approved):

PR #794, opened 2021-01-27
V2.1.5

timothysmith0609:
Contains fix in #793 (using wrong dry-run flag when kubectl client < 1.18 and server version >= 1.18)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 ## next
 
+## 2.1.5
+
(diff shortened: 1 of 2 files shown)


dturn (approved):
Do you want to leave a note in 2.1.4 that there is a bug

karanthukral (approved):

timothysmith0609:
>Do you want to leave a note in 2.1.4 that there is a bug

Not sure it's worth it. The same bug is also present in 2.1.3 (but for 1.19+ clusters)

PR #798, opened 2021-02-02
Account for mutating webhooks with side effects when batch dry-run server-side apply

timothysmith0609:
This PR adds some smarts to the batch dry-runner. In particular, it cross-references the supplied list of resources with those that have mutating admission webhooks _which are not dry-run safe_. Resources that we know will fail batch dry-running are thus pre-emptively added to the legacy dry-run flow.


lib/krane/deploy_task.rb
```diff
@@ -277,16 +277,42 @@ def validate_configuration(prune:)
...
+            versions = rule.dig('apiVersions')
+            kinds = rule.dig('resources').map(&:singularize)
+            groups.each do |group|
+              versions.each do |version|
+                kinds.each do |kind|
+                  individuals += resources.select do |r|
+                    (r.group == group || group == '*' || match_policy == "Equivalent") &&
+                    (r.version == version || version == '*' || match_policy == "Equivalent") &&
+                    (r.type.downcase == kind.downcase)
+                  end
```
timothysmith0609, line 298: This is a straightforward implementation of the rules defined [here](https://kubernetes.io/docs/reference/kubernetes-api/extend-resources/mutating-webhook-configuration-v1/#MutatingWebhookConfiguration).

test/integration-serial/serial_deploy_test.rb
```diff
@@ -525,12 +525,35 @@ def test_batch_dry_run_apply_failure_falls_back_to_individual_resource_dry_run_v
...
+    Krane::ClusterResourceDiscovery.any_instance.expects(:fetch_mutating_webhook_configurations).returns(resp)
+    Krane::KubernetesResource.any_instance.expects(:validate_definition).times(4) # resources we expect to have to individually validate
+    deploy_fixtures('hello-cloud', subset: %w(web.yml.erb secret.yml), render_erb: true) do |fixtures|
+      container = fixtures['web.yml.erb']['Deployment'][0]['spec']['template']['spec']['containers'][0]
+      container['volumes'] = [
+        name: 'secret',
+        secret: {
+          secretName: fixtures['secret.yml']["Secret"][0]['metadata']['name'],
+        },
+      ]
```
timothysmith0609: This _does not_ fail the dry-run, as I would have thought, even though the Secret resource is not part of the dry-run apply (and is not present on the cluster beforehand). I will try a few more failure scenarios to see if I can actually induce this type of transitive failure

lib/krane/deploy_task.rb
```diff
@@ -277,16 +277,42 @@ def validate_configuration(prune:)
...
-      Krane::Concurrency.split_across_threads(resources) do |r|
-        # No need to pass in kubectl (and do per-resource dry run apply) if batch dry run succeeded
-        if batch_dry_run_success
-          r.validate_definition(kubectl: nil, selector: @selector, dry_run: false)
-        else
-          r.validate_definition(kubectl: kubectl, selector: @selector, dry_run: true)
-        end
+      batchable_resources, individuals = partition_dry_run_resources(resources)
+      batch_dry_run_success = kubectl.server_dry_run_enabled? && validate_dry_run(batchable_resources)
+      individuals += batchable_resources unless batch_dry_run_success
```
timothysmith0609, line 313: As discussed, we still fall back to per-resource dry-running if the batch fails

lib/krane/kubernetes_resource.rb
```diff
@@ -226,6 +226,11 @@ def group
       version ? grouping : "core"
     end
 
+    def version
+      prefix, version = @definition.dig("apiVersion").split("/")
+      version ? version : prefix
```
dturn, line 231:
```suggestion
      version || prefix
```

lib/krane/deploy_task.rb
```diff
@@ -277,16 +277,42 @@ def validate_configuration(prune:)
     end
     measure_method(:validate_configuration)
 
+    def partition_dry_run_resources(resources)
+      individuals = []
+      mutating_webhooks = cluster_resource_discoverer.fetch_mutating_webhook_configurations
+      mutating_webhooks.each do |spec|
+        spec.dig('webhooks').each do |webhook|
```
dturn, line 284: This makes me think we want a `MutatingWebhookConfigurations` class instead of putting all of the logic in this function.

lib/krane/deploy_task.rb
dturn: Any idea how to ask the api server what GVKs it knows to convert between? Not saying we use them, but this adds to our list of things that break with non-unique kinds.

lib/krane/deploy_task.rb
```diff
@@ -277,16 +277,42 @@ def validate_configuration(prune:)
...
+            kinds = rule.dig('resources').map(&:singularize)
+            groups.each do |group|
+              versions.each do |version|
+                kinds.each do |kind|
+                  individuals += resources.select do |r|
+                    (r.group == group || group == '*' || match_policy == "Equivalent") &&
+                    (r.version == version || version == '*' || match_policy == "Equivalent") &&
+                    (r.type.downcase == kind.downcase)
+                  end
+                  resources -= individuals
```
dturn, line 299: I can never remember the pass by semantics of ruby but, Is resources a copy or does this impact `validate_resources`

dturn (commented):
Some small thoughts but looking like its on the right direction.

lib/krane/deploy_task.rb
timothysmith0609: >Any idea how to ask the api server what GVKs it knows to convert between?

From my understanding, the `Equivalent` condition simply disregards gvk (I'm tempted to open an upstream issue since this can have undesirable side-effects as we've found). I don't think this is something for Krane to deal with.

From the official reference:
>Equivalent: match a request if modifies a resource listed in rules, even via another API group or version. For example, if deployments can be modified via apps/v1, apps/v1beta1, and extensions/v1beta1, and "rules" only included apiGroups:["apps"], apiVersions:["v1"], resources: ["deployments"], a request to apps/v1beta1 or extensions/v1beta1 would be converted to apps/v1 and sent to the webhook.

lib/krane/deploy_task.rb
timothysmith0609: It's pass by reference, however the `-=` operator returns a new object, not a mutation of the old one. If we used, e.g. `<<`, then we would mutate the incoming object. As a matter of caution, I'll `dup` the `resources` array to avoid any issues in the future.

lib/krane/deploy_task.rb
timothysmith0609: I made a few attempts at this. Modelling it as a `KubernetesResource` didn't seem right, but neither did modelling it in the top-level namespace. If I had to make a choice, I'm tempted to do the latter.

dturn: Maybe its time for an `Internal` namespace for objects that are K8s objects but aren't designed for having deploy logic?

timothysmith0609: I'm loath to make the separation... ultimately this is still just a kubernetes object. For now, I've made it a subclass of `KubernetesResource` and added the appropriate overrides should one of these ever be deployed.

test/integration-serial/serial_deploy_test.rb
```diff
@@ -525,15 +525,66 @@ def test_batch_dry_run_apply_failure_falls_back_to_individual_resource_dry_run_v
...
+    assert_logs_match_all([
+      "Result: SUCCESS",
+      "Successfully deployed 4 resources",
+    ], in_order: true)
+  end
+
+  def test_resources_with_side_effect_inducing_webhooks_with_transitive_dependency_does_not_fail_batch_running
+    resp = mutating_webhook_fixture(File.join(fixture_path("mutating_webhook_configurations"), "secret_hook.json"))
+    Krane::ClusterResourceDiscovery.any_instance.expects(:fetch_mutating_webhook_configurations).returns(resp)
+    expected_dry_runs = 0
```
dturn, line 555: expected or actual?

dturn (commented):
Some questions about the tests, code looks good

test/integration-serial/serial_deploy_test.rb
```diff
@@ -525,15 +525,66 @@ def test_batch_dry_run_apply_failure_falls_back_to_individual_resource_dry_run_v
...
+    assert_deploy_success(result)
+    assert_logs_match_all([
+      "Result: SUCCESS",
+      "Successfully deployed 1 resource",
+    ], in_order: true)
+  end
+
+  def test_resources_with_side_effect_inducing_webhooks_are_not_batched_server_side_dry_run
+    resp = mutating_webhook_fixture(File.join(fixture_path("mutating_webhook_configurations"), "ingress_hook.json"))
+    Krane::ClusterResourceDiscovery.any_instance.expects(:fetch_mutating_webhook_configurations).returns(resp)
```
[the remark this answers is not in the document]
timothysmith0609: I didn't realize I could just set `failurePolicy: ignore` with a small timeout to circumvent the need for an actual webhook server. I've changed the tests to actually deploy a real `MutatingWebhookConfiguration` and use that to test things out. 

Worth noting that the partitioning logic is only really necessary for `v1beta1` resources, since `v1` only allows (and requires) either `None` or `NoneOnDryRun` to be specified for the `sideEffects` field.

test/integration-serial/serial_deploy_test.rb
```diff
@@ -525,15 +525,66 @@ def test_batch_dry_run_apply_failure_falls_back_to_individual_resource_dry_run_v
...
+    Krane::KubernetesResource.any_instance.expects(:validate_definition).with { |params| params[:dry_run] == false }
+    result = deploy_fixtures("hello-cloud", subset: %w(secret.yml))
+    assert_deploy_success(result)
+    assert_logs_match_all([
+      "Result: SUCCESS",
+      "Successfully deployed 1 resource",
+    ], in_order: true)
+  end
+
+  def test_resources_with_side_effect_inducing_webhooks_are_not_batched_server_side_dry_run
```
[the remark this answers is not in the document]
timothysmith0609: I've fleshed out this test a bit more

dturn (approved):
👍

karanthukral (approved):
sorry for the delayed review

PR #802, opened 2021-02-10
v2.1.6

timothysmith0609:


lib/krane/cluster_resource_discovery.rb
```diff
@@ -90,23 +90,6 @@ def resource_hash(path, namespaced, blob)
       }
     end
 
-    def gvk_string(api_versions, resource)
```
dturn, line 93: Thanks for noticing this.

dturn (commented):
I'm holding off on ✅  till we decide if we need to bump retires on https://github.com/Shopify/krane/pull/802/files#diff-3e12a0be9cc2e0923634da8886dcd780cbfb28fd6d6a1604285b1153dc422638R69

dturn (approved):

PR #803, opened 2021-02-11
Configure default line limit for container logs

jenshenny:
From https://github.com/Shopify/krane/pull/800, it was suggested to make `DEFAULT_LINE_LIMIT` configurable by introducing an environment variable.

**How is this accomplished?**
Turned the constant into a class method, if `ENV["LOG_LINE_LIMIT"]` (name open to suggestions) is not specified or cannot be converted to an integer, the number of lines will default to 25.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,11 @@
 ## next
 
 *Enhancements*
+- ENV["KRANE_LOG_LINE_LIMIT"] allows the number of container logs printed for failures to be configurable from the 25 line default [#803](https://github.com/Shopify/krane/pull/803).
+
+## 2.1.6
+
+*Enhancements*
 - Remove the need for a hard coded GVK overide list via improvements to cluster discovery [#778](https://github.com/Shopify/krane/pull/778)
 
 *Bug Fixes*
--- a/lib/krane/container_logs.rb
+++ b/lib/krane/container_logs.rb
@@ -3,7 +3,7 @@ module Krane
   class ContainerLogs
     attr_reader :lines, :container_name
 
-    DEFAULT_LINE_LIMIT = 25
+    DEFAULT_LINE_LIMIT = Integer(ENV.fetch('KRANE_LOG_LINE_LIMIT', 25))
 
     def initialize(parent_id:, container_name:, namespace:, context:, logger:)
       @parent_id = parent_id


CHANGELOG.md
```diff
@@ -2,6 +2,9 @@
 
 - Remove resources that are targeted by side-effect-inducing mutating admission webhooks from the serverside dry run batch [#798](https://github.com/Shopify/krane/pull/798)
 
+*Enhancements*
+- ENV["LOG_LINE_LIMIT"] allows the number of container logs printed for failures to be configurable from the 25 line default [#803](https://github.com/Shopify/krane/pull/803).
```
casperisfine, line 6: I'd prefix this with `KRANE_` -> `KRANE_DEFAULT_LINE_LIMIT`

casperisfine: Erf, `KRANE_LOG_LINE_LIMIT`

lib/krane/container_logs.rb
```diff
@@ -3,7 +3,9 @@ module Krane
   class ContainerLogs
     attr_reader :lines, :container_name
 
-    DEFAULT_LINE_LIMIT = 25
+    def self.default_line_limit
+      Integer(ENV["LOG_LINE_LIMIT"], exception: false) || 25
```
casperisfine, line 7:
```suggestion
      Integer(ENV.fetch('KRANE_LOG_LINE_LIMIT', 25))
```

lib/krane/container_logs.rb
```diff
@@ -3,7 +3,9 @@ module Krane
   class ContainerLogs
     attr_reader :lines, :container_name
 
-    DEFAULT_LINE_LIMIT = 25
+    def self.default_line_limit
```
casperisfine, line 6: I'm not sure it's worth converting to a method. It's only useful for testing, but it's not a high value test.

lib/krane/container_logs.rb
jenshenny: ` Integer(ENV.fetch('KRANE_LOG_LINE_LIMIT', 25))` would throw an exception if it was a string that can't be converted into an integer (eg. `ENV['KRANE_LOG_LINE_LIMIT']=lol`, are we fine with letting that happen since we're probably going to be the only ones using it?

casperisfine: Yes I know. But if you put garbage in that ENV var, we might as well let you know rather than silently getting back to the default.

casperisfine (approved):
LGTM.

CHANGELOG.md
```diff
@@ -6,6 +6,9 @@
 *Bug Fixes*
 - Remove resources that are targeted by side-effect-inducing mutating admission webhooks from the serverside dry run batch [#798](https://github.com/Shopify/krane/pull/798)
 
+*Enhancements*
+- ENV["KRANE_LOG_LINE_LIMIT"] allows the number of container logs printed for failures to be configurable from the 25 line default [#803](https://github.com/Shopify/krane/pull/803).
+
```
timothysmith0609, line 11: You'll just need to rebase on master since we just cut a release today

dturn (approved):

timothysmith0609 (approved):

Issue #808, opened 2021-02-23
2.1.6+ error deploying 


strowi:
# Bug report

[Description of the bug]

**Expected behavior:** successful deployment with v2.1.6+

**Actual behavior:** Deployment fails with:
```# testing k8s-connection:
Kubernetes master is running at https://rancher02.XYZ.net/k8s/clusters/c-testing
CoreDNS is running at https://rancher02.XYZ.net/k8s/clusters/c-testing/api/v1/namespaces/kube-system/services/kube-dns:dns/proxy
KubeDNSUpstream is running at https://rancher02.XYZ.net/k8s/clusters/c-testing/api/v1/namespaces/kube-system/services/kube-dns-upstream:dns/proxy
To further debug and diagnose cluster problems, use 'kubectl cluster-info dump'.
  - no global resources to deploy. 
# checking/creating namespace haproxy-exporter
NAME                              STATUS   AGE
haproxy-exporter   Active   461d
secret "ejson-keys" deleted
secret/ejson-keys created
# deploying to haproxy-exporter... 
[INFO][2021-02-23 14:32:22 +0000] 
[INFO][2021-02-23 14:32:22 +0000] ---------------------------------Phase 1: Initializing render task----------------------------------
[INFO][2021-02-23 14:32:22 +0000] Validating configuration
[INFO][2021-02-23 14:32:22 +0000] 
[INFO][2021-02-23 14:32:22 +0000] -----------------------------------Phase 2: Rendering template(s)-----------------------------------
[INFO][2021-02-23 14:32:22 +0000] Rendering 00_imagePullSecret_hub_xyz_net.yml...
[INFO][2021-02-23 14:32:22 +0000] Rendered 00_imagePullSecret_hub_xyz_net.yml
[INFO][2021-02-23 14:32:22 +0000] Rendering haproxy-exporter-deployment.yml.erb...
[INFO][2021-02-23 14:32:22 +0000] Rendered haproxy-exporter-deployment.yml.erb
[INFO][2021-02-23 14:32:22 +0000] 
[INFO][2021-02-23 14:32:22 +0000] ------------------------------------------Result: SUCCESS-------------------------------------------
[INFO][2021-02-23 14:32:22 +0000] Successfully rendered 2 template(s)
[INFO][2021-02-23 14:32:22 +0000] 
[INFO][2021-02-23 14:32:25 +0000] 
[...]

PR #809, opened 2021-02-28
Retrieve cluster_url from kubectl dynamically to allow for multiple c…

strowi:
just a quick fix for #808, not sure it qualifies shopify's standards though, as i am not a ruby pro.

**How is this accomplished?**
dynamically retrieve cluster_url from kube-context and use this for api_paths/fetch_api_path

**What could go wrong?**
not sure, retrieving wrong/no apis from cluster -> deployments fail


lib/krane/cluster_resource_discovery.rb
```diff
@@ -52,9 +52,16 @@ def fetch_mutating_webhook_configurations
 
     private
 
+    def cluster_url()
```
dturn, line 55: This will prevent us from making a kubectl call each time.
```
def cluster_url
  @cluster_url ||= being 
        raw_json, _, st = kubectl.run("config", "view", "--minify", "--output", "jsonpath={.clusters[*].cluster}", attempts: 5, use_namespace: false)
        st.success? ? JSON.parse(raw_json)["server"] : "/"

  end
end
```

lib/krane/cluster_resource_discovery.rb
```diff
@@ -66,7 +73,7 @@ def api_paths
 
     def fetch_api_path(path)
       @api_path_cache[path] ||= begin
-        raw_json, err, st = kubectl.run("get", "--raw", path, attempts: 2, use_namespace: false)
+        raw_json, err, st = kubectl.run("get", "--raw", cluster_url, attempts: 2, use_namespace: false)
```
dturn, line 76:
```suggestion
        raw_json, err, st = kubectl.run("get", "--raw", "#{cluster_url}/#{path}, attempts: 2, use_namespace: false)
```

dturn (commented):
This doesn't seem to work as expected for me:

```
>kubectl config view --minify  --output 'jsonpath={.clusters[*].cluster}' --context cluster-a
map[certificate-authority-data:DATA+OMITTED server:https://xx.xx.xx.xxx]%   
```

lib/krane/cluster_resource_discovery.rb
strowi: Hm interesting, maybe there is a difference between token + cert-based config? This is what i get:

```
~> kubectl config view --minify  --output 'jsonpath={}' --context ci|jq                                                         (ci/test)∞
{
  "apiVersion": "v1",
  "clusters": [
    {
      "cluster": {
        "server": "https://rancher02.xyz.net/k8s/clusters/c-xyz"
      },
      "name": "ci"
    }
  ],
  "contexts": [
    {
      "context": {
        "cluster": "ci",
        "namespace": "test",
        "user": "user-xyz"
      },
      "name": "ci"
    }
  ],
  "current-context": "ci",
  "kind": "Config",
  "preferences": {},
  "users": [
    {
      "name": "user-xyz",
      "user": {
        "token": "REDACTED"
      }
    }
  ]
}
```

lib/krane/cluster_resource_discovery.rb
strowi: right, seems more consistent. ;) As i said, definitely not a ruby-pro.

lib/krane/cluster_resource_discovery.rb
strowi: The jsonpath can/should be further limited to `jsonpath={.clusters[*].cluster.server}`, so we don't pull the cert-data.. that would also eliminate the need to json_parse ?

strowi: @dturn could you take another look? 
I updated the PR with your suggestions, and modified the retrieval to only getthe server url, not the whole context.
I'm still getting 2 calls to cluster_url, but i guess thats a lot less than before..:

```
[INFO][2021-03-05 09:49:04 +0100]	Discovering resources:
https://rancher02.XYZ.net/k8s/clusters/cxyz
[INFO][2021-03-05 09:49:08 +0100]	  - Deployment/nginx-deployment
[INFO][2021-03-05 09:49:08 +0100]	  - Service/nginx
https://rancher02.xyznet/k8s/clusters/c-xyz
```

regards,
strowi

dturn:
We're getting close. I believe you're getting  `"https://rancher02.xyz.net/k8s/clusters/c-xyz"` and I'm now just getting `https://xx.xx.xx.xx` but what we really want from your response is `/k8s/clusters/c-xyz` and either `nil` or `/` for mine. I'm guessing you'll need to parse the response with something like ` URI(raw_response).path`

strowi:
Ah i think i understand now, having a cert-based context uses a different approach without the https://domain..
With the above additions i was able to deploy to a rancher-url + a k3d cluster using cert-based auth.

strowi:
not to bug, but is there anything else needed?

timothysmith0609:
Hi @strowi I'm going to be taking a look at this today or tomorrow, sorry for the delay

timothysmith0609:
The code in and of itself looks good to me, but would like to have a test in place to validate everything is working as expected. I think just a simple unit test should suffice

strowi:
This might take me a while, since i haven't had a lot of contact with unit-tests in ruby.. if anyone got a  quick solution.. ;)

timothysmith0609:
If I have a chance I’ll try to come up with something

timothysmith0609:
Hi Strowi, hoping to write the tests before weeks end and get this merged

lib/krane/cluster_resource_discovery.rb
```diff
@@ -52,9 +52,24 @@ def fetch_mutating_webhook_configurations
 
     private
 
+    def cluster_url()
+      @cluster_url ||= begin
+        raw_response, err, st = kubectl.run("config", "view", "--minify", "--output", "jsonpath={.clusters[*].cluster.server}", attempts: 5, use_namespace: false)
+        uri = if st.success?
+          if URI(raw_response).path == ( nil || "/" )
```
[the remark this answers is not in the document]
strowi: @timothysmith0609 thx for working on this. As you might have noticed my ruby-skills are low to non-existent, so i made the changes i needed to get it working with rancher again.

Never observed, just wanted to cover all the possibilities.

lib/krane/cluster_resource_discovery.rb
```diff
@@ -52,9 +52,24 @@ def fetch_mutating_webhook_configurations
...
+            URI(raw_response).path
+          else
+            URI(raw_response).to_s
+          end
+        else
+          raise FatalKubeAPIError, "Error retrieving cluster url : #{err}"
+        end
+      end
```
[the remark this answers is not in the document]
strowi: looks good and still seems to work.:)

strowi:
has been solved by https://github.com/Shopify/krane/pull/827 ;)

PR #810, opened 2021-03-03
Deploy time instrumentation

timothysmith0609:
Not intended to be merged, for experimentation only


jerr (approved):
👍 the instrumentation part looks good to me, this will be very useful, thanks 😍

PR #811, opened 2021-03-09
Add support for SOPS-encrypted YAML templates

CodingAnarchy:

Allow users to encrypt their YAML templates with Mozilla SOPS and have `krane` be able to decrypt and use these upon deploy.

**How is this accomplished?**

Modifying `TemplateSets#templates` to recognize a `sops` key in the YAML as a sign it needs to decrypt a SOPS file, and giving it the logic to do so.

**What could go wrong?**

Users must have `sops` installed and configured on the deploying system in order to use this functionality. Encrypted SOPS files are otherwise invalid templates, and an error is raised if so.

### DIFF
--- a/README.md
+++ b/README.md
@@ -137,6 +137,8 @@ If you need to share a namespace with resources which are managed by other tools
 All templates must be YAML formatted.
 We recommended storing each app's templates in a single directory, `{app root}/config/deploy/{env}`. However, you may use multiple directories.
 
+Templates may be encrypted YAML produced from [Mozilla SOPS](https://github.com/mozilla/sops). To use these templates, ensure that SOPS is installed and configured with the appropriate decryption keys before running `krane deploy` or SOPS-encrypted templates will throw an invalid template error.
+
 If you want dynamic templates, you may render ERB with `krane render` and then pipe that result to `krane deploy -f -`.
 
 ### Customizing behaviour with annotations
--- a/lib/krane/template_sets.rb
+++ b/lib/krane/template_sets.rb
@@ -1,4 +1,5 @@
 # frozen_string_literal: true
+require 'open3'
 require 'krane/delayed_exceptions'
 require 'krane/ejson_secret_provisioner'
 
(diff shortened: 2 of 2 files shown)


timothysmith0609:
Hi @CodingAnarchy , thanks for the contribution! Krane has a philosophy of keeping as limited a scope as possible. In particular, this means staying focused on its core tasks (rendering/deploying) and offloading additional functionality to other, more focused, tools.

In the case of `sops`, is there anything to prevent you from piping the output from `krane render` to `sops` and finally to `krane deploy`? I see that `sops` outputs to `STDOUT` by default so it certainly seems doable.

CodingAnarchy:
@timothysmith0609 We've been running with this patch on an old version of `kubernetes-deploy`, so I'm not sure how `krane render` would modify the flow there.

My intention here was to solve two problems we currently have:

(1) The template directory we have is mixed with some standard YAML and some encrypted YAML - the only distinctions between the two are easily broken filename conventions and the actual content of the file (the `sops` keys in the YAML). Use of `krane render` and then piping this to `sops` doesn't really work for this case.

(2) Adding a new deployment file that is encrypted with `sops` requires extensive and time consuming looping through the template directory to decrypt and replace files with the decrypted version if we can't do it inline as in this patch, which has both performance implications for a CI build/deploy and some security implications (doing things inline allows for maintaining these values only in memory while the deploy task is running).

We are mainly trying to update to `krane` and cease maintaining our older forked version, but this patch seems necessary for us at this stage.

Also, `krane render` seems to be designed to parse ERB templates (per the documentation) and our templates - especially those encrypted by SOPS - are already standard YAML.

timothysmith0609:
Hi CodingAnarchy, please accept my apologies it's taken a while to respond. I'm going to consult with a colleague who's currently out of office, but my feeling is that we most likely won't accept this change. On a conceptual level, `krane deploy` is intended to take as input fully realized YAML manifests. The fact that we do accept things like templated ERB files is due to a backwards-compatibility grandfathering of a feature we've come to regret. In general, this means we try to offload the work of rendering templates (in whichever shape or form) to other processes. 

>(2) Adding a new deployment file that is encrypted with sops requires extensive and time consuming looping through the template directory to decrypt and replace files with the decrypted version if we can't do it inline as in this patch, which has both performance implications for a CI build/deploy and some security implications (doing things inline allows for maintaining these values only in memory while the deploy task is running).

If possible, could you elaborate on this point? How many decryptions do you need to do for a CI run? Naïvely, I would have expected you could partition the files by grepping for the `sops` key, decrypting those, and piping that output along with the other partition, but something tells me I'm taking too simplistic a view with that.

CodingAnarchy:
Perhaps I am misunderstanding something about the process recommendation in this context, but the basic setup we have is that all Kubernetes secrets files are encrypted with SOPs in our deploy template directory.

Our current workflow uses a modified fork of the old `kubernetes-deploy` to allow for passing the template directory as-is when deploying. This doesn't have to be how we are doing things, for sure, but I am not following how we can simply decrypt a portion of the files and then pass all the files to be deployed at once.

Alternatively, we could possibly alter the process to call `krane deploy` multiple times, but the idea here was to keep things simpler from a deployment perspective. In essence, the more times we have to call `krane deploy` the less benefit we get over just using the Kubernetes CLI commands directly.

timothysmith0609:
After thinking this over some more I'm going to close this PR. For a while, krane has been moving in a direction where we want its core functionality to be relatively straightforward. In essence, that means we don't want to be adding additional logic outside of what is required for deploying a _realized_ set of templates. From that point of view, adding an additional codepath that leverages tooling that _does not need to be_ in the critical path of a deploy feels out of place. I would like to thank you for taking the time to make the contribution and apologize that I won't be moving forward with it, here.

PR #813, opened 2021-03-09
Use same timeout everywhere in resource discovery

dturn:
Make krane work in high latency envs

close: https://github.com/Shopify/krane/issues/805

**How is this accomplished?**
Using the same timeout length everywhere.

**What could go wrong?**
Deploys get slower


timothysmith0609 (approved):
There and back again...

PR #822, opened 2021-04-30
ClusterResourceDiscovery: disable memoization

### LINKED ISSUE #821: Error when deploying - "error: no matches for kind "FrontendConfig""
# Bug report

Frequently, when deploying our app `spin-services`, we're seeing the error `no matches for kind "FrontendConfig" in version "networking.gke.io/v1"`: https://buildkite.com/shopify/spin-multi-deploy/builds/689#504d3cb7-dc28-43b6-9b3a-bbfbf26d0f98

```
[FATAL][2021-04-29 16:14:32 +0000][spin-us-central1-1][spin-services]	Unidentified error(s):
[FATAL][2021-04-29 16:14:32 +0000][spin-us-central1-1][spin-services]	    error: no matches for kind "FrontendConfig" in version "networking.gke.io/v1"
```

This seems to be exactly the same error as #761. I see that was replaced by #778, which was meant to fix this more generally, but it looks like it's recurring.

This is sometimes fixed by a retry, but not usually.

mistydemeo:
This is slightly a shot in the dark, but as an attempt to resolve #821 I've disabled the memoization that was introduced in #778 in case that's the source of the bug. I'm not sure this is really the solution, but it may be worth testing.

**How is this accomplished?**
I've simply removed instance variable assignments/reads.

**What could go wrong?**
Performance could degrade, although the equivalent calls were not cached before #778.

### DIFF
--- a/lib/krane/cluster_resource_discovery.rb
+++ b/lib/krane/cluster_resource_discovery.rb
@@ -53,26 +53,22 @@ module Krane
     private
 
     def api_paths
-      @api_path_cache["/"] ||= begin
-        raw_json, err, st = kubectl.run("get", "--raw", "/", attempts: 5, use_namespace: false)
-        paths = if st.success?
-          JSON.parse(raw_json)["paths"]
-        else
-          raise FatalKubeAPIError, "Error retrieving raw path /: #{err}"
-        end
-        paths.select { |path| %r{^\/api.*\/v.*$}.match(path) }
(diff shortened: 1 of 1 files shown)


timothysmith0609:
Hi Misty, we're convinced the problem lies in the fact that the `krane` executable in multi-deploy containers has a race condition which can cause an older version of `krane` to be in `$PATH`. We are exploring the fix for this now.

PR #826, opened 2021-05-13
Drop 1.15 support

timothysmith0609:
Dropping 1.15 support as it's quite old now and no longer supported. Most likely will drop 1.16 in the near future as well.

Also updates `dev.yml` to use 1.18.18 for local k8s cluster for development.


karanthukral (approved):
Thanks!

timothysmith0609:
This ended up being done during the 2.1.10 release.

PR #827, opened 2021-05-13
Dynamically discover cluster url for resource discovery

timothysmith0609:
Replaces #809 

Dynamically discovers the cluster_url to be used for the raw queries during cluster resource discovery (see #809 for further details). I dug through the original implementation a bit and fixed some test regressions that were introduced by the change.

cc @strowi

### DIFF
--- a/lib/krane/cluster_resource_discovery.rb
+++ b/lib/krane/cluster_resource_discovery.rb
@@ -52,8 +52,9 @@ module Krane
 
     private
 
-    def cluster_url
-      @cluster_url ||= begin
+    # During discovery, the api paths may not actually be at the root, so we must programatically find it.
+    def base_api_path
+      @base_api_path ||= begin
         raw_response, err, st = kubectl.run("config", "view", "--minify", "--output",
           "jsonpath={.clusters[*].cluster.server}", attempts: 5, use_namespace: false)
         raise FatalKubeAPIError, "Error retrieving cluster url: #{err}" unless st.success?
@@ -64,7 +65,7 @@ module Krane
 
     def api_paths
       @api_path_cache["/"] ||= begin
-        raw_json, err, st = kubectl.run("get", "--raw", cluster_url, attempts: 5, use_namespace: false)
+        raw_json, err, st = kubectl.run("get", "--raw", base_api_path, attempts: 5, use_namespace: false)
         paths = if st.success?
           JSON.parse(raw_json)["paths"]
         else
--- a/lib/krane/cluster_resource_discovery.rb
+++ b/lib/krane/cluster_resource_discovery.rb
@@ -58,7 +58,7 @@ module Krane
           "jsonpath={.clusters[*].cluster.server}", attempts: 5, use_namespace: false)
         raise FatalKubeAPIError, "Error retrieving cluster url: #{err}" unless st.success?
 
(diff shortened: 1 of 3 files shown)


test/unit/cluster_resource_discovery_test.rb
```diff
@@ -6,7 +6,7 @@ class ClusterResourceDiscoveryTest < Krane::TestCase
 
   def test_fetch_resources_failure
     crd = mocked_cluster_resource_discovery(success: false)
-    assert_raises_message(Krane::FatalKubeAPIError, "Error retrieving raw path /:") do
+    assert_raises_message(Krane::FatalKubeAPIError, "Error retrieving cluster url:") do
```
timothysmith0609, line 9: This is just an ordering issue / implementation detail, since we now make an additional kubectl call to discover what the base path is, that call fails before the 'raw path' error.

lib/krane/cluster_resource_discovery.rb
```diff
@@ -66,7 +76,7 @@ def api_paths
 
     def fetch_api_path(path)
       @api_path_cache[path] ||= begin
-        raw_json, err, st = kubectl.run("get", "--raw", path, attempts: 2, use_namespace: false)
+        raw_json, err, st = kubectl.run("get", "--raw", File.join(cluster_url, path), attempts: 2, use_namespace: false)
```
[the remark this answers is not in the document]
timothysmith0609: I've changed this again to just use `path`. Presumably your kubeconfig should already have the base URI. @strowi when you test this can you let me know if this fails; unfortunately I can't easily test against your situation

strowi:
Again thanks! The joining-part was where i got confused because i sometimes woud get an error when the url contains 2 slashes like "apis//...", but cant observe this anymore.
I can confirm that it's still possible to deploy to local minikube and rancher clusters. ;)

lib/krane/cluster_resource_discovery.rb
```diff
@@ -52,9 +52,20 @@ def fetch_mutating_webhook_configurations
 
     private
 
+    # During discovery, the api paths may not actually be at the root, so we must programatically find it.
+    def base_api_path
+      @base_api_path ||= begin
+        raw_response, err, st = kubectl.run("config", "view", "--minify", "--output",
```
manuelfelipe, line 58: so... how does this work when `kubeconfig` doesn't have a `current-context` set? or we are setting the currently config explicitly somewhere else before this runs?

manuelfelipe: looks like you are still able to provide `--context` when using the `config` command. ie, 

```
# using different random names
kubectl --context my-context-38 config view --minify --output jsonpath="{.clusters[*].cluster.server}"
https://35.x.x.x%

kubectl --context my-context-37 config view --minify --output jsonpath="{.clusters[*].cluster.server}"
https://34.x.x.x%
```

you get the gist, missing explicit argument basically.

manuelfelipe: guessing I first need to read the code before 😮. Looks like the `Kubectl` class already does this by default in https://github.com/Shopify/krane/blob/master/lib/krane/kubectl.rb#L116 unless explicitly disabled.

manuelfelipe (approved):

PR #836, opened 2021-09-01
Support restart task for statefulsets and daemonsets

timothysmith0609:
Closes https://github.com/Shopify/krane/issues/835
Closes https://github.com/Shopify/krane/issues/610

Adds support to the restart task for StatefulSets and DaemonSets. To maintain backwards compatibility, and simplify some implementation, this adds separate flags for `statefulsets` and `daemonsets`.

This ended up being a bigger PR than I thought since the assumption that only deploys would be restart targets was fairly embedded in the code. Because of this, it was more straightforward to duplicate some parts of the code (e.g. `#fetch_deployments|statefulsets|daemonsets`, etc.).  In addition, the old code directly altered an ENV var (`RESTARTED_AT`) of a pod's container to trigger a restart. I've switched this to use the convention used by `kubectl`, which is to add an annotation in `spec.template.metadata` of the target resource. 

I still need to add some tests to ensure StatefulSets and DaemonSets are safe against regressions, but I've tophatted it enough to open this up for review, now.

List of changes:

- [x] Restart task changes
- [x] Update README
- [x] Update tests so we hit codepaths for Deployment, StatefulSet, and Daemonset restarts
- [x] Add/Change some fixtures

### DIFF
--- a/lib/krane/restart_task.rb
+++ b/lib/krane/restart_task.rb
@@ -103,7 +103,7 @@ module Krane
 
     private
 
-    def tags(status, deployments)
+    def tags(status, deployments, statefulsets, daemonsets)
       %W(namespace:#{@namespace} context:#{@context} status:#{status} deployments:#{deployments.to_a.length}
          statefulsets:#{statefulsets.to_a.length} daemonsets:#{daemonsets.to_a.length}})
     end


test/integration/restart_task_test.rb
```diff
@@ -198,44 +199,6 @@ def test_restart_not_existing_namespace
       in_order: true)
   end
 
-  def test_restart_failure
```
timothysmith0609, line 201: I'm not sure how to go about forcing a failure in this case, since we are no longer pushing an env variable into a container (and I don't really see the need to do so, anyway).

lib/krane/restart_task.rb
```diff
@@ -171,18 +251,38 @@ def fetch_deployments(list)
...
+
+    def build_patch_payload(_deployment)
       {
         spec: {
           template: {
-            spec: {
-              containers: containers.map do |container|
-                {
-                  name: container.name,
-                  env: [{ name: "RESTARTED_AT", value: Time.now.to_i.to_s }],
```
stefanmb, line 183: I found an [internal reference to this variable here](https://github.com/Shopify/cloudbuddies/blob/master/buddies/appbuddy/state_config_map.go#L126).

test/integration/restart_task_test.rb
stefanmb: I think you could use the [downward api](https://kubernetes.io/docs/tasks/inject-data-application/downward-api-volume-expose-pod-information/#store-pod-fields). In that example they mount pod annotations into the local file system.

A more convoluted alternative would be to kubectl inside the pod and read the annotations off the well known deployment.

lib/krane/restart_task.rb
timothysmith0609: This looks like perhaps it was cargo culted from here since they're using it to simply trigger a restart in this case (and they don't particularly care about _reading_ its value). I'm tempted to change the implementation in the buddy to mimic this one (changing annotations on the template metadata). IMO having these processes/controllers mounting variables directly into containers doesn't strike me as the wisest idea.

README.md
```diff
@@ -451,7 +451,7 @@ resource to deploy.
 
 # krane restart
 
-`krane restart` is a tool for restarting all of the pods in one or more deployments. It triggers the restart by touching the `RESTARTED_AT` environment variable in the deployment's podSpec. The rollout strategy defined for each deployment will be respected by the restart.
+`krane restart` is a tool for restarting all of the pods in one or more deployments, statefuls sets, and/or daemon sets. It triggers the restart by patching template metadata with the `kubectl.kubernetes.io/restartedAt` annotation (with the value being an RFC 3339 representation of the current time). Note this is the manner in which `kubectl rollout restart` itself triggers restarts.
```
stefanmb, line 454: 👍 Indeed RFC 3339 is used by [kubectl](https://github.com/kubernetes/kubectl/blob/652881798563c00c1895ded6ced819030bfaa4d7/pkg/polymorphichelpers/objectrestarter.go#L41) too.

lib/krane/restart_task.rb
```diff
@@ -97,62 +104,135 @@ def tags(status, deployments)
       %W(namespace:#{@namespace} context:#{@context} status:#{status} deployments:#{deployments.to_a.length}})
     end
 
-    def identify_target_deployments(deployment_names, selector: nil)
-      if deployment_names.nil?
-        deployments = if selector.nil?
-          @logger.info("Configured to restart all deployments with the `#{ANNOTATION}` annotation")
-          apps_v1_kubeclient.get_deployments(namespace: @namespace)
+    def identify_target_workloads(deployment_names, statefulset_names, daemonset_names, selector: nil)
+      if deployment_names.blank? && statefulset_names.blank? && daemonset_names.blank?
```
stefanmb, line 108: Curious why `blank?` vs `nil`? Are we passing empty strings in one of the paths? Looks like the default param in `run!` is an empty list now instead of nil?

test/integration/restart_task_test.rb
```diff
@@ -212,7 +242,7 @@ def test_restart_failure
           "command" => [
             "/bin/sh",
             "-c",
-            "test $(env | grep -s RESTARTED_AT -c) -eq 0",
+            "test $(cat /etc/podinfo/annotations | grep -s kubectl.kubernetes.io/restartedAt -c) -eq 0",
```
stefanmb, line 245: 👍

lib/krane/restart_task.rb
```diff
@@ -22,6 +22,8 @@ def initialize(deployment_name, response)
     HTTP_OK_RANGE = 200..299
     ANNOTATION = "shipit.shopify.io/restart"
 
+    RESTART_TRIGGER_ANNOTATION = "kubectl.kubernetes.io/restartedAt"
```
stefanmb, line 25: Is there any specific benefit to reusing the same annotation name as kubectl or is this done just to reduce annotation clutter?

stefanmb (approved):
Overall LGTM - left a few small questions.

lib/krane/restart_task.rb
timothysmith0609: I'm fairly ambivalent about which one to use; this one cuts down noise, but can also be construed as somewhat misleading. I'm tempted to leave it as-is and change it if we get some complaints

lib/krane/restart_task.rb
timothysmith0609: While the default arg is the empty list, I can't guarantee that someone won't explicitly pass `nil` at some point. Using `blank?` seems preferable over `empty?` or `nil?` since it will catch either case.

peiranliushop (commented):
LGTM. I assume whoever uses older version of krane will still trigger restart through patching the env var.

PR #839, opened 2021-10-21
Switch to GitHub Actions for CI

DazWorrall:

Ref: https://github.com/Shopify/k8s-ci/issues/58

Switch to GitHub actions for CI, which has some upsides:

* Much better contributor experience - no need to wait for a Shopify org member to run CI on your behalf, and CI will work on forks etc
* Relies on community and GitHub maintained infra
* Helps Shopify retire some ❄️y infrastructure
* Removes some risk from Shopify of figuring out how to run code from third party contributors on our own infra

I have maintained the same matrix of Ruby version => Kubernetes version. I have also disabled warnings in CI - the logs are unintelligible at the moment.

**How is this accomplished?**

Removing the old config, and writing a new one! A notable change between the old and new infra is the cluster setup is done using `kind` rather than minikube.

**What could go wrong?**

CI times on Actions are a bit slower than our own infra, as runners have less compute resources available. I think this is a worthwhile trade off though - particularly for third party contributors, slower CI trumps unavailable CI.

This required some minor fixes to the tests to fix some assumptions about the previous infra, but there is still a couple of failing tests pertaining to PVC's that I would appreciate someone else helping with:

https://github.com/Shopify/krane/runs/3964240918?check_suite_focus=true#step:5:318
https://github.com/Shopify/krane/runs/3964240918?check_suite_focus=true#step:5:372

### DIFF
--- a/test/integration/krane_deploy_test.rb
+++ b/test/integration/krane_deploy_test.rb
@@ -117,7 +117,7 @@ class KraneDeployTest < Krane::IntegrationTest
       prune_matcher("persistentvolumeclaim", "", "hello-pv-claim"),
     ] # not necessarily listed in this order
     # 1.21 appears to list the ingress as belonging to extensions, not networking.k8s.io
-    if kube_server_version >= Gem::Version.new("1.17.0") && kube_server_version <= Gem::Version.new("1.21.0")
+    if kube_server_version >= Gem::Version.new("1.18.0") && kube_server_version < Gem::Version.new("1.21.0")
       expected_pruned << prune_matcher("ingress", "networking.k8s.io", "web")
     else
       expected_pruned << prune_matcher("ingress", "extensions", "web")


DazWorrall:
Happy infra freeze @Shopify/app-lifecycle 👋 This would be a great PR to get over the line during the chill: tl;dr moving CI to GitHub actions, but I have a failing test I need help with.

jpfourny:
My best guess is that the test is making the assumption that the "default" storage class has `Immediate` for `volumeBindingMode`. Perhaps that is not the case?

The code in `persistent_volume_claim.rb` is pretty clear about deploy not succeeding with `Pending` status unless `volumeBindingMode` is set to `WaitForFirstConsumer`.

```ruby
def deploy_succeeded?
      return true if status == "Bound"

      # if the StorageClass has volumeBindingMode: WaitForFirstConsumer,
      # it won't bind until after a Pod mounts it. But it must be pre-deployed,
      # as the Pod requires it. So 'Pending' must be treated as a 'Success' state
      if storage_class&.volume_binding_mode == "WaitForFirstConsumer"
        return status == "Pending" || status == "Bound"
      end
      false
    end
```

DazWorrall:
> My best guess is that the test is making the assumption that the "default" storage class has `Immediate` for `volumeBindingMode`.

Confirmed you're correct @jpfourny, minikube's default storage class:

```
NAME                 PROVISIONER                RECLAIMPOLICY   VOLUMEBINDINGMODE   ALLOWVOLUMEEXPANSION   AGE
standard (default)   k8s.io/minikube-hostpath   Delete          Immediate           false                  70s
```

... vs kind:

```
NAME                 PROVISIONER             RECLAIMPOLICY   VOLUMEBINDINGMODE      ALLOWVOLUMEEXPANSION   AGE
standard (default)   rancher.io/local-path   Delete          WaitForFirstConsumer   false                  15s
```

I'll propose a change here to remove the assumption about the default class.

jpfourny (approved):

PR #849, opened 2021-12-02
ccccccvrfurlvftjdccifenkiblnhdiccldekcltvnedRevert "Explicitly add date and cgi gems"

casperisfine:
...

**How is this accomplished?**
...

**What could go wrong?**
...


JamesOwenHall (commented):
Hmm, I may have hit the approve button too quickly. This is not just a library, it's an executable in its own right as well.

casperisfine:
Sure, but it still doesn't use `Date.parse` nor `GCI`.

arbourd (approved):

PR #854, opened 2021-12-17
Fix local (client) dry-runs with kubectl 1.23.0

### LINKED ISSUE #850: Krane failing with the latest ubuntu version on github actions
# Bug report

In the latest [version of the ubuntu image used for Github actions](https://github.com/actions/virtual-environments/blob/ubuntu20/20211209.3/images/linux/Ubuntu2004-README.md), they updated kubectl to version 1.23.0.

According to the [Kuberentes Changelog for this version](https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG/CHANGELOG-1.23.md) they have:

> Removed kubectl --dry-run empty default value and boolean values. kubectl --dry-run usage must be specified with --dry-run=(server|client|none). (#105327, @julianvmodesto)

This breaks Krane on version 2.1.5, but from what I can see in the Krane's changelog this hasn't been addressed yet. Sorry if this has actually been fixed in a more recent version of Krane, don't really have the time to test it (holidays starting in 5 minutes 👼).

The error that this generates is:

```
[FATAL][2021-12-15 16:30:06 +0000]	Invalid template: ConfigMap-application20211215-6158-o5fuyo.yml
[FATAL][2021-12-15 16:30:06 +0000]	> Error message:
[FATAL][2021-12-15 16:30:06 +0000]	    error: --dry-run flag without a value was specified. A value must be set: "none", "server", or "client".
```

### Steps to Reproduce

1. Setup AWS credentials in your workflow:
```yaml
- name: Configure AWS credentials
  uses: aws-actions/configure-aws-credentials@v1
    with:
      aws-access-key-id: ${{ secrets.AWS_ACCESS_KEY_ID }}
      aws-secret-access-key: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
      aws-region: ${{ secrets.AWS_DEFAULT_REGION }}
```
2. Set your kubeconfig with:
```yaml
- name: Setup kubeconfig
  run: |
    aws eks update-kubeconfig --name $CLUSTER_NAME --alias $CLUSTER_NAME
```
3. Installl krane within a Github workflow with a step similar to this:
```yaml
- name: Install krane
   run: |
     gem install krane -v 2.1.5
```
4. Try to run a deploy the usual way

# Workaround

For now, we have found a workaround by installing the previous version of kubectl and overriding the one installed by default with a step like this:

```yaml
- name: Install kubectl 1.22.4
  run: |
    sudo apt remove -y kubectl
    sudo apt-get update
    sudo apt-get install -y apt-transport-https ca-certificates curl
    sudo curl -fsSLo /usr/share/keyrings/kubernetes-archive-keyring.gpg https://packages.cloud.google.com/apt/doc/apt-key.gpg
    echo "deb [signed-by=/usr/share/keyrings/kubernetes-archive-keyring.gpg] https://apt.kubernetes.io/ kubernetes-xenial main" | sudo tee /etc/apt/sources.list.d/kubernetes.list
    sudo apt-get update
    sudo apt install kubectl=1.22.4-00
    sudo cp /usr/bin/kubectl /usr/local/bin/kubectl
```

arbourd:

Fix #850, which breaks all krane deployments if using kubectl `1.23.0`.

**How is this accomplished?**

By explicitly setting `--dry-run=client` instead of `--dry-run`

**What could go wrong?**

Older versions might not work, but this PR only sets `--dry-run=client` if the client is greater than or equal to 1.18 when the change (and deprecation notice) was added.

### DIFF
--- a/lib/krane/kubernetes_resource.rb
+++ b/lib/krane/kubernetes_resource.rb
@@ -584,7 +584,12 @@ module Krane
     # If the resource template uses generateName, validating with apply will fail
     def validate_with_local_dry_run(kubectl)
       verb = deploy_method == :apply ? "apply" : "create"
-      command = [verb, "-f", file_path, "--dry-run", "--output=name"]
+      command = if kubectl.client_version >= Gem::Version.new('1.18')
+        [verb, "-f", file_path, "--dry-run=client" "--output=name"]
+      else
+        [verb, "-f", file_path, "--dry-run" "--output=name"]
+      end
+
       kubectl.run(*command, log_failure: false, output_is_sensitive: sensitive_template_content?,
         retry_whitelist: [:client_timeout, :empty, :context_deadline], attempts: 3, use_namespace: !global?)
     end
--- a/test/unit/krane/kubernetes_resource/deployment_test.rb
+++ b/test/unit/krane/kubernetes_resource/deployment_test.rb
@@ -361,7 +361,7 @@ class DeploymentTest < Krane::TestCase
 
   def stub_validation_dry_run(out: "", err: "", status: SystemExit.new(0))
     kubectl.expects(:run)
-      .with('apply', '-f', anything, '--dry-run', '--output=name', anything)
+      .with('apply', '-f', anything, '--dry-run=client', '--output=name', anything)
       .returns([out, err, status])
   end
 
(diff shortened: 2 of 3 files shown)


jpfourny:
I merged this prematurely ... look like CI had some errors. I think it may be missing commas in the command array. Looking into it now ...

PR #856, opened 2021-12-17
Build for kubectl 1.22, 1.23

arbourd:

Test against GA kubernetes versions 1.22, 1.23.

**How is this accomplished?**

By adding the test to the matrix.

**What could go wrong?**

The test for 1.23 will fail because krane is broken with 1.23.0+


jpfourny:
I created a new branch from this PR and rebased with master. Here is my version: https://github.com/Shopify/krane/pull/858

PR #858, opened 2022-01-05
Build for kubectl 1.22, 1.23 (Rebased)

jpfourny:

Test against GA kubernetes versions 1.22, 1.23.
_Note: Branched from https://github.com/Shopify/krane/pull/856 and rebased on master._

**How is this accomplished?**

By adding the test to the matrix.

**What could go wrong?**

Nothing expected to go wrong - just adding to existing test matrix.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -10,6 +10,8 @@ jobs:
       fail-fast: false
       matrix:
         ruby:
+        - '3.0' # With k8s 1.23
+        - '3.0' # With k8s 1.22
         - '3.0' # With k8s 1.21
         - '3.0' # With k8s 1.20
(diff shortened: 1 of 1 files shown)


jpfourny:
@DazWorrall Would it be OK to label the test cases with both the Ruby and K8s version? It's hard to tell them apart, having only the Ruby version repeated 3 times 😄 .

JamesOwenHall (approved):

jpfourny:
The 2.6.6 tests are a little flaky. 3 runs, 3 different problems 😄

DazWorrall:
> Would it be OK (or even possible?) to label the test cases with both the Ruby and K8s version? 

I think this can be done yeah, I will follow up in another PR.

PR #859, opened 2022-01-05
Configure a more useful Actions job name

DazWorrall:
Ref: https://github.com/Shopify/krane/pull/858#issuecomment-1005896132

> Would it be OK (or even possible?) to label the test cases with both the Ruby and K8s version? It's hard to tell them apart, having only the Ruby version repeated 3 times 😄 .

@jpfourny is quite right.


jpfourny (approved):
❤️

PR #860, opened 2022-01-06
Fix CI version matrix

DazWorrall:
Follow on #858 

The multi-dimensional matrix config in actions did not work as I expected, #859 showed that we weren't running all the kube versions we wanted. I fix it here, which reveals that some of our 1.22 and 1.23 tests are [failing](https://github.com/Shopify/krane/runs/4725641323?check_suite_focus=true#step:6:83) with `no matches for kind "CustomResourceDefinition" in version "apiextensions.k8s.io/v1beta1"`. I tried updating the version to just `v1` for all the fixture CRD's, but that [fails](https://github.com/Shopify/krane/runs/4725888565?check_suite_focus=true#step:6:879) with `missing required field "versions" in io.k8s.apiextensions-apiserver.pkg.apis.apiextensions.v1.CustomResourceDefinitionSpec`.

Looks like the spec is different and the fixtures need more work. I'm sure someone with more context than me can help find the right fix here.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -44,7 +44,7 @@ jobs:
         - ruby: '2.6.8'
           kind_version: 'v0.11.1'
           kubernetes_version: '1.18.19'
-          kind_image: '1.18: kindest/node:v1.18.19@sha256:7af1492e19b3192a79f606e43c35fb741e520d195f96399284515f077b3b622c'
+          kind_image: 'kindest/node:v1.18.19@sha256:7af1492e19b3192a79f606e43c35fb741e520d195f96399284515f077b3b622c'
         - ruby: '2.6.7'
           kind_version: 'v0.11.1'
           kubernetes_version: '1.17.17'
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -11,24 +11,25 @@ jobs:
       fail-fast: false
       matrix:
         ruby:
-        - '3.0' # With k8s 1.23
-        - '3.0' # With k8s 1.22
-        - '3.0' # With k8s 1.21
+        # Use unique Ruby versions, or GitHub gets confused when building the rest of the matrix
+        - '3.0.3' # With k8s 1.23
+        - '3.0.2' # With k8s 1.22
+        - '3.0.1' # With k8s 1.21
         - '3.0' # With k8s 1.20
         - '2.7' # With k8s 1.19
-        - '2.6.6' # With k8s 1.18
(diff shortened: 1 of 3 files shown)


.github/workflows/ci.yml
```diff
@@ -11,24 +11,25 @@ jobs:
       fail-fast: false
       matrix:
         ruby:
-        - '3.0' # With k8s 1.23
-        - '3.0' # With k8s 1.22
-        - '3.0' # With k8s 1.21
+        # Use unique Ruby versions, or GitHub gets confused when building the rest of the matrix
```
jpfourny, line 14: Ouch.

DazWorrall:
@jpfourny any thoughts on who can help resolve this?

jpfourny:
🤔 We may need to maintain multiple versions of deploy fixtures selected based on the k8s version being deployed to.

DazWorrall:
`CustomResourceDefinition` has been available in the `v1` API [since 1.16](https://kubernetes.io/blog/2021/07/14/upcoming-changes-in-kubernetes-1-22/#what-to-do) apparently, so at least for those we may not need different resources for different cluster versions. The same page says `Ingress` is in `v1` since `1.19`.

jpfourny:
> `CustomResourceDefinition` has been available in the `v1` API [since 1.16](https://kubernetes.io/blog/2021/07/14/upcoming-changes-in-kubernetes-1-22/#what-to-do) apparently, so at least for those we may not need different resources for different cluster versions. The same page says `Ingress` is in `v1` since `1.19`.

Interesting ... I wonder if it's time to drop testing for really old K8s versions.

DazWorrall:
We should probably test against whatever minimum version we have in the wild at Shopify, or some minimum supported version our other tooling expects. I don't know if we have any data or opinions on that. Certainly testing against 7 different releases does feel like a stretch - the bigger the gap the harder support will be.

jpfourny:
> We should probably test against whatever minimum version we have in the wild at Shopify, or some minimum supported version our other tooling expects. I don't know if we have any data or opinions on that. Certainly testing against 7 different releases does feel like a stretch - the bigger the gap the harder support will be.

AFAIK, all Shopify-managed clusters are at least 1.20 and we need to be on 1.22 in the next few months (we are following Google's support schedule).

jpfourny:
WIP: https://github.com/Shopify/krane/pull/862
Will be sending out reviews tomorrow.

jpfourny:
This can be closed, since I included the changes into my last PR (which is now in master).

PR #861, opened 2022-01-10
isolated_execution_state active_support require for Rails 7+

timothysmith0609:
We need to explicitly require `active_support/isolated_execution_state`, otherwise a `NameError` is raised (see image)

image

**How is this accomplished?**
Add the requirement where we load all the other `active_support` dependencies

**What could go wrong?**
Could result in a similar error if we're not requiring everything we need, though that seems unlikely.


timothysmith0609:
Can we get rid of the conditional and just always include that sublibrary, then?

gigr (approved):

PR #866, opened 2022-01-11
fail deploy task when supplied with no non-empty resources

jpfourny:
Fix https://github.com/Shopify/krane/issues/865

**How is this accomplished?**
Assert that final resource list is non-empty in `discover_resources` method of `deploy_task.rb`.
Note: Empty YAML documents are already skipped during discovery.

**What could go wrong?**
Nothing I can think of. It is extremely unlikely that someone would intentional deploy an empty resource / zero resources.


timothysmith0609 (commented):
The only 2 edge cases I can think of are:

- We want to prune a namespace so we deploy an empty set. **This doesn't seem very likely so I'm not worried about it**
- Rendering fails, but we also include secrets, so the set of resources isn't empty. This is less a problem with `krane` and more a problem with how we pipe rendering/secret output directly into the deploy process. Having said that, this is exactly the scenario that occurred which triggered the incident leading to this PR.

jpfourny:
> The only 2 edge cases I can think of are:
> 
> * We want to prune a namespace so we deploy an empty set. **This doesn't seem very likely so I'm not worried about it**
> * Rendering fails, but we also include secrets, so the set of resources isn't empty. This is less a problem with `krane` and more a problem with how we pipe rendering/secret output directly into the deploy process. Having said that, this is exactly the scenario that occurred which triggered the incident leading to this PR.

For runtime-based apps, secrets are combined in `prod-platform-cli`  via `render-secrets` subcommand before being piped into the `deploy` subcommand. Both of these subcommands will bail if given empty resources as of https://github.com/Shopify/production-platform/pull/786.

lib/krane/version.rb
```diff
@@ -1,4 +1,4 @@
 # frozen_string_literal: true
 module Krane
-  VERSION = "2.3.6"
+  VERSION = "2.3.7"
```
lmunro-at-shopify, line 3: `2.3.7` has been released in #861 (without a CHANGELOG entry unfortunately).

peiranliushop:
This is good. But we are not out of woods yet even with this is change.
`production-platform-next` [render-secrets](https://github.com/Shopify/production-platform/blob/7ec026f522fc9e0fa46a06bcc550ef4e29055a5c/pkg/cli/deploy/secrets/cmd.go#L40) could also pass in empty yml piped in from `krane render`.

stdin to render-secrets was introduced in PPA, so its appropriate to fix it there, like @epk 's PPA [PR](https://github.com/Shopify/production-platform/pull/786/files#diff-94ea7d625c79d83666c3f576cfeae8d5e5280c0d215059f04415bb8c5bb759fc) and without the yaml formatter change.

epk:
> krane render -f config/deploy/production --current-sha SHA | production-platform-next render-secrets -f config/config.production.ejson -m -

> stdin to render-secrets was introduced in PPA, so its appropriate to fix it there, like @epk 's PPA PR and without the yaml formatter change.

Reverted the formatter change in https://github.com/Shopify/production-platform/pull/788

peiranliushop (approved):

PR #874, opened 2022-02-18
Ensure predeploy deploy timeouts actually raise a timeout error, not …

timothysmith0609:
Currently, the `Predeploying priority resources` phase can only raise a `FatalDeploymentError`. This is causing some pain during cluster bootstrap, as some priority resources are triggering timeout conditions. I don't see why we shouldn't add the ability for this phase to timeout, so I've added it, here.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 ## next
 
+*Bug fixes*
+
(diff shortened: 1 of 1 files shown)


jpfourny (approved):
Makes sense to me. This will require a krane bump in `gkrane-config` image to rollout. Let me know if you want me to take care of that for you.

Issue #873, opened 2022-02-07
Wrong CRD may be referenced by Custom Resource with duplicate Kind


timothysmith0609:
# Bug report

Currently, `DeployTask#discover_resources` is the following:

```ruby
    def discover_resources
      @logger.info("Discovering resources:")
      resources = []
      crds_by_kind = cluster_resource_discoverer.crds.group_by(&:kind)
      @template_sets.with_resource_definitions(current_sha: @current_sha, bindings: @bindings) do |r_def|
        crd = crds_by_kind[r_def["kind"]]&.first
        r = KubernetesResource.build(namespace: @namespace, context: @context, logger: @logger, definition: r_def,
          statsd_tags: @namespace_tags, crd: crd, global_names: @task_config.global_kinds)
        resources << r
        @logger.info("  - #{r.id}")
      end
```

For `crds_by_kind`, there is an error when the same `kind` appears in multiple groups. For example, if there are multiple `Certificate` resources, the `crd` passed into `KubernetesResource::build` will not match the underlying resource definition.

E.g.
```
(byebug) r_def["apiVersion"]
"stable.shopify.io/v1"
(byebug) crd.group
"cert-manager.io"
```

The solution is to group CRDs by their full GVK and match them with resource definitions that fully match it.

**Version(s) affected:** Observed in `krane 2.4.0`, but most likely present in all krane versions.

cc @jerr

PR #875, opened 2022-02-25
[Bug Fix] Grouping Cluster Resources by group and Kind

arunikadhal:
Fixes [#873](https://github.com/Shopify/krane/issues/873)

**How is this accomplished?**
Instead of grouping by `kind` in `DeployTask#discover_resources`, we now group by `group` and `kind` to avoid picking the wrong CRD when there are duplicates of the same `kind`. 

**What could go wrong?**
...

### DIFF
--- a/lib/krane/deploy_task.rb
+++ b/lib/krane/deploy_task.rb
@@ -240,9 +240,7 @@ module Krane
       resources = []
       crds_by_group_kind = cluster_resource_discoverer.crds.group_by(&:group_kind)
       @template_sets.with_resource_definitions(current_sha: @current_sha, bindings: @bindings) do |r_def|
-        grouping, version = r_def.dig("apiVersion").split("/")
-        group = version ? grouping : "core"
-        kind = r_def["kind"] ? r_def["kind"] : ""
+        group, kind = group_kind_for_r_def(r_def)
         crd = crds_by_group_kind[group + "/" + kind]&.first
         r = KubernetesResource.build(namespace: @namespace, context: @context, logger: @logger, definition: r_def,
           statsd_tags: @namespace_tags, crd: crd, global_names: @task_config.global_kinds)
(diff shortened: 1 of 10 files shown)


lib/krane/kubernetes_resource/custom_resource_definition.rb
```diff
@@ -41,9 +41,16 @@ def group_version_kind
       # New spec here: https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions/#create-a-customresourcedefinition
       # This is only used in testing, so we will simply take the first version we see:
       version = @definition.dig("spec", "versions", 0, "name")
+      kind = @definition.dig("spec", "names", "kind")
       "#{group}/#{version}/#{kind}"
     end
 
+    def group_kind
+      kind = @definition.dig("spec", "names", "kind")
```
peiranliushop, line 49: Could you reuse function `kind` below?

lib/krane/deploy_task.rb
```diff
@@ -238,9 +238,10 @@ def secrets_from_ejson
     def discover_resources
       @logger.info("Discovering resources:")
       resources = []
-      crds_by_kind = cluster_resource_discoverer.crds.group_by(&:kind)
+      crds_by_group_kind = cluster_resource_discoverer.crds.group_by(&:group_kind)
```
peiranliushop, line 241: Are there tests to cover this change?

d1egoaz: will be great to test this change, this is supposed to change the order of the items

timothysmith0609:
You'll also need to fix the CR logic in `global_deploy_task`, as well

lib/krane/deploy_task.rb
d1egoaz: yeah, sorry, I misread the issue in https://github.com/Shopify/krane/issues/873 and I thought a consequence of this PR is that it'll change the order of the items in some list, however, this change is about grouping differently. Sorry for the noise

arunikadhal:
> You'll also need to fix the CR logic in `global_deploy_task`, as well (and add test)

There seems to be a different issue in `global_deploy_task`, I have opened a new issue to fix the `kind` `name` discrepancy [here](https://github.com/Shopify/krane/issues/880).

PR #876, opened 2022-03-07
Resolve errors for StatefulSet restart with updateStrategy: OnDelete

stefanmb:

Resolve some logic errors in StatefulSet restart handling introduced in https://github.com/Shopify/krane/pull/840.

1. If a bare pod is present the current `delete_statefulset_pods` will crash. (See below)
1. The current success conditions for the restart cannot be met reliably due to:
   1.  [An unresolved bug in Kubernetes](https://github.com/kubernetes/kubernetes/pull/106059) that prevents the `currentRevision` from matching `updateRevision` 
   1.  For STS with `OnDelete` the `currentReplicas` field is sometimes nil.

stacktrace

**How is this accomplished?**
1. Introduce new test to trigger the `nil` error condition
1. Check for `nil` dereference
1. Remove readiness condition that is impossible to satisfy in STS with `OnDelete`

**What could go wrong?**
Until the [upstream fix](https://github.com/kubernetes/kubernetes/pull/106059) we'll still have to deploy with `--no-verify-result` for STS with `OnDelete`.

**Acknowledgments** 

Thanks to @epk for helping to debug these issues.


test/integration/restart_task_test.rb
```diff
@@ -36,7 +36,7 @@ def test_restart_by_annotation
   end
 
   def test_restart_statefulset_on_delete_restarts_child_pods
-    result = deploy_fixtures("hello-cloud", subset: "stateful_set.yml") do |fixtures|
+    result = deploy_fixtures("hello-cloud", subset: ["unmanaged-pod.yml", "stateful_set.yml"]) do |fixtures|
```
stefanmb, line 39: Deploying a bare pod without the fix produces the expected error:

```
Error:
RestartTaskTest#test_restart_statefulset_on_delete_restarts_child_pods:
NoMethodError: undefined method `find' for nil:NilClass
    /home/runner/work/krane/krane/lib/krane/restart_task.rb:216:in `block in delete_statefulset_pods'
    /opt/hostedtoolcache/Ruby/3.0.3/x64/lib/ruby/3.0.0/delegate.rb:349:in `select!'
    /opt/hostedtoolcache/Ruby/3.0.3/x64/lib/ruby/3.0.0/delegate.rb:349:in `block in delegating_block'
    /home/runner/work/krane/krane/lib/krane/restart_task.rb:215:in `delete_statefulset_pods'
    /home/runner/work/krane/krane/lib/krane/restart_task.rb:239:in `block in restart_statefulsets!'
    /home/runner/work/krane/krane/lib/krane/restart_task.rb:233:in `each'
    /home/runner/work/krane/krane/lib/krane/restart_task.rb:233:in `restart_statefulsets!'
    /home/runner/work/krane/krane/lib/krane/restart_task.rb:74:in `run!'
    /home/runner/work/krane/krane/lib/krane/restart_task.rb:49:in `run'
    /home/runner/work/krane/krane/test/integration/restart_task_test.rb:49:in `test_restart_statefulset_on_delete_restarts_child_pods'
```

lib/krane/kubernetes_resource/stateful_set.rb
```diff
@@ -27,11 +30,10 @@ def deploy_succeeded?
                        "Consider switching to rollingUpdate.")
           @success_assumption_warning_shown = true
         end
+      else
+        success &= desired_replicas == status_data['currentReplicas'].to_i
```
stefanmb, line 34: This property is not supported for STS with OnDelete:

```
kubectl rollout status sts/foo                                                                                                                                                                          
error: rollout status is only available for RollingUpdate strategy type
```

... therefore this condition can never be true in this case and krane will stall.

lib/krane/kubernetes_resource/stateful_set.rb
```diff
@@ -19,6 +19,9 @@ def status
     end
 
     def deploy_succeeded?
+      success = observed_generation == current_generation &&
+        desired_replicas == status_data['readyReplicas'].to_i &&
+        status_data['currentRevision'] == status_data['updateRevision'] &&
```
stefanmb, line 24: This check is currently broken in upstream Kubernetes, but if we remove the check the method will return immediately, before the pods are actually restarted. 

We could (re)implement some of the controller logic to check for individual pods but I believe running with --no-verify-result` until the upstream fix is the better option.

Upstream fix: https://github.com/kubernetes/kubernetes/pull/106059

lib/krane/restart_task.rb
```diff
@@ -213,7 +213,7 @@ def patch_daemonset_with_restart(record)
     def delete_statefulset_pods(record)
       pods = kubeclient.get_pods(namespace: record.metadata.namespace)
       pods.select! do |pod|
-        pod.metadata.ownerReferences.find { |ref| ref.uid == record.metadata.uid }
+        pod.metadata&.ownerReferences&.find { |ref| ref.uid == record.metadata.uid }
```
stefanmb, line 216: Fixes the `nil` errors encountered if bare pods are found in the namespaces. At Shopify nearly all app namespaces will have such a pod (e.g. secret seeder).

lib/krane/kubernetes_resource/stateful_set.rb
```diff
@@ -19,6 +19,9 @@ def status
     end
 
     def deploy_succeeded?
+      success = observed_generation == current_generation &&
+        desired_replicas == status_data['readyReplicas'].to_i &&
+        status_data['currentRevision'] == status_data['updateRevision']
```
stefanmb, line 24: This check is currently broken in upstream Kubernetes for `OnDelete`, but if we remove the check the method will return immediately, before the pods are actually restarted successfully.

We could (re)implement some of the controller logic to check for individual pods but I believe running with `--no-verify-result` until the upstream fix is the better option.

Upstream fix: https://github.com/kubernetes/kubernetes/pull/106059

test/integration/restart_task_test.rb
```diff
@@ -36,7 +36,8 @@ def test_restart_by_annotation
   end
 
   def test_restart_statefulset_on_delete_restarts_child_pods
-    result = deploy_fixtures("hello-cloud", subset: "stateful_set.yml") do |fixtures|
+    result = deploy_fixtures("hello-cloud", subset: ["configmap-data.yml", "unmanaged-pod-1.yml.erb",
+      "service-account.yml"], render_erb: true) do |fixtures|
```
stefanmb, line 40: Deploying a bare pod in the integration test without the fix produces the expected error:

```
Error:
RestartTaskTest#test_restart_statefulset_on_delete_restarts_child_pods:
NoMethodError: undefined method `find' for nil:NilClass
    /home/runner/work/krane/krane/lib/krane/restart_task.rb:216:in `block in delete_statefulset_pods'
    /opt/hostedtoolcache/Ruby/3.0.3/x64/lib/ruby/3.0.0/delegate.rb:349:in `select!'
    /opt/hostedtoolcache/Ruby/3.0.3/x64/lib/ruby/3.0.0/delegate.rb:349:in `block in delegating_block'
    /home/runner/work/krane/krane/lib/krane/restart_task.rb:215:in `delete_statefulset_pods'
    /home/runner/work/krane/krane/lib/krane/restart_task.rb:239:in `block in restart_statefulsets!'
    /home/runner/work/krane/krane/lib/krane/restart_task.rb:233:in `each'
    /home/runner/work/krane/krane/lib/krane/restart_task.rb:233:in `restart_statefulsets!'
    /home/runner/work/krane/krane/lib/krane/restart_task.rb:74:in `run!'
    /home/runner/work/krane/krane/lib/krane/restart_task.rb:49:in `run'
    /home/runner/work/krane/krane/test/integration/restart_task_test.rb:49:in `test_restart_statefulset_on_delete_restarts_child_pods'
```

test/integration/restart_task_test.rb
```diff
@@ -36,7 +36,8 @@ def test_restart_by_annotation
   end
 
   def test_restart_statefulset_on_delete_restarts_child_pods
-    result = deploy_fixtures("hello-cloud", subset: "stateful_set.yml") do |fixtures|
+    result = deploy_fixtures("hello-cloud", subset: ["configmap-data.yml", "unmanaged-pod-1.yml.erb",
```
stefanmb, line 39: Deploying a bare pod in the integration test without the fix produces the expected error:

```
Error:
RestartTaskTest#test_restart_statefulset_on_delete_restarts_child_pods:
NoMethodError: undefined method `find' for nil:NilClass
    /home/runner/work/krane/krane/lib/krane/restart_task.rb:216:in `block in delete_statefulset_pods'
    /opt/hostedtoolcache/Ruby/3.0.3/x64/lib/ruby/3.0.0/delegate.rb:349:in `select!'
    /opt/hostedtoolcache/Ruby/3.0.3/x64/lib/ruby/3.0.0/delegate.rb:349:in `block in delegating_block'
    /home/runner/work/krane/krane/lib/krane/restart_task.rb:215:in `delete_statefulset_pods'
    /home/runner/work/krane/krane/lib/krane/restart_task.rb:239:in `block in restart_statefulsets!'
    /home/runner/work/krane/krane/lib/krane/restart_task.rb:233:in `each'
    /home/runner/work/krane/krane/lib/krane/restart_task.rb:233:in `restart_statefulsets!'
    /home/runner/work/krane/krane/lib/krane/restart_task.rb:74:in `run!'
    /home/runner/work/krane/krane/lib/krane/restart_task.rb:49:in `run'
    /home/runner/work/krane/krane/test/integration/restart_task_test.rb:49:in `test_restart_statefulset_on_delete_restarts_child_pods'
```

lib/krane/restart_task.rb
timothysmith0609: 🤦 Thank you for fixing that

lib/krane/kubernetes_resource/stateful_set.rb
stefanmb: @timothysmith0609 Great question!

I had to look into this a bit. The brief answer is certain combinations of operations can cause the `currentReplicas` field become `nil`. The simple test as-is does not hit these conditions, so it passes.

I've pushed a version of this test that hangs (in the same way I saw in real life) here: https://github.com/Shopify/krane/commit/5c4acd050aa06205ae38706bb81a8fcfefad2bb8

You can observe locally this is the state this sequence of events leaves the stateful set in:

```
  status:
    collisionCount: 0
    currentRevision: stateful-busybox-6b5489d7d6
    observedGeneration: 2
    readyReplicas: 2
    replicas: 2
    updateRevision: stateful-busybox-96676bd54
    updatedReplicas: 2
```

Notice the absence of `currentReplicas`.

I did not push this version of the test here because it's hanging due to [the](https://github.com/Shopify/krane/pull/876/files#r821258485) so the only way to runt his test is with `verify_result: false` which doesn't make for a good test.

If you prefer, I can drop these changes from this PR since IRL I will have to run without verify result anyway until the previously mentioned bug is fixed.

timothysmith0609: >If you prefer, I can drop these changes from this PR since IRL I will have to run without verify result anyway for now.

May as well keep the changes for now, to avoid repeating the same work later.

PR #879, opened 2022-03-09
Cr group kind fixes

timothysmith0609:
Fixes the problem whereby custom resources can not discriminate between the same kind in different groups. This actually involved changes to `#predeploy_sequence` and `CustomResource(Definition)`

Note this doesn't solve the wider problem of multiple kinds among different standard API groups: that will require more changes than what's here

### DIFF
--- a/lib/krane/deploy_task.rb
+++ b/lib/krane/deploy_task.rb
@@ -74,7 +74,7 @@ module Krane
         Pod
       ).map { |r| [r, default_group] }
 
-      crs = cluster_resource_discoverer.crds.select(&:predeployed?).map { |cr| [cr.kind, { group: cr.group }] }
+      crs = cluster_resource_discoverer.crds.select(&:predeployed?).map { |cr| [cr.group_kind, { group: cr.group }] }
       Hash[before_crs + crs + after_crs]
     end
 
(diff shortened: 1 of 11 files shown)


lib/krane/deploy_task.rb
```diff
@@ -74,7 +74,7 @@ def predeploy_sequence
         Pod
       ).map { |r| [r, default_group] }
 
-      crs = cluster_resource_discoverer.crds.select(&:predeployed?).map { |cr| [cr.kind, { group: cr.group }] }
+      crs = cluster_resource_discoverer.crds.select(&:predeployed?).map { |cr| [cr.group_kind, { group: cr.group }] }
```
timothysmith0609, line 77: the redundancy of duplicating group information makes this fix much easier to implement, and is really just an internal consideration, anyway, so it felt like it was worth it.

lib/krane/deploy_task.rb
```diff
@@ -241,7 +241,7 @@ def discover_resources
       crds_by_group_kind = cluster_resource_discoverer.crds.group_by(&:group_kind)
       @template_sets.with_resource_definitions(current_sha: @current_sha, bindings: @bindings) do |r_def|
         group, kind = group_kind_for_r_def(r_def)
-        crd = crds_by_group_kind[group + "/" + kind]&.first
+        crd = crds_by_group_kind["#{kind}.#{group}"]&.first
```
timothysmith0609, line 244: I've adopted the convention of displaying group-kind information as `$KIND.$GROUP`, which mirrors how tools like kubectl, etc. work. e.g. `deployments.apps`

test/unit/krane/kubernetes_resource/custom_resource_definition_test.rb
```diff
@@ -96,6 +96,7 @@ def test_cr_instance_fails_validation_when_rollout_conditions_for_crd_invalid
       logger: @logger, statsd_tags: @statsd_tags, crd: crd,
       definition: {
         "kind" => "UnitTest",
+        "apiVersion" => "stable.example.io/v1",
```
timothysmith0609, line 99: `apiVersion` is required of any k8s resource, so it makes more sense to add it here than try to code around its absence

timothysmith0609 (commented):
The test fixes are just to deal with the fact that Custom Resources are now represented by `$kind.$group`, not just `kind`.

arunikadhal (approved):

PR #888, opened 2022-05-12
Use json format for kubectl version requests

KnVerey:
The "long" golang output format for `kubectl version` is being deprecated and removed. We should really be using json (or yaml) instead. https://github.com/kubernetes/kubernetes/pull/108987

**How is this accomplished?**
Updating the kubectl class and related tests.

**What could go wrong?**
If the JSON payload does not contain the expected keys, we'll raise a KubectlError. Previously, we would cache an empty hash. I think the explicit error is preferable to whatever would go wrong down the line with the empty hash.

### DIFF
--- a/test/unit/krane/ejson_secret_provisioner_test.rb
+++ b/test/unit/krane/ejson_secret_provisioner_test.rb
@@ -133,7 +133,7 @@ class EjsonSecretProvisionerTest < Krane::TestCase
   private
 
   def stub_server_dry_run_validation_request(attempts: 3)
-    stub_kubectl_response("apply", "-f", anything, "--server-dry-run", "--output=name",
+    stub_kubectl_response("apply", "-f", anything, "--dry-run=server", "--output=name",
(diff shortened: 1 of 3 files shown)


test/unit/krane/ejson_secret_provisioner_test.rb
```diff
@@ -133,7 +133,7 @@ def test_run_with_selector_does_not_raise_exception
   private
 
   def stub_server_dry_run_validation_request(attempts: 3)
-    stub_kubectl_response("apply", "-f", anything, "--server-dry-run", "--output=name",
+    stub_kubectl_response("apply", "-f", anything, "--dry-run=server", "--output=name",
```
KnVerey, line 136: This is because I changed the dummy data below to use a version > 1.18. Krane doesn't support 1.18 anymore, so the handling of the older flag could be removed entirely at this point (unrelated to this PR).

jpfourny (approved):

RRethy (approved):

PR #890, opened 2022-05-14
Fix `replace-force` deployment method override

jpfourny:
When trying to use `replace-force` value in `krane.shopify.io/deploy-method-override` annotation, `krane` will hit an error:

```
/usr/local/bundle/gems/krane-2.4.6/lib/krane/resource_deployer.rb:119:in `block in deploy_resources': Unexpected deploy method! (:"replace-force") (ArgumentError)
```

The logic in `kubernetes_resource.rb` is blindly trying to apply a method with symbol `replace-force`, but the method is actually `replace_force`. This PR attempts to fix this.

**How is this accomplished?**
Simply gsub `-` with `_` before converting the annotation value to a symbol.
Releasing version `2.4.7` of the GEM.

**What could go wrong?**

Nothing should go wrong.


tanner-bruce (approved):
LGTM, thanks for fixing it so quickly

jpfourny:
> LGTM, thanks for fixing it so quickly

It was easy enough. I will have to rollout `krane` into various other deploy tools next week before.

RRethy (approved):

PR #896, opened 2022-07-13
fix: initial commit for deploy fix when having duplicate kinds

### LINKED ISSUE #895: Krane is unable to fully deploy resources when there are duplicate Kinds
# Bug report
When a CRD installed with for kind "Deployment", Krane will be unable to deploy those resources if the scoping is different on the CRD. It will check the state of the wrong resource and in the wrong scope, cluster instead of namespace. 

**Expected behavior:** deployment to succeed.

**Actual behavior:** Krane constantly checks if deployment is succeeded but eventually times out.

**Version(s) affected:** probably all versions.

### Steps to Reproduce

1. Start a new KinD cluster.
2. Deploy the CRD in the collapsible below.
3. `kubectl create ns example`.
4. `DEBUG=1 bundle exec exe/krane deploy example kind-krane -f krane-manifest.yaml --global-timeout=1m`.
5. Krane does deploy the resource but will be unable to check the status. 

In debug mode you can see that it uses the wrong command to get the resource status:
```
kubectl get Deployment --chunk-size=0 --context=kind-kind-krane --output=json --request-timeout=30
```

<details>
  <summary>CRD Deployment</summary>
  
  ```yaml
---
apiVersion: apiextensions.k8s.io/v1
kind: CustomResourceDefinition
metadata:
  name: deployments.stable.example.com
spec:
  group: stable.example.com
  versions:
  - name: v1
    served: true
    storage: true
    schema:
      openAPIV3Schema:
        type: object
        properties:
          test:
            type: string
  scope: Cluster
  names:
    plural: deployments
    singular: deployment
    kind: Deployment
  ```
</details>

<details>
  <summary>Deployment</summary>
  
  ```yaml
---
kind: Deployment
apiVersion: apps/v1
metadata:
  name: nginx-deployment
spec:
  selector:
    matchLabels:
      deploy: example
  template:
    metadata:
      labels:
        deploy: example
    spec:
      containers:
        - name: nginx
          image: nginx
  ```
</details>

renescheepers:
- Dropped support for Ruby 2.7, has been EOL since 2019-12-25
- Dropped support for Kubernetes 1.19, has been EOL since 2021-10-28
- Fixed #895

### DIFF
--- a/lib/krane/api_resource.rb
+++ b/lib/krane/api_resource.rb
@@ -2,12 +2,14 @@
 
 module Krane
   class APIResource
-    attr_reader :group, :kind, :namespaced
+    attr_reader :group, :kind, :version, :namespaced, :verbs
 
-    def initialize(group, kind, namespaced)
+    def initialize(group, kind, version, namespaced, verbs)
       @group = group
       @kind = kind
+      @version = version
       @namespaced = namespaced
+      @verbs = verbs
     end
 
     def group_kind
--- a/lib/krane/cluster_resource_discovery.rb
+++ b/lib/krane/cluster_resource_discovery.rb
@@ -20,47 +20,29 @@ module Krane
 
     def prunable_resources(namespaced:)
       black_list = %w(Namespace Node ControllerRevision)
-      fetch_resources(namespaced: namespaced).map do |resource|
-        next unless resource["verbs"].one? { |v| v == "delete" }
-        next if black_list.include?(resource["kind"])
-        [resource["apigroup"], resource["version"], resource["kind"]].compact.join("/")
+      fetch_resources.map do |resource|
+        next unless namespaced == resource.namespaced
+        next unless resource.verbs.one? { |v| v == "delete" }
+        next if black_list.include?(resource.kind)
+
+        [
+          resource.group,
+          resource.version,
+          resource.kind,
+        ].compact.join("/")
       end.compact
     end
 
-    def fetch_resources(namespaced: false)
-      responses = Concurrent::Hash.new
-      Krane::Concurrency.split_across_threads(api_paths) do |path|
-        responses[path] = fetch_api_path(path)["resources"] || []
-      end
-      responses.flat_map do |path, resources|
-        resources.map { |r| resource_hash(path, namespaced, r) }
-      end.compact.uniq { |r| "#{r['apigroup']}/#{r['kind']}" }
-    end
-
-    def fetch_group_kinds
-      output, err, st = kubectl.run("api-resources", "--no-headers=true", attempts: 2, use_namespace: false)
-      if st.success?
-        output.split("\n").map do |l|
-          matches = l.scan(/\S+/)
-
-          if matches.length == 4
-            # name, api, namespaced, kind
-            ::Krane::APIResource.new(
-              ::Krane::KubernetesResource.group_from_api_version(matches[1]),
-              matches[3],
-              matches[2] == "true"
-            )
-          else
-            # name, shortname, api, namespaced, kind
-            ::Krane::APIResource.new(
-              ::Krane::KubernetesResource.group_from_api_version(matches[2]),
-              matches[4],
-              matches[3] == "true"
-            )
-          end
+    def fetch_resources
+      @fetch_resources ||= begin
+        responses = Concurrent::Hash.new
+        Krane::Concurrency.split_across_threads(api_paths) do |path|
+          responses[path] = fetch_api_path(path)["resources"] || []
         end
-      else
-        raise FatalKubeAPIError, "Error retrieving group kinds: #{err}"
+
+        responses.flat_map do |path, resources|
+          resources.map { |r| resource_hash(path, r) }
+        end.compact.uniq(&:group_kind)
       end
     end
 
@@ -114,23 +96,30 @@ module Krane
       end
     end
 
-    def resource_hash(path, namespaced, blob)
-      return unless blob["namespaced"] == namespaced
+    def resource_hash(path, blob)
       # skip sub-resources
       return if blob["name"].include?("/")
+
       path_regex = %r{(/apis?/)(?<group>[^/]*)/?(?<version>v.+)}
       match = path.match(path_regex)
-      {
-        "verbs" => blob["verbs"],
-        "kind" => blob["kind"],
-        "apigroup" => match[:group],
-        "version" => match[:version],
-      }
+      ::Krane::APIResource.new(
+        match[:group],
+        blob["kind"],
+        match[:version],
+        blob["namespaced"],
+        blob["verbs"]
+      )
     end
 
     def fetch_crds
-      raw_json, err, st = kubectl.run("get", "CustomResourceDefinition.apiextensions.k8s.io", output: "json", attempts: 5,
-        use_namespace: false)
+      raw_json, err, st = kubectl.run(
+        "get",
+        "CustomResourceDefinition.apiextensions.k8s.io",
+        output: "json",
+        attempts: 5,
+        use_namespace: false
+      )
+
       if st.success?
         JSON.parse(raw_json)["items"]
       else
--- a/lib/krane/deploy_task.rb
+++ b/lib/krane/deploy_task.rb
@@ -85,6 +85,7 @@ module Krane
       render_erb: false, kubeconfig: nil)
       @logger = logger || Krane::FormattedLogger.build(namespace, context)
       @template_sets = TemplateSets.from_dirs_and_files(paths: filenames, logger: @logger, render_erb: render_erb)
+
       @task_config = Krane::TaskConfig.new(context, namespace, @logger, kubeconfig)
       @bindings = bindings
       @namespace = namespace
@@ -211,14 +212,22 @@ module Krane
       @logger.info("Discovering resources:")
       resources = []
       crds_grouped = cluster_resource_discoverer.crds.group_by(&:cr_group_kind)
-      group_kinds = @task_config.group_kinds
+      api_resources = cluster_resource_discoverer.fetch_resources
 
       @template_sets.with_resource_definitions(current_sha: @current_sha, bindings: @bindings) do |r_def|
         group = ::Krane::KubernetesResource.group_from_api_version(r_def["apiVersion"])
 
         crd = crds_grouped[::Krane::KubernetesResource.combine_group_kind(group, r_def["kind"])]&.first
-        r = KubernetesResource.build(namespace: @namespace, context: @context, logger: @logger, definition: r_def,
-          statsd_tags: @namespace_tags, crd: crd, group_kinds: group_kinds)
+        r = KubernetesResource.build(
+          namespace: @namespace,
+          context: @context,
+          logger: @logger,
+          definition: r_def,
+          statsd_tags: @namespace_tags,
+          crd: crd,
+          api_resources: api_resources,
+        )
+
         resources << r
         @logger.info("  - #{r.pretty_id}")
       end
--- a/lib/krane/global_deploy_task.rb
+++ b/lib/krane/global_deploy_task.rb
@@ -160,15 +160,21 @@ module Krane
     def discover_resources
       logger.info("Discovering resources:")
       resources = []
-      group_kinds = @task_config.group_kinds
+      api_resources = cluster_resource_discoverer.fetch_resources
 
       crds_grouped = cluster_resource_discoverer.crds.group_by(&:group_kind)
       @template_sets.with_resource_definitions do |r_def|
         group = ::Krane::KubernetesResource.group_from_api_version(r_def["apiVersion"])
         crd = crds_grouped[::Krane::KubernetesResource.combine_group_kind(group, r_def["kind"])]&.first
 
-        r = KubernetesResource.build(context: context, logger: logger, definition: r_def,
-          crd: crd, group_kinds: group_kinds, statsd_tags: statsd_tags)
+        r = KubernetesResource.build(
+          context: context,
+          logger: logger,
+          definition: r_def,
+          crd: crd,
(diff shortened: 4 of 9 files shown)


lib/krane/cluster_resource_discovery.rb
```diff
@@ -37,12 +37,43 @@ def fetch_resources(namespaced: false)
...
+              "group_kind" => ::Krane::KubernetesResource.combine_group_kind(group, matches[3]),
+            }
+          else
+            group = ::Krane::KubernetesResource.group_from_api_version(matches[2])
+            # name, shortname, api, namespaced, kind
+            {
+              "namespaced" => matches[3] == "true",
+              "group" => group,
+              "kind" => matches[4],
+              "group_kind" => ::Krane::KubernetesResource.combine_group_kind(group, matches[4]),
```
timothysmith0609, line 62: Is it worth making a proper class (`APIResource`?) out of this, with something like a `::Parse` method. That way we can encapsulate the initialization as well as provide a way to unit test it.

test/fixtures/mutating_webhook_configurations/secret_hook.yaml
```diff
@@ -18,7 +18,7 @@ webhooks:
   reinvocationPolicy: Never
   rules:
   - apiGroups:
-    - core
+    - ""
```
timothysmith0609, line 21: Are these actually equivalent? Does `""` imply `core`. What broke that required this change?

lib/krane/deploy_task.rb
```diff
@@ -238,18 +210,22 @@ def secrets_from_ejson
     def discover_resources
       @logger.info("Discovering resources:")
       resources = []
-      crds_by_kind = cluster_resource_discoverer.crds.group_by(&:kind)
+      crds_by_kind = cluster_resource_discoverer.crds.group_by(&:cr_group_kind)
```
timothysmith0609, line 213: nit: `crds_by_kind` as a variable name is misleading now

lib/krane/kubernetes_resource.rb
```diff
@@ -44,38 +44,70 @@ class KubernetesResource
...
-      rescue NameError
-        nil
+        inst
       end
 
       def timeout
         self::TIMEOUT
       end
 
+      def group
```
timothysmith0609, line 78: The only edge case I see for `::group` and `::kind` is if we call it on `CustomResource`, in which case the result wouldn't make sense. I don't think that's going to be a problem in practice, though

timothysmith0609 (commented):
I need to go through the tests, but I'm on board with the solution in general.

At first I was apprehensive about explicitly specifying groups for the hardcoded resources, but on the whole I don't think it's a problem:

- They represent mature objects that are unlikely to radically change
- If they do change, we can simply define a new GroupKind and ride out the deprecation

It does tie our implementation to a specific set of kubernetes revisions, but as we already maintain a fairly tight skew, I don't see that as very problematic, overall.

renescheepers:
> I need to go through the tests, but I'm on board with the solution in general.
> 
> At first I was apprehensive about explicitly specifying groups for the hardcoded resources, but on the whole I don't think it's a problem:
> 
> * They represent mature objects that are unlikely to radically change
> * If they do change, we can simply define a new GroupKind and ride out the deprecation

If the group kind for an existing resources changes we could simply copy the class and extend the original one. 
> 
> It does tie our implementation to a specific set of kubernetes revisions, but as we already maintain a fairly tight skew, I don't see that as very problematic, overall.

This is already the case at the moment.

test/fixtures/mutating_webhook_configurations/secret_hook.yaml
renescheepers: I would need to create a copy of the Krane class specifically for this test. Since Secret.core isn't a valid resource anymore. This doesn't break for current master since it doesn't care about group anyways.

lib/krane/kubernetes_resource.rb
renescheepers: Why doesn't it make sense for the CR?

timothysmith0609: Because this method is generating its result based on _class name_ within `krane`. For custom resources this ends up being: `group(cr) == ""` and `kind(cr) == CustomResource`, but this isn't reflective of any actual object on the API-server.

Ultimately I don't think it's a big deal, but just wanted to point it out in case this violated an assumption somewhere

renescheepers: This only happens when you call it on the class, not on an actual instance of CustomResource.

jpfourny:
FYI: You will need to merge with this PR: https://github.com/Shopify/krane/pull/900
Version 2.4.9 will be cut from that, shortly.

test/integration/krane_deploy_test.rb
```diff
@@ -1826,6 +1826,14 @@ def test_succeeds_with_deploy_method_override
...
   end
 
+  def test_duplicate_kind_resource_definition
+    result = deploy_global_fixtures("crd", subset: ["deployment.yml"])
+    assert_deploy_success(result)
+
+    result = deploy_fixtures("crd", subset: ["web.yml"], global_timeout: 30)
+    assert_deploy_success(result)
+  end
+
```
timothysmith0609, line 1836: This test is passing just fine on master branch, I don't think it's forcing the failure you think it is

lib/krane/restart_task.rb
```diff
@@ -167,6 +167,12 @@ def identify_target_deployments(selector: nil)
         apps_v1_kubeclient.get_deployments(namespace: @namespace, label_selector: selector_string)
       end
       deployments.select { |d| d.dig(:metadata, :annotations, ANNOTATION) }
+      .map do |d|
+        d["apiVersion"] = "apps/v1"
+        d["kind"] = "Deployment"
```
KnVerey, line 172: why do we need to set/override these now?

lib/krane/deploy_task.rb
```diff
@@ -238,18 +210,22 @@ def secrets_from_ejson
     def discover_resources
       @logger.info("Discovering resources:")
       resources = []
-      crds_by_kind = cluster_resource_discoverer.crds.group_by(&:kind)
+      crds_grouped = cluster_resource_discoverer.crds.group_by(&:cr_group_kind)
+      group_kinds = @task_config.group_kinds
```
KnVerey, line 214: Within task config, you get this from an instance of cluster_resource_discoverer. Why not call it directly here? Since that class controls expensive operations, I'd think we should reuse a single instance as much as possible.

lib/krane/deploy_task.rb
```diff
@@ -238,18 +210,22 @@ def secrets_from_ejson
...
+
       @template_sets.with_resource_definitions(current_sha: @current_sha, bindings: @bindings) do |r_def|
-        crd = crds_by_kind[r_def["kind"]]&.first
+        group = ::Krane::KubernetesResource.group_from_api_version(r_def["apiVersion"])
+
+        crd = crds_grouped[::Krane::KubernetesResource.combine_group_kind(group, r_def["kind"])]&.first
         r = KubernetesResource.build(namespace: @namespace, context: @context, logger: @logger, definition: r_def,
-          statsd_tags: @namespace_tags, crd: crd, global_names: @task_config.global_kinds)
+          statsd_tags: @namespace_tags, crd: crd, group_kinds: group_kinds)
```
KnVerey, line 221: Same comment as in global_deploy_task about using cluster_resource_discoverer here instead of passing the list of group kinds to `build`.

KnVerey:
> [fix: all tests serial](https://github.com/Shopify/krane/pull/896/commits/55e59fd0c81b14647dcffaa820909741f40562c5)

I don't see this actually changing tests to run differently, but the commit message caught my eye. It's important that Krane be able to run multiple deploys currently, even when used as a gem. So just wanted to make super sure we are aware of that. 🙏

lib/krane/kubernetes_resource.rb
```diff
@@ -414,9 +455,12 @@ def self.go_template_for(kind, name)
...
           '(ne .reason "Pulled")',
         ]
-        condition_start = "{{if and #{and_conditions.join(' ')}}}"
+
+        and_conditions_string = and_conditions.map { |c| "{{if #{c}}}" }.join("")
+        ends = "{{end}}" * and_conditions.length
+
         field_part = FIELDS.map { |f| "{{#{f}}}" }.join(%({{print "#{FIELD_SEPARATOR}"}}))
-        %({{range .items}}#{condition_start}#{field_part}{{print "#{EVENT_SEPARATOR}"}}{{end}}{{end}})
+        %({{range .items}}#{and_conditions_string}#{field_part}{{print "#{EVENT_SEPARATOR}"}}#{ends}{{end}})
```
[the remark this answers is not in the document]
renescheepers: Yes, the older version of Kubectl (which we still support) are compiled with Go 1.17. Only since 1.18 it evaluates the statements correctly. 

https://tip.golang.org/doc/go1.18#:~:text=test.fuzzminimizetime.-,text/template,-Within%20a%20range

lib/krane/api_resource.rb
```diff
@@ -0,0 +1,17 @@
+# frozen_string_literal: true
+
+module Krane
+  class APIResource
```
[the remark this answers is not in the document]
renescheepers: Its only for containing the response of the api resources call, so I don't have to return a hash.

lib/krane/restart_task.rb
renescheepers: The response doesn't contain these values and we now depend on these values being set.

test/integration/krane_deploy_test.rb
renescheepers: Should be fixed now, the testing code applies a prefix to the Kind so in that case there is no duplicate.

lib/krane/kubernetes_resource.rb
```diff
@@ -44,38 +44,70 @@ class KubernetesResource
...
         self::TIMEOUT
       end
 
+      def group
+        # For example pods (Krane::Pod)
+        if name.scan(/::/).length == 1
+          return ""
+        end
+
+        # Converts Krane::ApiextensionsK8sIo::CustomResourceDefinition to apiextensions.k8s.io
```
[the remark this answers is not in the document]
renescheepers: The API version isn't always available, for example [here](https://github.com/Shopify/krane/blob/f0e0da7287947fa04756d834636e5081af748d22/lib/krane/kubernetes_resource/service.rb#L13). There is no instance available to fetch this information for.

lib/krane/kubernetes_resource.rb
```diff
@@ -44,38 +44,70 @@ class KubernetesResource
...
+            c_group.sub!(char, ".#{char.downcase}")
+          end
+        end
+
+        c_group
+      end
+
       def kind
-        name.demodulize
+        # Converts Krane::ApiextensionsK8sIo::CustomResourceDefinition to CustomResourceDefinition
```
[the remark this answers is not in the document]
renescheepers: Sometimes the class is called where is no instance is present. Here for example:

- https://github.com/Shopify/krane/blob/f0e0da7287947fa04756d834636e5081af748d22/lib/krane/kubernetes_resource/service.rb#L9
- https://github.com/Shopify/krane/blob/f0e0da7287947fa04756d834636e5081af748d22/lib/krane/kubernetes_resource/apps/deployment.rb#L8

lib/krane/global_deploy_task.rb
```diff
@@ -164,13 +160,17 @@ def validate_globals(resources)
...
+
+      crds_grouped = cluster_resource_discoverer.crds.group_by(&:group_kind)
       @template_sets.with_resource_definitions do |r_def|
-        crd = crds_by_kind[r_def["kind"]]&.first
+        group = ::Krane::KubernetesResource.group_from_api_version(r_def["apiVersion"])
+        crd = crds_grouped[::Krane::KubernetesResource.combine_group_kind(group, r_def["kind"])]&.first
+
         r = KubernetesResource.build(context: context, logger: logger, definition: r_def,
-          crd: crd, global_names: global_kinds, statsd_tags: statsd_tags)
+          crd: crd, group_kinds: group_kinds, statsd_tags: statsd_tags)
```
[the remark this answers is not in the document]
renescheepers: Yes, that does make sense. I tried to make the minimum amount of changes, so that most stays the same.

renescheepers:
> > [fix: all tests serial](https://github.com/Shopify/krane/pull/896/commits/55e59fd0c81b14647dcffaa820909741f40562c5)
> 
> I don't see this actually changing tests to run differently, but the commit message caught my eye. It's important that Krane be able to run multiple deploys currently, even when used as a gem. So just wanted to make super sure we are aware of that. 🙏

Yes, because the tests take a long time I went to each test-suite separately. One of those is `integration-serial` so while I didn't specifically touch any files there, I needed to do these fixes to have to specific tests pass.

lib/krane/api_resource.rb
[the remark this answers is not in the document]
KnVerey: Ah, in that case we shouldn't need it, because per my other comment we should be using the discovery information directly instead of re-making the calls via kubectl and parsing tabular output. But if that really isn't possible for some reason, I'd make this class private to the class that needs it, to avoid confusion.

lib/krane/restart_task.rb
KnVerey: Ah right, we got them from Kubeclient instead of kubectl here. Related issue: https://github.com/ManageIQ/kubeclient/issues/368

renescheepers:
Not planning on merging this pull request. This pull request was needed to resolve some issues that would happen when used in combination with Crossplane. However we stopped using Crossplane. Even though it still would be nice to fix, it is way too risky to deploy since this is used everywhere.

PR #887, opened 2022-04-17
Add Ruby 3.1.2 to CI.

simi:
**What are you trying to accomplish with this PR?**

Add support for Ruby 3 1.2.

**How is this accomplished?**

Updating GitHub Actions workflow to include Ruby 3.1.2 with latest kubectl.

**What could go wrong?**

Let's hope nothing :pray:.

PR #900, opened 2022-07-19
Update YAML.safe_load call to be compatible with Ruby 3.1

jhoos:

When krane is run under Ruby 3.1, and is invoked with an argument like `--bindings=@somefile`, it crashes, because the binding code is loading `somefile` using a deprecated call format for `YAML.safe_load` that is no longer valid in Ruby 3.1.

This is also causing unit test failures when the tests are run under Ruby 3.1.

**How is this accomplished?**

The `safe_load` call is fixed to use the current arguments, which are valid for Ruby 2.7 and later.

**What could go wrong?**

Probably nothing since there's test coverage for this. The PR also adds Ruby 3.1 to the GitHub action test matrix to cover this case.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -12,6 +12,7 @@ jobs:
       matrix:
         ruby:
         # Use unique Ruby versions, or GitHub gets confused when building the rest of the matrix
+        - '3.1.2' # With k8s 1.23
         - '3.0.3' # With k8s 1.23
         - '3.0.2' # With k8s 1.22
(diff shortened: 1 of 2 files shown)


jhoos:
This PR can replace #887, as it adds the same GitHub action configuration, and fixes the error that the author of that PR likely saw.

peiranliushop (approved):

jpfourny:
Are you going to release a new version?
See: https://github.com/Shopify/krane/blob/master/CONTRIBUTING.md#releasing-a-new-version-shopify-employees

PR #911, opened 2022-11-21
Kind fetcher locks are not fully thread-safe

mensfeld:
In case of extensive concurrent usage, the mutex handed over to two threads under same key may differ as illustrated below:

```ruby
require 'concurrent'

10000.times do
  kind_fetcher_locks = Concurrent::Hash.new { |hash, key| hash[key] = Mutex.new }

  refs = Set.new
  100.times.map do |i|
    Thread.new { refs << kind_fetcher_locks[i % 50].object_id }
  end.each(&:join)

  raise "Not 50 but #{refs.count}" unless refs.size == 50
end
```

this can lead to really weird issues.

Works when fixed as above:

```ruby
require 'concurrent'

10000.times do
  mutex = Mutex.new
  # kind_fetcher_locks = Concurrent::Hash.new { |hash, key| hash[key] = Mutex.new }

  kind_fetcher_locks = Concurrent::Hash.new do |hash, key|
    mutex.synchronize do
      break hash[key] if hash.key?(key)
      hash[key] = Mutex.new
    end
  end

  refs = Set.new
  100.times.map do |i|
    Thread.new { refs << kind_fetcher_locks[i % 50].object_id }
  end.each(&:join)

  raise "Not 50 but #{refs.count}" unless refs.size == 50
end
```

ref https://github.com/ruby-concurrency/concurrent-ruby/issues/970

...

**How is this accomplished?**
...

**What could go wrong?**
...

### DIFF
--- a/lib/krane/resource_cache.rb
+++ b/lib/krane/resource_cache.rb
@@ -8,8 +8,13 @@ module Krane
 
     def initialize(task_config)
       @task_config = task_config
-
-      @kind_fetcher_locks = Concurrent::Hash.new { |hash, key| hash[key] = Mutex.new }
+      @mutex = Mutex.new
+      @kind_fetcher_locks = Concurrent::Hash.new do |hash, key|
+        @mutex.synchronize do
+          break hash[key] if hash.key?(key)
+          hash[key] = Mutex.new
+        end
+      end
       @data = Concurrent::Hash.new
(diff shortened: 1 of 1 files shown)


mensfeld:
I don't have an option to trigger a re-run of CI, signed the CLA

PR #913, opened 2023-01-09
Optimize chance of users reading error messages

benlangfeld:

In our experience, it is common for the other elements of debug output (primarily pod logs) to be long enough that they fill the screen, pushing the part of the output that most needs to call attention passing into scrollback. It is also very common for users to not scroll back appropriately, assuming the important output about a failure will be at the bottom of the output.

**How is this accomplished?**

This moves the important information to the bottom to increase the chance that a user will read it.

**What could go wrong?**

Users might still not read log output.

### DIFF
--- a/lib/krane/kubernetes_resource.rb
+++ b/lib/krane/kubernetes_resource.rb
@@ -255,22 +255,6 @@ module Krane
 
     def debug_message(cause = nil, info_hash = {})
       helpful_info = []
-      if cause == :gave_up
-        debug_heading = ColorizedString.new("#{id}: GLOBAL WATCH TIMEOUT (#{info_hash[:timeout]} seconds)").yellow
-        helpful_info << "If you expected it to take longer than #{info_hash[:timeout]} seconds for your deploy"\
-        " to roll out, increase --global-timeout."
-      elsif deploy_failed?
-        debug_heading = ColorizedString.new("#{id}: FAILED").red
-        helpful_info << failure_message if failure_message.present?
-      elsif deploy_timed_out?
-        debug_heading = ColorizedString.new("#{id}: TIMED OUT (#{pretty_timeout_type})").yellow
-        helpful_info << timeout_message if timeout_message.present?
-      else
-        # Arriving in debug_message when we neither failed nor timed out is very unexpected. Dump all available info.
-        debug_heading = ColorizedString.new("#{id}: MONITORING ERROR").red
-        helpful_info << failure_message if failure_message.present?
-        helpful_info << timeout_message if timeout_message.present? && timeout_message != STANDARD_TIMEOUT_MESSAGE
-      end
 
       final_status = "  - Final status: #{status}"
       final_status = "\n#{final_status}" if helpful_info.present? && !helpful_info.last.end_with?("\n")
@@ -310,6 +294,23 @@ module Krane
             end
           end
         end
+      end
+
+      if cause == :gave_up
+        debug_heading = ColorizedString.new("#{id}: GLOBAL WATCH TIMEOUT (#{info_hash[:timeout]} seconds)").yellow
+        helpful_info << "If you expected it to take longer than #{info_hash[:timeout]} seconds for your deploy"\
+        " to roll out, increase --global-timeout."
+      elsif deploy_failed?
+        debug_heading = ColorizedString.new("#{id}: FAILED").red
+        helpful_info << failure_message if failure_message.present?
+      elsif deploy_timed_out?
+        debug_heading = ColorizedString.new("#{id}: TIMED OUT (#{pretty_timeout_type})").yellow
+        helpful_info << timeout_message if timeout_message.present?
+      else
+        # Arriving in debug_message when we neither failed nor timed out is very unexpected. Dump all available info.
+        debug_heading = ColorizedString.new("#{id}: MONITORING ERROR").red
+        helpful_info << failure_message if failure_message.present?
+        helpful_info << timeout_message if timeout_message.present? && timeout_message != STANDARD_TIMEOUT_MESSAGE
       end
 
       helpful_info.join("\n")


RRethy (commented):
> Users might still not read log output.

Exactly this.

I don't think changing the structure of log messages which have been in place for years without an evidence based reason is a good idea. It's hard to qualify some error messages as *important* since it's always situational.

benlangfeld:
> > Users might still not read log output.
> 
> Exactly this.
> 
> I don't think changing the structure of log messages which have been in place for years without an evidence based reason is a good idea. It's hard to qualify some error messages as _important_ since it's always situational.

I have evidence that this would significantly reduce the number of tickets I deal with from developers who don't know how far to scroll up to find the most important information. This is not situational conjecture, this is a concrete way to reduce the support load of users of this library. I don't believe that summary dismissal is justified.

peiranliushop:
hi @benlangfeld , since you mentioned that there are many support works caused by the existing logging structure, could you post some examples of logs and what is the proposed changes in the log, so we can compare the diff and evaluate whether it is beneficial? 

Since krane is a critical service, we don't roll out PR on a daily basis, but go through rigid procedure to release to ensure nothing breaks. Could you add more tests to verify the changed logs?

benlangfeld:
> since you mentioned that there are many support works caused by the existing logging structure, could you post some examples of logs and what is the proposed changes in the log, so we can compare the diff and evaluate whether it is beneficial?

Thanks for your question @peiranliushop . One example from today: https://gist.github.com/benlangfeld/0f75ba6e63b9a230695a6d12dc301dbe

Screenshot 2023-01-09 at 15 25 51
Screenshot 2023-01-09 at 15 25 44

If you believe other formatting modifications to be desirable, I'd be happy to prepare the appropriate implementation changes.

> Since krane is a critical service, we don't roll out PR on a daily basis, but go through rigid procedure to release to ensure nothing breaks. Could you add more tests to verify the changed logs?

Sure.

peiranliushop:
Are you using shipit to deploy? Is the red section `Deployment failed. Please classify failure reason` what you propose or is it something new in shipit? Can you give me the link to the UI?
You attached 2 images. What are the different between the 2 and which part is your proposed log change?

benlangfeld:
> Are you using shipit to deploy? Is the red section `Deployment failed. Please classify failure reason` what you propose or is it something new in shipit? Can you give me the link to the UI?

That red banner is a shipit plugin for gathering metrics on the reason for deployment failure. It is not open-source, since no-one has expressed any interest in using it.

> You attached 2 images. What are the different between the 2 and which part is your proposed log change?

There are two images because the logs are long enough to scroll for two full pages of the browser view. They are just two different scroll positions in order to demonstrate the poor UX.

My proposed log change is evident from the gist I linked to, which is a textual version of the same log as presented in the screenshots, a before and after version demonstrating the change.

benlangfeld:
> I compared the 2 log files, the only diff is the highlighted section that you want to move to the end of the debug message. I am not sure if it will help because its simply not so much logs. Screenshot 2023-01-09 at 3 48 29 PM

I think you misunderstood @peiranliushop . In this case, if the user had seen those logs, they would have resolved their issue for themselves. They didn't see those logs because users look at the *end* of the output and expect to see the most useful information there. Humans are lazy. My suggestion here is to put the most useful information where the user is most likely to see it and in so doing prevent them from opening useless tickets to which the answer is very often "scroll up to this specific point in the very long log output".

peiranliushop:
I see what you intend to solve.

From my experience, Krane deploy process is composed of the following these phases and they are logged in chronological order. Most relevant info is logged at the beginning of each phases and less important info / details follow through. This style is consistent among the phases.
```
Phase 1: Initializing deploy
Phase 2: Checking initial resource statuses
Phase 3: Predeploying priority resources
Phase 4: Deploying all resources
Result: SUCCESS or FAILED
```
My suggestion is to direct your support requests to a guide explaining how the deploy log is structure to help them find the info quickly.

benlangfeld:
@peiranliushop I find that as a rule, if one implementation option requires a manual, and another is self-evident, the self-evident one is better. Is your suggestion not to improve krane's logging format because you disagree that the current log format is deficient in usability or because you for some reason don't want to improve the usability of the tool? I really don't understand this resistance to improvement.

benlangfeld:
> > if one implementation option requires a manual, and another is self-evident, the self-evident one is better.
> 
> Determining which is which is subjective. I think the core difference of opinions here is due to the manner in which we use krane logs. From what I gather in your comments, you are showing users the end of the logs, and they are then not scrolling up to see the beginning of the error logs. Reference:

No, we're not actually auto-scrolling to the bottom of the log output. In my experience, when a human encounters several hundred lines of logs for some process they know to have failed, they immediately scroll to the end of the output expecting the thing that broke to be the last thing that happened and thus appear at the end of the log. Do you disagree with this observation?

> > They didn't see those logs because users look at the end of the output and expect to see the most useful information there. When they see nothing useful at the end, they throw their arms up and say they have no idea what's going on
> 
> We use krane logs different internally. When we have a deploy, we tell the users if it was successful or not along with a summary of the logs (e.g. which cluster had a failed deploy). We then offer the users a link to the logs served as a HTML page. Going to this page shows them the logs from the top, and lets them read them in a top-to-bottom/left-to-right manner. This means that the first error logs they see are those which are near the top, not the bottom.

I find it very hard to believe that anyone reads any substantial quantity of logs that have been filtered to include only the operation they are interested in (in this case, a single deployment process, but another example would be filtering webserver logs by client IP or similar) in a top-to-bottom fashion. I've never seen anyone actually do that in the real world. Do people actually do that, or is that just what you're expecting them to do?

> If we did merge this PR, it likely results in a worse user experience internally, but more importantly, it would result in a change to the log structure which we use many thousands of times per day.

I'm not sure I understand the distinction here, can you clarify?

> > because you for some reason don't want to improve the usability of the tool? I really don't understand this resistance to improvement.
> 
> This is getting close to breaking our [CODE_OF_CONDUCT](https://github.com/Shopify/krane/blob/c093e5f08b0c39fd0b9246d3f182090683213fff/CODE_OF_CONDUCT.md#our-standards) (if it hasn't already done so).

No, it doesn't break the code of conduct. The question is attempting to distinguish between disagreement that the proposal produces an actual improvement to the usability of the tool (which I very strongly believe it does, and until your points that I'm responding to at the beginning of this message it was not possible to determine that you disagreed with) and disagreement that improvements to the usability should be made (on the grounds, for example, of some tool you have which parses the krane logs and would break if the format were changed). What I wrote does not at all mean whatever you thought it meant that caused you to consult the code of conduct.

If you want to accuse me of anything, it's that I'm being insistent in the hope that you'll reverse your decision, and I'm doing so because I very strongly and genuinely believe that your decision on this proposed change does not prioritise the usability of the tool, and I was hoping that we could overcome the dismissiveness. I now see that the criteria by which suggestions are evaluated is other and I will no longer campaign for improvements.

> I'm going to repost examples of behaviour that contributes to creating a positive environment. I'm not going to lock this thread because I understand that people have bad days, but please keep these in mind for future discussions in our repos and in this thread or it will be locked.
> 
> ```
> * Using welcoming and inclusive language
> * Being respectful of differing viewpoints and experiences
> * Gracefully accepting constructive criticism
> * Focusing on what is best for the community
> * Showing empathy towards other community members
> ```

Indeed.

peiranliushop:
@benlangfeld

Hi Ben, thank you for bringing to our attention the logging issue you observed and offered a solution kindly. I also appreciate your patience and responses with detailed information over the past 2 days.

As mentioned in an early comment, we feel that the bottom position you suggested may not universally be more discoverable than right below the heading of deploy result. Also, if I read the code correctly, the proposed change in debug_message() may not get the effect you wanted in some cases. The function is called in iteration among all failed resources. If there is more than one failed or timed out resource, the message you want to be at the bottom will only be at the bottom for the very last resource.

That being said, I agree with you that finding relevant info in deploy logs is not easy, especially when the log outputs are long and embedded with events and errors. We're very much interested in your ideas for improving this, even if we don't agree with this solution in particular. It might help us to learn more about what in the logs is working well for you, and what is just noise your users always scroll past, so we can compare your experiences to our own. For example, are the logs for timeouts (as in your screenshot) specifically the main problem and are there other areas you think are ineffective? Are the logs overall useful? If you prefer to eliminate some noises in the logs, there is an undocumented [option](https://github.com/Shopify/krane/blob/c093e5f08b0c39fd0b9246d3f182090683213fff/lib/krane/kubernetes_resource.rb#L251-L253) to suppress pod events and logs via an environment variable. Do you think it is helpful? 

Rather than continuing the conversation here, I suggest to open an issue where we can discuss the root problem and explore potential solutions. For example, is it possible to have  a Shipit plugin to collapse lines [like Buildkite does](https://buildkite.com/docs/pipelines/managing-log-output) if Krane changed its output in a minor way? Should we add a brand new summary line at the very bottom? Should the logs include different debug information for timeouts vs failures?

Krane is an open source software and we count on everyone in the community to contribute in order to improve. We want all contributors to feel welcome, and I apologize for the frustrating experience you've had in this case. I hope my explanation is to your satisfaction and I am sincerely looking forward to your continued support.

PR #920, opened 2023-03-30
Update hpa version to v2 for kubernetes 1.25

micahvdk:
Trying to get krane to work with `1.25`

**How is this accomplished?**
Update the `kubectl_resource_type`

### DIFF
--- a/lib/krane/kubernetes_resource/horizontal_pod_autoscaler.rb
+++ b/lib/krane/kubernetes_resource/horizontal_pod_autoscaler.rb
@@ -16,7 +16,7 @@ module Krane
     end
 
     def kubectl_resource_type
-      'hpa.v2beta1.autoscaling'
+      'hpa.v2.autoscaling'
     end
 
     def status


lib/krane/kubernetes_resource/horizontal_pod_autoscaler.rb
```diff
@@ -16,7 +16,7 @@ def deploy_failed?
     end
 
     def kubectl_resource_type
-      'hpa.v2beta1.autoscaling'
+      'hpa.v2.autoscaling'
```
d1egoaz, line 19: I wonder if we can use `horizontalpodautoscalers.autoscaling` and avoid issues with the api version used.

d1egoaz (approved):
`horizontalpodautoscalers.autoscaling` it'll work, but looking at the history of the file https://github.com/Shopify/krane/pull/305 the reason to use `hpa.v2beta1.autoscaling` was that krane is really interested in the `v2beta1/autoscaling` condition statuses, which are also available in `v2/autoscaling`.

So, this PR looks ok to me as it is.

PR #921, opened 2023-04-10
Version 3.1.0

d1egoaz:
-------

Release a new version.

Fixes https://github.com/Shopify/krane/issues/918

**How is this accomplished?**
...

**What could go wrong?**
...


d1egoaz:
Some tests are failing due to the hpa change. I'll send a fix soon.

jpfourny (commented):
Should we add K8s 1.25 to the test matrix?

d1egoaz:
> Should we add K8s 1.25 to the test matrix?

Yes, but there are breaking changes in v1.25 that might need changes in Krane.
https://kubernetes.io/docs/reference/using-api/deprecation-guide/#v1-25

Those changes should be added in a separate PR.

PR #922, opened 2023-04-11
Fix test for hpa v2

d1egoaz:
Resource class was changed in https://github.com/Shopify/krane/pull/920.

But for some reason the CI tests weren't triggered in that PR and it was showing as passing all checks.

-------


kn100 (approved):
lgtm

PR #923, opened 2023-04-18
Remove the hard dependency on Oj

casperisfine:
`oj` frequently have stability issues, either crashes or memory leaks.

Based on the discussion in https://github.com/Shopify/krane/pull/390, I understand that krane parses lots of JSON, so instead of just using the stdlib I chose to use MultiJson so that it can leverage the faster JSON parser installed.

This way the Oj dependency isn't forced on the user.


orcephrye:
Perhaps I am confused but couldn't you change the add_dependency line to something like:
```
spec.add_dependency("oj", ">= 3.0")
```

casperisfine:
@orcephrye the goal here is to remove the dependency entirely, not relax it.

epk:
Sorry, I meant to rebase on master. 

Opened https://github.com/Shopify/krane/pull/923 instead

PR #926, opened 2023-06-02
Check for successful deploys of StatefulSets with an OnDelete update strategy

ayana-s:
## What are you trying to accomplish with this PR?
**🏗️ ISSUE: https://github.com/Shopify/krane/issues/912**

When checking for a successful deploy, if a StatefulSet's [updateStrategy](https://kubernetes.io/docs/concepts/workloads/controllers/statefulset/#update-strategies) is set to `OnDelete` (default is `RollingUpdate`), krane [does not check if the rollout is successful and instead prints a warning message](https://github.com/Shopify/krane/blob/master/lib/krane/kubernetes_resource/stateful_set.rb#L21-L37) ("can't monitor the update since it doesn't occur until deletion") and moves on.

**We want to enable krane to accurately detect successful deploys for StatefulSets with an `updateStrategy` of `OnDelete`.**

## How is this accomplished?
Update krane's `deploy_succeeded?` check to include StatefulSets with `updateStrategy: OnDelete`. 
* Currently, the method works for `updateStrategy: RollingUpdate` and checks that the following holds true:
```
observed_generation == current_generation
replicas == readyReplicas == currentReplicas
currentRevision == updateRevision
```
* Because of a [bug in Kubernetes](https://github.com/kubernetes/kubernetes/issues/106055), `onDelete`'s `currentRevision` value is not updated reliably. Also, `currentReplicas` relies on `currentRevision`, so neither of these values can be used.
* So, our added check will use `updatedReplicas` instead. If `updatedReplicas` doesn't match `replicas`, then pods have not been updated to the latest revision. `readyReplicas` can be used to check for their actual health.

As [confirmed over Slack](https://shopify.slack.com/archives/C05ABDX94GK/p1685974546899649), we're going with [Katrina's suggestion](https://github.com/Shopify/krane/issues/912#issuecomment-1341543337) of putting this behaviour behind an annotation to allow control on a per-resource basis using a style Krane users are already accustomed to. We're **extending the [`krane.shopify.io/required-rollout annotation`](https://github.com/Shopify/krane#customizing-behaviour-with-annotations) to be compatible with StatefulSet resources.**

Currently, for `Deployments` with annotations, the state of the `replicaSet` is checked to confirm a successful deploy. Because `StatefulSet` resources don't use `replicaSets` to create pods, we suggest only using `full` as an option for this annotation, with this definition:
- `full`: deployment is successful when all pods are ready

The logic we propose in this PR is as follows:
```
# For every successful deploy, check: 
observed_generation == current_generation
status_data['updateRevision'] == pod_data["metadata"]["labels"]["controller-revision-hash"]

# For updateStrategy: RollingUpdate, the rest of the checks remain the same: 
replicas == readyReplicas == currentReplicas && currentRevision == updateRevision

# For updateStrategy: OnDelete && required-rollout: full, check:
replicas == readyReplicas == updatedReplicas

# For updateStrategy: OnDelete with no specified required-rollout value, no additional checks are conducted. The code runs as it did previously.
```

## What should reviewers focus on?
1. Is this tested thoroughly enough?

## Testing
* Added unit tests. Existing CI & integration tests pass. (to be confirmed)
* Checked our clusters to ensure `updateRevision` (using `k -n [namespace] --context [context] get sts [namespace] -o yaml | tail -n 10`) and `controller-revision-hash` (using `k -n [namespace] --context [context] get pod [pod name] -o yaml | grep -i revision`) match after an update to our StatefulSet config (before the pod is marked as ready)
* Manually deployed a test runtime, nothing broken

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 ## next
 
+## 3.3.0
+
+*Enhancements*
+
+- Enable detection of successful StatefulSet deploys when the `updateStrategy` is `onDelete`, behind an annotation. [#926](https://github.com/Shopify/krane/pull/926)
+
 ## 3.2.0
 
 *Enhancements*
--- a/lib/krane/version.rb
+++ b/lib/krane/version.rb
@@ -1,4 +1,4 @@
 # frozen_string_literal: true
 module Krane
-  VERSION = "3.2.0"
+  VERSION = "3.3.0"
 end


lib/krane/kubernetes_resource/pod.rb
```diff
@@ -10,6 +10,7 @@ class Pod < KubernetesResource
     )
 
     attr_accessor :stream_logs
+    attr_reader :definition
```
ayana-s, line 13: Allows StatefulSet resource to access each Pod's `definition["metadata"]["labels"]["controller-revision-hash"]` value, a new addition to how we check for successful deploys

lib/krane/kubernetes_resource/stateful_set.rb
```diff
@@ -19,25 +22,34 @@ def status
     end
 
     def deploy_succeeded?
-      success = observed_generation == current_generation &&
-        desired_replicas == status_data['readyReplicas'].to_i &&
-        status_data['currentRevision'] == status_data['updateRevision']
+      success = observed_generation == current_generation
+
+      @pods.each do |pod|
+        success &= pod.definition["metadata"]["labels"]["controller-revision-hash"] == status_data['updateRevision']
```
ayana-s, line 28: This is a new check we're introducing that works for both updateStrategy types because a Pod's `controller-revision-hash` should update and match its StatefulSet's `updateRevision` value once the Pod is created (as opposed to waiting for the Pod to be ready)

test/fixtures/for_unit_tests/stateful_set_pod_test.yml
```diff
@@ -0,0 +1,98 @@
+apiVersion: v1
+kind: Pod
+metadata:
+  creationTimestamp: "2019-10-07T16:05:37Z"
+  generateName: test-ss-
```
stevenhope, line 5: nit: should this be `test-ss`?

ayana-s: good q! `generateName` is an optional prefix that's used to generate a name (in the absence of an actual `name` field), so we typically include a hyphen at the end. This is also done in the pod configs for the DaemonSet tests [(example)](https://github.com/Shopify/krane/blob/master/test/fixtures/for_unit_tests/daemon_set_pod.yml#L5)

lib/krane/kubernetes_resource/stateful_set.rb
```diff
@@ -19,25 +22,34 @@ def status
...
-        # Gem cannot monitor update since it doesn't occur until delete
         unless @success_assumption_warning_shown
           @logger.warn("WARNING: Your StatefulSet's updateStrategy is set to OnDelete, "\
-                       "which means updates will not be applied until its pods are deleted. "\
-                       "Consider switching to rollingUpdate.")
+                       "which means updates will not be applied until its pods are deleted.")
           @success_assumption_warning_shown = true
         end
+
+        if required_rollout == 'full'
```
sanaasy, line 38: can we use the constant you defined at the top of this file?
```suggestion
        if required_rollout == DEFAULT_REQUIRED_ROLLOUT
```

lib/krane/kubernetes_resource/stateful_set.rb
```diff
@@ -19,25 +22,34 @@ def status
...
+          success &= desired_replicas == status_data['readyReplicas'].to_i &&
+            desired_replicas == status_data['updatedReplicas'].to_i
+        end
       else
-        success &= desired_replicas == status_data['currentReplicas'].to_i
+        if required_rollout == 'full'
```
sanaasy, line 43:
```suggestion
        if required_rollout == DEFAULT_REQUIRED_ROLLOUT
```

lib/krane/kubernetes_resource/stateful_set.rb
```diff
@@ -19,25 +22,34 @@ def status
...
+          success &= status_data['currentRevision'] == status_data['updateRevision'] &&
+            desired_replicas == status_data['readyReplicas'].to_i &&
+            desired_replicas == status_data['currentReplicas'].to_i
```
sanaasy, line 46: can we do something like this instead?
```
success &= (status_data['currentRevision'] == status_data['updateRevision']) && 
(desired_replicas == status_data['readyReplicas'].to_i == status_data['currentReplicas'].to_i)
```
or did you opt to write it this way for better readability?

lib/krane/kubernetes_resource/stateful_set.rb
```diff
@@ -67,5 +79,9 @@ def parent_of_pod?(pod_data)
       pod_data["metadata"]["ownerReferences"].any? { |ref| ref["uid"] == @instance_data["metadata"]["uid"] } &&
       @instance_data["status"]["currentRevision"] == pod_data["metadata"]["labels"]["controller-revision-hash"]
     end
+
+    def required_rollout
+      krane_annotation_value("required-rollout") || DEFAULT_REQUIRED_ROLLOUT
```
sanaasy, line 84: is there a reason you chose to return the value instead of a boolean?

lib/krane/kubernetes_resource/stateful_set.rb
```diff
@@ -5,6 +5,9 @@ class StatefulSet < PodSetBase
     TIMEOUT = 10.minutes
     ONDELETE = 'OnDelete'
     SYNC_DEPENDENCIES = %w(Pod)
+    REQUIRED_ROLLOUT_ANNOTATION = "required-rollout"
+    REQUIRED_ROLLOUT_TYPES = %w(full none).freeze
```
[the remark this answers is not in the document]
ayana-s: yep! more specifically, the `freeze` method is what makes the array immutable -- the `%w` notation just creates the array of strings. we're using this syntax because most string arrays in krane seem to follow this pattern as well

lib/krane/kubernetes_resource/stateful_set.rb
sanaasy: oh actually you could do 

```
desired_replicas == (status_data['readyReplicas'].to_i == status_data['currentReplicas'].to_i)
```

but it's not as clear to other people who might look at the code and the readability level goes down with that 🤔

test/unit/krane/kubernetes_resource/stateful_set_test.rb
```diff
@@ -4,39 +4,90 @@
...
+    assert_predicate(ss, :deploy_succeeded?)
+  end
+
+  def test_deploy_does_not_succeed_when_replica_counts_do_not_match_for_rollingupdate_strategy_with_full_annotation
+    ss_template = build_ss_template(status: { "currentReplicas": 1 }, updateStrategy: "RollingUpdate", rollout: "full")
+    ss = build_synced_ss(ss_template: ss_template)
+    refute_predicate(ss, :deploy_succeeded?)
+  end
+
   def test_deploy_failed_ensures_controller_has_observed_deploy
```
null-sleep, line 49: Not sure whats being tested here

test/unit/krane/kubernetes_resource/stateful_set_test.rb
```diff
@@ -4,39 +4,90 @@
...
-    template = build_ss_template(status: { "observedGeneration": 1 })
-    ss = build_synced_ss(template: template)
+    ss_template = build_ss_template(status: { "observedGeneration": 1 })
+    ss = build_synced_ss(ss_template: ss_template)
     refute_predicate(ss, :deploy_succeeded?)
   end
 
-  def test_deploy_failed_not_fooled_by_stale_status
+  def test_deploy_failed_not_fooled_by_stale_status_for_rollingupdate_strategy
```
null-sleep, line 55: 👍

lib/krane/kubernetes_resource/stateful_set.rb
```diff
@@ -19,25 +22,34 @@ def status
...
         unless @success_assumption_warning_shown
           @logger.warn("WARNING: Your StatefulSet's updateStrategy is set to OnDelete, "\
-                       "which means updates will not be applied until its pods are deleted. "\
-                       "Consider switching to rollingUpdate.")
+                       "which means updates will not be applied until its pods are deleted.")
           @success_assumption_warning_shown = true
         end
+
+        if required_rollout == 'full'
+          success &= desired_replicas == status_data['readyReplicas'].to_i &&
```
null-sleep, line 39: We should certainly leave a comment with context as to why this is different. We can just like this issue tbh: https://github.com/Shopify/krane/issues/912

lib/krane/kubernetes_resource/stateful_set.rb
null-sleep: You can also do
```
success &= status_data['currentRevision'] == status_data['updateRevision']
success &= desired_replicas == status_data['readyReplicas'].to_i
success &= desired_replicas == status_data['currentReplicas'].to_i
```
No strong opinions

test/unit/krane/kubernetes_resource/stateful_set_test.rb
ayana-s: good catch - this was an existing test to confirm that `deploy_succeeded?` is **not** true when current and observed generations aren't equal. Will update the test name to be clearer

lib/krane/kubernetes_resource/stateful_set.rb
stevenhope: > I am guessing the pod definitions are updated each time the method is called?

Hey Dhruv, to clarify, are you asking to confirm if this method updates the pod definitions?

ayana-s: Yep, I believe so -- @stevenhope found that pod definitions should be updated each time `deploy_succeeded?` is called due to the [resource_watcher](https://github.com/Shopify/krane/blob/master/lib/krane/resource_watcher.rb#L34) running sync at an interval until it succeeds.

null-sleep (approved):

renescheepers (approved):

vvuibert:
blocked on https://github.com/Shopify/krane/pull/929

PR #934, opened 2023-08-04
Test against Kubernetes v1.27

cjoudrey:
As the PR title says, this modifies `ci.yml` so that we start testing `krane` against Kubernetes v1.27.


cjoudrey:
cc @RRethy for review when you have a minute.

PR #940, opened 2023-12-08
s/whitelist/allowlist

### LINKED ISSUE #917: kubectl 1.26 flag changes
In kubectl v1.26, the `--prune-whitelist` flag is deprecated in favour of `--prune-allowlist`. The old flag will be removed in v1.28 at the earliest. Krane's `kubectl apply` invocation will need to be adjusted accordingly when the relevant versions enter its supported skew.

Ref https://github.com/kubernetes/kubernetes/pull/113116
(FWIW the [ApplySet proposal](https://github.com/kubernetes/enhancements/tree/master/keps/sig-cli/3659-kubectl-apply-prune) will ultimately replace these flags entirely if it succeeds, but that is even further away)

timothysmith0609:
Closes #917 

**How is this accomplished?**
global search and replace all whitelist references and change to allowlist

**What could go wrong?**
From what I understand, `--prune-allowlist` is only available in 1.26+, so any clusters running older versions will not have access. This necessitates, at the very least, a minor version bump. I've added compatibility code so older k8s versions will still work with the old `--prune-whitelist` flag

**Note: Failing CI is due to flaky tests: I'm slowly rerunning them one-by-one to get 🟢**

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 # 3.4.0
 
-- Use `prune-allowlist` instead of `prune-whitelist`. This change requires kubernetes 1.26 and higher. [#940](https://github.com/Shopify/krane/pull/940)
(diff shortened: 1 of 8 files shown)


DazWorrall (approved):
LGTM

PR #943, opened 2024-01-19
Batch dry run but like for real this time

timothysmith0609:
Part of https://github.com/Shopify/infrastructure-tooling/issues/686

Related: https://github.com/Shopify/krane/issues/797

https://github.com/Shopify/krane/pull/781 implemented batched dry-run `kubectl apply` as part of the resource validation phase. For what I'm sure was a compelling reason at the time, if batched dry-run fails, we fallback to doing a resource-by-resource check. Unfortunately for larger resource sets, this means we end up making `N` `kubectl` calls as we individually check each resource: the very issue https://github.com/Shopify/krane/pull/781 was meant to solve.

It turns out `batch_dry_run` pathologically fails because we were attempting to `apply` non-applyable resources (e.g. Pods with `generateName`, etc.). The solution is to partition resources into applyable/non-applyable and batch dry-run only the former. A quick top hat proves this works.

With that out of the way, the next thing we should do is just trust the batch results and raise a `FatalDeploymentError` if it doesn't succeed.

### DIFF
--- a/lib/krane/resource_deployer.rb
+++ b/lib/krane/resource_deployer.rb
@@ -162,7 +162,7 @@ module Krane
           @prune_allowlist.each { |type| command.push("#{allow_list_flag}=#{type}") }
         end
 
-        command.push(kubectl.dry_run_flag) if dry_run
+        command.push("--dry-run=client") if dry_run
         output_is_sensitive = resources.any?(&:sensitive_template_content?)
         global_mode = resources.all?(&:global?)
         out, err, st = kubectl.run(*command, log_failure: false, output_is_sensitive: output_is_sensitive,


lib/krane/deploy_task.rb
```diff
@@ -285,17 +284,21 @@ def validate_configuration(prune:)
...
-      validate_globals(resources)
-      batch_dry_run_success = validate_dry_run(resources)
       resources.select! { |r| r.selected?(@selector) } if @selector_as_filter
-      Krane::Concurrency.split_across_threads(resources) do |r|
+      validate_globals(resources)
+      applyables, individuals = resources.partition { |r| r.deploy_method == :apply }
+      batch_dry_run_success = validate_dry_run(applyables)
+      resources_to_validate = batch_dry_run_success ? individuals : resources
+
+      Krane::Concurrency.split_across_threads(resources_to_validate) do |r|
```
timothysmith0609, line 293: Originally I wanted to to just trust the output of the batched dry run test and raise a `FatalDeploymentError`. However, that made it **very** difficult to give a consistent UX with regards to giving useful feedback: the main issue being it became necessary to suppress all feedback if a sensitive document/secret was contained in the apply set.

To that end, I've kept things pretty similar, just removing resources from the per-resource checks if batched dry-run succeeds. If it fails, since that's a less common case, we default to per-resource checking in order to provide more granular feedback and highlight what the actual issue it. This is the difference between "Something somewhere went wrong" and "THIS particular file and THIS particular line are wrong".

PR #945, opened 2024-01-29
smaller batch dry run blast

timothysmith0609:
Replaces #943 

- Shrinks down the size of the batched apply set (to a tiny amount, in reality)
- All other kubectl validation is done client side and should be very fast. If necessary let's try to batch the client side validations, but that requires a big refactor


timothysmith0609:
Replaced by #946

PR #950, opened 2024-04-02
Fix StatefulSet#deploy_succeeded? for OnDelete

JamesOwenHall:
In #926, a check was added to `StatefulSet#deploy_succeeded?` that requires all pods belonging to a StatefulSet to have the `controller-revision-hash` label set to the StatefulSet's `updateRevision` status. It ensures that all of the StatefulSet's pods have been updated to the StatefulSet's latest revision. This is fine for StatefulSets with `updateStrategy: RollingUpdate`, but doesn't work when it's `OnDelete` and we don't want to wait for some other mechanism to roll the pods. 

We're wrong to run this check for all StatefulSets.

**How is this accomplished?**
I've removed the check entirely since the StatefulSet's status should be sufficient to monitor rollout status. 

**What could go wrong?**
...

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,20 +1,24 @@
 ## next
 
-# 3.5.0
+## 3.5.1
+
+- Fixed successful deployment check for StatefulSets introduced in v3.3.0.
+
+## 3.5.0
 
 - Test against k8s 1.28
 - Drop support for k8s 1.23
 
-# 3.4.2
+## 3.4.2
 
 - Remove flag `--skip-dry-run` (see [#946](https://github.com/Shopify/krane/pull/946))
 - Remove support for batched server-side dry-run ([#946](https://github.com/Shopify/krane/pull/946))
 
-# 3.4.1
+## 3.4.1
 
 - Added flag `--skip-dry-run` to completely opt out of dry run validation.
 
-# 3.4.0
+## 3.4.0
 
 - Use `prune-allowlist` instead of `prune-whitelist` for 1.26+ clusters. Clusters running 1.25 or less will continue to use `--prune-whitelist`. [#940](https://github.com/Shopify/krane/pull/940)
 
--- a/lib/krane/version.rb
+++ b/lib/krane/version.rb
@@ -1,4 +1,4 @@
 # frozen_string_literal: true
 module Krane
-  VERSION = "3.5.0"
+  VERSION = "3.5.1"
 end


lib/krane/kubernetes_resource/stateful_set.rb
```diff
@@ -22,26 +22,16 @@ def status
     def deploy_succeeded?
       success = observed_generation == current_generation
 
-      @pods.each do |pod|
-        success &= pod.definition["metadata"]["labels"]["controller-revision-hash"] == status_data['updateRevision']
-      end
```
JamesOwenHall, line 27: This is the check I've removed because it's not needed as far as I can tell. This previously passed in some of our deploys due to https://github.com/kubernetes/kubernetes/issues/106055 not reliably setting `.status.currentRevision` properly.

I don't believe it should be necessary to look up the pods. The StatefulSet's status should be enough.

test/unit/krane/kubernetes_resource/stateful_set_test.rb
```diff
@@ -80,7 +81,7 @@ def build_ss_template(status: {}, updateStrategy: "RollingUpdate", rollout: "ful
     ss_fixture.dup.deep_merge(
       "status" => status,
       "spec" => {"updateStrategy" => {"type" => updateStrategy}},
-      "metadata" => {"annotations" => {"krane.shopify.io/rollout" => rollout}}
```
JamesOwenHall, line 83: This annotation wasn't set correctly.

lib/krane/kubernetes_resource/stateful_set.rb
```diff
@@ -22,26 +22,16 @@ def status
...
+                       "which means updates will not be applied until its pods are deleted.")
           @success_assumption_warning_shown = true
         end
-
-        if required_rollout == 'full'
-          success &= desired_replicas == status_data['readyReplicas'].to_i
-          success &= desired_replicas == status_data['updatedReplicas'].to_i
-        end
+      else
+        success &= status_data['currentRevision'] == status_data['updateRevision']
```
ayana-s, line 32: Should we explicitly check for RollingUpdate when requiring this condition? If I'm understanding correctly, this check would currently still be applied to OnDelete where `required_rollout == "full"`, but as you [mentioned above](https://github.com/Shopify/krane/pull/950/files#r1548064091), `currentRevision` [isn't reliably set](https://github.com/kubernetes/kubernetes/issues/106055) for the OnDelete strat

lib/krane/kubernetes_resource/stateful_set.rb
```diff
@@ -22,26 +22,15 @@ def status
...
-        success &= pod.definition["metadata"]["labels"]["controller-revision-hash"] == status_data['updateRevision']
-      end
-
-      if update_strategy == 'RollingUpdate'
-        success &= status_data['currentRevision'] == status_data['updateRevision']
-        success &= desired_replicas == status_data['readyReplicas'].to_i
-        success &= desired_replicas == status_data['currentReplicas'].to_i
-
-      elsif update_strategy == ONDELETE
+      if update_strategy == ONDELETE && required_rollout != "full"
```
timothysmith0609, line 25: Are there known cases of `update_strategy == DELETE` AND `required_rollout == "full"`? If those conditions hold, `krane` will time out unless something external deletes the pods, right? I suppose that's not a problem if that's what's been encoded in the manifest since that would signal that as the desired behaviour 🤔

ayana-s: I do know of [one such case](https://github.com/Shopify/kafka-infrastructure/blob/main/config/deploy/partials/kafka.yml.erb#L130) in `kafka-infrastructure`, where [this script](https://github.com/Shopify/zonal-deploy-controller/blob/main/pkg/controllers/statefulset_controller.go) in `zonal-deploy-controller` is used to delete the pods. I don't know that this change would break anything there, but tagging @Shopify/streaming-foundations for visibility

timothysmith0609 (approved):

ayana-s (approved):
I haven't tophatted (unsure if it's useful to at this point -- if a tophat is necessary, maybe it'd just involve manually deploying a test runtime and deleting/recreating an existing StatefulSet to confirm that changes introduced here aren't breaking changes?), but the code & logic look sound - thanks for fixing this, James!

PR #954, opened 2024-04-19
Fix statefulset pod selector

JamesOwenHall:
When deploying a StatefulSet, deploys could sporadically be marked failed due to Krane monitoring the health of the current revision's pods, not the updated revision. This PR updates the logic to monitor the updated revision.

**How is this accomplished?**
`StatefulSet#parent_of_pod?` is the method that filters pods. I've updated it to select pods matching `.status.updateRevision`.

**What could go wrong?**
...

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 ## next
 
+## 3.5.2
+
+- Fixed an issue where deploying StatefulSets monitored the health of the previous revision's pods instead of the updated one.
+
 ## 3.5.1
 
 - Fixed successful deployment check for StatefulSets introduced in v3.3.0.
--- a/lib/krane/version.rb
+++ b/lib/krane/version.rb
@@ -1,4 +1,4 @@
 # frozen_string_literal: true
 module Krane
-  VERSION = "3.5.1"
+  VERSION = "3.5.2"
 end


ilikeorangutans (approved):

lib/krane/kubernetes_resource/stateful_set.rb
```diff
@@ -65,7 +65,7 @@ def desired_replicas
     def parent_of_pod?(pod_data)
       return false unless pod_data.dig("metadata", "ownerReferences")
       pod_data["metadata"]["ownerReferences"].any? { |ref| ref["uid"] == @instance_data["metadata"]["uid"] } &&
-      @instance_data["status"]["currentRevision"] == pod_data["metadata"]["labels"]["controller-revision-hash"]
+      @instance_data["status"]["updateRevision"] == pod_data["metadata"]["labels"]["controller-revision-hash"]
```
devonboyer, line 68: I'm trying to figure out whether `updateRevision` will be empty the first time a `StatefulSet` is deployed.

devonboyer: I think it does the right thing:

https://github.com/kubernetes/kubernetes/blob/master/pkg/controller/statefulset/stateful_set_control.go#L271

JamesOwenHall:
🎩 'ed and seems to have fixed the issue.

PR #969, opened 2024-12-04
Allow more resources to be predeployed

c-gerke:
### What are you trying to accomplish with this PR?
This PR aims to allow operators of the Krane gem to designate more types of resources as being able to be "predeployed" than just CRs and CRDs, as was the previous behavior of the "predeployed" logic.

### How is this accomplished?
There are a few changes that have been made to facilitate this:

#### Documentation Updates:
* `CONTRIBUTING.md`: Added instructions for implementing the `predeployed?` method in custom Kubernetes resource classes and updated the references for adding new resources
* `README.md`: Expanded the description of the `krane.shopify.io/predeployed` annotation to include more resource types and clarified its default behavior

#### Code Enhancements:
* `lib/krane/kubernetes_resource.rb`: Introduced a `PREDEPLOYED_RESOURCE_TYPES` constant to hold all of the resources that were previously statically predeployed, and a `predeployed?` method to determine if a resource should be predeployed based on its type or annotation, which is inherited by all `lib/krane/kubernetes_resource` files
* `lib/krane/kubernetes_resource/custom_resource.rb`: Added a `predeployed?` method to handle custom resources, defaulting to true, similar to that in [`custom_resource_definition.rb`](https://github.com/Shopify/krane/blob/main/lib/krane/kubernetes_resource/custom_resource_definition.rb#L60-L63)
* `lib/krane/deploy_task.rb`: Updated the `predeploy_sequence` method to include additional resource types in the predeploy sequence

#### Deployment Process:
* `lib/krane/resource_deployer.rb`: Modified the `predeploy_priority_resources` method to filter resources based on the `predeployed?` method (see questions section below for a clarification request)
* `lib/krane/resource_deployer.rb`: Added a StatsD measurement for the sync duration in the `deploy_resources` method

#### Tests:
* `test/helpers/mock_resource.rb`: Modified to allow the test MockResource to be able to be predeployed
* `test/integration-serial/serial_deploy_test.rb`: Added multiple tests to verify that resources with the `krane.shopify.io/predeployed` annotation are correctly predeployed
* `test/unit/krane/resource_deployer_test.rb`: Modified unit tests to ensure the `predeploy_priority_resources` method respects the predeploy list and the `predeployed?` annotation, and that the unit tests know how to operate in a universe where the `krane.shopify.io/predeployed` annotation can be deployed to almost any resource

### What could go wrong?
This could allow operators of Krane to deploy resources out of sequence, causing deployments to fail and requiring troubleshooting.

An example of this failure mode would be adding the annotation to a `Service`, causing it to be predeployed before a `Deployment` that the `Service` would target. In this case where the `krane.shopify.io/predeployed: "true"` annotation is set on the `Service` without something like the `krane.shopify.io/skip-endpoint-validation: "true"` annotation, the `Service` will fail citing there are no endpoints for the `Service` and the deployment is likely invalid.

In essence, most of the observed "cons" are giving operators of Krane a gun that could be used to shoot themselves in the foot.

#### `lib/krane/resource_deployer.rb` and `lib/krane/deploy_task.rb`
In `lib/krane/resource_deployer.rb`, we are getting all of the resources and filtering afterwards for those with the `predeployed` annotation. <sup>[1](https://github.com/powerhome/krane/pull/5/files#diff-dbe1eb9fcee116e148d1f7634d0a6b8a42c7863dd3e103b73ecf220f198a4cd9R52-R53)</sup> This is a different behavior than that for CRs in [`lib/krane/deploy_task.rb`](https://github.com/Shopify/krane/blob/main/lib/krane/deploy_task.rb#L77) because we don't have access in `lib/krane/deploy_task.rb` at that point in the process to read the files and determine if they need to be predeployed.

This causes us to always hit 'phase 3 - predeploy' of the deployment process (thus the change in line 347 of `test/integration-serial/serial_deploy_test.rb`), which seems a little inefficient. Is there a better way that we should do this so we can skip hitting 'phase 3 - predeploy' if there aren't any resources that will need to be predeployed?

### DIFF
--- a/1.0-Upgrade.md
+++ b/1.0-Upgrade.md
@@ -7,7 +7,7 @@
 * There are breaking changes in the public API (such as the renaming of the `KubernetesDeploy` namespace to `Krane`, and the change in default values for different arguments of the public interface).
 * StatsD metrics will now be generated with the `krane` prefix.
 * `krane deploy` now considers all namespaced resources eligible for pruning, including
-custom resources. See [blacklist](https://github.com/Shopify/krane/blob/main/lib/krane/cluster_resource_discovery.rb#L22) for exceptions.
+custom resources. See [blacklist](https://github.com/Shopify/krane/blob/main/lib/krane/cluster_resource_discovery.rb#L20) for exceptions.
 * `kubernetes-deploy` (now `krane deploy`) / `DeployTask` can no longer deploy global (non-namespaced) resources. A new command called `krane global-deploy` and a related class called `GlobalDeployTask` were added to replace that feature.
 * `krane deploy` will not render erb templates. Use `krane render | krane deploy --stdin` to reproduce this functionality.
 * If you attempt to install two gems that have conflicting executables, `gem install` will warn you but the most recently installed one will win.
--- a/README.md
+++ b/README.md
@@ -167,7 +167,7 @@ before the deployment is considered successful.
     - Percent (e.g. 90%): The deploy is successful when the number of new pods that are ready is equal to `spec.replicas` * Percent.
   - _Compatibility_: StatefulSet
     - `full`: The deployment is successful when all pods are ready.
-- `krane.shopify.io/predeployed`: Causes a Custom Resource or [other resources](https://github.com/powerhome/krane/tree/main/lib/krane/kubernetes_resource) to be deployed in the pre-deploy phase.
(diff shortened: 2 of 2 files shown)


timothysmith0609:
Hi @c-gerke is this PR ready for a proper review? Thanks for the contribution!

PR #972, opened 2025-01-15
Set last-applied-label during create or replace operation atomically

timothysmith0609:
For resources that are BOTH 
- Created via `create` and/or `replace`
- Prunable

`krane` currently first issues the `create/replace` command, then subsequently calls `apply`. This is done to ensure ongoing cleanup of resources like DB-Migrate pods that would otherwise just hang around.

Unfortunately, this subsequent `apply` breaks if a mutating admission controller modified an otherwise immutable field. The document presented by `krane` now conflicts with the API server and the deploy fails.

Thankfully, `kubectl` supports adding the `last-applied-configuration` annotation (which is the signal used by k8s to marks resources as **prunable**). Using the exact same logic we were using to add individual resources to the `applyables` list, we can just conditionally set the `--save-config` flag on `kubectl create/replace`


jpfourny (approved):
Nice!

jpfourny (approved):

assuntaw (approved):

PR #974, opened 2025-02-11
Add support for --partials-dir

frison:
Prior to this change, partials were discovered by first checking `./partials` and then `../partials`. However, we can unlock easier sharing of partials with the ability to provide your own partials directory that is checked first.

The ability to include partials from non-standard/hard-coded partial paths (ie: `./partials` and `../partials`) via a command-line parameter of `--partials-dir`. 

**How is this accomplished?**

When a `--partials-dir` is included, it's prepended to the `@partials_dirs` in the `Renderer`, and consequently is used first (if the partials are present). 

**What could go wrong?**

🤔  -- the test coverage in this area is pretty good, and I've maintained all existing test behaviours. I think the worst case, is failing to render templates.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,7 @@
 ## next
 
+- Add a `--partials-dir` parameter. When provided, this path will be check for partials first, prior to both `./partials` and `../partials` being checked.
+
 ## 3.7.0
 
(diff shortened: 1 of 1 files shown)


jpfourny:
`3.7.1` should be released to rubygems.org soon.

PR #976, opened 2025-02-28
All in one

timothysmith0609:
...

**How is this accomplished?**
...

**What could go wrong?**
...

### DIFF
--- a/lib/krane/template_sets.rb
+++ b/lib/krane/template_sets.rb
@@ -96,7 +96,7 @@ module Krane
     private_constant :TemplateSet
 
     class << self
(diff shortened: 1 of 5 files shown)


lib/krane/template_sets.rb
```diff
@@ -96,7 +96,7 @@ def templates(filename:, raw:)
     private_constant :TemplateSet
 
     class << self
-      def from_dirs_and_files(paths:, logger:, render_erb: true)
+      def from_dirs_and_files(paths:, logger:, render_erb: true, partition: false)
```
timothysmith0609, line 99: we can remove `partition` key

PR #977, opened 2025-03-05
Release v3.7.1

jpfourny:

Release v3.7.1

### DIFF
--- a/lib/krane/version.rb
+++ b/lib/krane/version.rb
@@ -1,4 +1,4 @@
 # frozen_string_literal: true
 module Krane
-  VERSION = "3.7.0"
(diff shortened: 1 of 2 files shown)


charlescng (approved):

lib/krane/version.rb
```diff
@@ -1,4 +1,4 @@
 # frozen_string_literal: true
 module Krane
-  VERSION = "3.7.0"
+  VERSION = "3.7.1"
```
timothysmith0609, line 3: I'd argue this could be 3.8.0 since there's an additive change to the public CLI API, but I don't feel _that_ strongly

timothysmith0609 (approved):

PR #993, opened 2025-12-10
fix: validate that contexts exists before trying to build client

philippeboyd:

If the first file in environment variable `KUBECONFIG` is "empty" with simply the following structure

```yaml
apiVersion: v1
clusters: null
contexts: null
current-context: some-cluster
kind: Config
preferences: {}
users: null
```

`krane` will crash with the following error [when trying to access `contexts` in gem `kubeclient`](https://github.com/ManageIQ/kubeclient/blob/c6e29f9ce9bff133099da01e239dfbdd9710d95a/lib/kubeclient/config.rb#L45) because `contexts` is set to `null` in the empty config. 

```
'Kubeclient::Config#contexts': undefined method 'map' for nil (NoMethodError)

      @kcfg['contexts'].map { |x| x['name'] }
```

**How is this accomplished?**
By adding validation when creating `@kubeclient_configs` to simply skip files in `$KUBECONFIG` that have `null` contexts. This does parse the file twice tough (once in `krane` for validation then in `kubeclient`).

Is this the right approach? 

**What could go wrong?**
Nothing really, file with `null` contexts will be ignored when building the kube client.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,9 @@
 ## next
 
+- Skip kubeconfig files with no contexts defined instead of crashing when building the Kubernetes client
+
 ## 3.9.1
+
 - Test against k8s 1.34
 
 ## 3.9.0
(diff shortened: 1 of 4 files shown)


Mangara (approved):
LGTM once test failures are addressed.

ngc (approved):

PR #1001, opened 2026-09-14
Test against Kubernetes v1.35, drop v1.29

timothysmith0609:
Adds Kubernetes 1.35 to the CI test matrix and drops 1.29, which is out of upstream support.

### Changes

- **`.github/workflows/ci.yml`** — added `1.35.8` with its kind node image; removed `1.29.4` and its image
- **`dev.yml`** — bumped the local kind cluster to 1.35.8, matching the newest tested version (same as the last two bumps)
- **`README.md`** — added the 1.35 compatibility row, removed 1.29
- **`CHANGELOG.md` / `lib/krane/version.rb`** — 3.9.1 → 3.9.2

### On the patch version

Used **1.35.8**, the exact image published in [kind v0.33.0's release notes](https://github.com/kubernetes-sigs/kind/releases/tag/v0.33.0) (the current kind release):

```
kindest/node:v1.35.8@sha256:07b2536e30b803ed61d1677a79df6115f798ce64c80f9e22f6ed45afd09323c0
```

Verified before use:
- Digest matches both kind's release notes and Docker Hub. Cross-checking Docker Hub's digest for the existing `1.34.3` entry reproduced the committed value exactly, confirming the digest source is correct.
- `https://dl.k8s.io/release/v1.35.8/bin/linux/amd64/kubectl` returns HTTP 200, since CI downloads kubectl by the interpolated version string.
- Both YAML files parse, and every matrix version pairs with a correctly-matching kind image. Matrix stays at 96 jobs.

### Two notes for reviewers

1. **`MIN_KUBE_VERSION` is still `1.22.0`** (`lib/krane/common.rb:23`). Dropping 1.29 from the test matrix doesn't change what the gem enforces at runtime, and the previous deprecation commit (fd1d8b2, which dropped 1.26–1.28) didn't touch it either — so it's left alone to match precedent. If the intent is to stop *supporting* older versions rather than just stop testing them, that constant is a separate deliberate decision.

2. **The 1.29 README row was removed outright** rather than flipped to `No`, following #994, which deleted the 1.26–1.28 rows the same way. The "Last officially supported in gem version" column has been `--` for everything since 1.25, so it appears to no longer be tracked — but happy to record `3.9.1` there instead if preferred.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


renescheepers (approved):

renescheepers:
Add 1.36 as well

